PR #16, opened 2021-03-25
fix io cmakelist

beomki-yeo:
After `make install`, the following error occurs.

```
CMake Error at io/cmake_install.cmake:54 (file):
  file INSTALL cannot find "/home/beomki/projects/traccc/traccc/io/csv": No
  such file or directory.
```

So I changed the directory of csv installation.


asalzburger:
Hi, great - I really need to get a CI working for this.

PR #17, opened 2021-03-31
WIP: Cuda clusterizaion

beomki-yeo:
Currently working on implementing cuda for clusterization part. @asalzburger  @HadrienG2 

Major changes
1. separate algorithms folder out of core directory. (algorithms/cpu, algorithms/cuda)
2. cpu and cuda still shares same edm, geometry, definitions
3. vecmem was used for memory management

There are still lots of things to handle though: The algorithm is not completed yet and testing with massive data set is required


CMakeLists.txt
```diff
@@ -6,8 +6,13 @@ enable_testing()
 
 option(TRACCC_UNIT_TESTS "Enable unit tests" On)
 option(TRACCC_BENCHMARKS "Enable benchmark tests" On)
+option(TRACCC_BUILD_CUDA "Build CUDA" ON)
```
HadrienG2, line 9: This should probably be OFF by default. Not everyone who is interested in traccc has a working CUDA setup, and builds that fail / require exotic deps by default are not nice.

CMakeLists.txt
```diff
@@ -6,8 +6,13 @@ enable_testing()
...
 
-set(CMAKE_CXX_STANDARD 17)
+if(TRACCC_BUILD_CUDA)
+  set(CMAKE_CXX_STANDARD 17)
```
HadrienG2, line 12: I think that instead of making this conditional on TRACCC_BUILD_CUDA, we should make it configurable with a default value irrespective of whether CUDA is enabled or not, and have the if(TRACCC_BUILD_CUDA) part check that it's set to a standard that is compatible with the CUDA code.

algorithms/cpu/include/detail/.#sparse_ccl.hpp
```diff
@@ -0,0 +1 @@
+beomki@localhost.localdomain.1631329:1616872382
```
HadrienG2, line 1: You probably don't want to commit this.

algorithms/CMakeLists.txt~
```diff
@@ -0,0 +1,23 @@
...
+
+install(
+  DIRECTORY cpu/include/algorithms
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+
+install(
+  DIRECTORY cuda/include/algorithms
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+
+target_link_libraries(traccc_algorithms PUBLIC vecmem::core vecmem::cuda traccc::core)
```
HadrienG2, line 23: You probably don't want to commit this.

algorithms/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+add_library(traccc_algorithms SHARED ${CMAKE_CURRENT_SOURCE_DIR}/cuda/src/clusterization/component_connection_kernels.cu)
```
HadrienG2, line 1: Shouldn't this be conditional on TRACCC_BUILD_CUDA being on?

algorithms/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+  ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}
+  )
+
+install(
+  DIRECTORY cpu/include/algorithms
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
+
+install(
+  DIRECTORY cuda/include/algorithms
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```
HadrienG2, line 21: Same question as above.

algorithms/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+
+target_link_libraries(traccc_algorithms PUBLIC vecmem::core vecmem::cuda traccc_core)
```
HadrienG2, line 23: I suspect that we'll somehow want to ensure that vecmem::cuda is only linked to if CUDA is enabled.

core/include/definitions/algebra.hpp~
```diff
@@ -0,0 +1,495 @@
...
+        std::array<scalar, 3> normalize(const std::array<scalar, 3> &v)
+        {
+            scalar oon = 1. / std::sqrt(dot(v, v));
+            return {v[0] * oon, v[1] * oon, v[2] * oon};
+        }
+
+    } // namespace vector
+
+} // end of namespace
+
```
HadrienG2, line 495: You probably don't want to commit this.

core/include/definitions/algebra.hpp
```diff
@@ -8,6 +8,8 @@
 #pragma once
 
 #include "definitions/primitives.hpp"
+#include <cuda.h>
+#include <cuda_runtime.h>
```
HadrienG2, line 12: These includes should be conditional on CUDA being enabled.

core/include/definitions/algebra.hpp
```diff
@@ -17,41 +19,59 @@
 namespace traccc
 {
 
+#ifdef __CUDACC__      
+  __device__ __host__
+#endif
```
HadrienG2, line 24: You'll probably want to factor out those repeating three lines into a macro.

core/CMakeLists.txt
```diff
@@ -1,18 +1,18 @@
...
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-    $<INSTALL_INTERFACE:include>
-)
-
-install(TARGETS traccc_core
-        PUBLIC_HEADER
-          DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/core
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:include>
+  ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}
```
HadrienG2, line 11: You'll probably want to make this conditional on TRACCC_BUILD_CUDA.

algorithms/cuda/src/clusterization/component_connection_kernels.cu~
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  for (auto ms: ms_per_module){
+    //ms.local = 1./ms.total_weight * ms.local;
+
+    //printf("%f %f \n", ms.local[0], ms.local[1]);
+  }
+  
+}
+
```
HadrienG2, line 186: You probably didn't want to commit this.

algorithms/cuda/include/clusterization/component_connection_kernels.cuh~
```diff
@@ -0,0 +1,34 @@
...
+  void sparse_ccl_cuda(const vecmem::data::jagged_vector_view< cell >& cell_per_event,
+		  vecmem::data::jagged_vector_view< unsigned int >& label_per_event,
+		  vecmem::data::vector_view<unsigned int> num_labels);
+  
+  void sp_formation_cuda(const vecmem::data::jagged_vector_view< cell >& cell_per_event,
+		    const vecmem::data::jagged_vector_view< unsigned int >& label_per_event,
+		    const vecmem::data::vector_view<unsigned int> num_labels,
+		    vecmem::data::jagged_vector_view< measurement >& ms_per_event,
+		    vecmem::data::jagged_vector_view< spacepoint>& sp_per_event);
+}
```
HadrienG2, line 34: You probably didn't want to commit this.

core/CMakeLists.txt
```diff
@@ -23,12 +23,41 @@ install(
...
+  
+#  target_link_libraries(CudaAlgorithms PUBLIC vecmem::core vecmem::cuda traccc::core)
+#  target_include_directories(
+#    CudaAlgorithms
+#    PUBLIC
+#    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+#    $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+#    ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}
+#    )  
+#endif()
```
HadrienG2, line 61: This is clearly under active rework, so I'll not comment.

core/include/edm/measurement.hpp~
```diff
@@ -0,0 +1,33 @@
...
+        event_id event = 0;
+        geometry_id module = 0;
+        std::vector<measurement> items;
+
+        transform3 placement = transform3{};
+        
+    };
+
+    using measurement_container = std::vector<measurement_collection>;
+}
```
HadrienG2, line 33: You probably didn't want to commit this.

io/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ install(TARGETS traccc_io
 target_link_libraries(traccc_io INTERFACE traccc::core dfelibs)
 
 install(
-  DIRECTORY csv
+  DIRECTORY include/csv
```
HadrienG2, line 17: Nice catch!

core/include/edm/measurement.hpp
```diff
@@ -15,8 +15,9 @@ namespace traccc {
     
     /// A measurement definition, fix to two-dimensional here
     struct measurement {
-        point2 local = {0., 0.};
-        variance2 variance = { 0., 0.};
+      point2 local = {0., 0.};
+      variance2 variance = { 0., 0.};
+      scalar total_weight = 0;
```
HadrienG2, line 20: For my understanding, why is this new weight needed ?

algorithms/cuda/src/clusterization/component_connection_kernels.cu
```diff
@@ -0,0 +1,185 @@
...
+		       vecmem::data::vector_view<unsigned int> _num_labels,
+		       vecmem::data::jagged_vector_view< measurement > _ms_per_event,
+		       vecmem::data::jagged_vector_view< spacepoint > _sp_per_event){
+  int gid = blockDim.x * blockIdx.x + threadIdx.x;
+
+  vecmem::jagged_device_vector< cell > cell_per_event(_cell_per_event);
+  vecmem::jagged_device_vector< unsigned int > label_per_event(_label_per_event);  
+  vecmem::device_vector< unsigned int > num_labels(_num_labels);
+  vecmem::jagged_device_vector< measurement > ms_per_event(_ms_per_event);
+  vecmem::jagged_device_vector< spacepoint > sp_per_event(_sp_per_event);
```
HadrienG2, line 156: Note for future vecmem dev: it would be nice if the API could be adjusted so that there is no need for such "redeclaration".

HadrienG2 (commented):
Thanks, it's super nice to have a GPU-based vecmem usage example like this. This will certainly come in handy for us at IJCLab as we intend to start doing a SyCL port soon.

Although I'm well aware that this PR is WIP, I left a bunch of comments on things which I'm pretty sure you'll want to change sooner or later, just in case you didn't catch some of them.

core/include/edm/measurement.hpp~
asalzburger: We should add a rule in `.gitignore`.

algorithms/cuda/src/clusterization/component_connection_kernels.cu
beomki-yeo: Certainly. I am not sure if it is possible technically though. 
I think it is gonna be nice if we can ping attila here?

core/include/edm/measurement.hpp
beomki-yeo: Depending on how to design the kernel function, we can skip this.

In the cpu side, measurement is created only when the sum of hit weights > 0, so we don't need to have this feature.
In the cuda side, I created the measurement regardless of the weight sum - why? If I want to follow the same strategy of cpu code, I have to write one more kernel function for measurement creation and record the number of measurements per module. This was thing that I didn't want to do.

But I also think putting a feature that is not used by cpu side is not desirable at the same time. I will look into this more to find a better solution

core/include/definitions/algebra.hpp
krasznaa: Wait... does this actually work? 😕  I was convinced that not even `std::array` could be used directly in device code. Can it? 😕  (I don't know, maybe the latest CUDA version allows for this?...)

Similar to [llama::Array](https://github.com/alpaka-group/llama/blob/develop/include/llama/Array.hpp), I added [vecmem::static_vector](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/static_vector.hpp) exactly for this type of usage. To handle small objects that are created directly in device code. (And can even be passed from the host to the device or vice-versa as they are.) I've had the plan for a while to introduce a `vecmem::static_array` as well, which would pretty much just be a simplified version of `vecmem::static_vector`...

tests/algorithms/component_connection_tests_cuda.cpp
```diff
@@ -0,0 +1,91 @@
...
+  
+  // fill jagged_vector
+  cell_per_event.push_back(cell_per_module);
+  label_per_event.push_back(label_per_module);
+
+  vecmem::vector< unsigned int> num_labels(cell_per_event.size(), 0 , &managed_resource);
+
+  // run sparse_ccl
+  vecmem::data::jagged_vector_data< traccc::cell> cell_data(cell_per_event,&managed_resource);
+  vecmem::data::jagged_vector_data< unsigned int > label_data(label_per_event, &managed_resource);
```
krasznaa, line 58: You could simplify these calls to:

```c++
auto cell_data = vecmem::get_data( cell_per_event, &managed_resource );
auto label_data = vecmem::get_data( label_per_event, &managed_resource );
```

At least as long as you use the "managed resource". For explicit memory movements the code will unfortunately need to become a bit more elaborate. But that's something that we're still in the process of figuring out ourselves... (https://github.com/acts-project/vecmem/pull/60)

krasznaa (commented):
Very interesting work Beomki! We should really find some time right after Easter with you and @stephenswat to sit down for a bit of a discussion on how to split these developments between all of us.

I do very much like the direction that you started out in with this PR. 😄

algorithms/cuda/src/clusterization/component_connection_kernels.cu
```diff
@@ -0,0 +1,185 @@
...
+
+  void sparse_ccl_cuda(
+		  const vecmem::data::jagged_vector_view< cell >& cell_per_event,
+		  vecmem::data::jagged_vector_view< unsigned int>& label_per_event,
+		  vecmem::data::vector_view<unsigned int> num_labels){    
+    sparse_ccl_kernel<<< 1, cell_per_event.m_size >>>(cell_per_event,
+						    label_per_event,
+						    num_labels);        
+    VECMEM_CUDA_ERROR_CHECK(cudaGetLastError());
+    VECMEM_CUDA_ERROR_CHECK(cudaDeviceSynchronize());
```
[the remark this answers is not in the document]
krasznaa: Never mind... Apparently "WIP:" is meaningful here as well. It's just quite confusing that there is a separate mechanism in place as well for declaring a PR as a draft... 😕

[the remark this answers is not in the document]
paulgessinger: Yeah, one of them is through a Bot that's added to the org and the other is GH native. We use the bot on the ACTS main repo to prevent merging WIP PRs. I think I can disable the Bot for this repo if you prefer this.

algorithms/cuda/src/clusterization/component_connection_kernels.cu
paulgessinger: I suppose you could make the types convertible from one to another so improve the ergonomics. This could of course cause problems in cases where you don't want that to occur.

core/include/definitions/algebra.hpp
krasznaa: Hmm... Apparently CUDA 11.2 (didn't try older versions...) **can** be forced into using [std::array](https://en.cppreference.com/w/cpp/container/array). :thinking: I wrote this little test for it:

```cuda
#include <cuda_runtime.h>

#include <array>
#include <iostream>
#include <vector>

__global__
void arrayTestKernel( int size, int* ptr ) {

   const int i = blockIdx.x * blockDim.x + threadIdx.x;
   if( i >= size ) {
      return;
   }

   const std::array< int, 2 > constants = { 2, 3 };
   ptr[ i ] = ptr[ i ] * constants.at( 0 ) + constants.back();
   return;
}

int main() {

   std::vector< int > host_vec = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
   int* device_array = nullptr;
   cudaMalloc( &device_array, host_vec.size() * sizeof( int ) );
   cudaMemcpy( device_array, host_vec.data(), host_vec.size() * sizeof( int ),
               cudaMemcpyHostToDevice );

   arrayTestKernel<<< 1, host_vec.size() >>>( host_vec.size(), device_array );

   cudaMemcpy( host_vec.data(), device_array, host_vec.size() * sizeof( int ),
               cudaMemcpyDeviceToHost );
   cudaFree( device_array );

   const std::vector< int > expected_result = {
      5, 7, 9, 11, 13, 15, 17, 19, 21, 23 };
   if( host_vec != expected_result ) {
      std::cerr << "What?!?" << std::endl;
      return 1;
   }
   std::cout << "All good!" << std::endl;

   return 0;
}
```

Which works like:

```
[bash][Elrond]:test > nvcc -o test --expt-relaxed-constexpr array_test.cu 
[bash][Elrond]:test > ./test 
All good!
[bash][Elrond]:test >
```

So even with CUDA 11.2 one needs to use an experimental feature for this. But it does work... So it's a nay on the `vecmem::static_array` implementation... :stuck_out_tongue:

P.S. [std::vector](https://en.cppreference.com/w/cpp/container/vector) is of course still a no-go in device code.

```
/data/projects/vecmem/vecmem/tests/cuda/array_test.cu(16): error: calling a __host__ function("std::allocator<int> ::allocator") from a __global__ function("arrayTestKernel") is not allowed

/data/projects/vecmem/vecmem/tests/cuda/array_test.cu(16): error: identifier "std::allocator<int> ::allocator" is undefined in device code

/data/projects/vecmem/vecmem/tests/cuda/array_test.cu(16): error: calling a __host__ function("std::vector<int, ::std::allocator<int> > ::vector") from a __global__ function("arrayTestKernel") is not allowed

/data/projects/vecmem/vecmem/tests/cuda/array_test.cu(16): error: identifier "std::vector<int, ::std::allocator<int> > ::vector" is undefined in device code

/data/projects/vecmem/vecmem/tests/cuda/array_test.cu(16): error: calling a __host__ function("std::allocator<int> ::~allocator") from a __global__ function("arrayTestKernel") is not allowed
...
```

So `vecmem::static_vector` will continue having a use-case. :wink:

beomki-yeo: @HadrienG2 Sure. I will do it
@krasznaa Good point. I used the `--expt-relaxed-constexpr` to make it work. 
The general discussion point is whether to write `algebra.hpp` for cpu and cuda separately. If we want to separate them, the featurer of `static::array` or `static::vector` (or even `static::tuple`) will be definitely useful.

tests/algorithms/component_connection_tests_cuda.cpp
beomki-yeo: Is that right? I got the following error: 
```
error: no matching function for call to ‘get_data(vecmem::data::jagged_vector_data<unsigned int>&, vecmem::cuda::managed_memory_resource*)’
```

stephenswat: The `get_data` method is designed to work with jagged standard vectors, so either `std::vector<std::vector<T>>` or `vecmem::vector<vecmem::vector<T>>`. If you already have a `jagged_vector_data` there should be no reason to use them. :slightly_smiling_face:

algorithms/cuda/src/clusterization/component_connection_kernels.cu
krasznaa: The types are already convertible.
  - https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/device_vector.hpp#L63-L65
  - https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/jagged_device_vector.hpp#L82-L85

But if you want to make use of a "vector type" you do need to declare its concrete type somewhere in your code.

In the meantime I also thought a bit more about the helper function idea that I brought up before. But that won't really work. Since there is no one-to-one association from data to "interactive" types in the vecmem code. (At the moment `vecmem::data::vector_view` can be used as an input to both `vecmem::device_vector` and `vecmem::device_array`.)

tests/algorithms/component_connection_tests_cuda.cpp
krasznaa: I'm confused though. 😦  The code that we see here has `cell_per_event` and `label_per_event` both declared as `vecmem::jagged_vector` types. So I'm not sure how you ended up calling `vecmem::get_data(...)` on a `vecmem::data::jagged_vector_data` type here... 😕

PR #19, opened 2021-04-09
Another CUDA implementation of connected component labeling

stephenswat:
This merge request (which for now is a complete mess to be cleaned up later) adds some CUDA connected component labeling to traccc.

### DIFF
--- a/device/cuda/include/traccc/cuda/cca/component_connection.hpp
+++ b/device/cuda/include/traccc/cuda/cca/component_connection.hpp
@@ -1,0 +1,21 @@
+/*
+ * TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include "traccc/edm/cell.hpp"
+#include "traccc/edm/measurement.hpp"
+#include "traccc/utils/algorithm.hpp"
+
+namespace traccc::cuda {
+struct component_connection
+    : algorithm<host_measurement_container(const host_cell_container& cells)> {
+    host_measurement_container operator()(
+        const host_cell_container& cells) const;
+};
+}  // namespace traccc::cuda
--- a/device/cuda/src/cca/component_connection.cu
+++ b/device/cuda/src/cca/component_connection.cu
@@ -1,0 +1,581 @@
+/*
+ * TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include "traccc/cuda/cca/component_connection.hpp"
+#include "traccc/cuda/utils/definitions.hpp"
+#include "traccc/edm/cell.hpp"
+#include "traccc/edm/cluster.hpp"
+#include "traccc/edm/measurement.hpp"
+#include "vecmem/containers/vector.hpp"
+#include "vecmem/memory/allocator.hpp"
+#include "vecmem/memory/binary_page_memory_resource.hpp"
+#include "vecmem/memory/cuda/managed_memory_resource.hpp"
+
+namespace {
+static constexpr std::size_t MAX_CELLS_PER_PARTITION = 2048;
+static constexpr std::size_t THREADS_PER_BLOCK = 256;
+using index_t = unsigned short;
+}  // namespace
+
+namespace traccc::cuda {
+namespace details {
+struct ccl_partition {
+    std::size_t start;
+    std::size_t size;
+};
+
+struct cell_container {
+    std::size_t size = 0;
+    channel_id* channel0 = nullptr;
+    channel_id* channel1 = nullptr;
+    scalar* activation = nullptr;
+    scalar* time = nullptr;
+    geometry_id* module_id = nullptr;
+};
+
+struct measurement_container {
+    unsigned int size = 0;
+    scalar* channel0 = nullptr;
+    scalar* channel1 = nullptr;
+    scalar* variance0 = nullptr;
+    scalar* variance1 = nullptr;
+    geometry_id* module_id = nullptr;
+};
+
+__device__ bool is_adjacent(channel_id ac0, channel_id ac1, channel_id bc0,
+                            channel_id bc1) {
+    unsigned int p0 = (ac0 - bc0);
+    unsigned int p1 = (ac1 - bc1);
+
+    return p0 * p0 <= 1 && p1 * p1 <= 1;
+}
+
+__device__ void reduce_problem_cell(cell_container& cells, index_t tid,
+                                    unsigned char& adjc, index_t adjv[]) {
+    /*
+     * The number of adjacent cells for each cell must start at zero, to
+     * avoid uninitialized memory. adjv does not need to be zeroed, as
+     * we will only access those values if adjc indicates that the value
+     * is set.
+     */
+    adjc = 0;
+
+    channel_id c0 = cells.channel0[tid];
+    channel_id c1 = cells.channel1[tid];
+    geometry_id gid = cells.module_id[tid];
+
+    /*
+     * First, we traverse the cells backwards, starting from the current
+     * cell and working back to the first, collecting adjacent cells
+     * along the way.
+     */
+    for (index_t j = tid - 1; j < tid; --j) {
+        /*
+         * Since the data is sorted, we can assume that if we see a cell
+         * sufficiently far away in both directions, it becomes
+         * impossible for that cell to ever be adjacent to this one.
+         * This is a small optimisation.
+         */
+        if (cells.channel1[j] + 1 < c1 || cells.module_id[j] != gid) {
+            break;
+        }
+
+        /*
+         * If the cell examined is adjacent to the current cell, save it
+         * in the current cell's adjacency set.
+         */
+        if (is_adjacent(c0, c1, cells.channel0[j], cells.channel1[j])) {
+            adjv[adjc++] = j;
+        }
+    }
+
+    /*
+     * Now we examine all the cells past the current one, using almost
+     * the same logic as in the backwards pass.
+     */
+    for (index_t j = tid + 1; j < cells.size; ++j) {
+        /*
+         * Note that this check now looks in the opposite direction! An
+         * important difference.
+         */
+        if (cells.channel1[j] > c1 + 1 || cells.module_id[j] != gid) {
+            break;
+        }
+
+        if (is_adjacent(c0, c1, cells.channel0[j], cells.channel1[j])) {
+            adjv[adjc++] = j;
+        }
+    }
+}
+
+__device__ void fast_sv_1(index_t* f, index_t* gf, unsigned char adjc[],
+                          index_t adjv[][8], unsigned int size) {
+    /*
+     * The algorithm finishes if an iteration leaves the arrays unchanged.
+     * This varible will be set if a change is made, and dictates if another
+     * loop is necessary.
+     */
+    bool gfc;
+
+    do {
+        /*
+         * Reset the end-parameter to false, so we can set it to true if we
+         * make a change to our arrays.
+         */
+        gfc = false;
+
+        /*
+         * The algorithm executes in a loop of three distinct parallel
+         * stages. In this first one, we examine adjacent cells and copy
+         * their cluster ID if it is lower than our, essentially merging
+         * the two together.
+         */
+        for (index_t tst = 0, tid;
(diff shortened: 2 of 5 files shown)


stephenswat:
Okay, I've updated this bit of CUDA code to provide one-event-one-kernel semantics. On my GPU, which is an NVIDIA GeForce 1660 Ti, this code processes an entire event in roughly 3.5 milliseconds. Note that this kernel is a CCA kernel, not just a CCL kernel, which means that it does cluster finding and measurement creation in one go.

core/include/definitions/algebra.hpp
```diff
@@ -17,42 +17,42 @@
 namespace traccc
 {
 
-    std::array<scalar, 2> operator*(const std::array<scalar, 2> &a, scalar s)
+    inline std::array<scalar, 2> operator*(const std::array<scalar, 2> &a, scalar s)
```
paulgessinger, line 20: This is something we should probably harmonize with `detray`. @niermann999 @asalzburger

device/cuda/src/component_connection.cpp
```diff
@@ -0,0 +1,77 @@
...
+            total_cells += data.items.at(i).size();
+        }
+
+        vecmem::vector<cell> cells(&mem);
+        cells.reserve(total_cells);
+
+        vecmem::vector<unsigned int> blocks(data.headers.size() + 1, &mem);
+        blocks[0] = 0;
+
+        vecmem::vector<float> out(4 * MAX_CLUSTERS_PER_MODULE * data.headers.size(), &mem);
```
paulgessinger, line 41: Probably I overlooked it, but where is `MAX_CLUSTERS_PER_MODULE` defined?

device/cuda/src/component_connection.cpp
```diff
@@ -0,0 +1,77 @@
...
+        //     std::cout << std::endl;
+        //     for (std::size_t j = 0; j < 16; ++j) {
+        //         std::cout << i << ", " << j << " (" << out[4 * 128 * i + 4 * j] <<
+        //         ", " << out[4 * 128 * i + 4 * j + 1] << ") with variance (" <<
+        //         out[4 * 128 * i + 4 * j + 2] << ", " << out[4 * 128 * i + 4 * j + 3] <<
+        //         ")" << std::endl;
+        //     }
+        // }
+
+        return {};
```
paulgessinger, line 75: I guess ultimately this should return something, right?

device/cuda/src/details/sparse_ccl.cu
```diff
@@ -0,0 +1,343 @@
...
+
+        /*
+         * Seek the correct cell region in the input data. Again, this is all a
+         * contiguous block of memory for now, and we use the blocks array to
+         * define the different ranges per block/module. At the end of this we
+         * have the starting address of the block of cells dedicated to this
+         * module, and we have its size.
+         */
+        const cell * cells = &_cells[blocks[blockIdx.x]];
+        const unsigned int size = blocks[blockIdx.x + 1] - blocks[blockIdx.x];
```
paulgessinger, line 45: Does `blocks[blockIdx.x + 1]` not go out of bounds for the last block?

device/cuda/src/details/sparse_ccl.cu
```diff
@@ -0,0 +1,343 @@
...
+
+            unsigned int tc0 = cells[tid].channel0;
+            unsigned int tc1 = cells[tid].channel1;
+
+            /*
+             * First, we traverse the cells backwards, starting from the current
+             * cell and working back to the first, collecting adjacent cells
+             * along the way.
+             */
+            for (unsigned int j = tid - 1; j < tid; --j) {
```
paulgessinger, line 138: `j >= 0`? Guessing this works because `unsigned int` wraps around, in fact becoming greater than `tid`.

device/cuda/src/details/sparse_ccl.cu
```diff
@@ -0,0 +1,343 @@
...
+
+        /*
+         * These arrays are the meat of the pudding of this algorithm, and we
+         * will constantly be writing and reading from them which is why we
+         * declare them to be in the fast shared memory. Note that this places a
+         * limit on the maximum activations per module, as the amount of shared
+         * memory is limited. These could always be moved to global memory, but
+         * the algorithm would be decidedly slower in that case.
+         */
+        __shared__ unsigned int f[MAX_ACTIVATIONS_PER_MODULE], gf[MAX_ACTIVATIONS_PER_MODULE];
```
paulgessinger, line 82: You could be a bit more explicit (in the comment or the variables names) what `f` and `gf` **are**.

device/cuda/src/details/sparse_ccl.cu
```diff
@@ -0,0 +1,343 @@
...
+            __syncthreads();
+
+            /*
+             * To determine whether we need another iteration, we use block
+             * voting mechanics. Each thread checks if it has made any changes
+             * to the arrays, and votes. If any thread votes true, all threads
+             * will return a true value and go to the next iteration. Only if
+             * all threads return false will the loop exit.
+             */
+        } while (__any_sync(__activemask(), gfc));
```
paulgessinger, line 267: I can imagine what `__any_sync` does, but what does `__activemask()` to?

device/cuda/src/details/sparse_ccl.cu
```diff
@@ -0,0 +1,343 @@
...
+         * move onto the actual processing part.
+         */
+        __syncthreads();
+
+        /*
+         * The algorithm finishes if an iteration leaves the arrays unchanged.
+         * This varible will be set if a change is made, and dictates if another
+         * loop is necessary.
+         */
+        bool gfc;
```
paulgessinger, line 194: Is this `bool` thread local?

paulgessinger (commented):
From what I can tell, this looks sensible.

core/include/definitions/algebra.hpp
stephenswat: Sure thing. I have found these need to be inlined to avoid link-time conflicts, but if there is a better solution that is fine by me.

device/cuda/src/component_connection.cpp
stephenswat: It's defined in the header file!

device/cuda/src/component_connection.cpp
stephenswat: Yes, the only reason this doesn't return anything is that the EDM is still in flux, but once that dust settles we can just return the data here.

device/cuda/src/details/sparse_ccl.cu
stephenswat: The blocks vector is defined as:

```cpp
vecmem::vector<unsigned int> blocks(data.headers.size() + 1, &mem);
```

So it's one larger than the number of blocks. :slightly_smiling_face:

device/cuda/src/details/sparse_ccl.cu
stephenswat: `i >= 0` on decrementing unsigned integer loops is an anti-pattern and will keep looping forever. This is how I've always been taught to do loops for unsigned integers.

device/cuda/src/details/sparse_ccl.cu
stephenswat: Yes, this is absolutely true. Good point, thank you! :smile:

device/cuda/src/details/sparse_ccl.cu
stephenswat: The `__any_sync` function takes a thread mask, which is a 32-bit number where the _i_th bit indicates whether the _i_th thread of the 32-thread warp is currently executing. Remember, when a GPU thread skips a conditional block it doesn't just skip the program counter to the end, it just marks itself as inactive until the end of the block. Active mask gives you that mask. Requesting a response in the `__any_sync` from all the threads unconditionally can lead to dead-lock.

device/cuda/src/details/sparse_ccl.cu
stephenswat: Yes! The sharing of whether any element of the union-find set was changed is shared between threads only at the end, when the `__any_sync` fires.

beomki-yeo:
just passing by - @stephenswat what is your plan for updating the cuda clusterization code?

device/cuda/src/details/sparse_ccl.cuh
```diff
@@ -0,0 +1,23 @@
...
+
+#include <vector>
+
+#include "cuda/algorithms/component_connection.hpp"
+#include "edm/cell.hpp"
+#include "edm/measurement.hpp"
+#include "vecmem/memory/memory_resource.hpp"
+
+#define MAX_ACTIVATIONS_PER_PARTITION 2048
+#define MAX_CLUSTERS_PER_PARTITION 512
```
beomki-yeo, line 11: This seems not being used

beomki-yeo (commented):
Sorry for too late review and thanks for wonderful job!
I didn't fully follow the `fast_sv` algorithm in detail but it should be okay if you get the same result with the CPU version - The final output should be unique regardless of the algorithm

Here are the general comments but you don't have to resolve them in this PR:
1. It seems you used GPU-friendly SoA data structure. Thing is that a certain algorithm can benefit from it when the thread divergence is minimized and data access is coalesced. Regarding that, some parts of `fast_sv` in this PR seem not obeying the requirements. (Especially in for loops)  We might need to know which between data conversion and coalesced data access has more impact in speed.

2. I should have let you know this earlier. Aggregation here is quite smart way to skip the cluster EDM. However it can be problematic in the future when we need to split the cluster formed by multiple particles. (In ATLAS the cluster is resolved by NN) You don't have to take it seriously for now but just keep it in mind.

+ Could you update PR so we can merge it later

device/cuda/src/details/sparse_ccl.cu
```diff
@@ -0,0 +1,450 @@
...
+#include "vecmem/memory/memory_resource.hpp"
+
+namespace {
+using index_t = unsigned short;
+}
+
+namespace traccc {
+namespace cuda {
+namespace details {
+__device__ bool is_adjacent(const cell& a, const cell& b) {
```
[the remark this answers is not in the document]
stephenswat: Well spotted, I have removed it for now. :+1:

device/cuda/src/details/sparse_ccl.cuh
stephenswat: Also removed. :smile:

device/cuda/src/details/sparse_ccl.cu
```diff
@@ -0,0 +1,450 @@
...
+         * To determine whether we need another iteration, we use block
+         * voting mechanics. Each thread checks if it has made any changes
+         * to the arrays, and votes. If any thread votes true, all threads
+         * will return a true value and go to the next iteration. Only if
+         * all threads return false will the loop exit.
+         */
+    } while (__syncthreads_or(gfc));
+}
+
+__device__ void fast_sv_2(index_t* f, index_t* gf, unsigned char adjc[],
```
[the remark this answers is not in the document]
stephenswat: Let's keep this around for now, we can use it to do some performance measurements and comparisons down the line, but at some point we should remove it and stick to a single algorithm.

stephenswat:
This now depends on #153.

stephenswat:
Are there any further comments on this? Otherwise I think we can move towards merging it in.

device/cuda/src/cca/component_connection.cu
```diff
@@ -0,0 +1,583 @@
+/*
+ * TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include <iostream>
```
beomki-yeo, line 9: I think this is a left over?

beomki-yeo (approved):
Let's merge after some trivial changes

device/cuda/include/traccc/cuda/cca/component_connection.hpp
```diff
@@ -0,0 +1,21 @@
+/*
+ * TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
```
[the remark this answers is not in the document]
stephenswat: Hah, well spotted!

device/cuda/src/cca/component_connection.cu
stephenswat: Indeed, good catch! :smile:

stephenswat:
Good call, I've resolved the issues.

stephenswat:
@beomki-yeo can you re-approve?

beomki-yeo (approved):

PR #23, opened 2021-04-12
Cell EDM Update, main branch (2021.04.12.)

krasznaa:
So... This is how I imagine that we should try to organise the EDM in traccc using [vecmem](https://github.com/acts-project/vecmem).

The design of the updated `traccc::cell_collection` and `traccc::cell_container` structs is the same as what I outlined a while ago in https://github.com/krasznaa/traccc/commit/6a75c4915d49c97fe69763e2c1475ebfd1f93c03, and what @stephenswat did in #19. It makes `traccc::cell_container` into a struct that holds a "simple vector" of "header information" for each detector module in the event, and a "jagged vector" that holds all of the cells belonging to all of those detector modules.

It's just that, unlike the previous 2 examples, this time around I did all of this using the VecMem types in a (hopefully) clever templated way. Abstracting the vector types out of the container types. Plus I added a bunch of code for handling the data from the updated `traccc::cell_container` in a GPU friendly way. But none of that is exercised in the code at this point yet.

Then I went and updated all of the algorithmic code to use the updated EDM correctly. Most of this meant relatively simple changes. Making sure that the CSV reading code would make use of `vecmem::memory_resource` and that "cell" and "module" information would be handled in separate objects.

The only really tricky part (in my mind at least) is what I did with `traccc::component_connection`. I was desperately trying to re-write that code in a templated way in which the compiler could correctly instantiate the type's operators based on what specialisation of `traccc::cell_collection` they receive. But as I explained in a detailed e-mail on the `atlas-sw-core` mailing list over the weekend (sorry for those not in ATLAS :frowning:), this doesn't work quite as well as I was hoping it would. I can give a longer explanation in the discussion if needed.

Instead I created a separate public and private interface for `traccc::component_connection`. The public interface provides functions specifically for the "host" and "device" EDM types, while the templated implementation was made private. This way the user can make use of the class with simple calls, while internally the algorithm is still only implemented once. This is not as elegant as I was hoping it to be, but I thought for now it should suffice.

Note that basically all functions will eventually need to be re-written never to return EDM objects, but just to operate on existing ones. So that the algorithms would not have to decide themselves how the memory management of the objects would be set up. (Avoiding the change in this PR for instance, where `traccc::read_cells(...)` needs to receive a `vecmem::memory_resource` parameter.) But I didn't want to do that already in this PR.

For a proper implementation of the clusterization on GPUs, the cluster EDM will need to be re-written in a similar fashion as well. We should discuss in the PR whether that should be added to this PR, or be done separately once/if this PR goes in.

Pinging @stephenswat, @paulgessinger, @asalzburger, @beomki-yeo and @SylvainJoube.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,31 +1,40 @@
-cmake_minimum_required(VERSION 3.9)
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2021 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
 
+# Set up the project.
+cmake_minimum_required(VERSION 3.9)
 project(traccc VERSION 0.0 LANGUAGES CXX)
 
-enable_testing()
+# Standard CMake include(s).
+include( CTest )
+include( GNUInstallDirs )
 
-option(TRACCC_UNIT_TESTS "Enable unit tests" On)
-option(TRACCC_BENCHMARKS "Enable benchmark tests" On)
+# Explicitly set the output directory for the binaries. Such that if this
+# project is included by another project, the main project's configuration would
+# win out.
+set( CMAKE_RUNTIME_OUTPUT_DIRECTORY
+   "${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_BINDIR}" CACHE PATH
+   "Directory for the built binaries" )
+set( CMAKE_LIBRARY_OUTPUT_DIRECTORY
+   "${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}" CACHE PATH
+   "Directory for the built libraries" )
+set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY
+   "${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_LIBDIR}" CACHE PATH
+   "Directory for the built static libraries" )
 
-set(CMAKE_CXX_STANDARD 17)
+# Include the traccc CMake code.
+list( APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake" )
+include( traccc-compiler-options )
 
-set(CMAKE_INSTALL_INCLUDEDIR ${CMAKE_INSTALL_PREFIX}/include)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_INSTALL_PREFIX}/lib)
-set(CMAKE_INSTALL_LIBDIR ${CMAKE_INSTALL_PREFIX}/lib)
-
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin")
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
-
-set(TF_BUILD_EXAMPLES OFF)
-set(TF_BUILD_TESTS OFF)
-set(INSTALL_GTEST OFF)
-set(BENCHMARK_ENABLE_INSTALL OFF)
-set(BENCHMARK_ENABLE_TESTING OFF)
-
+# Build the necessary externals.
 add_subdirectory(extern)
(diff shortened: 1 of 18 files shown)


krasznaa:
In case anybody actually had a look...

I spent the last few hours re-writing the cluster EDM to use VecMem as well. That turned out to be even more difficult, since the current implementation uses a vector of a vector of a vector type.

But finally I did manage to update all of the existing code to work (seemingly) correctly. So in principle this sort of design for the EDM code should be possible to extend to all of the existing EDM types.

It's worth noting though that the CPU code did become noticeably slower after my EDM modifications. :frowning: In an optimised build, after the cell **and** cluster modifications, the `seq_example` test became ~20% slower. While a debug build became ~5x slower. :confused: Of course I modified the algorithms in a pretty careless way, only caring about their correctness, and not their performance. Still, this does look at least a little worrisome... At least that's "only" due to my latest changes with the cluster EDM. The performance of the code as it is in this PR, is the same as the current HEAD, within statistical fluctuations.

paulgessinger:
I want to look at this in some detail, but I'm not sure when I'll get around to do it, so don't wait for it before merging.

HadrienG2:
> It's worth noting though that the CPU code did become noticeably slower after my EDM modifications. frowning In an optimised build, after the cell **and** cluster modifications, the `seq_example` test became ~20% slower. While a debug build became ~5x slower. confused Of course I modified the algorithms in a pretty careless way, only caring about their correctness, and not their performance. Still, this does look at least a little worrisome... At least that's "only" due to my latest changes with the cluster EDM. The performance of the code as it is in this PR, is the same as the current HEAD, within statistical fluctuations.

Indeed, this is surprising, considering how straightforward your changes look. It would be worthwhile to throw a profiler at this code just in case the compiler is doing something silly like not inlining a small vecmem utility function...

If nothing of the sort is happening, one possible explanation would be that because the performance of CCL is fully bound by memory accesses, the new layout could somehow introduces new indirections or data dependencies that weren't there before?

beomki-yeo:
I think this MR is good to be merged as it works perfecty with cuda as well

asalzburger:
Hi, I had a look through it and I agree with @beomki-yeo that this could go in to my mind.

Not sure how we do agreed to do approvals on `traccc`, at the moment we certainly can commont on PRs still, but probably we should use the approval functionality?

beomki-yeo:
I suggest similar policy that acts main branch uses could go

beomki-yeo:
As mentioned above, the data reading with new EDM seems a bit slower, especially with managed_memory of cuda.
@HadrienG2 For ccl algorithm, two different edm w/ cpu show similar time at least.

original edm
==> Elapsed time... 
- read: 3.70073
- ccl: 0.226442

new edm
==> Elapsed time... 
- reading time |  cpu: 3.94691 |  cuda: 5.52805
- ccl time        |  cpu: 0.22153 |  cuda: 0.0195265

krasznaa:
Hi Beomki,

Thanks for the numbers!

I wouldn't be worried about the "reading time" going up in your setup all too much. Additional profiling will be useful to do on this, but we do know that CUDA memory allocations are expensive. And at the moment your code is not doing any "fancy" memory management with VecMem yet (I guess). So during the file reading I expect that many more CUDA memory allocations and deallocations are executed at the moment than absolutely necessary.

It's also very reassuring to see that the algorithm time didn't change in your tests either. Of course there any slowdown would only come from an inefficient memory layout in my PR. Since the host memory allocated by VecMem is exactly the same (for the purposes of the CPU code at least) as the one allocated by `malloc(...)`.

So, for the reading I think we'll be able to improve on things with using the "clever" (high level) memory resources provided by VecMem, and also by just adding a couple of `std::vector::reserve(...)` calls to the code... :stuck_out_tongue:

Cheers,
        Attila

krasznaa (commented):
Could somebody finally merge this in?... Naturally I can't approve this myself...

beomki-yeo:
@asalzburger @paulgessinger @HadrienG2 I also don't know who has a permission for approval so pinging random people

HadrienG2 (approved):
I can approve, but not merge

PR #24, opened 2021-04-13
Add a simple clusterization example program

stephenswat:
The aim of this little example program is to augment the existing sequence example, but it only runs the connected component labeling algorithm. It gives us some statistics about the event file read, which I have found insightful. It also gives some information about the runtime of the clusterization algorithm.

```
stephen $ ./build/bin/ccl_example data/tml_pixels/event000000000-cells.csv                              
Running ./build/bin/ccl_example on data/tml_pixels/event000000000-cells.csv

Non-zero pixels per module
  Min     Max | Count
    0 -     0 |     0 
    1 -     1 |    41 ***
    2 -     2 |    65 *****
    3 -     3 |    70 *****
    4 -     5 |   206 *************
    6 -     7 |   280 ******************
    8 -    10 |   549 ***********************************
   11 -    15 |   792 **************************************************
   16 -    22 |   484 *******************************
   23 -    31 |   337 **********************
   32 -    44 |   266 *****************
   45 -    63 |   196 *************
   64 -    90 |   160 **********
   91 -   127 |   125 ********
  128 -   180 |    57 ****
  181 -   255 |    12 *
  256 -   361 |    34 ***
  362 -   511 |   103 *******
  512 -   723 |    73 *****
  724 -  1023 |     9 *
 1024 -  1447 |     0 
 1448 -  2047 |     0 
 2048 -  2895 |     0 
 2896 -  4095 |     0 
 4096 -  5792 |     0 
 5793 -  8191 |     0 
 8192 -   inf |     0 

CPU budget allocation
    Component |       Runtime
 Data loading |    668.480 ms
   Statistics |      0.357 ms
   Clustering |     38.880 ms
        Total |    707.717 ms
```


krasznaa:
Looks useful. I just wonder if for performance measurements we should add multiple execution of the algorithm to the code.

I'm just thinking of what we discussed yesterday. That depending on the memory management of the application, the first execution may be noticeably slower than the consequent ones.

Maybe we should not use this particular executable to measure such a thing, but as part of the eventual performance profiling, I think it would be good to do something along those lines. (Being able to process the same event, or possibly multiple events, over and over again a set number of times.)

stephenswat:
Good call. I've modified the executable to run the CCL once cold, then it will run it ten times to heat up any caches, and then it will benchmark again with everything hot:

```
CPU budget allocation
    Component |       Runtime
 Data loading |    695.594 ms
   Statistics |      0.339 ms
     Cold run |     39.010 ms
      Heating |    385.630 ms
      Hot run |     38.913 ms
        Total |   1159.487 ms
```

examples/cpu/ccl_example.cpp
```diff
@@ -0,0 +1,136 @@
...
+    for (std::size_t i = 0; i < 10; ++i) {
+        for (const traccc::cell_collection & module : modules) {
+            traccc::cluster_collection clusters_per_module = cc(module);
+        }
+    }
+
+    auto time_process_p3 = std::chrono::high_resolution_clock::now();
+
+    for (const traccc::cell_collection & module : modules) {
+        traccc::cluster_collection clusters_per_module = cc(module);
```
krasznaa, line 120: :thinking: I wonder if the code should explicitly do something with the output variable to make sure that the code is getting executed. As a very aggressive compiler optimisation may decide not to execute this line at all, since its output is not needed anywhere later on.

Doing something as silly as

```c++
std::size_t dummy = 0;
for (const traccc::cell_collection & module : modules) {
       dummy += cc(module).size();
}
...
std::cout << ( dummy % 2 ? ' ' : ' ' );
```

should force the compiler to not be able to "optimise out" this code. (Of course we would adjust this once/if the cluster EDM is updated...)

krasznaa (approved):
Let's try to move things along finally. Stephen, if you think my comment is worth doing something about, then please let us know. Otherwise I'm happy with this executable as-is.

examples/cpu/ccl_example.cpp
HadrienG2: This optimization barrier is too simple, the compiler can optimize it out into `std::cout << ' ';` and then proceed to remove the computation of `dummy`.

Here are the compiler optimization barriers that I typically use. These will work on all compilers that support the GNU dialect of inline assembler (i.e. anything but MSVC, which we probably don't care about here). You can do everything with the `assume_accessed_global` barrier, but the `_local` barriers may lead to nicer codegen when applied judiciously.

```cpp
  // Forces compiler to assume that any data directly or indirectly targeted by
  // the reference has been read and written. Easy to use, but may affect compiler
  // memory access optimizations for data behind unrelated pointers...
  template <typename T>
  void assume_accessed_global(const T& data) {
    asm("" : : "g"(&data) : "memory");
  }

  // Forces compiler to assume that a configurable number of elements behind the
  // target pointer have been read and written, will not account for transitive
  // pointers but will not affect compiler optimizations for other memory regions.
  template <typename T, size_t ACCESSED_ELEMS=1>
  void assume_accessed_local(const T* ptr) {
    const auto array_ptr = reinterpret_cast<T(*)[ACCESSED_ELEMS]>(const_cast<T*>(ptr));
    asm("" : "+m"(*array_ptr) : : );
  }

  // Forces compiler to assume that a configurable number of elements behind the
  // target pointer have been read, will not account for transitive pointers but
  // will not affect compiler optimizations for other memory regions.
  template <typename T, size_t ACCESSED_ELEMS=1>
  void assume_read_local(const T* ptr) {
    const auto array_ptr = reinterpret_cast<const T(*)[ACCESSED_ELEMS]>(ptr);
    asm("" : : "m"(*array_ptr) : );
  }

  // Forces compiler to assume that a configurable number of elements behind the
  // target pointer have been written, will not account for transitive pointers
  // but will not affect compiler optimizations for other memory regions.
  template <typename T, size_t ACCESSED_ELEMS=1>
  void assume_written_local(const T* ptr) {
    const auto array_ptr = reinterpret_cast<T(*)[ACCESSED_ELEMS]>(const_cast<T*>(ptr));
    asm("" : "=m"(*array_ptr) : : );
  }
  ```

Comment (stephenswat) [reply examples/cpu/ccl_example.cpp]: I'd say for now it's probably fine as it is. A compiler would have to have _very_ good static analysis capabilities to decide to optimize out this function call, and my gcc is leaving it as it. This program is also not really meant to be a benchmark, but rather an example of how the CCL code might be used. We can always look back at this later, if it turns out to be an issue. :slightly_smiling_face:

Comment (krasznaa) [reply examples/cpu/ccl_example.cpp]: I'm fine with that. :+1:

For the "final" benchmarks in the project we should come up with some clever code for this, but in this particular executable it's not a priority. Not just yet.

Hadrien, I would prefer if we could find some clever way of doing this without relying on platform specific tricks. I think that would be possible as well. (Maybe not in a completely generic way, but at least in a "return value specific" one.) But let's table that discussion until we really need it...

P.S. Let's see if the thumbs-up emoji has an effect on this PR's approval status... :thinking:

Comment (HadrienG2) [reply examples/cpu/ccl_example.cpp]: @krasznaa Note that these tricks are only compiler specific, not CPU specific (because the ASM block body is empty, only the input/output/clobbers are used as an optimization barrier). But yeah, portable optimization barriers are hard.

Without using compiler-specific extensions, the best trick I know is to escape a pointer to a separate compilation unit and write in big red bold letters in your documentation WHATEVER YOU DO, DO NOT ENABLE LTO, and then someday somebody needs LTO, and the overhead of that trick is much higher than that of inline assembly trick. So given that pretty much everyone but MSVC supports the GNU C inline assembly syntax well enough for my needs, I usually don't bother.

File: examples/cpu/ccl_example.cpp [hunk 2]
```diff
@@ -0,0 +1,140 @@
...
+#include <algorithm>
+#include <iomanip>
+#include <chrono>
+
+namespace {
+    double delta_ms(
+        std::chrono::high_resolution_clock::time_point s,
+        std::chrono::high_resolution_clock::time_point e
+    ) {
+        return std::chrono::duration_cast<std::chrono::microseconds>(e - s).count() / 1000.0;
```
krasznaa, line 25: This is only meant as an "educational corner" here, for this executable I don't care about this too much.

You should **never** perform a division with a constant value in your code. You should always rather perform a multiplication with the inverse of the constant. So in this case you would rather write:

```c++
return std::chrono::duration_cast<std::chrono::microseconds>(e - s).count() * 1E-3;
```

In more complicated cases you can leave it up to the compiler to perform the division at compile time by using `constexpr` expressions.

As I started, this is not really relevant here. But in other parts of this project we should all be very aware of how much slower divisions are than multiplications. :wink:

krasznaa (approved):
I vote for putting this in finally. In the "final" executables we should be careful with choosing between `at(...)` function and `operator[]` operator calls in performance critical parts, but here we can be a bit more lenient for the time being...

examples/cpu/ccl_example.cpp
krasznaa: I don't know. That's like calling the passing of objects by references instead of by value (in a place where passing by reference is appropriate) a premature optimisation. :thinking: In any numerical calculation not dividing by a constant should be as much of a basic code writing pattern as any other.

And this is why I wanted to bring it up on this example, as an educational comment.

stephenswat: I'm with you that, generally, in performance-sensitive code, it's a good idea to multiply by the reciprocal rather than divide, but I think a blanket statement that division is _always_ bad is actually quite harmful. There's a few reasons why one might want to divide rather than multiply. The most obvious, of course, is numerical accuracy. There is a very good reason why compilers will not (by default) make the transformation between `x / N` to `x * (1.0 / N)`: while it's valid in "proper" mathematics, the two are not equivalent in IEEE 754 arithmetic. Sure, the difference is usually not huge, and the result is biased either way, but the fact of the matter is that the two approaches will give you two different results. Whether the difference is acceptable is application-specific, and usually in scientific computing it's fine. And then there is the problem of different handling of subnormal numbers and other such IEEE 754 oddities.

Writing code using division is often a lot clearer than code written using multiplication with the reciprocal, because the programmer's intent is immediately clear. Reciprocal multiplication usually involves declaring additional variable, and it makes the code harder to read. I'll actually go out on a limb here and say that if you're going to be lax in your IEEE compliance anyway you might as well turn on `-funsafe-math-optimizations`, at least locally: that will leave your code in a more readable state and the compiler will probably be able to do a better job than most humans anyway. In this particular case, we're working with nice, round, powers of 10, so the two approaches are equivalent in readability. But in general, I find code written in the form `a / 27.0` much more readable than `a * 0.037037037`.

And most importantly, making this transformation is not even really guaranteed to speed up your code. If your kernel is memory-bound, for example, a transformation like really just makes your code harder to read, potentially less numerically accurate, for no discernible speed gain.

Just to be clear, once again, that you are absolutely correct that replacing division by multiplication is _often_ a good idea (although I would argue to let the compiler do it for you). I am just not a fan of making blanket statements like this. :slightly_smiling_face:

stephenswat:
We've had a lot of useful comments on this merge request, but I don't think there is any immediate action that we need to take right now. I would like to move forward with merging this in, if that is okay with everyone. :smile:

beomki-yeo:
I think this guys (@paulgessinger @HadrienG2 ) have permission to approve?

HadrienG2 (approved):

paulgessinger:
This seems to be good to go, so I'm merging this now.

PR #29, opened 2021-05-02
Edm with generic container

beomki-yeo:
1. The EDM for cell, measurement and spacepoint are defined based on the generic container (`core/include/edm/container.hpp`) which contains relavant header and items in the type of `vecmem::vector` and `vecmem::jagged_vector`, respectively. 

2. The calculation of local position error is also added in `measurement_creation` using the weighted average and pdf of local position.


beomki-yeo:
pinging @krasznaa @stephenswat

core/CMakeLists.txt
```diff
@@ -30,7 +25,14 @@ install(
...
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
-add_library(traccc::core ALIAS traccc_core)
\ No newline at end of file
+if (TRACCC_BUILD_cuda)
+  find_library(CUDART_LIBRARY cudart ${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES})
+  target_include_directories(traccc_core
+    PUBLIC
+    ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
+endif()
```
krasznaa, line 36: I'm confused about this part. Why do you want to link `libcudart` publicly against `traccc::core`? :confused: At a quick glance I don't see that you would've added any CUDA code to the headers in the core library...

I also don't understand why using `PUBLIC` for the include directory type actually works. :confused: Since `traccc_core` is an `INTERFACE` library, I would've bet money that you would only be allowed to declare include directories and libraries with the `INTERFACE` keyword in `target_include_directories(...)` and `target_link_libraries(...)` for `traccc_core`...

All in all, as long as you just use VecMem types in `traccc_core`, it should be enough to just link it against `vecmem::core`, and not worry about any direct CUDA/HIP/SYCL dependency on this library. (Those explicit dependencies should only be declared for the libraries holding the CUDA/HIP/SYCL specific algorithms.)

core/CMakeLists.txt
```diff
@@ -30,7 +25,14 @@ install(
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
 install(
-  DIRECTORY include/utils
+  DIRECTORY include/algorithms
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
```
krasznaa, line 29: I should've done this myself, but the code is unnecessarily complicated here. There should really only be a single `install(...)` command for the headers. This one:

```cmake
install( DIRECTORY include/
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR} )
```

I.e. the `/` at the end of `include/` tells CMake to consider the contents of the directory, and not to copy the `include` directory itself as well.

This way we won't have to keep adjusting this technical code whenever a new header sub-directory is added.

krasznaa (commented):
I find your code proposal **very** intriguing. I didn't consider that we could get away with using a single high-level container type for this project. :thinking:

I will look at your code more carefully today. Since I still wonder if we're not giving up too much for this generality.

But I'm very intrigued. If we can get away with this way of defining the EDM for the project, that would be **very** cool. :wink:

stephenswat:
So, the idea here is to move from hand-written product types of headers and rows to a more general approach where the types are generated automatically using templates, such that the user specifies the header type and the row type (and optionally a container type), and out comes a type for a specific header-row EDM? That seems sensible enough to me. You note a second change as well, is that related to the EDM change or has that sort of snuck in with the main changes in this MR?

krasznaa:
Ahh, now I get it. This PR doesn't touch the "cluster EDM". This makes sense. Since that's the really difficult part here. That's where we need to deal with a vector-of-vector-of-vector type.

Still... If we can make use of a single declaration of all of the other types, that's already a big win. So I'm in favour of implementing the code like this. :+1: However we'll still need to find a good way of dealing with the clusters...

krasznaa:
Wait... what happened? How come the PR became "approved" all of a sudden? :confused: I'm definitely very much in favour of these changes in general, but I didn't hit any approve button yet myself. And I don't see anybody else having done that either. :confused:

paulgessinger:
Hm. Policybot says you 'approved' but I think you only added a review comment, which shouldn't trigger this. Maybe it's in fact the thumbs up? We never had this problem before, but maybe that is a default 'feature'. I'll try to find out.

krasznaa:
The reason I suspect the emoji is because the PR was only "approved" after my second comment. After posting my review it stayed as it should've.

But as you see in #24, leaving a thumbs-up in a discussion about a piece of code made no difference. So maybe it only has an effect in "top level comments"? :confused: If so, this would be a very fragile setup...

paulgessinger:
Not sure if a comment event will unapprove now, let's see.

core/CMakeLists.txt
beomki-yeo: At least in the PR, we don't need to link libcudart to traccc::core. I just didn't push some codes that are defined with the cuda type qualifiers (`__host__ __device` ), which needs cuda library. For example, in my latest code, the functions `definitions/algebras.hpp` has the cuda type qualifiers to be used inside kernel. It will be useful in case we want to reuse cpu tracking helper function in gpu kernel.

About the INTERFACE keyword, I will make a proper change soon

krasznaa: I see. But for just using the `__host__` and `__device__` keywords you should not make the CUDA runtime library's headers visible to the compilation. The CUDA headers have nothing to do with what `__host__` and `__device__` do.

In VecMem we also put these declarations on (template) functions in `vecmem::core`. Without `vecmem::core` including any CUDA headers. That works fine, since when the headers are parsed by `nvcc`, it makes use of these keywords. While the C\+\+ compiler just doesn't see them.

Long story short, you should just use `__device__` and `__host__` without worrying about the build system. Only the code that calls `cuda<Foo>(...)` functions needs to interact with the CUDA header(s).

beomki-yeo:
> You note a second change as well, is that related to the EDM change or has that sort of snuck in with the main changes in this MR?

It's not related to EDM change. The error calculation was missing in traccc so I've just added few lines in `measurement_creation`

core/CMakeLists.txt
beomki-yeo: Um. seems you are right after trying to compile without linking libraries. Gonna fix it

core/CMakeLists.txt
beomki-yeo: Good point. Will fix it

beomki-yeo:
I certainly don't have clever answer about how to deal with cluster-like objects. For the moment, I am bypassing it by defining labeling container which is not a fundamental solution though. I am afraid similar problem may occur in other tracking algorithm such as CKF, but we have see how it goes

core/include/edm/container.hpp
```diff
@@ -0,0 +1,129 @@
...
+        vecmem::data::vector_view< header_t > headers;
+
+	/// View of the data describing the items	
+        vecmem::data::jagged_vector_view< item_t > items;
+
+    };
+
+    /// Helper function for making a "simple" object out of the container
+    template< typename header_t, typename item_t >
+    __inline__
```
krasznaa, line 123: If you insist on inlining this function (which would happen for a template function by default anyway), please use the standard `inline` keyword instead of the non-standard `__inline__`.

examples/cpu/seq_example.cpp
```diff
@@ -59,45 +59,57 @@ int seq_run(const std::string& detector_file, const std::string& cells_dir, unsi
...
+	    
+            spacepoints_per_event.items.push_back(std::move(spacepoints_per_module));
+	    spacepoints_per_event.headers.push_back(module.module);
         }
 
         traccc::measurement_writer mwriter{std::string("event")+event_number+"-measurements.csv"};
-        for (const auto& measurements_per_module : measurements_per_event){
-            auto module = measurements_per_module.module;
-            for (const auto& measurement : measurements_per_module.items){
+	for (int i=0; i<measurements_per_event.items.size(); ++i){
```
krasznaa, line 98: You should change `i` into an `std::size_t` variable. As the build currently emits the following warning:

```
/data/ssd-1tb/projects/traccc/traccc/examples/cpu/seq_example.cpp:98:17: warning: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::vector<traccc::measurement, std::pmr::polymorphic_allocator<traccc::measurement> >, std::pmr::polymorphic_allocator<std::vector<traccc::measurement, std::pmr::polymorphic_allocator<traccc::measurement> > > >::size_type’ {aka ‘long unsigned int’} [-Wsign-compare]
   98 |  for (int i=0; i<measurements_per_event.items.size(); ++i){
      |                ~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

(The "most exact" thing would be to make use of the `size_type` definition of the type that you're using. But let's not get crazy here. `std::size_t` will work fine.)

examples/cpu/seq_example.cpp
```diff
@@ -59,45 +59,57 @@ int seq_run(const std::string& detector_file, const std::string& cells_dir, unsi
...
-                mwriter.append({ module, local[0], local[1], 0., 0.});
+                mwriter.append({ module.module, local[0], local[1], 0., 0.});
             }
         }
 
         traccc::spacepoint_writer spwriter{std::string("event")+event_number+"-spacepoints.csv"};
-        for (const auto& spacepoint_per_module : spacepoints_per_event){
-            auto module = spacepoint_per_module.module;
-            for (const auto& spacepoint : spacepoint_per_module.items){
+	for (int i=0; i<spacepoints_per_event.items.size(); ++i){
```
krasznaa, line 108: Same here, please change `int` to `std::size_t`.

krasznaa (changes requested):
Other than the minor technical comments (which should still be done...), I think we should go ahead with this one. I checked myself as well that the update does not change the results from the existing tests, and if anything, it may make them marginally faster. (My results are not statistically significant though.)

So let's not leave this one in limbo for too long. :wink:

examples/cpu/seq_example.cpp
beomki-yeo: good comment. changed to `size_t` for now

krasznaa (approved):
From my side this is ready to be included in the main branch.

paulgessinger:
Then let's do it.

PR #30, opened 2021-05-03
Configure policybot to not interpret comments as approval

paulgessinger:
@krasznaa I think this should remove the behavior we saw in #24.


krasznaa (approved):
Let's give it a try. I guess it shouldn't hurt us...

PR #33, opened 2021-05-06
Implement container element views

stephenswat:
The container classes were recently refactored into a more generalized class, which allows us to write generalized element views for them. Currently, we are accessing them directly through the inner vectors, but this is dangerous and makes it extremely difficult to enforce invariants that we would like to have for containers. This change introduces an accessor view for header-vector elements in the containers, which allows us to more safely and more ergonomically access the inner elements. Effectively, the semantics are very similar to those of array-of-structs, even though the underlying layout is a struct-of-arrays. We also introduce several helper methods to make working with containers a lot simpler.


core/include/edm/container.hpp
```diff
@@ -51,9 +84,163 @@ namespace traccc {
...
+        container(
+            header_vector && hv,
+            item_vector && iv
+        )
+            : headers(hv)
+            , items(iv)
+        {
+            if (headers.size() != items.size()) {
+                throw std::logic_error("Header and item length not equal.");
+            }
```
krasznaa, line 132: I'm not sure how such a constructor will fit into making this type usable in both device and host code at the same time...

If a constructor like this does work in device code, we would still need to use an

```c++
assert( headers.size() == items.size() );
```

statement in it, as exceptions are definitely a no go.

core/include/edm/container.hpp
```diff
@@ -51,9 +84,163 @@ namespace traccc {
...
+        /**
+         * @brief Bounds-checking mutable element accessor.
+         */
+        element_view
+        at(
+            const size_type & i
+        ) {
+            if (i >= size()) {
+                throw std::out_of_range("Index out of range.");
+            }
```
krasznaa, line 160: No exceptions in device code...

core/include/edm/container.hpp
```diff
@@ -51,9 +84,163 @@ namespace traccc {
...
+        /**
+         * @brief Bounds-checking immutable element accessor.
+         */
+        const element_view
+        at(
+            const size_type & i
+        ) const {
+            if (i >= size()) {
+                throw std::out_of_range("Index out of range.");
+            }
```
krasznaa, line 174: No exceptions in device code...

core/include/edm/container.hpp
```diff
@@ -51,9 +84,163 @@ namespace traccc {
...
+            : headers(a...)
+            , items(a...)
+        {}
+
+        /**
+         * @brief Bounds-checking mutable element accessor.
+         */
+        element_view
+        at(
+            const size_type & i
```
krasznaa, line 156: 😕  Does the standard library use size type variables like this? I don't think we should be using constant references for these arguments. The compiler should have a much easier time with size variables passed simply by value.

core/include/edm/container.hpp
```diff
@@ -51,9 +84,163 @@ namespace traccc {
...
+        operator[](
+            const size_type & i
+        ) {
+            return {headers.at(i), items.at(i)};
+        }
+
+        /**
+         * @brief Immutable element accessor.
+         */
+        const element_view
```
krasznaa, line 192: This is not how this should be implemented.

You should rather define a separate `const_element_view` type, like:

```c++
using const_element_view = container_element< const header_t, const item_t >;
```

This would ensure that the caller does not get non-const access to the variables. Since the const-ness of the returned value is lost as soon as the user does something like:

```c++
my_container::element_view el = foo[ i ];
```

core/include/edm/container.hpp
```diff
@@ -51,9 +84,163 @@ namespace traccc {
...
+         */
+        size_type
+        size(
+            void
+        ) const {
+            #ifndef NDEBUG
+            if (headers.size() != items.size()) {
+                throw std::logic_error("Header and item length not equal.");
+            }
+            #endif
```
krasznaa, line 213: No exceptions in device code...

core/include/edm/container.hpp
```diff
@@ -51,9 +84,163 @@ namespace traccc {
...
+        }
+
+        /**
+         * @brief Push a header and a vector into the container.
+         */
+        template<typename h_prime, typename v_prime>
+        void
+        push_back(
+            h_prime && new_header,
+            v_prime && new_items
```
krasznaa, line 236: There should be a "copy version" of this function as well, if we want to have this functionality.

krasznaa (changes requested):
I do think that we need to make the container type smarter than it is now. For the cluster container this will be inevitable, since that one will have a less-than-trivial internal.

To make this properly work in device code, we'll have to add `__device__` and `__host__` declarations on all of these functions. For my renewed look at the cluster container, I've introduced `TRACCC_HOST`, etc. macros for doing this.

Adding all of those declarations can be done in a follow-up PR as well. Here I just want to make sure that the "core functionality" is one that will translate well to device code as well.

core/include/edm/container.hpp
```diff
@@ -16,6 +16,39 @@
...
+     * In order to enforce certain invariants on the @c container header-vector
+     * type, we access elements (which, of course, are product types of a header
+     * and a vector) through this wrapper class. This class provides
+     * low-overhead access to the struct-of-arrays container class to emulate an
+     * array-of-structs architecture.
+     *
+     * @tparam header_t The type of the header object.
+     * @tparam vector_t The fully qualified vector type.
+     */
+    template<typename header_t, typename vector_t>
```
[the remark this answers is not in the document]
stephenswat: But `vector_t` here would be an actual vector type, not the elements inside that vector. These container elements are product types of a header and a _vector_ of items, not a header and a single item.

core/include/edm/container.hpp
```diff
@@ -51,9 +84,163 @@ namespace traccc {
...
+         * jagged vector have the same size type, so they can be indexed using
+         * the same type.
+         */
+        static_assert(
+            std::is_same_v<
+                typename header_vector::size_type,
+                typename item_vector::size_type
+            >,
+            "Size type for container header and item vectors must be the same."
+        );
```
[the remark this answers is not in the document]
stephenswat: It is not strictly necessary, but good practice to ensure that these types are the same. If course if one is `std::size_t` and the other is an unsigned integer, it's unlikely to be a problem, but the two types should at least be convertible.

core/include/edm/container.hpp
stephenswat: Hmm, maybe I have misunderstood Beomki's intention here, but these containers are not designed to be used in device code at all, are they? :confused:

core/include/edm/container.hpp
stephenswat: I disagree here, implicitly copying data here is a performance anti-pattern. If the user does want to copy, they can do so explicitly without our code having to care about it.

core/include/edm/container.hpp
stephenswat: In which way are `const container_element<A, B>` and `container_element<const A, const B>` not equivalent here?

core/include/edm/container.hpp
krasznaa: Have a look at:

https://github.com/acts-project/traccc/blob/main/core/include/edm/container.hpp#L68-L71

The idea, which I had with my code for the cell container, and which Beomki took for this generalisation as well, is that we just specialise this template class with either the host or device vector types of VecMem, and that's how we get the types used on the host and on the device. So all code present in this class would either have to be usable on a device, or be explicitly marked as only being usable on the host.

The latter is not necessarily a bad thing to do. For instance we can definitely have some constructors be host-only. That's fine. But the "regular" functions should be usable in device code as well.

core/include/edm/container.hpp
stephenswat: I see your point, but I would propose that we instead add mutable and immutable accessor methods to the `element_view` class. That is, in my opinion, a far more elegant solution than to add these typedefs.

krasznaa: 😕  This is exactly what we talked about with @scott-snyder in a Core SW meeting about a month ago. When you create such view objects, it's very easy to expose non-const references to vector elements from inside of const accessor functions.

Have another look at the example I gave above. That would expose non-const references from a possibly const container.

core/include/edm/container.hpp
stephenswat: The standard library uses both const references and pass-by-value depending on the container. Since const references are more universal than pass by value, I figure that this would be a more general solution.

core/include/edm/container.hpp
stephenswat: I assume `foo` is a container here? If that container is const, the only `operator[]` available for it will be the const one, returning a const `element_view`. Then, this `element_view` would only have const accessor methods available, which in turn can only return const references to the contained objects. I fail to see how you would get a non-const reference in this way, can you explain?

core/include/edm/container.hpp
krasznaa: Let's just follow what `std::vector` does. And that just passes size variables around by value.

https://en.cppreference.com/w/cpp/container/vector/at

We don't want to use anything other than integer type indices on these containers.

core/include/edm/container.hpp
krasznaa: You don't return a const **reference** from this function. You return a const **value**. As such, my example line shows how the const-ness of that value goes away just by receiving it into a non-const value.

(In general returning a const value of anything from a function has not much of a meaning in C(\+\+)...)

core/include/edm/container.hpp
stephenswat: Ah, I guess I misunderstood the design here then... Although this really just seems like important a lot of the annoying constraints of the device programming world into the host world. But sure, we can mark the host-only methods as such. In that case, I would say the bounds-checked accessor methods should be host only, as should of course the `reserve` and `push_back` methods (unless we want to use the device growing methods for this). Alternatively it might be worth defining a `device_container` class instead, similar to the design of vecmem.

core/include/edm/container.hpp
stephenswat: Sure! If only we had C++20 and we could enforce a concept equal to the SequenceContainer named requirement... :cry:

core/include/edm/container.hpp
stephenswat: Of course. :man_facepalming:

core/include/edm/container.hpp
beomki-yeo: It's going to definitely useful if we can use them on the device side. I am afraid defining host-only and device-only methods would increase code complexity. Most of all I don't see the use cases for this constructor.

paulgessinger:
Discussions are still open, and have been for a while. How do we go ahead with this one @stephenswat @krasznaa ?

beomki-yeo:
@paulgessinger  Can you merge it unless you have objection? 
I don't want this kind of PR for low-level structure to hang so long time.

beomki-yeo:
Oh wait. Looks like `__device__ __host__` has not been added yet as @krasznaa mentioned. What is your plan? @stephenswat

stephenswat:
Rebased, and qualifiers added. :smile:

beomki-yeo:
I think we can merge this? @paulgessinger

paulgessinger:
Right now there's a conflict that needs to be resolved. Otherwise I suppose this is good to go.

stephenswat:
The conflict has been resolved. I'll wait for an explicit approval, and then we can merge this. :smiley:

PR #34, opened 2021-05-07
Swap from std::array to a GPU friendly array type

stephenswat:
We use `std::array` in a lot of places in the EDM, and obviously this type is not GPU-friendly; it cannot be used in CUDA kernels, for example. To solve this problem, I have implemented a new, basic, GPU-friendly array type to replace it. This merge request does a few things:

1. It fixes a bug where the two-dimensional norm was silently returning a three-dimensional vector.
2. It changes most uses of `std::array` to the typedefs, like `traccc::vector3`.
3. It introduces a new `traccc::array` type which is a little simpler than `std::vector`, a little stricter, but can be used on the GPU.
4. It changes the typedefs like `traccc::vector3` to use the new `traccc::array` type.

### DIFF
--- a/core/include/definitions/algebra.hpp
+++ b/core/include/definitions/algebra.hpp
@@ -445,7 +445,7 @@ namespace traccc
          * 
          * @param v the input vector
          **/
-        std::array<scalar, 3> normalize(const std::array<scalar, 2> &v)
+        std::array<scalar, 2> normalize(const std::array<scalar, 2> &v)
         {
             scalar oon = 1. / std::sqrt(dot(v, v));
             return {v[0] * oon, v[1] * oon};
--- a/core/include/definitions/algebra.hpp
+++ b/core/include/definitions/algebra.hpp
@@ -17,42 +17,42 @@
 namespace traccc
 {
 
-    std::array<scalar, 2> operator*(const std::array<scalar, 2> &a, scalar s)
+    vector2 operator*(const vector2 &a, scalar s)
     {
         return {a[0] * s, a[1] * s};
     }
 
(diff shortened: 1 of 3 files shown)


krasznaa (commented):
I'm desperately trying to find where we discussed about this with @beomki-yeo. I think it was one of the PRs, but I can just not find it now.

I didn't want to introduce such a custom array type in VecMem because:
  - SYCL can already deal with `std::array` in device code;
  - CUDA is able to deal with `std::array` in device code starting with version 11 (as long as one uses a specific compile time flag).

So creating a `vecmem::static_array` type (remember, we already have a `vecmem::array`, but that behaves differently...) seemed like wasted effort. And it still kind of seems that way... 😦

stephenswat:
You mean the `--expt-relaxed-constexpr` flag right? My understanding is that that isn't full support, just a more general assumption that `constexpr` methods are also `__device__`. But if we're green to start using that flag, and CUDA 11, then that's fine too! As long as I can make measurement objects on the device side I will be happy. :smile:

krasznaa:
Hmm... I've been flip-flopping on this a couple of times.

"Simple" SYCL code can make use of `std::array` variables fine on both Intel and NVidia hardware. But once I try to do something a bit more complicated, the code starts misbehaving badly. (On both hardware actually.)

So in the end, let's go ahead with the "`vecmem::static_array` project". We'll hear this afternoon anyway how Llama did the same...

stephenswat:
Okay! I'll drop the code that is here into vecmem as is, close this merge request, and take the bugfix commits to a new merge request.

PR #35, opened 2021-05-11
Port ACTS binning algorithm for CPU

beomki-yeo:
@paulgessinger Can you merge this after merging Acts [PR872](https://github.com/acts-project/acts/pull/872)

-------------------------------------------------------------------------------------------------------------------------------------
I ported ACTS binning algorithm to traccc.
Most of codes are same from ACTS one,  but there is a few differences, especially for `spacepoint_grid` which is a container for binned spacepoints.

In this implementation:
```c++
using spacepoint_grid = grid< internal_spacepoint< spacepoint > ,
                             traccc::axis<AxisBoundaryType::Closed>, traccc::axis<AxisBoundaryType::Bound> >;
```           

```c++
template < typename T, class... Axes >
class grid{
 ...
 private:
    std::tuple< Axes... > m_axes;
    vecmem::jagged_vector<T> m_values; // T = internal_spacepoint< spacepoint > 
};
```
                  
In ACTS:
```c++
template <typename external_spacepoint_t>
using SpacePointGrid =
    detail::Grid<std::vector<std::unique_ptr<
                     const InternalSpacePoint<external_spacepoint_t>>>,
                 detail::Axis<detail::AxisType::Equidistant,
                              detail::AxisBoundaryType::Closed>,
                 detail::Axis<detail::AxisType::Equidistant,
                              detail::AxisBoundaryType::Bound>>;
```

```c++
template <typename T, class... Axes>
class Grid final {
   ...
 private:
  /// set of axis defining the multi-dimensional grid                                 
  std::tuple<Axes...> m_axes;
  /// linear value store for each bin                                                 
  std::vector<T> m_values; // T = std::vector<std::unique_ptr<const InternalSpacePoint<external_spacepoint_t>>>
  };
```
Template parameter of `external_spacepoint_t`  is removed as we don't need it for demonstrator.
The value type for `grid` is also changed from `std::vector< std::unique_ptr< const InternalSpacePoint< external_spacepoint_t> > >` to `internal_spacepoint< spacepoint>`. 
The reason for not using vector type is to utilize `vecmem::jagged_vector< internal_spacepoint<spacepoint > >` in `grid.hpp`. Regarding `unique_ptr`, I removed it because I was not sure if it works for device code. Please tell me if it is okay to use smart pointer in the device code (@krasznaa  @stephenswat)

For the upcoming seed finding in GPU, I expect `binned_spgroup` object can be passed to the kernel directly so we can iterate over bottom/middle/top spacepoints within device code: We will have to add `__device__ __host__` qualifiers for every iterator class but I think it is worth doing it if we can put everything on the plate.

Actually binning algorithm itself can also be ported on GPU (the constructor of `binned_spgroup.hpp` seems highly parallelizable). Leave a comment if you think it is better to do that.

### Update (May 13)

1. The EDM for `internal_spacepoint<spacepoint>` is introduced. Instead, `grid` no longer contains the `internal_spacepoint`:

- header: global bin index, neighborhood bin indices
- item: internal spacepoints

One issue is that the number of neighborhood indides is supposed to be variant depending on the number of neighborhoods per axis, and the number of axes. Currently this is not handled and fixed the max size of index vector to 9 (3x3) which is for two axes and one neighbor bins.

2. All algorithms for binning is written in  `spacepoint_grouping.hpp`:

- neighborhood bin finding
- spacepoint binning


beomki-yeo:
Um. It's obvious that the binned_sp_group can not be passed to the kernel. 
I will try to make `internal_spacepoint` edm usable within the kernel after some refactoring

krasznaa:
I think we're getting to the meat of this project with this one...

My hope is that we can make use of the existing seed finding code **not** by copy-pasting it into traccc. But rather by:
  - Implementing the "experiment specific EDM" used by the seed finding in the traccc repo;
  - Updating the seedfinding code in Acts to use VecMem itself;
  - Pull in the Acts code into traccc similar to how we pull in VecMem already.

This will be difficult, no doubt about it. But in my mind this was always meant to be the really "interesting" part of how we make traccc work.

(This way we could make use of both the existing CUDA and SYCL code in Acts without having to copy-paste everything into here.)

core/include/algorithms/seeding/seeding_config.hpp
```diff
@@ -0,0 +1,122 @@
...
+#include "utils/axis.hpp"
+#include "utils/grid.hpp"
+
+namespace traccc{
+
+// define spacepoint_grid
+using spacepoint_grid = grid<traccc::axis<AxisBoundaryType::Closed>,
+			     traccc::axis<AxisBoundaryType::Bound> >;
+    
+struct seedfinder_config{
```
paulgessinger, line 19: This struct exists twice?

beomki-yeo: That's a silly mistake. I removed them

beomki-yeo (commented):
>     * Pull in the Acts code into traccc similar to how we pull in VecMem already.

Did you mean we put acts libraries into traccc? Well. I doubt because that will put more constraints in development. Some parts of ACTS codes need to be modified based on our needs, and I believe that's one of the reasons that we are developing the demonstrator.

core/include/edm/internal_spacepoint.hpp
```diff
@@ -0,0 +1,146 @@
...
+    /// Convenience declaration for the internal_spacepoint container view type to use in host code
+    using internal_spacepoint_container_view
+    = container_view< bin_information, internal_spacepoint<spacepoint> >;
+
+    size_t find_vector_id_from_global_id(size_t global_bin, vecmem::vector<bin_information>& headers){
+	auto location
+	    = std::find_if(headers.begin(), headers.end(),
+			   [&global_bin](const bin_information& bin_info)
+			   {return bin_info.global_index==global_bin;})
+	    - headers.begin();
```
paulgessinger, line 121: `std::distance`?

paulgessinger:
So, do we put this in with the code from ACTS core for now, and then try to rearchitect as suggested by @krasznaa?

beomki-yeo (commented):
> So, do we put this in with the code from ACTS core for now, and then try to rearchitect as suggested by @krasznaa?

What @krasznaa  suggested was like: modifying ACTS -> putting the changes into tracc. But I support the opposite way :) ...

asalzburger:
> I think we're getting to the meat of this project with this one...
> 
> My hope is that we can make use of the existing seed finding code **not** by copy-pasting it into traccc. But rather by:
> 
> * Implementing the "experiment specific EDM" used by the seed finding in the traccc repo;
> * Updating the seedfinding code in Acts to use VecMem itself;
> * Pull in the Acts code into traccc similar to how we pull in VecMem already.
> 
> This will be difficult, no doubt about it. But in my mind this was always meant to be the really "interesting" part of how we make traccc work.
> 
> (This way we could make use of both the existing CUDA and SYCL code in Acts without having to copy-paste everything into here.)

Yes, indeed - we should have a discussion how we do that: probably we can try to split some core functionality off into repositories (like grids, etc) that then can be used across projects?

beomki-yeo:
CI definitely needs to have Acts...
```
 /usr/bin/ld: cannot find -lActsCore
collect2: error: ld returned 1 exit status
```

paulgessinger:
So this is ready to go now?

paulgessinger (approved):
Looks good to me at least. The Grid copy is gone. Merging this.

PR #40, opened 2021-05-14
CPU parallel example

georgi-mania:
Create CPU parallel example based on the sequential example extended with OpenMP pragmas. 
Time-wise performance gain of ~5x is obtained (in comparison to the seq_example).

### DIFF
--- a/examples/cpu/CMakeLists.txt
+++ b/examples/cpu/CMakeLists.txt
@@ -6,3 +6,10 @@ target_link_libraries (seq_example LINK_PUBLIC traccc::core traccc::io vecmem::c
 
 add_executable (ccl_example ccl_example.cpp)
 target_link_libraries (ccl_example LINK_PUBLIC traccc::core traccc::io)
+
+# find OpenMP
+find_package(OpenMP REQUIRED)
+set(GCC_OMP "-fopenmp")
+set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} ${GCC_OMP}")
+add_executable (par_example par_example.cpp)
+target_link_libraries (par_example LINK_PUBLIC traccc::core traccc::io vecmem::core OpenMP::OpenMP_CXX)
--- a/examples/cpu/CMakeLists.txt
+++ b/examples/cpu/CMakeLists.txt
@@ -6,10 +6,3 @@ target_link_libraries (seq_example LINK_PUBLIC traccc::core traccc::io vecmem::c
 
 add_executable (ccl_example ccl_example.cpp)
 target_link_libraries (ccl_example LINK_PUBLIC traccc::core traccc::io)
-
-# find OpenMP
-find_package(OpenMP REQUIRED)
-set(GCC_OMP "-fopenmp")
-set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} ${GCC_OMP}")
-add_executable (par_example par_example.cpp)
-target_link_libraries (par_example LINK_PUBLIC traccc::core traccc::io vecmem::core OpenMP::OpenMP_CXX)
--- a/examples/cpu/par_example.cpp
+++ b/examples/cpu/par_example.cpp
@@ -1,0 +1,158 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include "edm/cell.hpp"
+#include "edm/cluster.hpp"
+#include "edm/measurement.hpp"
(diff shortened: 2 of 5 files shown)


examples/cpu/par_example.cpp
```diff
@@ -0,0 +1,158 @@
...
+            traccc::host_measurement_collection measurements_per_module = mt(clusters_per_module, module);
+            traccc::host_spacepoint_collection spacepoints_per_module = sp(module, measurements_per_module);
+            // The algorithmnic code part: end
+
+            n_cells += cells_per_event.items[i].size();
+            n_clusters += clusters_per_module.items.size();
+            n_measurements += measurements_per_module.size();
+            n_space_points += spacepoints_per_module.size();
+
+#pragma omp critical
```
stephenswat, line 95: Personally I find it a lot clearer to indent OpenMP pragmas on the same level as the construct they're modifying, but that's up to you.

examples/cpu/par_example.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+        // Output containers
+        traccc::host_measurement_container measurements_per_event;
+        traccc::host_spacepoint_container spacepoints_per_event;
+        measurements_per_event.headers.reserve(cells_per_event.headers.size());
+        measurements_per_event.items.reserve(cells_per_event.headers.size());
+        spacepoints_per_event.headers.reserve(cells_per_event.headers.size());
+        spacepoints_per_event.items.reserve(cells_per_event.headers.size());
+
+#pragma omp parallel for
```
stephenswat, line 77: Is your intention here to spawn a nested thread team? I believe you would probably be better off leaving this pragma out and relying only on the outer parallel pragma. I think that would give you better performance.

examples/cpu/par_example.cpp
georgi-mania: That is also my preference but I chose to format the code (with clang-format using the "llvm" style) before pushing it. Happy to use any style you recommend me so that it is inline with the existing code.

examples/cpu/par_example.cpp
georgi-mania: Yes, it was intentionally. On my machine, I saw no performance difference between these 2 choices. But I can double check with much larger number of events and threads and see how that goes.

paulgessinger:
@georgi-mania can you rebase? I added a formatting CI job that causes conflicts here.

examples/cpu/par_example.cpp
georgi-mania: Tested with 1000 parallel events, with and without the nested team, and there was no impact on performance.

georgi-mania:
Rebased.

examples/cpu/CMakeLists.txt
```diff
@@ -6,3 +6,10 @@ target_link_libraries (seq_example LINK_PUBLIC traccc::core traccc::io vecmem::c
 
 add_executable (ccl_example ccl_example.cpp)
 target_link_libraries (ccl_example LINK_PUBLIC traccc::core traccc::io)
+
+# find OpenMP
+find_package(OpenMP REQUIRED)
+set(GCC_OMP "-fopenmp")
+set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} ${GCC_OMP}")
+add_executable (par_example par_example.cpp)
+target_link_libraries (par_example LINK_PUBLIC traccc::core traccc::io vecmem::core OpenMP::OpenMP_CXX)
```
stephenswat, line 15: Perhaps we can make OpenMP optional? That is to say, check if it's there, and compile the OpenMP examples only then.

stephenswat (changes requested):
I think we should move along with this merge request because it has been waiting for a while. However, I would like to suggest a few more (hopefully relatively minor) changes before we put this in:

1. I think we should move all of this code out of the `cpu` subdirectories and into new `openmp` subdirectories. While the sequential code and the parallel code obviously run on the same platform, they are two different platforms. I would suggest we move `examples/cpu/par_example.cpp` to `examples/openmp/par_example.cpp`, and we move the changes to `examples/cpu/CMakeLists.txt` to a new `examples/openmp/CMakeLists.txt`.
2. We should make it so that the project can still be compiled when OpenMP is not installed. I think the best way to do this would be to modify `examples/CMakeLists.txt` to first check for OpenMP, and then to `add_subdirectory(openmp)` only if `OpenMP_CXX_FOUND`.
3. Depending on whether this or #33 goes in first, we'll need to rebase again. :cry:

examples/cpu/CMakeLists.txt
georgi-mania: Rebased & implemented the required changes.

stephenswat (approved):
Looks great! :smile:

PR #44, opened 2021-06-16
Introduce a pre-commit hook for clang-format

stephenswat:
As it stands, there are a few ways to ensure your code conforms to `clang-format`. You could run the tool manually, you could run the bash script in the `.github` directory, you could use a CICD simulator like `act`, or you could just push and see what Github actions things of the code. In my opinion, none of these are really ideal, and I think one of the most powerful tools for enforcing formatting standards like this is the pre-commit hook.

This merge request adds a pre-commit hook that will check the user's index for bad formatting, and it will refuse the commit if anything is wrong.

The hook is designed to look only at the index, which has a few advantages. Firstly, it won't reject commits for existing bad formatting in other files. Secondly, it allows users to partially stage files without having to worry about messing up the formatting.

Note that this hook is not automatically installed, any developer that wants to use them has to enable them manually (good, otherwise RCE exploits would be very easy). Instructions on how to do that are included in the `.githooks/README.md` file.

### DIFF
--- a/.githooks/README.md
+++ b/.githooks/README.md
@@ -1,0 +1,21 @@
+# traccc git hooks
+
+This directory stores some useful hooks for traccc developers. To install one of
+these hooks, run the following command (from the repository root):
+
+    ln -s $(pwd)/.githooks/pre-commit .git/hooks
+
+If you are using a modern version of coreutils and you need relative paths, for
+example if your mount points may change, use:
+
+    ln -rs .githooks/pre-commit .git/hooks
+
+To install all hooks in this repository, you can also run:
+
+    git config --local core.hooksPath .githooks
+
+## Requirements
+
+These hooks require clang-format to be installed, and findable. Either
+`clang-format` needs to be a binary findable by bash, or the
+`CLANG_FORMAT_BINARY` environment variable must be set to a clang-format binary.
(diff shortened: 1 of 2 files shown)


paulgessinger:
Ok, I don't use personally `pre-commit` hooks because they're always buggy and kill my workflow :) 

I can run `.githooks/pre-commit` instead of the sensibly names `check_format.sh`, that's ok for me. I do need it to always print the version of clang-format though, because you get incorrect formatting if it's not the correct clang-format binary. (That's also a  reason for not running it as a pre-commit hook, but whatever suits you ;) )

stephenswat:
Oh no don't get me wrong this is not designed to replace `check_format.sh` at all, but they do different things. `check_format.sh` will actually do the formatting for you, which is great, and this is just a check to see if all the formatting is okay. Very different use cases as far as I see it.

paulgessinger:
Sure! 

For real though, just have your editor format on save, and enjoy world peace. 😆

stephenswat:
I agree with you that synchronizing some things between this hook and `check_format.sh` would be a good idea...

stephenswat:
There is also an additional problem with `check_format.sh` which is that it takes non-staged changes into account. Not a problem if you format automatically on save, I will admit, but makes it rather hard to integrate into a workflow otherwise. :frowning:

paulgessinger:
@stephenswat do you want to update this as well?

paulgessinger (approved):

Issue #47, opened 2021-07-07
error from vecmem when building tip of main


cgleggett:
When trying to build tip of main (dfcd7e7e91e1e8e4fb12d27824f9ad489c8f140b), I'm getting this error:

```
In file included from /home/leggett/work/traccc/src/core/include/definitions/algebra.hpp:14,
                 from /home/leggett/work/traccc/src/core/include/edm/cell.hpp:11,
                 from /home/leggett/work/traccc/src/core/include/algorithms/detail/sparse_ccl.hpp:10,
                 from /home/leggett/work/traccc/src/core/include/algorithms/component_connection.hpp:10,
                 from /home/leggett/work/traccc/src/examples/cpu/seq_single_module.cpp:10:
/home/leggett/work/traccc/src/core/include/definitions/primitives.hpp:13:10: fatal error: vecmem/containers/static_array.hpp: No such file or directory
 #include <vecmem/containers/static_array.hpp>
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

the version of vecmem that it's pulling in is 2e06a5167e0927e6e6ed6a4f1abd5b5f552f059f

cgleggett:
ahh - `cmake/traccc-vecmem.cmake` calls for GIT_TAG `2e06a5167e0927e6e6ed6a4f1abd5b5f552f059f`. 

This should be updated. It builds fine if you use the tip of vecmem (currently `1ccfd3a65e90b67339792ebd95aa87072bd41b64`) but I don't know if this is desirable.

paulgessinger:
Maybe it would make sense to use a submodule for vecmem as well for now?

krasznaa:
A submodule would be one way of doing that. Though I don't think that would make it either simpler or more complicated to change the version of VecMem than the system that we have now...

stephenswat:
This should be solved with the merging of #48.

PR #48, opened 2021-07-08
VecMem Version Update, main branch (2021.07.08.)

krasznaa:
Updated the project to build [VecMem v0.2.0](https://github.com/acts-project/vecmem/releases/tag/v0.2.0) by default.

As I discussed with @stephenswat today, we desperately need to introduce some CI checks for the repository. As the current main branch doesn't build out of the box for instance. (Because the VecMem version built by it by default is too old.)


stephenswat:
Solves #47.

stephenswat:
We should really merge this in soon, I think. Currently, the `main` branch of traccc does not compile by itself.

paulgessinger (approved):
Right. We also *need* CI for this repository. Anyway, this fixes the build.

PR #50, opened 2021-07-15
Add a simple CI build workflow

stephenswat:
We've had an issue recently where a commit slipped through the CI system and broke the entire `traccc` build. This is not necessarily unexpected, because our current CI system doesn't actually do any checks whether the code still builds. This commit aims to resolve that situation by adding a workflow that tries to build the code, and fails if there is a compiler error. This workflow will build the code in `Release` and `Debug` builds, and it will use a standard CPU image as well as an image with the CUDA compiler.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1,6 +1,7 @@
 name: Builds
 
 on:
+  push:
   pull_request:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


paulgessinger (approved):

stephenswat:
Ah, we can't fetch the data folder from the LFS hosted on CERN Gitlab. That could be a problem.

paulgessinger (approved):

PR #53, opened 2021-07-19
Add macros for host and device qualifiers

stephenswat:
As with some of the other ACTS projects, we will want to have some macros to add host and device (or both) qualifiers to functions without having the non-CUDA compiler implode on itself.


paulgessinger (approved):
Makes sense to me.

PR #54, opened 2021-07-20
Add Python-based CCL data generator tool

stephenswat:
Like I presented a few weeks ago in the parallelization meeting, I have been using my own data generator to test CCL and CCA implementations. In case it is useful to anyone else, I wanted to keep it around in the traccc repository. The data generated by this tool is devoid of any and all physical meaning, but it is specifically designed to expose the performance-sensitive parameters of the input data. In this case, that is the number of modules, the size of each modules, and the parameters to the log-normal distribution governing the number of hits per module, and the number of activations per hit.

### DIFF
--- a/extras/README.md
+++ b/extras/README.md
@@ -1,0 +1,6 @@
+# traccc extras
+
+This directory is designed to hold miscellaneous bits and bobs that are
+relevant to the traccc project. Examples include small scripts, write-ups,
+small data files, and whatever else which may be useful to the traccc team, but
+which does not fit in any of the other directories.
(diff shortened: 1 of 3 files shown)


krasznaa (approved):
I know my review was not asked for, but here it goes anyway. :stuck_out_tongue:

I think we should absolutely keep this script here. Organisation-wise this would be the smartest thing to do, to collect all useful code that we wrote for this R&D project, in the project's main repository.

I myself would probably make the Python script executable, using the standard

```sh
#!/usr/bin/env python
```

syntax, but that's not a hill I would die on either. :stuck_out_tongue:

Also, maybe adding a 1-line `README.md` file for the new `extras/` directory may make sense. Just to explain that we will hold "extra code" in that directory.

stephenswat:
I've made the Python file executable, and I've added a shebang to it. Also, there is now a `extras/README.md` file to inform users of what the directory is for!

paulgessinger (approved):

PR #55, opened 2021-07-20
Add Acts as a dependency through FetchContent

stephenswat:
As we have seen in #35, we need to somehow include ACTS as a dependency for traccc so we can compile it easily, including on the CI. To make this possible, this pull request adds some CMake code to fetch and build the ACTS core in the same way that we build the vecmem code right now. That is to say, we see if it is installed on the system. If it is, we're done. If it isn't, CMake fetches the code from source for us, and adds it into the tree for building. This should give us a flexible solution to the dependency problem, at least for now.

Thanks Attila for laying most of the groundwork for this in your vecmem CMake file. :smiley:

### DIFF
--- a/cmake/traccc-acts.cmake
+++ b/cmake/traccc-acts.cmake
@@ -1,0 +1,31 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2021 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# Guard against multiple includes.
+include_guard(GLOBAL)
+
+# Try to find ACTS installed on the system.
+find_package(Acts QUIET)
+
+# If it was found, then we're finished.
+if(Acts_FOUND)
+   # Call find_package again, just to nicely print where it is picked up from.
+   find_package(Acts)
+   return()
+endif()
+
+# Tell the user what's happening.
+message(STATUS "Building ACTS as part of the traccc project")
+
+# Declare where to get ACTS from.
+FetchContent_Declare(
+   Acts
+   GIT_REPOSITORY "https://github.com/acts-project/acts.git"
+   GIT_TAG "v9.2.0"
+)
+
+# Make the ACTS source available to the rest of the build system.
+FetchContent_MakeAvailable(Acts)
--- a/cmake/traccc-acts.cmake
+++ b/cmake/traccc-acts.cmake
@@ -7,7 +7,7 @@
 # Guard against multiple includes.
 include_guard(GLOBAL)
 
-# Try to find ACTS installed on the system.
+# Try to find Acts installed on the system.
 find_package(Acts QUIET)
 
 # If it was found, then we're finished.
@@ -18,14 +18,15 @@ if(Acts_FOUND)
 endif()
 
 # Tell the user what's happening.
-message(STATUS "Building ACTS as part of the traccc project")
+message(STATUS "Building Acts as a dependency of traccc")
 
-# Declare where to get ACTS from.
+# Declare where to get Acts from.
 FetchContent_Declare(
    Acts
    GIT_REPOSITORY "https://github.com/acts-project/acts.git"
    GIT_TAG "v9.2.0"
 )
 
-# Make the ACTS source available to the rest of the build system.
-FetchContent_MakeAvailable(Acts)
+# Make the Acts source available to the rest of the build system.
+FetchContent_Populate(Acts)
+add_subdirectory("${acts_SOURCE_DIR}" "${acts_BINARY_DIR}" EXCLUDE_FROM_ALL)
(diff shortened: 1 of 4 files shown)


cmake/traccc-acts.cmake
```diff
@@ -0,0 +1,31 @@
...
+
+# If it was found, then we're finished.
+if(Acts_FOUND)
+   # Call find_package again, just to nicely print where it is picked up from.
+   find_package(Acts)
+   return()
+endif()
+
+# Tell the user what's happening.
+message(STATUS "Building ACTS as part of the traccc project")
```
krasznaa, line 21: :smile: I think the Acts team would object to calling Acts "ACTS". :rofl:

cmake/traccc-acts.cmake
```diff
@@ -0,0 +1,31 @@
...
+
+# Declare where to get ACTS from.
+FetchContent_Declare(
+   Acts
+   GIT_REPOSITORY "https://github.com/acts-project/acts.git"
+   GIT_TAG "v9.2.0"
+)
+
+# Make the ACTS source available to the rest of the build system.
+FetchContent_MakeAvailable(Acts)
```
krasznaa, line 31: :thinking: I wonder if we may want to use the same formalism that I used for VecMem.

```cmake
FetchContent_Populate( VecMem )
add_subdirectory( "${vecmem_SOURCE_DIR}" "${vecmem_BINARY_DIR}"
   EXCLUDE_FROM_ALL )
```

The advantage of that is that only those parts of VecMem are built that are explicitly needed by traccc. So in VecMem's case the unit tests are not built by default.

I know that Acts works a little differently. It doesn't build unit tests by default, you have to explicitly turn them on. Still, I liked this setup quite a bit, where the "included project" would not add any targets to `ALL` itself. It would just provide the targets for components in this main project to use.

However I don't feel super strongly about it in the end...

krasznaa (commented):
I think this should work fine. At one point we will need to start worrying about synchronising how each project builds their own GPU code. To make sure that acts, vecmem and traccc would consistently turn CUDA/SYCL/etc. support on or off.

But since I'm not entirely sure yet myself how we should do that exactly, this should be fine for now. :wink:

I would just like to hear your opinion about using `add_subdirectory(...)` explicitly.

cmake/traccc-acts.cmake
stephenswat: Wait, is "ACTS" not the proper capitalization? Now I'm confused! The paper (https://arxiv.org/abs/2106.13593) does capitalize it that way! :sob:

paulgessinger: Ugh. This has been a bit of back and forth. We've converged on referring to it as *Acts* in the code since we follow (I believe it's) Google's guideline to not write all-caps acronyms when part of names. When referring to the project, we decided to use *ACTS* in the paper.

cmake/traccc-acts.cmake
stephenswat: So if I understand what you're saying, using the `FetchContent_MakeAvailable` formalism adds every target defined by vecmem (which is not explicitly disabled by a CMake flag) to `ALL`, and as such running `make` without any specifiers would do extra work building those unnecessary parts, while your approach does not implicitly add them to `ALL` and, as a result, parts of the code base are only built if they are actually required by one of the targets within traccc as a dependency? If that is the case, I agree it makes sense to do it your way.

cmake/traccc-acts.cmake
stephenswat: Alright, I'll change it to read "Acts" then! :smile:

cmake/traccc-acts.cmake
stephenswat: Yes, empirical proof (me testing it on my laptop) shows significantly faster builds when Acts is never actually used as a dependency anywhere in the code. It's unlikely that this would actually ever happen in practice, but I agree it is a better solution.

That said it seems like very strange design that `FetchContent_MakeAvailable` does not have a `EXCLUDE_FROM_ALL` option. Perhaps we should open a merge request on CMake...

paulgessinger: Doesn't `FetchContent_MakeAvailable` literally do the same thing as @krasznaa wrote above? If so, we could just use manual `Populate` and `add_subdirectory` with `EXCLUDE_FROM_ALL`.

stephenswat: _Almost_, but not quite. The `FetchContent_MakeAvailable` macro is defined here: https://github.com/Kitware/CMake/blob/master/Modules/FetchContent.cmake#L1183, and the subdirectory is added in line 1211. It does not specify the `EXCLUDE_FROM_ALL` flag, and I really don't understand why. Anyhow, I've updated the pull request to use the formalism suggested by Attila.

paulgessinger (approved):
That's what I meant in my comment.

PR #56, opened 2021-07-22
Implement a fast GPU-friendly map for module IDs

stephenswat:
Right now, we are using `std::map` for the mappings between module IDs and transformation matrices. This works on the CPU, but it does not work for the GPU or other heterogeneous platforms. For one, `std::map` is not supported at all in CUDA, but even if it were it most likely would not perform well.

We implement a new map that relies on the fact that our list of module IDs has a lot of contiguous sub-ranges. The map works in two steps. Firstly, it selects a contiguous sub-range using a balanced binary search tree, and then it gets the value itself through a constant-time linear access in an array.

This class is GPU-friendly because it stores all its data in a contiguous fashion, and it is designed to reduce thread divergence as
much as possible. Track divergence performance loss is determined by the ratio between the number of steps taken by the thread with most work, compared to the thread with least work. For the TrackML detector, this ratio is 15:1 for `std::map`, but it is only 7:1 for this map.

In principle, the map is more generally usable and it is templated to support any key and value type, but in practice the performance of it depends very, very heavily on the contiguous nature of sub-ranges in the key set. Therefore, it is likely only useful for module IDs.

As it stands, this map only supports creation, it does not support any post-creation updating like insertion or deletion. Constructing these maps is done by constructing a `std::map` first, and converting it later, all in one go.

The performance of this class is fairly good. For the TrackML detector, the `std::map` approach takes about 7.01 milliseconds to look up every single module. Meanwhile, this approach takes only 1.15 milliseconds to do the same, in my simple benchmarks!

### DIFF
--- a/core/include/geometry/module_map.hpp
+++ b/core/include/geometry/module_map.hpp
@@ -1,0 +1,395 @@
+/**
+ * TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <iostream>
+
+#include "definitions/primitives.hpp"
+
+namespace traccc {
+/**
+ * @brief A fast and GPU-friendly map for consecutive module IDs to values.
+ *
+ * Using std::map to model the relationship between geometry identifiers and
+ * transformation matrices works only on the CPU. std::map is not supported at
+ * all on CUDA, and either doesn't work or is inefficient on other platforms.
+ *
+ * This class is a more efficient and more GPU-friendly implementation of the
+ * geometry ID to transformation mapping concept.
+ *
+ * The implementation of this class relies on the idea that module IDs exist in
+ * contiguous sub-ranges, and we use this fact to get more efficient accesses.
+ * First, we reduce the range of module IDs to a set of contiguous sub-ranges.
+ * In the TrackML detector, there are roughly 18,000 modules, but there are
+ * only 71 contiguous sub-arrays. Because we can index a contiguous range in
+ * constant time, this presents possible speed-up.
+ *
+ * Thus, the map is essentially two-layered. First, we determine which
+ * contigous sub-range a given value must fall in, and then we get the value
+ * from that sub-range in constant time.
+ *
+ * The first layer is implemented as a simple balanced binary search tree,
+ * which we implement in a GPU-friendly fashion. This means that the cost of
+ * finding the relevant sub-range is O(log_2(n)), with n the number of
+ * sub-ranges. For the TrackML detector, this means we can find our value in
+ * only 7 steps!
+ *
+ * @note In theory, this class can be used for any mapping of key to value, but
+ * it relies quite heavily (for performance) on the contiguous nature of the
+ * keys inserted into it.
+ *
+ * @note This map supports only construction. You can't update it in any way.
+ * To build one, construct a std::map first, and then convert it.
+ */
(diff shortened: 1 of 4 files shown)


krasznaa (commented):
Very interesting little class.

Did you have a chance to test it actually running on a GPU? What I mean is, we need to think a bit about how this data would be transferred to the GPU, how the object could be constructed on the GPU, etc. Because if we can't answer those questions then all this is only making our CPU code faster. Which is definitely not nothing, but still...

Now... I can absolutely see ways in which VecMem could be used in this class. I would imagine that it could have the current constructor marked with `__host__`, and another constructor that's based on some "data object", marked with `__device__ __host__`.

Though... :thinking: We may not be able to get away with a single type. Because the issue is the same as with all of the VecMem types. That on the host the class needs to own its memory, while on the device it doesn't. But with some clever templating around the VecMem vector types, this could all fall into place.

So again, I wonder if a demonstration could be provided for how this would be used in GPU code in practice...

stephenswat:
I haven't applied this data structure on the GPU in practice yet, but it is designed from the ground up to be GPU friendly. As far as data transfer is concerned, the structure is designed to consist of two contiguous vectors, and nothing more. I did this to make transferring it to the GPU as easy as possible. All accesses are handled via indices, not pointers, so the code is independent of the position in memory of these vectors (to the maximum possible extent). The actual look-up algorithm might need to change a bit. Currently it is defined recursive because as is often the case this algorithm is most naturally defined via recursion. How well this will translate to CUDA, I am not sure. It is fully tail-recursive, so the compiler could do a great job compiling it in a GPU-friendly way. But otherwise it would be trivial to write a different implementation of lookup.

core/include/geometry/module_map.hpp
```diff
@@ -0,0 +1,392 @@
...
+        std::sort(keys.begin(), keys.end());
+
+        /*
+         * This vector will hold our nodes. We'll update it as we go, and we
+         * start out with an incomplete first node, which represents the single
+         * value at the start of the key vector. We will add to this later!
+         */
+        std::vector<module_map_node> nodes;
+        nodes.push_back(module_map_node{
+            .start = keys.front(), .size = 1, .index = values.size()});
```
paulgessinger, line 281: Wait, isn't this C++20?

core/include/geometry/module_map.hpp
```diff
@@ -0,0 +1,392 @@
...
+         * the last key, and store them if so.
+         */
+        for (std::size_t i = 1; i < keys.size(); ++i) {
+            /*
+             * If the key is not adjacent to the last one, we push a new
+             * partial node onto the node vector.
+             */
+            if (keys.at(i) != keys.at(i - 1) + 1) {
+                nodes.push_back(module_map_node{
+                    .start = keys.at(i), .size = 0, .index = values.size()});
```
paulgessinger, line 295: And this.

paulgessinger (commented):
Looks pretty clever to me.
Small comment about usage of designated initializers.

core/include/geometry/module_map.hpp
stephenswat: Heh, you're right. I keep forgetting that this isn't widely accepted C++ yet, I use it in C code all the time because it's been valid there for a long time. :sweat_smile:

Good catch, I'll change it.

paulgessinger: Oh, I can't wait either :D

stephenswat:
I've removed all the uses of designated initializers. :smile:

paulgessinger (approved):
Great!

PR #57, opened 2021-07-22
Add a fixture base class for tests using data

stephenswat:
We have a lot of tests which need to access data from the data directory, which is set via the `TRACCC_TEST_DATA_DIR` environment variable. This raises a lot of boilerplate code though, so we can abstract some of this away by defining a fixture base class which automatically grabs the data directory.


paulgessinger (approved):
Good idea.

PR #59, opened 2021-07-23
Reorganizing the algorithm directory

beomki-yeo:
I moved clusterization algorithms files into `core/include/clusterization` and `algorithms.hpp` into `core/include/utils` directory. This is just to follow Acts core style. @stephenswat is this okay for you?


paulgessinger:
Agreed that @stephenswat should comment here.

My 2 cents: I think `algorithms/algorithm.hpp` and `algorithms/clusterization/*.hpp` would be more intuitive.

stephenswat:
It's very good that you remind me, thanks. In general I am totally on board with this merge request. The only thing that comes to mind is that we might want to move all of these CPU algorithms into their own `/algorithms/cpu/` or `/device/cpu/` directory, to more clearly separate them from CUDA (and other platform) code. Just to be more future-proof. What do you two think?

beomki-yeo:
core directory itself is for cpu as far as I feel. If we want to make `core/algorithms/cpu`, we also have to make `core/algorithms/cuda` which is quite weird to me. `device/cpu` makes things complicated as well: We have to determine which part should be in `core` or `device/cpu`, and I think It will take one year for us to discuss ^^;;. Most of all I think It is better to keep following the Acts structure unless there is a good reason to change it.

But I am happy with `device/cuda` or `device/sycl` for gpu APIs

stephenswat (approved):
I agree. Let's not daddle on this for too long. :smile:

PR #60, opened 2021-07-26
Decouple I/O from traccc's algorithm

georgi-mania:
Add new example (io_dec_par_example) which works as function composition: `write o run o read`. In detail:

1. read the input CSV files and store the data in memory (`demonstrator_input`)
2. run the traccc::algorithms and stores the results in memory (`demonstrator_output`)
3. writes output CSV based on in-memory temporary results

**Note 1:** The code uses OpenMP parallel loops if OMP support is enabled. The `reader` and `writer` header files should compile without OpenMP support too.
**Note 2:** The `vecmem::resource` is passed by reference to all the functions that require it in order to ensure a state-less model (as agreed in ACTS Parallelization meeting on 04.06.2021) 
**Note 3:** Performance results for this implementation were shown in ACTS Parallelization meeting on 23.07.2021.

### DIFF
--- a/io/include/writer.hpp
+++ b/io/include/writer.hpp
@@ -45,4 +45,4 @@ namespace traccc {
         }
     }
 
-} // end namespace
+    }  // namespace traccc
--- a/io/include/writer.hpp
+++ b/io/include/writer.hpp
@@ -1,0 +1,48 @@
+#pragma once
+
+#include "edm/cell.hpp"
+#include "edm/cluster.hpp"
+#include "edm/measurement.hpp"
+#include "edm/spacepoint.hpp"
+#include "csv/csv_io.hpp"
+#include "tmp_edm.hpp"
+
+namespace traccc {
+
+    void write_measurements(size_t event, const traccc::host_measurement_container &measurements_per_event) {
+        traccc::measurement_writer mwriter{std::string("event") + std::to_string(event) + "-measurements.csv"};
+        for (size_t i = 0; i < measurements_per_event.items.size(); ++i) {
+            auto measurements_per_module = measurements_per_event.items[i];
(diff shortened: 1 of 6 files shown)


io/include/writer.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include "edm/spacepoint.hpp"
+#include "tmp_edm.hpp"
+
+namespace traccc {
+
+void write_measurements(
+    size_t event,
+    const traccc::host_measurement_container &measurements_per_event) {
+    traccc::measurement_writer mwriter{
+        std::string("event") + std::to_string(event) + "-measurements.csv"};
```
paulgessinger, line 16: Maybe this could be zero-padded to fixed with (not really important).

io/include/writer.hpp
```diff
@@ -0,0 +1,54 @@
...
+            mwriter.append({module.module, local[0], local[1], 0., 0.});
+        }
+    }
+}
+
+void write_spacepoints(
+    size_t event,
+    const traccc::host_spacepoint_container &spacepoints_per_event) {
+    traccc::spacepoint_writer spwriter{
+        std::string("event") + std::to_string(event) + "-spacepoints.csv"};
```
paulgessinger, line 31: If above, then here as well.

paulgessinger (approved):
Looks sensible to me. Let me know what you think of the minor comment I added.

io/include/writer.hpp
stephenswat: Yeah I agree with Paul on this one, that would be more consistent with the padding that we have everywhere else (data files, Fatras, etc). Perhaps we could handle this in the `traccc::measurement_writer` instead? Do the zero padding there, and accept just an event number as argument to it?

PR #62, opened 2021-08-05
Remove loud include of Acts package

stephenswat:
This `find_package` is doing essentially the same thing that [`cmake/traccc-acts.cmake`](https://github.com/acts-project/traccc/blob/77cc01fccc28940100ffe369c053ff38b8a93bff/cmake/traccc-acts.cmake) is doing, which means that we get some loud (and rather scary) error messages if we do not have Acts installed in the development. The previously mentioned CMake file should be more than sufficient to find Acts in virtually all cases, so we do not need to look for it in the main CMake file as well.


krasznaa (approved):
Yepp, fully aggree.

PR #63, opened 2021-08-05
Clean up unused variables in CSV module

stephenswat:
Nothing major, but these two unused variables were spamming my builds up with unused variable warnings, so I'd like to get rid of them.

### DIFF
--- a/io/include/csv/csv_io.hpp
+++ b/io/include/csv/csv_io.hpp
@@ -277,12 +277,10 @@ std::vector<cluster_collection> read_truth_clusters(
 host_spacepoint_container read_hits(
     fatras_hit_reader& hreader, vecmem::memory_resource& resource,
     unsigned int max_hits = std::numeric_limits<unsigned int>::max()) {
-    uint64_t reference_id = 0;
     host_spacepoint_container result = {
         host_spacepoint_container::header_vector(&resource),
         host_spacepoint_container::item_vector(&resource)};
 
-    bool first_line_read = false;
     unsigned int read_hits = 0;
     csv_fatras_hit iohit;
 


krasznaa (approved):
Looks simple enough.

Should we introduce `-Werror` for one of the CI builds? :thinking: I like the setup which we have in VecMem, that it uses `-Werror` for Debug builds... Though we could also just add that flag explicitly in the CI build configuration file as well.

stephenswat:
Yes, I like the idea of having a `-Werror` CI build a lot, definitely something we should look into (should be a one-line change).

paulgessinger:
> Looks simple enough.
> 
> Should we introduce `-Werror` for one of the CI builds? 🤔 I like the setup which we have in VecMem, that it uses `-Werror` for Debug builds... Though we could also just add that flag explicitly in the CI build configuration file as well.

For ACTS core we do this, but we set it explicitly in the CI CMake call. I don't think it's a great idea to have CMake add this on every Debug build, even local ones.

krasznaa:
>>> For ACTS core we do this, but we set it explicitly in the CI CMake call. I don't think it's a great idea to have CMake add this on every Debug build, even local ones.

It **can** be annoying at times.

My thinking with turning it on for Debug builds was that if somebody is going about debugging VecMem, they should be prepared to write warning-free code. While somebody who is only interested in using VecMem in release mode (possibly with a compiler that we didn't test before), should not be hampered by this. (I've been bitten by XRootD using `-Werror` all the time. Which, when Apple moves to new versions of its compiler, usually means that old XRootD versions just do not build anymore...)

So... I'm still in favour of keeping the current setup for VecMem. But adding the flag (though `CXXFLAGS`) just in the CI, is fine for this project.

cgleggett:
A caveat for using `-Werror` is Eigen - it includes some deprecated CUDA headers, so there are always tons of warnings about that. I think it's fine for VecMem itself, but I believe it will screw up traccc if it's added at the top level.

PR #66, opened 2021-08-09
Various improvements to the new module map type

stephenswat:
As requested by @paulgessinger, this pull requests makes the following changes to the `traccc::module_map` type:

1. All accessor methods (and other methods) are now `const`.
1. Additional methods have been added to the class:
a. The `contains` method determines whether a key can be found in the map.
b. The `empty` method determines whether the map is empty, which it should never be.
1. Additional tests have been added for the new class methods.
1. A header inclusion error was fixed.


paulgessinger (approved):
👍

PR #68, opened 2021-08-09
Centralize per-event filename, data directory, clean up string usage

paulgessinger:
Changes to use `std::stringstream` to do the padding.

### DIFF
--- a/examples/cpu/seq_example.cpp
+++ b/examples/cpu/seq_example.cpp
@@ -29,13 +29,6 @@
 int seq_run(const std::string& detector_file, const std::string& cells_dir,
             unsigned int events) {
 
-    const char* env_d_d = std::getenv("TRACCC_TEST_DATA_DIR");
-    if (env_d_d == nullptr) {
-        throw std::ios_base::failure(
-            "Test data directory not found. Please set TRACCC_TEST_DATA_DIR.");
-    }
-    std::string data_directory = env_d_d + std::string("/");
-
     // Read the surface transforms
     auto surface_transforms = traccc::read_geometry(detector_file);
 
--- a/examples/cpu/seq_example.cpp
+++ b/examples/cpu/seq_example.cpp
@@ -28,12 +28,13 @@
 
 int seq_run(const std::string& detector_file, const std::string& cells_dir,
             unsigned int events) {
-    auto env_d_d = std::getenv("TRACCC_TEST_DATA_DIR");
+
+    const char* env_d_d = std::getenv("TRACCC_TEST_DATA_DIR");
     if (env_d_d == nullptr) {
         throw std::ios_base::failure(
             "Test data directory not found. Please set TRACCC_TEST_DATA_DIR.");
     }
-    auto data_directory = std::string(env_d_d) + std::string("/");
+    std::string data_directory = env_d_d + std::string("/");
 
     // Read the surface transforms
     auto surface_transforms = traccc::read_geometry(detector_file);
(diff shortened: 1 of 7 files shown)


tests/include/tests/data_test.hpp
```diff
@@ -25,17 +27,7 @@ class data_test : public ::testing::Test {
...
-
-        if (env_d_d == nullptr) {
-            throw std::ios_base::failure(
-                "Test data directory not found. Please set "
-                "TRACCC_TEST_DATA_DIR.");
-        }
-
-        data_directory = std::string(env_d_d) + std::string("/");
-    }
+    virtual void SetUp() override { data_directory = traccc::data_directory(); }
```
stephenswat, line 30: This isn't really as useful anymore, we could consider getting rid of it.

examples/openmp/par_example.cpp
```diff
@@ -19,18 +19,15 @@
 #include "edm/spacepoint.hpp"
 #include "geometry/pixel_segmentation.hpp"
 #include "omp.h"
+#include "reader.hpp"
+
+using namespace std::literals::string_literals;
```
stephenswat, line 24: Why do we need this? I don't seem to be able to spot any use of these literals?

examples/cpu/seq_example.cpp
```diff
@@ -25,17 +26,13 @@
 #include "seeding/spacepoint_grouping.hpp"
 //#include "seeding/seed_finding.hpp"
 
+using namespace std::literals::string_literals;
```
stephenswat, line 29: Why do we need this here?

io/include/reader.hpp
```diff
@@ -14,22 +15,27 @@ using namespace std::placeholders;
...
 
-const std::string data_directory() {
-    auto env_d_d = std::getenv("TRACCC_TEST_DATA_DIR");
-    if (env_d_d == nullptr) {
-        throw std::ios_base::failure(
-            "Test data directory not found. Please set TRACCC_TEST_DATA_DIR.");
-    }
-    std::string data_dir = std::string(env_d_d);
-    return data_dir.append("/");
+const std::string &data_directory() {
```
stephenswat, line 18: I'm not sure I am too thrilled about returning a static reference here. It shouldn't be a big problem for now, though.

stephenswat (commented):
Looks good in general, just have a few changes. I think we might also want to look into using `std::filesystem::path` for these paths, as using `std::string` for this purpose can get sloppy. That can be a different, later change though.

examples/openmp/par_example.cpp
paulgessinger: Right, became irrelevant.

io/include/reader.hpp
paulgessinger: Hm, what's your concern? Lifetime should be fine, and we don't have to copy a string that's not supposed to change over any executable lifetime.

examples/cpu/seq_example.cpp
paulgessinger: Should be gone.

io/include/reader.hpp
stephenswat: I guess if we only ever have one data directory per executable lifetime (which does make sense), then it's fine.

stephenswat (approved):

krasznaa:
Just a quick comment about `std::filesystem`. We should be careful with fully embracing it... Unfortunately even with the latest versions of GCC and Clang it can be a pain to use... :frowning: In [Gaudi](https://gitlab.cern.ch/gaudi/Gaudi) we had to come up with a very painful setup for its use. @paulgessinger, I think we even talked about this for [Acts](https://github.com/acts-project/acts), didn't we? That it is easier to just use [boost::filesystem](https://www.boost.org/doc/libs/1_76_0/libs/filesystem/doc/index.htm) in all situations for now, instead of doing the complication that Gaudi does.

For a bit of history, see:
  - https://gitlab.cern.ch/gaudi/Gaudi/-/merge_requests/987
  - https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/GaudiPluginService/CMakeLists.txt#L20-38

paulgessinger:
@krasznaa exactly. We switched **to** `std::filesystem` and then **back to** `boost::filesystem`. The corresponding PRs are 
- https://github.com/acts-project/acts/pull/242
- https://github.com/acts-project/acts/pull/423

PR #69, opened 2021-08-09
Make a few methods `inline` that have to be

paulgessinger:
Also remove a global `using namespace` in a header.


stephenswat (approved):
Looks great. :+1:

stephenswat (approved):

PR #71, opened 2021-08-09
Run tests and examples in the CI

paulgessinger:
Run the unit tests and the examples in the CI. The examples are not checked for correctness, just that they don't error out.

I suspect they will run very slowly in the Debug build, so I might have to tweak that. But let's see.


run_examples.sh
```diff
@@ -0,0 +1,17 @@
...
+
+
+set -e
+set -x
+
+$build_dir/bin/ccl_example $TRACCC_TEST_DATA_DIR/tml_pixels/event000000000-cells.csv
+$build_dir/bin/io_dec_par_example tml_detector/trackml-detector.csv tml_pixels 10
+$build_dir/bin/seq_example tml_detector/trackml-detector.csv tml_pixels 10
+$build_dir/bin/par_example tml_detector/trackml-detector.csv tml_pixels 10
+$build_dir/bin/seq_single_module
```
krasznaa, line 17: :thinking: Instead of using a BASH-requiring script, I wonder if we shouldn't simply use [add_test(...)](https://cmake.org/cmake/help/latest/command/add_test.html) to declare all of these to [CTest](https://cmake.org/cmake/help/latest/manual/ctest.1.html). That would also have the upside of allowing the tests to run in parallel if desired. And would also avoid hardcoding the build directory layout (the `bin/` directory...) into the script.

Note that we could use a separate label for the "slow" tests/examples, if they turn out to run for much longer than what we'd like from unit tests. (So that we could explicitly filter them out/in depending on which tests we want to run.)

krasznaa (commented):
I'm always in favour of more tests. I would just avoid BASH scripting if it's not absolutely necessary. (For that ever-elusive Mac/Windows compatibility.)

.github/workflows/builds.yml
```diff
@@ -28,4 +28,8 @@ jobs:
       - name: Configure
         run: cmake -DCMAKE_BUILD_TYPE=${{ matrix.build }} -B build -S .
       - name: Build
-        run: cmake --build build --
+        run: >
+          cmake --build build --
+          && cmake --build build -- test
+          && ./run_examples.sh build
```
stephenswat, line 34: At the risk of sounding pedantic, could we split this into three different jobs? Keep the build job as it is, then run a `Tests` job to execute just `cmake --build build -- test`, and then an `Examples` job that runs `./run_examples.sh build`?

stephenswat (changes requested):

run_examples.sh
paulgessinger: Right, but we're running this only in the CI, it should run unmodified on macOS, and I'm not sure we should make decisions based on wanting to run on Windows, honestly.

I feel like adding them to the CTest test only makes sense if we are asserting the correctness of the examples, rather than just running them. Therefore I think this is the way to go until we potentially restructure what the example executables do / look like.

.github/workflows/builds.yml
paulgessinger: @stephenswat because then you start having to deal with artifacts to ship the build output between jobs, which in my opinion is not worth it.

@krasznaa if you care about the output, I would probably prefer setting `CTEST_OUTPUT_ON_FAILURE=1` on the job, like we do in ACTS core.

paulgessinger: @stephenswat what I can do though is split it into steps, that way the interfaces shows them separately.

stephenswat: Sorry - I didn't mean different jobs but rather different steps of the same job.

paulgessinger: @stephenswat ok done.

run_examples.sh
krasznaa: I'm not sure I see the difference. :confused:

I actually really appreciate it when a project provides CTest recipes for how its executables could run. Yes, your script is a reasonable recipe as well, it's really just a matter of where we want to define that we want to run these as tests.

Since in the end you could also just fully forego this script, and just put these commands into the `builds.yml` file as well. It's purely a question of implementation how we want to do this exactly...

paulgessinger: @stephenswat can you break the tie? 😄

stephenswat: I'm with Paul on this one, I don't personally think support for Windows should be a deciding factor in how we set this up, especially as the CI is virtually guaranteed to not be running on Windows.

Right now none of the examples are really in a spot where we can comprehensively test them for correctness, so that is something we can explore at a later point, I think. For now I don't see much value in anything other than simple running the examples and checking if they don't crash and burn.

.github/workflows/builds.yml
krasznaa: I'm okay with using `CTEST_OUTPUT_ON_FAILURE`. But since the test is failing at the moment, that quite strongly underlines my point of output being necessary in the CI on failure...

run_examples.sh
paulgessinger: As another comment on Windows: I believe GitHub Action's Windows runners come with bash preinstalled.

krasznaa: Fine. But then, do we need the script? Why don't we just extend `builds.yml`?

As for CTest: It literally just "runs something", which is supposed to return 0 on success, and something else on failure. Don't mix up CTest and GTest! So CTest is quite appropriate for testing whether something runs, or fails to run.

krasznaa:
I wonder... Are the GitHub runners providing us with more than I CPU core? :thinking: If so, both the builds and the tests should take advantage of that. Since by now the builds are becoming fairly long. If we're just leaving performance on the table, that's not great...

run_examples.sh
krasznaa: I thought of one more argument. :stuck_out_tongue: If the "examples" are run by CTest, then we could easily see all of the failures at once.

Right now some of the tests are failing. So the examples are not even run. If everything would be run by CTest, we would see all the failures (possibly some of them from the examples), at once. :wink:

paulgessinger:
I think you get 2 cores. I can switch the build to run make with more jobs, or switch to ninja. What would you prefer?

run_examples.sh
stephenswat: Now that is a good point. :sweat_smile:

paulgessinger:
I don't actually understand the job failure even with more output. Locally the tests succeed.

stephenswat:
My first instinct would be that the LFS data files are not checked out correctly.

stephenswat:
You might want to add `lfs: true` to the checkout step of the job, that might be it.

paulgessinger:
Fine. I add the examples to ctest using `add_test`.
I don't exactly know where this is screwed up in the cmake config, but CMake/CTest is unable to extract the excutable location from the name automatically, so I have to specify `${CMAKE_RUNTIME_OUTPUT_PATH}/EXECUTABLE` which is a bit annoying, but ok.

stephenswat:
:tada:

stephenswat (approved):
More than happy to sign off on this now, and the LFS issue seems to be fixed.

PR #72, opened 2021-08-10
CMake Cleanup/Updates, main branch (2021.08.10.)

krasznaa:
The main trigger for this PR was #71, but once I started going, I changed a couple of things. :stuck_out_tongue:

So the original point was to replace the simple `add_test(<test name> <command>)` calls with `add_test(NAME <test name> COMMAND <command>)` ones. This is necessary because unfortunately CMake does not expand target names with the first formalism into full path names to the built executables. But it does do so with the second formalism. (This is not documented anywhere, and I consider it a bug. I just never took the time to report it to the developers... :frowning:)

While updating the OpenMP tests, I noticed that global variables were being updated for no reason whatsoever. Since the code was already correctly using the `OpenMP::OpenMP_CXX` imported target. Which, by itself, is enough to set up the build of the OpenMP code correctly. So I simply removed those lines. (Note by the way that `${OpenMP_EXE_LINKER_FLAGS}` is **not** an existing variable. It was never set to anything... :stuck_out_tongue:)

Finally, as long as I was tweaking the tests anyway, I tried to make the project respect the `BUILD_TESTING` flag a bit more. This is a flag automatically set up by [CTest](https://cmake.org/cmake/help/latest/module/CTest.html). And projects are advised to use it in the setup of their own tests. (With `BUILD_TESTING=FALSE` the `add_test(...)` function does not do anything. So it's a good idea to not even build the test executables in such a case.) The examples are still being built with `BUILD_TESTING=FALSE` as well, they are just not set up as tests for CTest in that case.
  - I had to re-order the setup of the externals and traccc's own code for this to work. Since in `traccc-vecmem.cmake` we set `BUILD_TESTING=FALSE` explicitly to avoid VecMem building its tests. Which would interfere with how traccc builds GoogleTest. So it seemed easiest to just put that forceful setting of `BUILD_TESTING` at the end of the configuration, to have everything else get `BUILD_TESTING` from the cache variable set up by CTest, as it should.

Pinging @paulgessinger, @georgi-mania.

P.S. I also put a couple of newlines at the end of the files that were missing it. :stuck_out_tongue:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,12 +29,14 @@ set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY
 list( APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake" )
 include( traccc-compiler-options )
 
-# Build the necessary externals.
-add_subdirectory(extern)
-include(traccc-dependencies)
-
 # Build the traccc code.
 add_subdirectory(core)
 add_subdirectory(io)
 add_subdirectory(examples)
-add_subdirectory(tests)
+if(BUILD_TESTING)
+   add_subdirectory(tests)
+endif()
+
+# Build the necessary externals.
+add_subdirectory(extern)
+include(traccc-dependencies)
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,10 +1,7 @@
 add_subdirectory(cpu)
 
-find_package(OpenMP)
-if (OPENMP_FOUND)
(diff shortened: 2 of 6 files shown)


stephenswat (approved):
Looks harmless enough. Good to go in when the CI passes as far as I'm concerned. :smile:

paulgessinger (approved):
Yeah, fine with me. I feel like `BUILD_TESTING` is a bit of an anti-pattern for exactly the reasons you outline, namely that you have to reorder configuration to enable only the correct testing suites. If it works for us this way, I'm ok.

Issue #49, opened 2021-07-09
Move to use algebra-plugins


asalzburger:
In order to harmonise the R&D repositories, `traccc` should not contain a copy of the aglebra-plugins, but use it as a submodule as `detray` does.

beomki-yeo:
This can be closed now

PR #74, opened 2021-08-12
Integrating algebra plugins

beomki-yeo:
from [issue #49](https://github.com/acts-project/traccc/issues/49)

A lot of things are brought from detray plugin directory. Thank @asalzburger @niermann999 for groundwork... :)
I made the algebra work for only one of plugin types (default: array). 

Reviewing request to: @paulgessinger @asalzburger


asalzburger:
That sounds about right, I don't foresee parallel plugins needed at any stage.

stephenswat:
Am I understanding correctly that this brings us back to using `std::array` in device code? :confused:

beomki-yeo:
Unfortunately yes. I think the algebra-plugin repository should integrate vecmem to replace `std::array` with `vecmem::static_array`

stephenswat:
Yeah, I think that would be a fine solution! :smile: Do we have any algebra plugins that are out-of-the-box device compatible right now?

beomki-yeo:
Are you talking about this or something else? https://github.com/acts-project/algebra-plugins
I would not say that it is fully device compatible but `std::array` manages to work in device

stephenswat:
Sadly `std::array` only works on the device with modern compilers under an experimental relaxed constexpr regime, so I am not sure we should go back to using it in device code. That is why we added `vecmem::device_array` in the first place.

beomki-yeo:
Hmm that sounds right. Then we may have two options - 1) Leave this PR until algebra plugin has vecmem array 2) or just close it and get by with current implementation. Of course I prefer the first option.

niermann999:
Hi,
let me have a look. In principle it should not be complicated to include in the plugins, if we sort out the cmake integration. I could for the moment just set a flag to only build vecmem for the plugins when in testing.

stephenswat:
I think having a `vecmem::static_array` plugin would be amazing. I know some of the plugins might work on the GPU (thinking of Eigen, and the `std::array` under the right circumstances), but something transitional that uses `vecmem::static_array` would be nice. The good thing is that the code written for `std::array` should translate (almost) one-to-one to the VecMem one.

beomki-yeo:
Okay. I modified a few things conjunction with [algebra_plugins PR 12](https://github.com/acts-project/algebra-plugins/pull/12) to avoid fetch_populating vecmem multiple times.

beomki-yeo:
@paulgessinger it's rebased now.

paulgessinger:
So what's the status on the algrebra plugins? On https://github.com/acts-project/algebra-plugins/pull/14 you suggested not using it after all?

beomki-yeo:
I am using the algebra plugins here. What is the concern??

Also, Wasn't it the conclusion of [PR14](https://github.com/acts-project/algebra-plugins/pull/14) that plugins cannot be included at the same time? [See the comments on PR14](https://github.com/acts-project/algebra-plugins/pull/14#discussion_r696671756)

beomki-yeo:
@paulgessinger I updated to use `__plugin` for types - is that what you meant?

paulgessinger (approved):
With the caveat that these plugins are now mutually exclusive (and I think will give linker errors if the cmake targets are linked together), we discussed by email to merge this, and potentially address this in a follow-up PR.

PR #75, opened 2021-08-15
Implement collection class

beomki-yeo:
Collection is similar with container but only with items
It follows the structure in [PR33](https://github.com/acts-project/traccc/pull/33)
I didn't add the element_view for collection since it seems not required for this vector type class

### DIFF
--- a/core/include/clusterization/component_connection.hpp
+++ b/core/include/clusterization/component_connection.hpp
@@ -27,9 +27,6 @@ namespace traccc {
 class component_connection
     : public algorithm<
           std::pair<const host_cell_collection&, const cell_module&>,
-          cluster_collection>,
-      public algorithm<
-          std::pair<const device_cell_collection&, const cell_module&>,
           cluster_collection> {
 
     public:
@@ -51,7 +48,7 @@ class component_connection
         const override {
         const host_cell_collection& cells = i.first;
         const cell_module& module = i.second;
-        return this->operator()<vecmem::vector>(cells, module);
+        return this->operator()(cells, module);
     }
 
     /// Callable operator for the connected component, based on one single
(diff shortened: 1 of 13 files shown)


examples/cpu/ccl_example.cpp
```diff
@@ -76,8 +76,10 @@ void print_statistics(const traccc::host_cell_container& data) {
 void run_on_event(traccc::component_connection& cc,
                   const traccc::host_cell_container& data) {
     for (std::size_t i = 0; i < data.headers.size(); ++i) {
+        traccc::host_cell_collection cells_per_module(data.items[i]);
+
         traccc::cluster_collection clusters_per_module =
-            cc({data.items.at(i), data.headers.at(i)});
+            cc({cells_per_module, data.headers.at(i)});
```
stephenswat, line 82: Here, for example. This just seems like more work and more overhead.

stephenswat (commented):
In general I don't feel very strongly either way, but I must say I don't really understand the added value of this. It seems like this just adds more boilerplate and more overhead for no discernible gain. Maybe you could help me understand the motivation behind this PR a little better?

beomki-yeo:
The motivation is from hackathon study where we used vector collection rather than header-vector container for trajectory edm. 
If we are going to use only container for edm, this has no values as you mentioned. Do you think it is okay to use vector-like collection for edm? or should we stick to container edm

stephenswat:
Ah, so you want to move away from the jagged vector design to a contiguous vector with offset indices? Yes I agree that is absolutely a good idea for GPU programming. :+1:

beomki-yeo:
Noooot really... I just used a vector because I didn't find a good reason to use jagged vector for trajectory container. 
But I still suspect It can be changed to jagged vector in the future only if there is a good reason for it

beomki-yeo:
It's rebased now @stephenswat can you review later on?

PR #76, opened 2021-08-17
Group sub-algorithms into higher level algorithms

beomki-yeo:
It's for grouping the algorithms into higher algorithm class. I think it is better since we have had to modify all example files whenever there is a change in the codes.

Currently the PR is in very WIP-ish state. Also need to think of how to include openmp code in algorithm codes... (@georgi-mania any suggestion?)

**Update**

Currently sub-algorithms are grouped based on its species:
`examples/algorithms/cpu/include/clusterization/clusterization_algorithm.hpp`
`examples/algorithms/cpu/include/track_finding/seeding_algorithm.hpp`
`examples/algorithms/cuda/include/cuda/track_finding/seeding_algorithm.hpp`

Those algorithms have been applied to `seq_example` and `seq_example_cuda` but not for the `par_example` or `io_dec_par_example` as I didn't decide how to put openmp codes into algorithm files.  We can discuss it in the next parallelization meeting.


examples/algorithms/include/clusterization/clusterization_algorithm.hpp
```diff
@@ -0,0 +1,82 @@
...
+        // output containers
+        auto& measurements_per_event = o.first;
+        auto& spacepoints_per_event = o.second;
+
+        // reserve the vector size
+        measurements_per_event.headers.reserve(cells_per_event.headers.size());
+        measurements_per_event.items.reserve(cells_per_event.headers.size());
+        spacepoints_per_event.headers.reserve(cells_per_event.headers.size());
+        spacepoints_per_event.items.reserve(cells_per_event.headers.size());
+
```
georgi-mania, line 50: To use the OpenMP (if the code was compiled with OpenMP support), the pragmas are needed:
`#if defined(_OPENMP)
#pragma omp parallel for
#endif`
The issue is with the containers that aggregate the results (e.g. `spacepoints_per_event.items.push_back`). They need synchronization mechanisms to ensure the consistency of the result. The open issue is still where would it be better to handle the synchronization? (a) at the user level, i.e. wrap the `container.push_back()` function into a critical region OR (b) at the container level, i.e. the container's implementation of the push_back() function already handles this case. There are pros and cons for each one. Maybe we can discuss more in the next parallelization meeting.

beomki-yeo: Thanks for reply. Maybe you can have a look into `examples/algorithms/cpu/include/clusterization/clusterization_algorithm.hpp` and `examples/algorithms/cpu/include/track_finding/seeding_algorithm` to catch up. This algorithms are applied to `seq_example` but not to `par_example` and `io_dec_par_example` as I didn't decide how to put openmp codes into algorithm files. We can also talk about how to do it in the next meeting

> if we know an approximate number of space points and measurements coming from each cell, we could allocate a fixed region in the container for each cell and in the end, just remove the empty spaces.

I think I've done it in `traccc/examples/algorithms/cpu/include/clusterization/clusterization_algorithm.hpp` by removing `push_back()`.

There are also several things that I want to point out: 
1) In my opinion, we have too many execution files: `seq_example` , `par_example` and `io_dec_par_example`. Once we determine which of either normal container or aggregated one (io_decoupled) is better, we may be able to merge two openmp run files into single one. We also have to decide whether to merge `seq_example` and openmp run file into one or not. It is possible if we put openmp codes into `clusterization_algorithm` or `seeding_algorithm`.
2) When checking the `par_example`, I found that there are two `omp parallel for`: one for events and the other for container. Having this two level parallelizations is better than just parallelizing only for events?

georgi-mania: The removal of the `push_back()` calls should be fine since `vecmem::jagged_vector` needs to guarantee that the memory addresses of its elements do not overlap, so we should not have race conditions. Maybe @stephenswat can confirm this. 

Regarding the 2 imbricated `parallel` loops: the answer depends on (a) the size of the loops and (b) the amount of computation done in each iteration. For the track fitting part in ACTS, it makes sense to have 2 imbricated loops (it reduced the memory consumption by half without impacting the wall clock execution time). If we have a smaller use case here, we can remove the container level loop. Do you have some estimates on the container sizes and maybe a rough estimate of the FLOPS done inside the loop?

I am happy to merge several examples if needed. Maybe we can discuss more in the next meeting.

beomki-yeo: container size is the number of modules so I think it is O(10^3) for tml detector and O(10^4) for atlas detector, respectively. And I have no idea about FLOPS.

> I am happy to merge several examples if needed

Yup. But I think it is okay to leave them until we have a complete tracking chain

examples/run/cuda/seq_example_cuda.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+int seq_run(const std::string& detector_file, const std::string& cells_dir,
+            unsigned int skip_events, unsigned int events, bool skip_cpu,
+            bool skip_write) {
+    auto env_d_d = std::getenv("TRACCC_TEST_DATA_DIR");
+    if (env_d_d == nullptr) {
+        throw std::ios_base::failure(
+            "Test data directory not found. Please set TRACCC_TEST_DATA_DIR.");
+    }
+    auto data_directory = std::string(env_d_d) + std::string("/");
```
georgi-mania, line 33: There is a helper function in [utils](https://github.com/acts-project/traccc/blob/556542b39feef4b3cda06d44f2becb6adc7fda0f/io/include/io/utils.hpp#L8) that handles the data_directory.

examples/run/cuda/seq_example_cuda.cpp
```diff
@@ -0,0 +1,282 @@
...
+                auto measurements_per_module = measurements_per_event.items[i];
+                auto module = measurements_per_event.headers[i];
+                for (const auto& measurement : measurements_per_module) {
+                    const auto& local = measurement.local;
+                    mwriter.append({module.module, "", local[0], local[1], 0.,
+                                    0., 0., 0., 0., 0., 0., 0.});
+                }
+            }
+
+            traccc::spacepoint_writer spwriter{
```
georgi-mania, line 193: There are also some writer functions in [writer.hpp](https://github.com/acts-project/traccc/blob/556542b39feef4b3cda06d44f2becb6adc7fda0f/io/include/io/writer.hpp#L13).

examples/run/cuda/seq_example_cuda.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+        // Read the cells from the relevant event file
+
+        /*-----------------------------
+              Read the cell data
+          -----------------------------*/
+
+        /*time*/ auto start_file_reading_cpu = std::chrono::system_clock::now();
+
+        std::string event_string = "000000000";
```
georgi-mania, line 80: There are some utils regarding reading in [reader.hpp](https://github.com/acts-project/traccc/blob/556542b39feef4b3cda06d44f2becb6adc7fda0f/io/include/io/reader.hpp#L26)

examples/run/cuda/seq_example_cuda.cpp
beomki-yeo: Fixed it now

core/include/edm/cell.hpp
```diff
@@ -64,7 +64,7 @@ struct cell_module {
     channel_id range0[2] = {std::numeric_limits<channel_id>::max(), 0};
     channel_id range1[2] = {std::numeric_limits<channel_id>::max(), 0};
 
-    pixel_segmentation pixel;
+    pixel_segmentation pixel = pixel_segmentation{-8.425, -36.025, 0.05, 0.05};
```
paulgessinger, line 67: I know this is pedantic, but why not
```suggestion
    pixel_segmentation pixel{-8.425, -36.025, 0.05, 0.05};
```

examples/run/cuda/seq_example_cuda.cpp
beomki-yeo: Didn't know that it exists! Thanks

examples/run/cuda/seq_example_cuda.cpp
beomki-yeo: Fixed it

core/include/edm/cell.hpp
beomki-yeo: makes sense. :p

examples/algorithms/include/clusterization/clusterization_algorithm.hpp
beomki-yeo: Instead, can you consider updating `par_example` and `io_dec_par_example` as a follow-up PR?

paulgessinger (approved):
Then let's get it in.

PR #78, opened 2021-08-20
CI build for CUDA

beomki-yeo:
This adds `-DTRACCC_BUILD_CUDA` option to CI build for cuda

@paulgessinger 
I have found two issues while working on this:
1) `ubuntu1804_cuda` seems support the `checkout@v2` due to the old git version
2) cuda 10 of `ubuntu1804_cuda` makes an build error in vecmem (@stephenswat @krasznaa is it supposed to happen?)

```
device/cuda/include/cuda/seeding/CMakeFiles/traccc_cuda_seeding.dir/build.make:82: recipe for target 'device/cuda/include/cuda/seeding/CMakeFiles/traccc_cuda_seeding.dir/__/__/__/src/seeding/doublet_counting.cu.o' failed

CMakeFiles/Makefile2:1634: recipe for target 'device/cuda/include/cuda/seeding/CMakeFiles/traccc_cuda_seeding.dir/all' failed
/__w/traccc/traccc/build/_deps/vecmem-src/core/include/vecmem/memory/deallocator.hpp(28): error: identifier "memory_resource" is undefined

/__w/traccc/traccc/build/_deps/vecmem-src/core/include/vecmem/memory/deallocator.hpp(46): error: identifier "memory_resource" is undefined

/__w/traccc/traccc/build/_deps/vecmem-src/core/include/vecmem/containers/data/jagged_vector_buffer.hpp(112): error: namespace "vecmem::details" has no member "deallocator"
```
Can we get an updated docker file with ubuntu20 and cuda 11?


beomki-yeo:
@paulgessinger do you have any comment on updating docker for recent cuda version?

paulgessinger:
Hey @beomki-yeo, I hadn't seen this, sorry. I'll have a look today at adding a more recent ubuntu+cuda image.

beomki-yeo:
Okay. I remember that cuda 11.4 also doesn't work for the seed finding, which I have to debug later (11.0 - 11.3 was fine as far as I remember)

beomki-yeo:
Hmm Still doesn't work after updating the docker image.

```
/__w/traccc/traccc/build/_deps/vecmem-src/core/include/vecmem/memory/memory_resource.hpp(21): error: name followed by "::" must be a class or namespace name

/__w/traccc/traccc/build/_deps/vecmem-src/core/include/vecmem/memory/polymorphic_allocator.hpp(22): error: name followed by "::" must be a class or namespace name

/__w/traccc/traccc/build/_deps/vecmem-src/core/include/vecmem/memory/polymorphic_allocator.hpp(22): error: expected a ";"
```

When I follow [memory_resource.hpp](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/memory/memory_resource.hpp),  the compiler is pointed to `experimental/memory_resource` rather than `memory_resource`. Anyone has an idea why it happens? @paulgessinger @krasznaa @stephenswat

beomki-yeo:
When I tested with my laptop (centos8, gcc 9.2 and cuda 11.3), the compilation was OK.

paulgessinger:
That's odd. I thought the `__has_include` macros would give an explicit error if that header was not found. It seems to include something, but then it's not in the correct namespace, or similar. This might be a difference in the standard libraries between C8 and Ubuntu?

stephenswat:
I have a fix for this behaviour which I will open a PR for on vecmem in just a moment.

stephenswat:
I think that, for a more solid solution, we will need to ensure that both the host-host compiler, the NVIDIA host compiler, and the NVIDIA CUDA compiler, have their C++ version set to the minimum version required for each of those three parts. In this particular case that would mean either using C++14 for the entire traccc project, or moving to C++17 for the NVIDIA compiler.

stephenswat:
The CI issues should be resolved with the introduction of #83. Rerunning the jobs to make sure that that is the case.

krasznaa:
While Stephen is still debugging the build on his own machine, at least the failure in the CI is understood. It turns out that CMake 3.17, which is in the Docker image, does not support C\+\+17 for CUDA yet.

https://gitlab.kitware.com/cmake/cmake/-/blob/v3.17.5/Modules/Compiler/NVIDIA-CUDA.cmake

So... @paulgessinger... we'd need a newer version of CMake in that Docker image. Let's just go for the current very latest one, 3.21.2, for some future-proofing.

paulgessinger:
I see, we could bump this I just used what we have for the other images. I'll be off for a bit, so I won't get around to do it very soon. You can open PR to the machines repo, and I can move that along.

stephenswat:
> I thought the same thing but it doesn't resolve the issue (...) Possibly CMake version or vecmem itself is the reason as mentioned above

Yeah work is in progress, it turned out to be a big of a deeper problem than I initially thought. Hope to have it fixed by tonight!

beomki-yeo:
@krasznaa @stephenswat can you make a new version tag for vecmem?

krasznaa:
> @krasznaa @stephenswat can you make a new version tag for vecmem?

Unfortunately the changes that we made in vecmem recently will not help with your use case. I'll make a new tag today, but I don't think that will help you in any way. Since the header that you ended up including in your `.cu` file, must be processed in C\+\+17 mode. This may or may not work with NVCC. Generally, "device compilers" are not meant to see that header at all...

krasznaa:
Here it is: https://github.com/acts-project/vecmem/releases/tag/v0.3.0

I'll open a PR for updating traccc to that version, after some further tests...

stephenswat:
Good news! The new CI images have been built and uploaded. Let's retry the CI on this one.

stephenswat:
It seems like Github Actions is out to frustrate me this week. Turns out the container registry has not (yet?) registered the new version as the new `latest` build: https://github.com/acts-project/machines/pkgs/container/ubuntu2004_cuda

beomki-yeo:
Okay now it works. thanks guys

.github/workflows/builds.yml
```diff
@@ -8,33 +8,41 @@ on:
...
-            container: ghcr.io/acts-project/ubuntu1804_cuda:v11
+            container: ghcr.io/acts-project/ubuntu2004_cuda:v13
         build:
           - Release
           - Debug
     steps:
       - name: Dependencies
         run: apt-get install -y git-lfs
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v1
```
stephenswat, line 30: Is there a reason to downgrade this step to version 1? Does v2 not do what we need it to do?

.github/workflows/builds.yml
```diff
@@ -8,33 +8,41 @@ on:
...
       - name: Dependencies
         run: apt-get install -y git-lfs
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v1      
         with:
           submodules: true
           lfs: true
       - name: Configure
-        run: cmake -DCMAKE_BUILD_TYPE=${{ matrix.build }} -B build -S .
+        run: >
```
stephenswat, line 35: I don't think it is ideal to do this with bash scripting, but for now it's fine as it is. Let's think of a more robust way to do this at a later point.

.github/workflows/builds.yml
```diff
@@ -8,33 +8,41 @@ on:
...
+          if ${{ matrix.platform.name == 'CPU'}}; then
+            cmake -B build -S . \
+              -DCMAKE_BUILD_TYPE=${{ matrix.build }} 
+          elif ${{ matrix.platform.name == 'CUDA'}}; then
+            cmake -B build -S . \
+              -DCMAKE_BUILD_TYPE=${{ matrix.build }} \
+              -DCMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc \
```
stephenswat, line 42: I would recommend `$(which nvcc)` here to make the finding a little more robust if we add more platforms later.

.github/workflows/builds.yml
```diff
@@ -8,33 +8,41 @@ on:
...
+              -DTRACCC_BUILD_CUDA=ON 
+          fi
       - name: Build
         run: cmake --build build -- -j$(nproc)
       - name: Run tests & examples
-        run: CTEST_PARALLEL_LEVEL=$(nproc) cmake --build build -- test
-          
+        run: CTEST_PARALLEL_LEVEL=$(nproc) cmake --build build -- test
```
stephenswat, line 48: Minor thing, but can we get the EOF newline back?

stephenswat (changes requested):
Looks good overall, just have a few questions. Once those are resolved I am happy to merge.

stephenswat (approved):

stephenswat (approved):

PR #80, opened 2021-08-26
Reject warnings in CI builds

### LINKED ISSUE #64: Warning-rejecting builds for the CI system
This is a continuation of the discussion in #63 (@paulgessinger, @krasznaa). We want to add a CI build which has `-Werror` enabled so we can reject code that generates compiler warnings. I see three options for doing this:

1. Add the `-Werror` flag to `Debug` builds on a CMake level.
2. Add the `-Werror` flag to the CI system via the environment variables.
3. Add a new `CI` build type to CMake, which has `-Werror`.

Thoughts?

stephenswat:
As we have discussed in #64, we want to reject warnings in CI builds in order to prevent potential code smells from creeping into our code base. This commit enables the -Werror flag on all CI builds, and adds some CMake logic to ensure that the dependencies can still be built correctly.

I noticed that my compiler was throwing a few warnings when building the Acts core, and those warnings became errors too. Obviously that's not ideal, so I added some CMake hacks to disable the warning in the local scope for both ways we build dependencies. I'd love another pair of eyes on this to make sure I am not doing something stupid there.

Resolves #64.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -35,15 +35,16 @@ jobs:
         run: >
           if ${{ matrix.platform.name == 'CPU'}}; then
             cmake -B build -S . \
-              -DCMAKE_BUILD_TYPE=${{ matrix.build }} 
+              -DCMAKE_BUILD_TYPE=${{ matrix.build }} \
+              -DCMAKE_CXX_FLAGS="-Werror"
           elif ${{ matrix.platform.name == 'CUDA'}}; then
             cmake -B build -S . \
               -DCMAKE_BUILD_TYPE=${{ matrix.build }} \
               -DCMAKE_CUDA_COMPILER=$(which nvcc) \
-              -DTRACCC_BUILD_CUDA=ON 
+              -DTRACCC_BUILD_CUDA=ON \
+              -DCMAKE_CXX_FLAGS="-Werror"
           fi
       - name: Build
         run: cmake --build build -- -j$(nproc)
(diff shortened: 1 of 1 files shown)


stephenswat:
So at this moment the build is failing because of an warning (promoted to an error) in one of our dependencies. Perhaps @krasznaa has a suggestion on how we can make this work, while also being able to compile our dependencies?

paulgessinger:
We could make it a flag and only set `-Werror` on the targets we control based on it?

krasznaa:
I've been thinking about a good way for silencing that @#% Vc warning for weeks now. :frowning: I think we'll need to shove a flag like `-Wno-old-style-cast` down the throat of Vc. The problem is that they explicitly turn on that flag. :scream:

https://github.com/VcDevel/Vc/blob/1.4/cmake/VcMacros.cmake#L248

So we'll need to make sure that "our flag" would win out...

They add it just for Debug builds. But there is no cache variable setting that could be used to turn it off "on the CMake level" for a Debug build. So I also considered forcing a Release build for Vc all the time. Even when the parent project is built in Debug mode.

All in all, this will have to be fixed in [algebra-plugins](https://github.com/acts-project/algebra-plugins).

https://github.com/acts-project/algebra-plugins/tree/main/extern/vc

I fear for now we'll have to delay this PR still a little longer. :frowning:

stephenswat:
That's fine, it is not particularly high priority.

PR #81, opened 2021-08-31
Changing parameter pack container constructor into explicit parameter ones

beomki-yeo:
Current traccc cuda seedfinding is compilable but doesn't run since the parameter pack container constructor causes an ambiguity with other constructors. I struggled to find an exact reason but still don't understand why it happens. :(((

I just removed parameter pack container constructor and replaced it with explicit ones, which makes cuda codes run normally.

@stephenswat  @paulgessinger Can you guys review?


stephenswat (approved):
Looks good to me! :smile:

PR #82, opened 2021-09-01
use `unsigned int` instead of `size_t` for seed finding 

beomki-yeo:
`size_t` in cuda device code is pretty expensive, and I don't think it is necessary to use `size_t` because the number of spacepoints/doublet/triplet or bin size never exceeds maximum values of `unsigned int`

There are meaningful difference in speedup when we use `unsigned int`


krasznaa:
I didn't know that `size_t` would affect performance so much if you didn't want to use atomic operations on it. :confused: Because yes, `unsigned int` is definitely a much safer type if you want to use any CUDA-provided functions.

There is a bit of "controversy" around this in vecmem as well. (https://github.com/acts-project/vecmem/issues/96) So in general I'm on your side, `unsigned int` is generally a good choice in device code. I just didn't know that in these cases there would be an actual performance difference. But I'll trust you on that one...

stephenswat:
CUDA devices don't have any 64-bit integer hardware as far as I know, so they emulate it using their 32-bit integer silicon. The amount of time that takes is very dependent on what operations you are trying to run. Addition and subtraction are trivial (just perform two 32-bit operations and carry), multiplication is a little more expensive, stuff like division can be quite pricey.

So it depends on your workload how much slower it is to use 64-bit integers compared to 32-bit ones. I would normally say that for the usual floating point-heavy workloads it doesn't matter at all, but for irregular workloads like seedfinding it could feasibly make a difference.

beomki-yeo:
I think `size_t` is slower than `unsigned int` when it comes to global memory writing due to its larger size.
As you can see, the changes I made in this PR are only for the indices, which are written into global memory during the seed finding. and it is also true that global memory writing time is significant in cuda seed finding. I don't think they affect the computation meaningfully.

stephenswat:
Are you sure? Every change I see in this PR looks like it would affect the register file, not the global memory. But of course I am not intimately familiar with your seeding implementation. So if you say this is faster then I'll take your word for it. :slightly_smiling_face:

beomki-yeo:
`sp_location` with indices (which is member of doublet and triplet) is written into global memory, and that's why

cgleggett:
interesting question - what determines the size of `size_t` on a cuda device? What happens if the host is 32-bit? Is it the size of size_t on the host that sets it, or is it something defined by the architecture of the device itself?

krasznaa:
> interesting question - what determines the size of `size_t` on a cuda device? What happens if the host is 32-bit? Is it the size of size_t on the host that sets it, or is it something defined by the architecture of the device itself?

It is `nvcc` that needs to decide what `std::size_t` is. (Though it does need to be compatible with the host compiler along the way...) I think CUDA stopped supporting 32-bit hosts a while ago. (Didn't it...?) All in all, I'm pretty sure that `std::size_t` is 64-bit-wide in any reasonable situations these days.

stephenswat:
I'm pretty sure that nvcc still supports 32-bit architectures (both on the host side and the device side). As far as I know both compilers will import their own implementation of the standard library headers, and then decide on the size depending on whether the compilation is 32-bit or 64-bit. I strongly suspect that this is decided by the host side, although that allows you to come up with some very esoteric scenarios where your host is 32-bit and your device is 64-bit. Since nvcc is a compiler driver and not a compiler itself it may in that case decide to use 64-bit integers for its host code and its device code, while the actual host compiler might default to 32-bits.

In reality 64-bit computing is so ubiquitous now that I don't think this is ever a problem, though.

krasznaa:
I'm 99.99% sure that 32-bit OS support from CUDA was dropped a long-long time ago...

Screen Shot 2021-09-02 at 5 42 55 PM

I'd have to double-check, but probably around CUDA 6 or 7...

core/include/edm/container.hpp
```diff
@@ -162,7 +162,7 @@ class container {
     /**
      * @brief Default Constructor
      */
-    TRACCC_HOST container() = default;
+    container() = default;
```
stephenswat, line 165: This looks like an accidentally included change, or is this intended?

stephenswat (commented):
As interesting as this discussion has been I think we're almost good to go ahead, I have one quick question about one of the container changes. Once that's been solved I am happy to merge.

core/include/edm/container.hpp
krasznaa: I don't know if it was accidental or not, but it's definitely the right way to go! Not for CUDA, but rather for SYCL.

In vecmem I had to remove all host/device specifiers from the default constructors of the "data types" to make them work with oneAPI/SYCL.
  - https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/data/vector_view.hpp#L49
  - https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/data/jagged_vector_view.hpp#L56

It does mean that one can't use the default constructor of such types in device code, but the default constructor of `container` was not available in device code to begin with anyway. But this way at least it may be possible to pass such objects to a SYCL kernel.

Though most probably this was **not** Beomki's thinking behind this change...

beomki-yeo: I removed the qualifier since there were warnings from it:
```
/home/beomki/projects/traccc/traccc/core/include/edm/container.hpp(165): warning: __host__ annotation is ignored on a function("container") that is explicitly defaulted on its first declaration
```
I didn't think that far ahead as atilla explained - but anyway it is nice if it can help SYCL

stephenswat: Great, in that case we can just leave it! :smile:

stephenswat (approved):

PR #83, opened 2021-09-03
Set c++17 cuda standard

beomki-yeo:
This tiny PR forces to use c++17 for cuda .

I want to use some c++17 features (e.g. std::optional) for track parameter estimation which I am working on.
Due to the reason stated in [mattermost link](https://mattermost.web.cern.ch/acts/pl/jbdtcby6kbgexcxixngjrmm39h), the default vecmem c++ cuda standard is 14, but I think it's okay to use c++17 because we are not using DPC++ AFAIK.

Anyway we may need this because acts highly relies on c++17...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,7 @@ endif()
     
 # Set GPU-related options
 if( TRACCC_BUILD_CUDA )
+  set( CMAKE_CUDA_STANDARD 17 CACHE STRING "The (CUDA) C++ standard to use" )
   add_subdirectory( device/cuda )
 endif()
 


beomki-yeo:
@krasznaa How do you think?

krasznaa:
> @krasznaa How do you think?

Sorry for the delay... On top of what we are discussing on Mattermost, I think I would rather have this setting in [cmake/traccc-compiler-options.cmake](cmake/traccc-compiler-options.cmake) instead. Just to collect all similar settings into a single place. (Note that setting `CUDA_CXX_STANDARD`, with CUDA not actually used, is **not** a problem. It's a bit overdoing things, but it's not a problem per se.)

In the end the important thing is to set this cache variable as early as possible. Since [vecmem](https://github.com/acts-project/vecmem) has its own settings for the C\+\+ standards.
  - https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-compiler-options-cpp.cmake
  - https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-compiler-options-cuda.cmake
  - https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-compiler-options-hip.cmake
  - https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-compiler-options-sycl.cmake

And any other external project may have such settings too. The current setup **is** correct, but still, explicitly putting these settings in this project as early as possible, is definitely a good idea...

stephenswat:
Perhaps it's worth using `std::experimental::optional` instead? You could typedef using a mechanism similar to the one we use to include PMR types into vecmem.

stephenswat (approved):
Alright, I've chatted with Attila a little about this and we've decided to merge this in as an emergency fix because right now the main branch is, for all intents and purposes, broken. We'll look at how to improve the structure of the CMake build in a later iteration.

beomki-yeo:
Thanks. yeah I feel I have messed up CMakeLists.txt with the last several PRs. Need to clean it out at some point :p

stephenswat:
Haha it's fine you've done a great job with it. :+1:

PR #84, opened 2021-09-06
Require CMake 3.18 or higher for builds with CUDA

stephenswat:
As it turns out CMake versions 3.17 and lower are not able to build CUDA correctly with the C++17 standard. To ensure that builds work correctly, I am adding a check to ensure that the minimum version is 3.18, which doe support C++17.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,8 @@ endif()
     
 # Set GPU-related options
 if( TRACCC_BUILD_CUDA )
+  # TODO: Tuck all of this away neatly in a CUDA configuration file.
+  cmake_minimum_required(VERSION 3.18)
   set( CMAKE_CUDA_STANDARD 17 CACHE STRING "The (CUDA) C++ standard to use" )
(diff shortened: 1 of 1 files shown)


krasznaa (commented):
Having multiple `cmake_minimum_required(...)` statements in the same file will work, I checked. So technically this works.

I'm just not sure why we'd want to merge in yet another messy piece of code. Adding this line here doesn't fix anything. It just makes the error more clear with the current Docker image.

I would argue that we need to fix the Docker image with a newer CMake version, which I understood you would be taking care of. (If not, I can do it myself.) That shall fix the CI in #78. We can include this CMake 3.18 requirement in a larger "cleanup PR", which shall make a lot of things nicer all at once.

stephenswat:
The new image is already being built by the machines CI as we speak: acts-project/machines#38

Adding this right now protects people running this on their own machine if they have an older version of CMake, but for sure it doesn't matter if we are planning a clean-up of the CMake code soon anyway.

Closing!

PR #85, opened 2021-09-07
Track parameter estimation for cpu and cuda

beomki-yeo:
[traccc_pr_75](https://github.com/acts-project/traccc/pull/75) should be merged first before this PR

This is for track parameter estimation from found seeds.
The algorithm is almost copy-and-paste of [EstimateTrackParamsFromSeed](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp) in Acts - a different thing is that the local position on surface is obtained from `measurement` directly while the acts one obtains it with global-to-local transformation and surface object.

There are a couple of features and issues that people might be interested

1. The size of eigen-based class/structure is measurement differently by cpu and cuda compiler which makes data transfer unstable and wrong. So the data size is synchronized by adding a data align qualifier (see `core/include/definitions/qualifiers.hpp`)
2. I initially put `std::optional< measurement >` member in spacepoint to refer it when I obtain the local track parameter, but it turned out that `std::optional` doesn't work in cuda device code - or anyone knows how to make it work??
3. The algebras used in track parameter estimation utilize algebra-plugin. But it seems it doesnt work with array plugin since the functions in algebra-plugin do not have __host__ __device__ qualifiers

Meanwhile I have observed that seed finding doesn't work with eigen_plugin and double scalar type ... I will look into that and make another PR to resolve it

**TO-DO list:**
1.  measure the time for cpu and cuda
2.  optimization

### Update on Sep 22
This can be merged after #92


core/include/edm/collection.hpp
```diff
@@ -0,0 +1,180 @@
...
+template <typename item_t>
+struct collection_data {
+    vecmem::data::vector_view<item_t> items;
+};
+
+/// Structure holding (all of the) data about the collection in host code
+template <typename item_t>
+struct collection_buffer {
+    vecmem::data::vector_buffer<item_t> items;
+};
```
krasznaa, line 147: I should have looked carefully at #75, sorry for not doing that. :frowning:

I don't see the point in this class... :confused: What does `collection` give us over:

```c++
template <typename item_t>
using host_collection = vecmem::vector<item_t>;

template <typename item_t>
using device_collection = vecmem::device_vector<item_t>;
```

? It literally just wraps an existing vector type, without adding any new feature to it (as far as I can tell...).

The same is true for the `collection_data` and `collection_buffer` types. What sort of abstraction do these give to the traccc code over using the vecmem types directly?

core/include/seeding/track_params_estimation_helper.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+// acts include
+#include "Acts/Definitions/Units.hpp"
+
+namespace traccc {
+
+// helper function used for both cpu and gpu
+struct track_params_estimation_helper {
+
+    static inline TRACCC_HOST_DEVICE bound_vector seed_to_bound_vector(
```
krasznaa, line 22: In this case I would've simply gone for a namespace, and a "simple" function inside of that namespace. Using `static` seems to be misused a bit in this case...

device/cuda/include/cuda/seeding/seed_finding.hpp
```diff
@@ -55,7 +55,7 @@ struct seed_finding {
     /// Callable operator for the cuda seed finding
     ///
     /// @return seed_container is the vecmem seed container
-    host_seed_container operator()(
+    host_seed_container& operator()(
```
krasznaa, line 58: Wait... what? Are you returning a permanent object all of a sudden? This does not seem thread-safe at all... :thinking:

device/cuda/include/cuda/seeding/track_params_estimation.hpp
```diff
@@ -0,0 +1,41 @@
...
+class track_params_estimation {
+    public:
+    using input_type =
+        std::pair<host_measurement_container&, host_seed_container&>;
+    using output_type = host_bound_track_parameters_collection;
+
+    track_params_estimation(vecmem::memory_resource* mr)
+        : m_mr(mr), params(mr) {}
+
+    output_type& operator()(input_type& i) {
```
krasznaa, line 25: Again, returning a non-const reference to a member object seems terribly thread-unsafe. I would've just opted for copy elision on return in this function...

examples/algorithms/cuda/include/cuda/track_finding/seeding_algorithm.hpp
```diff
@@ -22,7 +22,9 @@ class seeding_algorithm {
     using output_type =
         std::pair<host_internal_spacepoint_container, host_seed_container>;
 
-    seeding_algorithm(vecmem::memory_resource* mr = nullptr) : m_mr(mr) {
+    seeding_algorithm(vecmem::memory_resource* mr)
```
krasznaa, line 25: I guess this is not the only place where this pattern shows up, but this is where I'll comment about it...

We should only pass the memory resources around by pointer if it is allowed to pass a null pointer. In this case it's not really allowed anymore. So it would make for a better interface (in my opinion) to receive the memory resource as a reference. (`vecmem::memory_resource&`)

It is indeed annoying that the std classes receive memory resources through pointers. Since they do allow the pointers to be invalid. (In which case they fall back on a default resource.) But with vecmem in most cases we just need to pass explicitly created memory resources around. (Otherwise compilation on macOS can fail. As the Apple Clang compiler doesn't handle "default memory resources" quite correctly. Which provides a pretty good debugging tool for vecmem in the end. :stuck_out_tongue:)

examples/algorithms/cuda/include/cuda/track_finding/track_params_estimation_algorithm.hpp
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    output_type& operator()(input_type& i) {
+        params = tp->operator()(i);
+        return params;
+    }
+
+    private:
+    output_type params;
+    std::shared_ptr<traccc::cuda::track_params_estimation> tp;
```
krasznaa, line 35: Is it really necessary to use a shared pointer here? Remember, these things are not cheap.

krasznaa (commented):
I already made a lot of comments, I would stop here for now. I think we'd need to sort out what we want to do with the container classes first. As right now I just don't think that they add any value over using the underlying vecmem types directly.

core/include/edm/collection.hpp
beomki-yeo: No advantage really. I just wanted to keep the same behavior of container class for vector-based class as well. Of course it is not mandatory thing so If you and @stephenswat  don't like it, I can remove this class and close PR75

stephenswat: In my opinion, since the work is already there, I don't see any reason to remove it. If it works, it works. I might not completely understand why the code is there exactly but you're the one writing this algorithm so if your judgement is that this is helpful then I am personally happy to go along with it.

core/include/seeding/track_params_estimation_helper.hpp
beomki-yeo: ok makes sense

core/include/edm/collection.hpp
krasznaa: In this case I fear that this design unnecessarily complicates things. As this intermediate layer between the vecmem types and the traccc algorithmic code does not add any features as far as I can see.

The `traccc::container` type is quite different in this respect. That does add a non-trivial amount of logic on top of the vecmem types, so that one I really like... I just feel that this type just makes the code a little harder to read... 🤔

examples/algorithms/cuda/include/cuda/track_finding/seeding_algorithm.hpp
beomki-yeo: Okay!

examples/algorithms/cuda/include/cuda/track_finding/track_params_estimation_algorithm.hpp
stephenswat: I don't think this is necessarily a problem here; this is a pointer to a function object, so it is unlikely to be replaced very often. Since `std::shared_ptr` has zero de-referencing overhead, I don't foresee any performance degradation here.

Whether it is really necessary is debatable, but I can see some benefit to this approach given that function objects in C++ are often assumed to be copyable. Using shared pointers here can feasibly reduce the overhead of that copying. But perhaps Beomki had different reasons in mind?

core/include/edm/collection.hpp
beomki-yeo: I agree. It apparently looks better after removing collection class. I will close PR75

device/cuda/include/cuda/seeding/seed_finding.hpp
beomki-yeo: The reason for this change was to use the return output of certain algorithm as an input of next algorithm directly. With copy method, the memory resource is missing so we cannot use it directly for the next algorithm.
```
auto output = cuda_algorithm(input) // output cannot be used as an input for the next algorithm bec
``` 

But as you mentioned that it is not thread-safe, I reverted it to copy method and do some modification:
```
algorithm::output_type output(mr);
output = cuda_algorithm(input)
```
Is it okay to you?

device/cuda/include/cuda/seeding/track_params_estimation.hpp
beomki-yeo: answered above

examples/algorithms/cuda/include/cuda/track_finding/track_params_estimation_algorithm.hpp
beomki-yeo: @stephenswat is right - This algorithm objects are generated only once before the actual algorithm runs. It could also be normal objects but I was too lazy to make a copy operator for algorithm classes. And if it doesn't introduce any overheads, we can just keep this way.

examples/algorithms/cuda/include/cuda/track_finding/seeding_algorithm.hpp
beomki-yeo: Hey wait. If the input is reference, it's not polymorphic anymore for different `memory_resource`.

beomki-yeo: Do you just want it to be `(vecmem::cuda::memory_resource& mr)` for cuda algorithm and `(vecmem::sycl::memory_resource& mr)` for sycl algorithm?

beomki-yeo:
I update the algebra_plugins to the latest commit: the rotate function of eigen algebra is missing in the current version

beomki-yeo:
Current PR compiles, but there is a runtime error in cuda track parameter estimation.
It must work when [algebra_plugin_PR17](https://github.com/acts-project/algebra-plugins/pull/17) is merged, which adds inline to eigen functions

beomki-yeo:
Naive estimation shows that the averaged track parameter estimation time is:
cpu:  ~ 0.0009 sec
cuda: ~ 0.0012 sec

So cuda is slightly slower than cpu unfortunately. It's not a serious problem since the cuda seed finding time is usually 0.05 sec.
But let me profile and optimize it a bit for better performance.

examples/algorithms/cuda/include/cuda/track_finding/seeding_algorithm.hpp
stephenswat: References in C++ are the same in terms of polymorpism to pointers, so this should not be a problem. :slightly_smiling_face:

stephenswat:
If we're talking about 1.2 milliseconds here we're basically teetering on the border of what is achievable with memory transfer overhead and launch delays. Just following Amdahl's law I wouldn't necessarily recommend spending too much time optimizing this part of the code. Great job! :smile:

beomki-yeo:
@stephenswat OK. I will checkout whether the results are the same with Acts track parameter estimation before this PR goes in. Will let you know once it is done.

examples/algorithms/cuda/include/cuda/track_finding/seeding_algorithm.hpp
beomki-yeo: OK thanks. The changes are applied

beomki-yeo:
I added a test file `tests/cpu/algorithms/compare_with_acts_seeding.cpp`, which checks the results between acts and traccc seeding algorithm. Since I needed a hit file data so opened a [traccc-data_MR1](https://gitlab.cern.ch/beyeo/traccc-data/-/merge_requests/1) in traccc-data repository.

@asalzburger or anyone who has write access to traccc-data can review the MR?

beomki-yeo:
Other than than [traccc-MR](https://gitlab.cern.ch/beyeo/traccc-data/-/merge_requests/1), I also need [algebra-plugin_PR17](https://github.com/acts-project/algebra-plugins/pull/17) to be merged to make this PR work (@niermann999 could you have a look??)

beomki-yeo:
Are there more comments on this?

examples/algorithms/cuda/include/cuda/track_finding/track_params_estimation_algorithm.hpp
krasznaa: Just as for `traccc::track_params_estimation_algorithm`, I don't see what's the purpose of this class. :confused: What interface does it add on top of `traccc::cuda::track_params_estimation`?

Also, why does it not inherit from the `traccc::algorithm` base class? :confused:

examples/run/cpu/seq_example.cpp
```diff
@@ -28,46 +29,69 @@ int seq_run(const std::string& detector_file, const std::string& cells_dir,
...
+        /*----------------------------
+          Track params estimation
+          ----------------------------*/
+
+        traccc::track_params_estimation_algorithm::input_type tp_input(seeds);
+        traccc::track_params_estimation_algorithm::output_type tp_output;
+
+        tp_output = tp(tp_input);
+
+        auto& params = tp_output;
```
krasznaa, line 83: With the updated algorithms semantics this could surely be simplified quite a bit. Most of these lines should not be needed anymore.

tests/cpu/algorithms/compare_with_acts_seeding.cpp
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// Google Test can be run manually from the main() function
+// or, it can be linked to the gtest_main library for an already
+// set-up main() function primed to accept Google Test test cases.
+int main(int argc, char** argv) {
+    ::testing::InitGoogleTest(&argc, argv);
+
+    return RUN_ALL_TESTS();
+}
```
krasznaa, line 349: The test should rather be linked against `GTest::gtest_main` than to implement the `main()` function for ourselves.

krasznaa (commented):
I'm still very much in favour of this development, but I think the code could be simplified a bit in a number of places...

tests/cpu/algorithms/compare_with_acts_seeding.cpp
beomki-yeo: Okay

device/cuda/include/cuda/seeding/seed_finding.hpp
```diff
@@ -96,56 +95,52 @@ struct seed_finding {
...
     }
 
     private:
-    bool first_alloc;
     const seedfinder_config m_seedfinder_config;
     const seedfilter_config m_seedfilter_config;
     multiplet_estimator m_estimator;
     seed_filtering m_seed_filtering;
 
+    vecmem::memory_resource& m_mr;
```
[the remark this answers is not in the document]
beomki-yeo: Okay. definitely reference_wrapper would be better

device/cuda/include/cuda/seeding/track_params_estimating.cuh
```diff
@@ -0,0 +1,26 @@
+/** TRACCC library, part of the ACTS project (R&D line)
```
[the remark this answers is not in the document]
beomki-yeo: I see if that's a usual convention. I modified all .cuh into .hpp

device/cuda/include/cuda/seeding/track_params_estimation.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#pragma once
+
+#include <cuda/seeding/track_params_estimating.cuh>
+
+namespace traccc {
+namespace cuda {
+
+/// track parameter estimation for cuda
+class track_params_estimation {
```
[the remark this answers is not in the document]
beomki-yeo: Right. I removed `track_params_estimating` since it is confusing as you mentioned. Instead, I directly implement operator() in `track_params_estimation.cu`. Now it will look much better.

For the algorithm part, I will explain in the below as well; I removed the `track_params_estimation_algorithm` as it seems not needed for the moment. The initial motivation of algorithm one is for setting the configuration parameter rather than touching the original track_parameter_estimation in core directory. but It is apparent that we don't need the algorithm version because there is not any configuration parameter for track parameter estimation for the moment.

device/cuda/include/cuda/seeding/track_params_estimating.cuh
```diff
@@ -0,0 +1,26 @@
...
+
+/// Forward declaration of track parameter estimating fuction
+/// The bound track parameters at the bottom spacepoints are obtained
+///
+/// @param seeds seeds found by seed finding
+/// @param params vector of bound track parameters at the bottom spacepoints
+/// @param resource vecmem memory resource
+void track_params_estimating(host_seed_container& seeds,
+                             host_bound_track_parameters_collection& params,
+                             vecmem::memory_resource* resource);
```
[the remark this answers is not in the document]
beomki-yeo: Right. Anyway this file is removed

examples/algorithms/cpu/include/track_finding/track_params_estimation_algorithm.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include "seeding/track_params_estimation.hpp"
+
+namespace traccc {
+
+class track_params_estimation_algorithm
```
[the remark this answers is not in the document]
beomki-yeo: As I explained above, these are removed now

examples/run/cpu/seq_example.cpp
beomki-yeo: Done

beomki-yeo:
I still didn't edit `spacepoint_grouping` for the updated algorithm semantics but we can leave it as it is. Because I am going to remove it and rewrite with `detray grid` once acts-project/detray#96 is merged

tests/cpu/algorithms/compare_with_acts_seeding.cpp
beomki-yeo: I simply removed `int main(..)` since it was already linked to gtest. Not sure if it is what you meant though

beomki-yeo:
Rebased with main branch. 
The seq_example now runs for both eigen and array plugin.
However, This PR does not generate the execution files for different plugins with a single compilation - need to change cmake options to get execution file for different plugin. (Default option prioritizes the eigen plugin)

Event though it is not very desirable, let's make this happen in future PR... :)

pinging @krasznaa @stephenswat 

p.s. overall run time does not change a lot between array and eigen plugin

core/include/seeding/track_params_estimation_helper.hpp
```diff
@@ -0,0 +1,135 @@
...
+    // local2[0], local2[1], local2[2]);
+
+    // Lambda to transform the coordinates to the (u, v) space
+    auto uvTransform = [](const vector3& local) -> vector2 {
+        vector2 uv;
+        scalar denominator = local[0] * local[0] + local[1] * local[1];
+        uv[0] = local[0] / denominator;
+        uv[1] = local[1] / denominator;
+        return uv;
+    };
```
stephenswat, line 65: This function doesn't actually act as a closure - maybe we can just define it as a free function? Or maybe move it the algebra plugins?

device/cuda/src/seeding/track_params_estimation.cu
```diff
@@ -0,0 +1,80 @@
...
+    auto seeds_view = get_data(seeds, &m_mr.get());
+    auto params_view = vecmem::get_data(params);
+
+    // -- Num threads
+    // The dimension of block is the integer multiple of WARP_SIZE (=32)
+    unsigned int num_threads = WARP_SIZE * 2;
+
+    // -- Num blocks
+    // The dimension of grid is number_of_seeds / num_threads + 1
+    unsigned int num_blocks = seeds.get_headers()[0] / num_threads + 1;
```
stephenswat, line 39: Very minor, but this has a chance of creating grids that are larger than necessary. Perhaps try:

```suggestion
    unsigned int num_blocks = (seeds.get_headers()[0] + num_threads - 1) / num_threads;
```

device/cuda/src/seeding/track_params_estimation.cu
```diff
@@ -0,0 +1,80 @@
...
+    seed_container_view seeds_view,
+    vecmem::data::vector_view<bound_track_parameters> params_view) {
+
+    // Get device container for input parameters
+    device_seed_container seeds_device({seeds_view.headers, seeds_view.items});
+    device_bound_track_parameters_collection params_device({params_view});
+
+    // vector index for threads
+    unsigned int gid = threadIdx.x + blockIdx.x * blockDim.x;
+    const auto& n_seeds = seeds_device.get_headers()[0];
```
stephenswat, line 62: Is it intentional that you're always grabbing the first element here? Does the rest of the code guarantee that only this element is interesting?

examples/algorithms/cuda/include/cuda/track_finding/seeding_algorithm.hpp
```diff
@@ -16,13 +16,13 @@
...
 
-class seeding_algorithm {
-    public:
-    using input_type = host_spacepoint_container&;
-    using output_type =
-        std::pair<host_internal_spacepoint_container, host_seed_container>;
+class seeding_algorithm
+    : public algorithm<
+          std::pair<host_internal_spacepoint_container, host_seed_container>(
+              host_spacepoint_container&&)> {
```
stephenswat, line 22: Are these changes related to the track parameter seeding in some way or have they snuck into this PR?

examples/run/cuda/seq_example_cuda.cpp
```diff
@@ -98,52 +98,77 @@ int seq_run(const std::string& detector_file, const std::string& cells_dir,
...
-        n_measurements += measurements_per_event.total_size();
-        n_spacepoints += spacepoints_per_event.total_size();
-
         /*----------------------------
-          Seeding algorithm
+             Seeding algorithm
           ----------------------------*/
 
-        // cuda
+        /// CUDA
```
stephenswat, line 105: Using triple slashes here could potentially confuse a documentation generator.

stephenswat (commented):
Lots of good work here overall! It feels like there is a lot of unrelated work in this PR though - is that here for a reason or did it get included by accident? I also have some other comments, but otherwise I think this should be close to ready. :smile:

core/include/clusterization/measurement_creation.hpp
```diff
@@ -98,7 +98,7 @@ struct measurement_creation
                 // normalize the cell position
                 m.local = mean;
                 // normalize the variance
-                m.variance = var / totalWeight;
+                m.variance = 1 / totalWeight * var;
```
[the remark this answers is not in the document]
beomki-yeo: It is just becuase operator / is not defined for array in algebra-plugins

device/cuda/include/cuda/seeding/seed_finding.hpp
```diff
@@ -10,15 +10,16 @@
...
-#include <cuda/seeding/seed_selecting.cuh>
-#include <cuda/seeding/triplet_counting.cuh>
-#include <cuda/seeding/triplet_finding.cuh>
-#include <cuda/seeding/weight_updating.cuh>
+#include <cuda/seeding/doublet_counting.hpp>
+#include <cuda/seeding/doublet_finding.hpp>
+#include <cuda/seeding/seed_selecting.hpp>
+#include <cuda/seeding/triplet_counting.hpp>
+#include <cuda/seeding/triplet_finding.hpp>
+#include <cuda/seeding/weight_updating.hpp>
```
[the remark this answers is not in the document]
beomki-yeo: I thought so. But they don't have any cuda related code inside .cuh so decided to use .hpp extension for them.

device/cuda/include/cuda/seeding/seed_finding.hpp
```diff
@@ -36,28 +39,33 @@ struct seed_finding {
...
+          triplet_counter_container(m_sp_grid->size(false), &m_mr.get()),
+          triplet_container(m_sp_grid->size(false), &m_mr.get()),
+          seed_container(1, &m_mr.get()) {}
+
+    /// Callable operator for the seed finding
+    ///
+    /// @return seed_collection is the vector of seeds per event
+    output_type operator()(
+        host_internal_spacepoint_container&& i) const override {
+        std::lock_guard<std::mutex> lock(*mutex.get());
```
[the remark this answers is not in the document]
beomki-yeo: I also didn't want to make this change. The only way to remove `mutable` internal members whould be taking them as external parameters - but this is also coarse-grained.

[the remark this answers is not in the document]
beomki-yeo: And I don't think we have to achieve more parallelization in cuda `operator()`, which is grid bin level.

device/cuda/include/cuda/seeding/seed_finding.hpp
```diff
@@ -128,24 +134,24 @@ struct seed_finding {
...
+    std::reference_wrapper<vecmem::memory_resource> m_mr;
+
+    // mutable internal objects for multiplets
+    mutable std::shared_ptr<std::mutex> mutex{std::make_shared<std::mutex>()};
+    mutable host_doublet_counter_container doublet_counter_container;
+    mutable host_doublet_container mid_bot_container;
+    mutable host_doublet_container mid_top_container;
+    mutable host_triplet_counter_container triplet_counter_container;
+    mutable host_triplet_container triplet_container;
+    mutable host_seed_container seed_container;
```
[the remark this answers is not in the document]
beomki-yeo: I know.
Without these, I have to create host_doublet/triplet_container in operator() function for every event which makes entire algorithm much slower (it's almost 50% slowdown). Do you have a better idea on this?

device/cuda/include/cuda/seeding/track_params_estimation.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+    /// Callable operator for track_params_esitmation
+    ///
+    /// @param input_type is the seed container
+    ///
+    /// @return vector of bound track parameters
+    output_type operator()(host_seed_container&& seeds) const override;
+
+    private:
+    std::reference_wrapper<vecmem::memory_resource> m_mr;
```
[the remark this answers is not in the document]
beomki-yeo: It's something I don't understand well. what is normal wrapper?

device/cuda/src/seeding/track_params_estimation.cu
beomki-yeo: It doesn't have any problem. seed container has only one element (so it is just vector of seed). The reason that I used container is that I needed to save the number of seeds in header.

But I will make a change in seed finding  later so we don't use container but a single vecmem vector. Possibly need to add seed_counting algorithm to make it happen

examples/algorithms/cuda/include/cuda/track_finding/seeding_algorithm.hpp
beomki-yeo: They snuck.

core/include/clusterization/measurement_creation.hpp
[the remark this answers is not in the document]
beomki-yeo: Will mark with TODO for later change

core/include/seeding/track_params_estimation.hpp
```diff
@@ -0,0 +1,65 @@
...
+    ///
+    /// void interface
+    ///
+    /// @return vector of bound track parameter
+    void operator()(const host_seed_container& i, output_type& o) const {
+        // input for seed vector
+        const auto& seeds = i.get_items()[0];
+
+        // convenient assumption on bfield and mass
+        vector3 bfield = {0, 0, 2};
```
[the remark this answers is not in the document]
beomki-yeo: Got it

examples/run/cuda/seq_example_cuda.cpp
beomki-yeo: Will fix

device/cuda/include/cuda/seeding/weight_updating.hpp
```diff
@@ -28,7 +28,7 @@ void weight_updating(const seedfilter_config& filter_config,
                      host_internal_spacepoint_container& internal_sp_container,
                      host_triplet_counter_container& triplet_counter_container,
                      host_triplet_container& triplet_container,
-                     vecmem::memory_resource* resource);
+                     vecmem::memory_resource& resource);
```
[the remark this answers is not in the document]
beomki-yeo: Not really. Yeah sorry for unrelated change in this PR

device/cuda/include/cuda/seeding/seed_finding.hpp
```diff
@@ -36,28 +39,33 @@ struct seed_finding {
...
     }
 
     /// Callable operator for the cuda seed finding
     ///
     /// @return seed_container is the vecmem seed container
-    host_seed_container operator()(
-        host_internal_spacepoint_container& isp_container) {
-
+    void operator()(host_internal_spacepoint_container& isp_container,
+                    output_type& seed_container) const {
```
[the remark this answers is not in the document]
beomki-yeo: No. Will fix it

device/cuda/include/cuda/seeding/seed_finding.hpp
```diff
@@ -128,24 +134,24 @@ struct seed_finding {
...
-
-        return seed_container;
     }
 
     private:
-    bool first_alloc;
     const seedfinder_config m_seedfinder_config;
     const seedfilter_config m_seedfilter_config;
     multiplet_estimator m_estimator;
+    std::shared_ptr<spacepoint_grid> m_sp_grid;
```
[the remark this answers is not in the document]
beomki-yeo: Possibly yes. Anyway `m_sp_grid` will be removed in #95 which is written on top of this PR.

device/cuda/src/seeding/track_params_estimation.cu
beomki-yeo: makes sense. Will fix it

core/include/seeding/track_params_estimation_helper.hpp
beomki-yeo: Hmm I don't recommend moving it into algebra plugin.
I will just make a free function in the same file

beomki-yeo:
> is that here for a reason or did it get included by accident?

No good reason for that... :( I will try to make PR more clear in the future

device/cuda/include/cuda/seeding/seed_finding.hpp
krasznaa: Sure. (Sorry for the late answer... :frowning:)

device/cuda/include/cuda/seeding/seed_finding.hpp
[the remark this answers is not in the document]
krasznaa: This was one of my recommendations to use `.hpp` on headers that can be included by not only `.cu` files, but other C\+\+ source files as well.

Since as I explained before, I like to use `.cuh` as a "contagious" extension. Marking something as `.cuh` should be done only when that particular header has code for launching a kernel, or when it includes some other `.cuh` file. In all other cases we should just use `.hpp`.

beomki-yeo:
Are there more comments on this? Otherwise I want it to get merged in

beomki-yeo:
@stephenswat @krasznaa Are there more issues that I need to address?

core/include/seeding/track_params_estimation.hpp
```diff
@@ -0,0 +1,66 @@
...
+            bound_track_parameters track_params;
+            track_params.vector() =
+                seed_to_bound_vector(seed, bfield, PION_MASS_MEV);
+
+            o.push_back(track_params);
+        }
+    }
+
+    private:
+    std::reference_wrapper<vecmem::memory_resource> m_mr;
```
stephenswat, line 63: Not a huge deal, but I still don't see the need to use reference wrappers here as opposed to normal wrappers.

stephenswat (commented):
Looks fine to me - just two small things I would like to follow up on. When that is done we can check whether Attila or anyone else has any comments, and then I would be in favour of moving to merge. :smile:

core/include/definitions/primitives.hpp
```diff
@@ -35,4 +35,14 @@ using vector3 = __plugin::point3;
...
+// Fixme: Need to utilize algebra plugin for vector and matrix
+template <unsigned int kSize>
+using traccc_vector = Eigen::Matrix<scalar, kSize, 1>;
+
+template <unsigned int kRows, unsigned int kCols>
+using traccc_matrix = Eigen::Matrix<scalar, kRows, kCols>;
+
+template <unsigned int kSize>
+using traccc_sym_matrix = Eigen::Matrix<scalar, kSize, kSize>;
+
```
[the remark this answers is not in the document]
beomki-yeo: I agree that It's not ideal. 
But algebra-plugin doesn't support the general vector or matrix right now (it's only for 4x4 matrix and 3-vector). It should be developed there in the future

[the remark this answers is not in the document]
stephenswat: Ah yes, that makes sense. Thanks. I'll open a quick issue on the algebra-plugins!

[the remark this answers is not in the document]
stephenswat: Never mind you've done so already.

core/include/seeding/track_params_estimation.hpp
beomki-yeo: That's something I am not sure - it is recommended by @krasznaa. I think he can explain us the motivation for this.

stephenswat (approved):
I'll sign off on this for now, but I will leave just a bit of time for others to comment if they want.

core/include/definitions/primitives.hpp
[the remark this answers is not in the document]
beomki-yeo: No I haven't. But there are some folks including @georgi-mania  working on it I guess. you can ping them when you file an issue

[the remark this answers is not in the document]
stephenswat: https://github.com/acts-project/algebra-plugins/issues/40 right?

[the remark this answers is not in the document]
beomki-yeo: Oh there is indeed. Sorry I forgot..

beomki-yeo:
@stephenswat looks like there is no comment anymore. Can you get this in? We can come back to the wrapper thing at any time in the future

beomki-yeo:
it's squashed into one commit

stephenswat (approved):
Let's do it! :+1:

PR #87, opened 2021-09-13
Implement improved algorithm semantics

### LINKED ISSUE #77: `const` input type of `algorithm` class for `non-const` device input
One issue has been raised by [PR46](https://github.com/acts-project/traccc/pull/46#discussion_r685825190) where we can not use `algorithm` class for GPU because non-const vecmem function (`get_data()`) is called for const input data types. It might be very convenient if we just remove `const` qualifier in input data type but not sure if it is the best way for this

stephenswat:
Our current algorithm semantics, defined in #36, are not meeting our requirements. Firstly, they have an unnecessary output parameter method. Secondly, they do not allow us to specify the movement of affine data, of which we have a lot.

This commit changes the existing algorithm semantics in several ways. Firstly, multiple arguments are now more elegantly represented using a template parameter pack instead of a tuple. Secondly, the template is now in a more readable function type syntax. Thirdly, the output parameter version has been removed. Fourthly and lastly, the new algorithm semantics allow us to properly distinguish between different types of data.

Algorithms are now allowed to have affine data types as their arguments (rvalue references in templating), as well as immutable regular types (represented by const lvalue references). These two types should allow us to represent all the types of data we need.

I've also included some handy composition methods which should make it easier to compose smaller algorithms to build larger ones, and it should also make testing our algorithms easier in the long run.

This commit closes #77.

There are also some new tests to make sure these semantics work properly.

### DIFF
--- a/core/include/clusterization/component_connection.hpp
+++ b/core/include/clusterization/component_connection.hpp
@@ -25,12 +25,10 @@ namespace traccc {
 /// implementation internally.
 ///
 class component_connection
-    : public algorithm<
-          std::pair<const host_cell_collection&, const cell_module&>,
-          cluster_collection>,
-      public algorithm<
-          std::pair<const device_cell_collection&, const cell_module&>,
-          cluster_collection> {
+    : public algorithm<cluster_collection(const host_cell_collection&,
+                                          const cell_module&)>,
+      public algorithm<cluster_collection(const device_cell_collection&,
+                                          const cell_module&)> {
 
     public:
     /// @name Operators to use in host code
@@ -46,30 +44,9 @@ class component_connection
     /// c++20 piping interface:
     /// @return a cluster collection
     ///
-    cluster_collection operator()(
-        const std::pair<const host_cell_collection&, const cell_module&>& i)
-        const override {
-        const host_cell_collection& cells = i.first;
-        const cell_module& module = i.second;
+    cluster_collection operator()(const host_cell_collection& cells,
+                                  const cell_module& module) const override {
         return this->operator()<vecmem::vector>(cells, module);
-    }
-
-    /// Callable operator for the connected component, based on one single
-    /// module
-    ///
-    /// @param cells are the input cells into the connected component, they are
-    ///              per module and unordered
-    /// @param module The description of the module that the cells belong to
-    /// @param clusters[in,out] are the output clusters
-    ///
-    /// void interface
-    ///
-    void operator()(
-        const std::pair<const host_cell_collection&, const cell_module&>& i,
-        cluster_collection& clusters) const override {
-        const host_cell_collection& cells = i.first;
-        const cell_module& module = i.second;
-        this->operator()<vecmem::vector>(cells, module, clusters);
     }
 
     /// @}
@@ -89,30 +66,9 @@ class component_connection
     /// c++20 piping interface:
     /// @return a cluster collection
     ///
-    cluster_collection operator()(
-        const std::pair<const device_cell_collection&, const cell_module&>& i)
-        const override {
-        const device_cell_collection& cells = i.first;
-        const cell_module& module = i.second;
+    cluster_collection operator()(const device_cell_collection& cells,
+                                  const cell_module& module) const override {
         return this->operator()<vecmem::device_vector>(cells, module);
-    }
-
-    /// Callable operator for the connected component, based on one single
-    /// module
-    ///
-    /// @param cells are the input cells into the connected component, they are
-    ///              per module and unordered
-    /// @param module The description of the module that the cells belong to
-    /// @param clusters[in,out] are the output clusters
-    ///
-    /// void interface
-    ///
-    void operator()(
-        const std::pair<const device_cell_collection&, const cell_module&>& i,
-        cluster_collection& clusters) const override {
-        const device_cell_collection& cells = i.first;
-        const cell_module& module = i.second;
-        this->operator()<vecmem::device_vector>(cells, module, clusters);
     }
 
     private:
--- a/core/include/utils/algorithm.hpp
+++ b/core/include/utils/algorithm.hpp
@@ -9,6 +9,15 @@
 #pragma once
 
 namespace traccc {
+template <typename T>
+using rvalue_or_const_lvalue = std::disjunction<
+    std::is_rvalue_reference<T>,
+    std::conjunction<std::is_lvalue_reference<T>,
+                     std::is_const<std::remove_reference_t<T>>>>;
+
+template <typename T>
+class algorithm {};
+
 /**
  * @brief Unified algorithm semantics which convert an input to an output.
  *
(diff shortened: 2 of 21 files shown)


stephenswat:
Pinging @beomki-yeo because he might be interested and might want to provide feedback.

beomki-yeo:
Looks good to me. I wish we could define `input_type` in algorithm, but that's not very necessary. Will you apply it to the cuda algorithms as well?

stephenswat:
Sadly there is not really a convenient way to store a parameter pack in C++, so we can't really use the parameter pack in a named way. :sob:

core/include/utils/algorithm.hpp
```diff
@@ -18,35 +27,41 @@ namespace traccc {
...
+}
+
+template <typename A, typename B, typename C>
+std::function<B(A&&)> side_effect(std::function<B(A)> f,
+                                  std::function<void(const C&)> s) {
+    return [=](A&& i) -> B {
+        s(i);
+        return f(std::forward<A>(i));
+    };
+}
```
krasznaa, line 66: Do these two functions need to be part of `algorithm.hpp`? From what I see, they are only used in testing the code, no? Shouldn't they be defined in the test code, in a "hidden" way instead?

core/include/clusterization/component_connection.hpp
```diff
@@ -46,32 +44,11 @@ class component_connection
     /// c++20 piping interface:
     /// @return a cluster collection
     ///
-    cluster_collection operator()(
-        const std::pair<const host_cell_collection&, const cell_module&>& i)
-        const override {
-        const host_cell_collection& cells = i.first;
-        const cell_module& module = i.second;
+    cluster_collection operator()(host_cell_collection&& cells,
```
krasznaa, line 47: 😕 Is it really a good idea to force the user to hand over their collection to the algorithm? What if they would like to run multiple algorithms on the same `host_cell_collection`?

I just don't see why the previous `const host_cell_collection&` argument is not better than moving the object into the operator... Did you want to make sure that the object would be deleted as early as possible? But again, what if the user wants to use that collection after having run the algorithm?

krasznaa (commented):
The code looks nice, I like the `R(A...)` formalism. I'm just afraid of the change in the component connection algorithm... I would prefer to keep the arguments on that algorithm the same as they were.

core/include/clusterization/component_connection.hpp
stephenswat: I think this is a fair point; it is true that this kind of formalism can potentially get in the way if you're thinking about data re-use, but I would argue that traccc is a great place to try to incorporate affine typing (move semantics-ish) into our code. This, for me, is partly based in practical considerations, but also partially in more long-term programming language design.

traccc, in my view, is inherently a pipeline project, where we take data about pixel activations and turn them into tracks. At each step, we transform the data in some way, and we move forward through our chain of algorithms. It is a use case like this were move semantics are particularly powerful, since they allow us to quite literally model that movement of data. Do we really need to keep track of all of our old data all the time? Perhaps in some cases that is valuable, but surely in most cases we simply want to move on to the next algorithm in our chain, right? In cases where we really do need that old data, we can always make a copy of it, but I don't think that storing a copy should be the default approach.

Move semantics also serve a particularly useful purpose in high-memory usage heterogeneous computing, because they can guarantee at a type level (although the implementation is still somewhat iffy sometimes) that we can re-use memory that we no longer need. If you program your heterogeneous code cleverly that allows you to avoid a lot of allocations, since you can simply write your output to where your input used to be! You and I have talked about asynchronous allocation as well recently, and that becomes significantly easier with move semantics, too, since you can delegate the management of your memory to the kernel which is already running asynchronously. If you use move semantics properly you can get away with using less memory for your kernels, or conversely with running larger payloads on the same GPU!

Substructural typing is something we should not be ignoring, especially working on the border of computing and physics. I genuinely believe that more advanced type systems (especially affine and linear ones) which are now mostly academic toys will explode in popularity in the near future, if you consider for example how neatly the no-cloning theorem of quantum computing maps onto these systems! Of course modern C++ is not close to capturing a rich enough type theory to accomplish this, but we should not blindly follow the tried-and-tested "pass a const reference to everything" strategy, if we don't have to!

That said, I am a little bit of a dreamer in these regards. If you think passing these as const references for now I am more than happy to change the code to do that instead. :smile:

core/include/utils/algorithm.hpp
stephenswat: Hmm, I planned to make quite extensive use of these higher-order tools to write tests, for example. I think they will be useful to keep around.

stephenswat:
In spite of my own obnoxious programming practices I've changed all the rvalue references by const lvalue references (for now!). :slightly_smiling_face:

beomki-yeo:
I think it is OK to be merged if @krasznaa is also OK - I can try to apply this to cuda algorithms by myself and report later

examples/run/cpu/seq_example.cpp
```diff
@@ -52,20 +52,21 @@ int seq_run(const std::string& detector_file, const std::string& cells_dir,
...
-        auto ca_result = ca(cells_per_event);
-        auto& measurements_per_event = ca_result.first;
-        auto& spacepoints_per_event = ca_result.second;
-
         n_modules += cells_per_event.size();
         n_cells += cells_per_event.total_size();
+
+        auto ca_result = ca(std::move(cells_per_event));
+        auto measurements_per_event = std::move(ca_result.first);
+        auto spacepoints_per_event = std::move(ca_result.second);
```
paulgessinger, line 60: Would this to the same thing? 

```cpp
auto [measurements_per_event, spacepoints_per_event] = ca(std::move(cells_per_event));
```

asalzburger:
Updated this branch, I will have a look through as well!

core/include/utils/algorithm.hpp
```diff
@@ -18,35 +27,41 @@ namespace traccc {
...
```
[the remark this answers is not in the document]
stephenswat: The arguments to `compose` can (sadly) have side-effects, but this function is specifically defined to add explicit side effects (in the forms of functions of type _A_ → void) to a pure algorithm. So, for example, if I want to print some string representing the type _A_, I could write a side-effecting function _f_ : _A_ → void, which I could not compose normally. Because for any function _g_ : _A_ → _B_, _g_ ∘ _f_ does not compose, and for another function _h_ : _A_ → _A_, _f_ ∘ _h_ : _A_ → void, which means that we cannot continue our computation meaningfully, because all we have is the a value of the unit type.

So what this function allows you to do is take a function _g_ : _A_ → _B_ and a function _f_ : _A_ → void, and essentially glue them into a new function _g'_ : _A_ → _B_ which incorporates the side-effects defined by _f_ into _g_ while still having the same type as _g_ originally had.

If you wanted to define the same behaviour using composition alone, you would need to define our side effecting function _f_ as _A_ → _A_, which works but is not as ergonomic (in my opinion) as the side-effect mechanism. In reverse this is much easier, because you can construct a function of exactly that type by gluing your side-effects to the identity function, which gives you exactly that type. But doing it this way is much easier that doing it the other way around! :smile:

Is it used anywhere except the tests, no, not yet! But my plan was to take some of our code like the measurement statistics and life them into a void-returning algorithm, so you could re-use that code in many different places very easily.

stephenswat:
I got rid of some vestigial move statements, which should be more in line with the new const lvalue semantics.

core/include/utils/algorithm.hpp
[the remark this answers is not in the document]
paulgessinger: Ok, sure if this is what you want. I'll out myself as a simpleton here, and state that I don't see the benefit, really, but let's go.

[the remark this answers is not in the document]
stephenswat: Without a shadow of a doubt you are correct that I am over-complicating things. :see_no_evil:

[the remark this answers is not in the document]
krasznaa: I'm with Paul here. This code seems to be a bit more complicated than what we would actually need in this project in practice. :stuck_out_tongue: But in the end it's just a couple of lines of helper functions, so I don't object to them either.

krasznaa (approved):
From my side this PR still looks good. Since it touches all of the algorithms, it would be good to merge in, as other changes will very likely produce conflicts with it soon.

PR #89, opened 2021-09-14
Add detray dependency

beomki-yeo:
This is for adding detray dependency to traccc.
I am planning to use grid & axis classes of detray for spacepoint grouping.
I would like to ask if people are happy with this change? 

Also there was one issue that detray doesn't have package config module so we can not use `find_package(...)` yet


stephenswat (approved):

beomki-yeo:
Hmm I will make another PR to add detray with commit tag rather than origin/main

PR #90, opened 2021-09-16
Improve the numerical stability of measurement variance calculation

stephenswat:
In the current measurement creation algorithm, the variance of measurement positions is calculated using a naive single-pass algorithm. This algorithm is extremely numerically unstable because it is prone to catastrophic cancellation. Especially when using floats (for GPU processing), this algorithm will not provide the necessary accuracy and will likely lead to nonsensical variance values being calculated.

In this commit, I update the measurement creation algorithm to use a weighted variation of Welford's algorithm, which is a numerically stable single-pass algorithm which can work for large numbers of activations as well as high coordinate ranges. This should improve the accuracy of the variance calculation significantly.

### DIFF
--- a/core/include/clusterization/measurement_creation.hpp
+++ b/core/include/clusterization/measurement_creation.hpp
@@ -61,6 +61,15 @@ struct measurement_creation
         for (const auto &cluster : clusters.items) {
             scalar totalWeight = 0.;
 
+            // To calculate the mean and variance with high numerical stability
+            // we use a weighted variant of Welford's algorithm. This is a
+            // single-pass online algorithm that works well for large numbers
+            // of samples, as well as samples with very high values.
+            //
+            // To learn more about this algorithm please refer to:
+            // [1] https://doi.org/10.1080/00401706.1962.10490022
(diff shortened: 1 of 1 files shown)


beomki-yeo:
I think this can be merged

paulgessinger:
This is in *preparation* for CCA unit tests, yes? 
Generally looks good to me.

stephenswat:
Partially, yes. The main reason I want this to go in before the test suite is because the tests won't pass with the old variance code, because it doesn't give the correct results. :smile:

paulgessinger:
Right. Would you be ok to link out or at least name the weighted variation of Welford's algorithm in the comment so readers can look it up if need be?

stephenswat:
Great suggestion, I've added a comment block that links to the original paper, as well as a chapter of Knuth's _The Art of Computer Programming_ that explains this algorithm as well. :slightly_smiling_face:

krasznaa (approved):
Just because it seems that my approval gets PRs merged this morning, I think this one is ripe for a merge as well. :wink:

PR #91, opened 2021-09-17
Add a test suite for connected component analysis

stephenswat:
This commit adds a new data-driven test suite for connected component analysis, which accepts hit files as well as CCA truth files (which are not the same as the usual track truth files), and will compare the CCA algorithm on each hit-truth file pair.

This cannot be merged until #87, #90, and [merge request !2 on the traccc-data repository](https://gitlab.cern.ch/acts/traccc-data/-/merge_requests/2) are merged.

- [x] Wait for #87 to be merged
- [x] Wait for #90 to be merged
- [x] Wait for merge request !2 on the `traccc-data` repository to be merged
- [x] Rebase this merge request on top of the new main branch

### DIFF
--- a/tests/common/include/tests/cca_test.hpp
+++ b/tests/common/include/tests/cca_test.hpp
@@ -1,0 +1,126 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <gtest/gtest.h>
+
+#include <functional>
+
+#include "definitions/primitives.hpp"
+#include "edm/cell.hpp"
+#include "edm/cluster.hpp"
+#include "tests/data_test.hpp"
+
+using cca_function_t = std::function<
+    std::map<traccc::geometry_id, std::vector<traccc::measurement>>(
+        const traccc::host_cell_container &)>;
+
+class ConnectedComponentAnalysisTests
+    : public traccc::tests::data_test,
+      public testing::WithParamInterface<
+          std::tuple<cca_function_t, std::string>> {
+    public:
+    struct cca_truth_hit {
+        uint64_t geometry_id = 0;
+        uint64_t hit_id = 0;
+        uint64_t num_cells = 0;
+        traccc::scalar channel0 = 0;
+        traccc::scalar channel1 = 0;
+        traccc::scalar variance0 = 0.;
+        traccc::scalar variance1 = 0.;
+
+        DFE_NAMEDTUPLE(cca_truth_hit, geometry_id, hit_id, num_cells, channel0,
+                       channel1, variance0, variance1);
+    };
+
+    using cca_truth_hit_reader = dfe::NamedTupleCsvReader<cca_truth_hit>;
+
+    inline static std::string get_test_name(
+        const testing::TestParamInfo<ParamType> &info) {
+        return std::get<1>(info.param);
+    }
+
+    inline static std::vector<std::string> get_test_files(void) {
+        const std::vector<std::pair<std::string, std::size_t>> cases = {
+            {"dense", 100},
+            {"multiple_module_single_hit", 100},
+            {"single_module_multiple_hit_single", 100},
+            {"single_module_multiple_hit_single_sparse", 100},
+            {"single_module_single_hit", 100},
+            {"very_dense", 100},
+            {"trackml_like", 30},
+        };
+        std::vector<std::string> out;
+
+        for (const std::pair<std::string, std::size_t> &c : cases) {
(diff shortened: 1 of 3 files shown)


tests/algorithms/cca_tests.cpp
```diff
@@ -0,0 +1,170 @@
...
+                const traccc::host_cell_container &)>
+            f) {
+        std::string file_hits =
+            get_datafile("cca_test/" + file_prefix + "_hits.csv");
+        std::string file_truth =
+            get_datafile("cca_test/" + file_prefix + "_truth.csv");
+
+        traccc::cell_reader creader(file_hits);
+
+        vecmem::host_memory_resource resource;
```
krasznaa, line 79: For `vecmem::host_memory_resource` it doesn't really matter. But since I was preaching recently to @beomki-yeo and @niermann999 that memory resources should have long lifetimes (as some of the smart ones really only shine in such a setup), I think it would be better to put this memory resource outside of this function. I'd just make it into a `static` variable inside of the file.

tests/algorithms/cca_tests.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    return out;
+}
+}  // namespace
+
+class ConnectedComponentAnalysisTests
+    : public traccc::tests::data_test,
+      public testing::WithParamInterface<std::string> {
+    public:
+    void test_connected_component_analysis(
```
krasznaa, line 66: Why is this function separated from the `CpuSparseCCLAlgorithm` function? Helper functions like this mostly make sense if they are used from multiple places. But since this is only used in that one test, why not just implement one big function?

tests/algorithms/cca_tests.cpp
```diff
@@ -0,0 +1,170 @@
...
+           &info) { return info.param; });
+
+// Google Test can be run manually from the main() function
+// or, it can be linked to the gtest_main library for an already
+// set-up main() function primed to accept Google Test test cases.
+int main(int argc, char **argv) {
+    ::testing::InitGoogleTest(&argc, argv);
+
+    return RUN_ALL_TESTS();
+}
```
krasznaa, line 170: As I keep pestering @beomki-yeo as well, we should not be implementing `main(...)` functions for our tests. If we don't, if we just use `GTest::gtest_main`, then we can freely choose in the CMake configuration how we want to link our tests into executables.

In this setup this test must always be compiled into a standalone executable, it can't be combined with other tests into a larger executable. (Which is a design that I really like about VecMem's tests...)

krasznaa (changes requested):
Just a set of technical comments...

We should also make `add_traccc_test(...)` a little nicer. But I'll look into that myself in a separate PR.

tests/algorithms/cca_tests.cpp
stephenswat: The idea here is that these tests can be applied to any CCA algorithm, not just the CPU ones. That is why the design is so generic, and the idea of having this function be separate is that we can more easily apply it to, for example, a GPU implementation.

tests/algorithms/cca_tests.cpp
stephenswat: That's fair, I think we have a lot of standardization to do for traccc tests in general, something we can look into as a chore.

tests/algorithms/cca_tests.cpp
stephenswat: Can do. Not sure it really matters much for tests, but I don't mind either way.

stephenswat: This has been resolved (or, arguably, made worse) because this resource is now used in a header file, where defining translation-unit scope variables is a lot more awkward...

stephenswat:
This PR has been mostly rewritten from scratch, but with consideration for your comments. Please feel free to take a look. :smile:

stephenswat:
Shamelessly pinging @krasznaa and @paulgessinger, I'd love to get this merged in if either of you has some time to look over it and stamp it. :smile:

krasznaa (approved):
I'm ashamed to admit that I don't know every detail of this code anymore. :frowning:

Technically the thing looks okay. Although since the "test class" is not templated, I would've preferred going lighter on the `inline`-s, and putting the class's implementation into a `.cpp` file. But for a test that's currently only included in a single place I'm not opening a battle-front over that... :stuck_out_tongue:

Technically, I thought that the `TEST_P` declaration could've also gone into the "test class"'s `.cpp` file, and only `INSTANTIATE_TEST_SUITE_P` would've needed to go into the "test source file". But this is also something that we can generalise when/if we need to start using the test class in multiple places.

All in all, let's just go forward with it as-is.

stephenswat:
Thanks! Apparently the policy bot is not counting your review, though? :fearful:

PR #92, opened 2021-09-22
Remove explicit vecmem and algebra-plugin dependency

beomki-yeo:
They can be removed as they are already included in detray-dependency


cmake/traccc-vecmem.cmake
```diff
@@ -1,40 +0,0 @@
-# TRACCC library, part of the ACTS project (R&D line)
-#
-# (c) 2021 CERN for the benefit of the ACTS project
-#
-# Mozilla Public License Version 2.0
-
-# Guard against multiple includes.
-include_guard( GLOBAL )
```
krasznaa, line 8: I'm not really in favour of this... I would much rather change this line (and the equivalent line in detray/algebra-plugins) to:

```cmake
if( TARGET "vecmem::core" )
   return()
endif()
```

, and keep all of them in place. So that whoever is called first, wins.

Note that this can actually be absolutely necessary in case we don't want to build VecMem as part of this project, but rather pick up a pre-built version of it. Since remember, `find_package(vecmem)` does not declare the VecMem library targets globally! It only declares them in the scope in which `find_package(vecmem)` was called. (We had to realise this in https://github.com/acts-project/acts/pull/998 with @konradkusiak97.) So one of those submodules setting up VecMem for itself through `find_package(vecmem)` will not make the libraries available in this main project.

krasznaa (changes requested):
I don't like this. :frowning: I believe the key to making all projects work with each other robustly is to rather put a consistent set of CMake modules into them, which set up these "externals" in a way that allows the submodules to be included in whatever order.

Having removed algebra-plugins as a submodule, is a good thing. But we should now make sure that detray and traccc include it with a consistent set of CMake code. Similar to what I wrote down as a comment on `traccc-vecmem.cmake`.

beomki-yeo:
Okay then I will close this PR, but opened a PR in algebra-plugin (acts-project/algebra-plugins#22) to avoid the current issue with duplicate population of vecmem

> But we should now make sure that detray and traccc include it with a consistent set of CMake code. 
Makes sense. I think we have to open another PR for this later

PR #94, opened 2021-10-05
Add a minimal CUDA magnetic field storage

stephenswat:
This commit adds a minimum viable product version of the magnetic field storage for CUDA GPUs. It is based on texture memory, and the application shown generates a _z_-slice of a magnetic field map given as a text file.

Given a magnetic field map (as a `.txt` file), and optionally a _z_-index and an image size, this will render an image of the magnetic field at that coordinate, with the colours representing the magnitude of the magnetic field.

The implementation performs trilinear interpolation _using the GPU's build-in hardware_, so there is no software overhead to the interpolation. Also, the data is stored using a spatially efficient Z-curve, which means that locality is preserved very nicely.

Performance is quite reasonable. For an 8192 × 8192 image (that's about 67 million lookups), the total kernel time on my NVIDIA GTX 1660 Ti is 18.7 milliseconds, including the time needed to calculate the magnitude of the magnetic field at each point. That corresponds to a throughput of 3.57 billion magnetic field lookups per second.

However, please note that the CUDA array is very optimized for this kind of job, and the locality here is excellent. Real-life performance for propagation will most likely be lower.

When run, the program spits out an `output.bmp` file which are the images shown below.

This is specifically designed to work with the format given in https://gitlab.cern.ch/acts/acts-data/-/blob/master/MagneticField/ATLAS/ATLASBField_xyz.txt.

Sample image of the ATLAS magnetic field at _z_ = 0mm (1024 × 1024): 
atlas_0_1024_1024

Sample image of the ATLAS magnetic field at _z_ = 0mm (1024 × 512):
atlas_0_1024_512

Sample image of the ATLAS magnetic field at _z_ = 0mm (128 × 128):
atlas_0_128_128

Sample image of the ATLAS magnetic field at _z_ = 7500mm (1024 × 1024):
atlas_7500_1024_1024

### DIFF
--- a/examples/run/cuda/bfield_cuda.cu
+++ b/examples/run/cuda/bfield_cuda.cu
@@ -1,0 +1,625 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include <cmath>
+#include <fstream>
+#include <iostream>
+#include <limits>
+
+#include "cuda/utils/definitions.hpp"
+
+/*
+ * This structure contains all the information we need to use the magnetic field
+ * on the GPU, at least in this simple context. It contains the texture in GPU
+ * memory, the offsets in each direction (because texture indexing starts at
+ * zero, but we can have negative coordinates), and the total size of the
+ * magnetic field volume.
+ */
+struct MagneticField {
+    /*
+     * This is a CUDA object referring to a texture, the details are opaque and
+     * we are not allowed to know what is inside.
+     */
+    cudaTextureObject_t texture;
+
+    /*
+     * The offsets for coordinates in the B-field volume.
+     */
+    float offset_x, offset_y, offset_z;
+
+    /*
+     * The total sizes of the B-field volume.
+     */
+    float size_x, size_y, size_z;
+};
+
+/*
+ * This function takes the output of our kernel and writes it to a file, adding
+ * the necessary headers to turn it into a bitmap image.
+ */
+void render_bitmap(char* img, unsigned int w, unsigned int h,
+                   std::string fname) {
+    /*
+     * Start by opening our output file in binary format.
+     */
+    std::ofstream bmp(fname, std::ios::out | std::ios::binary);
+
+    /*
+     * The image size is the size of the image component (excluding the
+     * headers), which is simply the product of the width and the height
+     * because we are building a one-byte-per-pixel image.
+     *
+     * The width must be padded due to the way BMP images work.
+     */
+    unsigned int imgsize = (w + (4 - (w % 4)) % 4) * h;
+
+    /*
+     * The image data starts at byte 1078, after 54 bytes for the header and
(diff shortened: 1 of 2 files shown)


examples/run/cuda/bfield_cuda.cu
```diff
@@ -0,0 +1,625 @@
...
+    /*
+     * This is a CUDA object referring to a texture, the details are opaque and
+     * we are not allowed to know what is inside.
+     */
+    cudaTextureObject_t texture;
+
+    /*
+     * The offsets for coordinates in the B-field volume.
+     */
+    float offset_x, offset_y, offset_z;
```
beomki-yeo, line 32: I think we have to use `scalar` instead of `float`

examples/run/cuda/bfield_cuda.cu
```diff
@@ -0,0 +1,625 @@
...
+ * doing trilinear interpolation.
+ *
+ * Returns a 4-vector, but only the first 3 values are used.
+ */
+__device__ float4 read_bfield(MagneticField bf, float x, float y, float z) {
+    /*
+     * We take the coordinates, offset them as appropriate, and add 0.5 which
+     * is required by the interpolation hardware.
+     */
+    return tex3D<float4>(bf.texture, x + bf.offset_x + 0.5,
```
beomki-yeo, line 257: If I understand correctly, you added 0.5 for triggering interpolation? (0.5 possibly half of b field cell interval?)
If so, I wonder what the interpolation algorithm is. 
Actually I want to have the same interpolation algorithm used by the simulation stage.  
If the interpolation algorithm is different, we may get the unexpected result after Kalman filtering

beomki-yeo (commented):
I left some comments. 
One general question possibly out of the scope of this PR: Is it possible to make this work in the context of vecmem?

examples/run/cuda/bfield_cuda.cu
beomki-yeo: Possibly texture doesnt work with double?

stephenswat: Correct, this only works with single-precision indices.

examples/run/cuda/bfield_cuda.cu
beomki-yeo: Okay good. the linear filtering may be what we also have in Acts (someone needs to confirm though)

stephenswat:
I think we should make a decision on whether we actually want this in traccc. Right now it's just a stand-alone example without much reusable value to it. We can merge it in just for the sake of having more GPU example code, or we could close it for now and I can come back with this baked into a more usable format.

beomki-yeo:
I think it is good to merge this in - anyway it is only in example directory and we can make a decision later anytime.

beomki-yeo:
I prefer the bfield interface to be in extra directory though. But doesn't really matter

stephenswat:
I am marking this as a draft for now, because it does not currently fit into traccc. I am working on creating some abstractions around these ideas that will make them more suitable for incorporation into our project. Leaving the code here for anyone to study, though.

stephenswat:
Closing this merge request for now.

PR #99, opened 2021-10-25
Fix several outstanding compiler warnings

stephenswat:
We've had a few compiler warnings sneak into our build again, which I would like to get rid of. This commit fixes all the ones I could find.


krasznaa (approved):
Let's formally approve this finally. I didn't realise that this was still sitting here... :frowning:

PR #101, opened 2021-11-03
Fix external dependency tag for detray

stephenswat:
As it stands, the detray dependency points towards the main branch of the detray repository. This is dangerous, because it means an update to detray can potentially break the compilation of traccc, and this is exactly what happens. The HEAD of traccc does not currently build. As a temporary fix (until we can start putting out numbered releases of detray) I am proposing we fix the version of detray to a given commit.


krasznaa (approved):
Let me just fix the wording a bit. The latest version of [detray](https://github.com/acts-project/detray) does not build against the current version of this project. That project, with the right setup, works just fine.

But yes, this project absolutely can not use the latest version of [detray](https://github.com/acts-project/detray) just yet.

PR #102, opened 2021-11-04
Allow the use of a system detray installation

stephenswat:
Currently, our Acts and vecmem dependencies allow the user to supply their own, system-wide installation of those packages. traccc will try to find them, and use them if it succeeds. It only uses the submodules included if the system installation is not found.

This behaviour doesn't exist for detray, which makes the build system a little inconsistent. In this commit, I introduce the same checks that we have for Acts and vecmem, to make sure that we can use system installations of detray if desired.


stephenswat:
Closing in favour of #103.

PR #103, opened 2021-11-05
External Build Reorganisation, main branch (2021.11.05.)

krasznaa:
Here I'm updating this project to handle its externals in the same way in which [Algebra Plugins](https://github.com/acts-project/algebra-plugins) and [Detray](https://github.com/acts-project/detray) are now doing it, after the following updates:
  - https://github.com/acts-project/algebra-plugins/pull/27
  - https://github.com/acts-project/detray/pull/129

As described in those PRs as well, the idea here is to remove all "exploratory" `find_package(...)` calls from the code. Configuring it explicitly through cache variables exactly what needs to be found as existing already, and what needs to be built by this project. (Or by one of the projects that this project builds.)

There are a couple of imperfections in the PR still unfortunately. :frowning:
  - I want to make a numbered release of [Detray](https://github.com/acts-project/detray) after https://github.com/acts-project/detray/pull/148 goes in (@niermann999 ? :smile:). For now I've set up the code to take Detray from a particular Git hash.
  - The seed finding code does not work with the latest version of [Acts](https://github.com/acts-project/acts). So while I opened a PR into that project (https://github.com/acts-project/acts/pull/1071) to allow using it as a sub-project in this one, that will not help us too much right now. So I decided to just make a `v10.0.1` patch tag with the necessary changes myself. :stuck_out_tongue: (https://github.com/krasznaa/acts/releases/tag/v10.0.1) This PR picks up the [Acts](https://github.com/acts-project/acts) source code from that `v10.0.1` release.

I also went and made the absolute minimal changes required in the C\+\+ code to make it compatible with the latest version of [Algebra Plugins](https://github.com/acts-project/algebra-plugins). Though I intend to do a much bigger cleanup in those libraries/headers later on.

This is meant to replace #102. (Pinging @stephenswat.) As this will allow "builds of any complexity", as long as you provide just the right set of flags for CMake.

With this included, it should also become possible to start using VecMem's SYCL code in not too long. Hold out @konradkusiak97. :smile:


cmake/traccc-compiler-options.cmake
```diff
@@ -10,16 +10,24 @@ include_guard( GLOBAL )
...
-      traccc_add_flag( CMAKE_CXX_FLAGS_${mode} "-Wextra" )
-   endforeach()
+   traccc_add_flag( CMAKE_CXX_FLAGS "-Wall" )
+   traccc_add_flag( CMAKE_CXX_FLAGS "-Wextra" )
+
 endif()
+
+# Set the CUDA architecture to build code for.
+set( CMAKE_CUDA_ARCHITECTURES "52" CACHE STRING
+   "CUDA architectures to build device code for" )
```
stephenswat, line 29: This is only supported in CMake 3.18 as far as I know, perhaps we should ensure that the minimum CMake version is at least 3.18? I know we require 3.9 elsewhere. Also, is it really a smart idea to set defaults for this? I would rather throw an error if this is not set to a sane value by the user.

extern/acts/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+# CMake include(s).
+cmake_minimum_required( VERSION 3.11 )
+include( FetchContent )
+
+# Tell the user what's happening.
+message( STATUS "Building Acts as part of the TRACCC project" )
+
+# Declare where to get Acts from.
+set( TRACCC_ACTS_SOURCE
+   "URL;https://github.com/krasznaa/acts/archive/refs/tags/v10.0.1.tar.gz;URL_MD5;fcd2f2f7b5810e34b8faa12634043be3"
```
stephenswat, line 16: Is there a reason we use version 10.0.1 specifically? I know Acts is up to 14.1.0, is this the last working version, or did we pick this arbitrarily?

extern/acts/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+   CACHE STRING "Source for Acts, when built as part of this project" )
+mark_as_advanced( TRACCC_ACTS_SOURCE )
+FetchContent_Declare( Acts ${TRACCC_ACTS_SOURCE} )
+
+# Options used in the build of Acts.
+set( ACTS_SETUP_EIGEN3 FALSE CACHE BOOL
+   "Do not set up Eigen in the Acts code, we do it in this project" )
```
stephenswat, line 23: Are we certain that Eigen3 is always set up in traccc? I thought it was possible to disable that, no?

extern/algebra-plugins/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+# CMake include(s).
+cmake_minimum_required( VERSION 3.14 )
+include( FetchContent )
+
+# Tell the user what's happening.
+message( STATUS "Building Algebra Plugins as part of the TRACCC project" )
+
+# Declare where to get Algebra Plugins from.
+set( TRACCC_ALGEBRA_PLUGINS_SOURCE
+   "URL;https://github.com/acts-project/algebra-plugins/archive/refs/tags/v0.3.0.tar.gz;URL_MD5;7ffcfe299ebf654074b196e949afb010"
```
stephenswat, line 16: I think we might eventually want to store these URLs and hashes somewhere outside of the CMake configuration as a single source of truth, but this is fine for now.

extern/detray/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+# CMake include(s).
+cmake_minimum_required( VERSION 3.14 )
+include( FetchContent )
+
+# Tell the user what's happening.
+message( STATUS "Building Detray as part of the TRACCC project" )
+
+# Declare where to get Detray from.
+set( TRACCC_DETRAY_SOURCE
+   "GIT_REPOSITORY;https://github.com/acts-project/detray.git;GIT_TAG;3aafe70"
```
stephenswat, line 16: Any reason why we fetch this via Git instead of downloading it as a tarball?

extern/detray/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+set( DETRAY_SETUP_VECMEM FALSE CACHE BOOL
+   "Do not set up VecMem as part of Detray" )
+set( DETRAY_SETUP_ALGEBRA_PLUGINS FALSE CACHE BOOL
+   "Do not set up Algebra Plugins as part of Detray" )
+set( DETRAY_SETUP_DFELIBS FALSE CACHE BOOL
+   "Do not set up dfelibs as part of Detray" )
+set( DETRAY_SETUP_GOOGLETEST FALSE CACHE BOOL
+   "Do not set up GoogleTest as part of Detray" )
+set( DETRAY_SETUP_BENCHMARK FALSE CACHE BOOL
+   "Do not set up Google Benchmark as part of Detray" )
```
stephenswat, line 38: This seems fragile. Not in a way we can really control on the traccc side, but perhaps we could move to a more sane automatic detection mechanism for this? Or maybe disable these by default?

extern/dfelibs/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+mark_as_advanced( TRACCC_DFELIBS_SOURCE )
+FetchContent_Declare( dfelibs ${TRACCC_DFELIBS_SOURCE} )
+
+# Options used in the build of dfelibs.
+set( dfelibs_BUILD_EXAMPLES FALSE CACHE BOOL
+   "Turn off the build of the dfelibs examples" )
+set( dfelibs_BUILD_UNITTESTS FALSE CACHE BOOL
+   "Turn off the build of the dfelibs unit tests" )
+set( dfelibs_ENABLE_INSTALL TRUE CACHE BOOL
+   "Have dfelibs be installed together with the rest of the project" )
```
stephenswat, line 27: Is this necessary? I thought dfelibs was header only, why would we install it with the rest?

extern/vecmem/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+# Declare where to get VecMem from.
+set( TRACCC_VECMEM_SOURCE
+   "URL;https://github.com/acts-project/vecmem/archive/refs/tags/v0.6.0.tar.gz;URL_MD5;d741067ff148d2df3c4e6d4587d65fdc"
+   CACHE STRING "Source for VecMem, when built as part of this project" )
+mark_as_advanced( TRACCC_VECMEM_SOURCE )
+FetchContent_Declare( VecMem ${TRACCC_VECMEM_SOURCE} )
+
+# Options used in the build of VecMem.
+set( VECMEM_BUILD_TESTING FALSE CACHE BOOL
+   "Turn off the build of the VecMem unit tests" )
```
stephenswat, line 23: Same thing as before, maybe building the tests should be disabled by default on the vecmem side.

CMakeLists.txt
```diff
@@ -29,36 +28,136 @@ set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY
...
+set( TRACCC_BUILD_CUDA_DEFAULT FALSE )
+if( CMAKE_CUDA_COMPILER )
+   set( TRACCC_BUILD_CUDA_DEFAULT TRUE )
 endif()
 
-# option for algebra scalar type 
-set(TRACCC_CUSTOM_SCALARTYPE float)
-if(TRACCC_CUSTOM_SCALARTYPE)
-    message(STATUS "Building with custom scalar type: " ${TRACCC_CUSTOM_SCALARTYPE})
-    set(ALGEBRA_PLUGIN_CUSTOM_SCALARTYPE ${TRACCC_CUSTOM_SCALARTYPE})
```
stephenswat, line 44: Is there a reason we're getting rid of this functionality?

CMakeLists.txt
```diff
@@ -29,36 +28,136 @@ set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY
...
+option( TRACCC_BUILD_CUDA "Build the CUDA sources included in traccc"
+   ${TRACCC_BUILD_CUDA_DEFAULT} )
+option( TRACCC_BUILD_TESTING "Build the (unit) tests of traccc" TRUE )
+
+# Clean up.
+unset( TRACCC_BUILD_CUDA_DEFAULT )
+
+# Set up VecMem.
+option( TRACCC_SETUP_VECMEM
+   "Set up the VecMem target(s) explicitly" TRUE )
```
stephenswat, line 49: Does it make sense to have this be a configurable parameter? Surely if you turn this off there is no way your code will ever compile, right?

CMakeLists.txt
```diff
@@ -29,36 +28,136 @@ set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY
...
+
 # Build the traccc code.
-add_subdirectory(core)
-add_subdirectory(io)
-add_subdirectory(examples)
-add_subdirectory(plugins)
-if(BUILD_TESTING)
-  add_subdirectory(tests)
+add_subdirectory( core )
+if( TRACCC_BUILD_CUDA )
```
stephenswat, line 152: Maybe we should shove this logic into a CMakeLists file for `device/` itself?

stephenswat (commented):
Seems sane and sensible for the most part. I have a few things which I would like to discuss, but this is absolutely a step in the right direction. I can't comment much on the algebra related changes, because I am not fully familiar with how that is supposed to work yet.

cmake/traccc-compiler-options.cmake
krasznaa: It's a cache variable that we set here. If we set `CXXGPU30_ARCHITECTURES` to some value (assuming that in the c\+\+30 standard there will be a GPU architecture definition necessary), none of the build would care. Setting this variable for CMake <3.18 has absolutely no effect whatsoever. While with CMake 3.18+ if we don't set it, we get a warning from CMake.

I believe setting a default is a good idea. We've been doing that in VecMem since the beginning as well.

https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-compiler-options-cuda.cmake#L17

If the user wants to override this conservative default, they can do it from the command line with `-DCMAKE_CUDA_ARCHITECTURES="52;60;70"`, or something similar. The point is, with a default set, they just don't have to do it all... the... time...

extern/acts/CMakeLists.txt
krasznaa: I wrote about this in the description of the PR... Currently traccc is set up to build v10.0.0. I tried building the current seed-finding code against v14.1.0, but it didn't. We would need to update the code in this repository to make it compatible with the latest version of Acts.

I was not going to do that in this PR... Once https://github.com/acts-project/acts/pull/1071 is accepted, and a new Acts release is made, we can update to that in a separate PR.

extern/acts/CMakeLists.txt
krasznaa: It is. But by default it **is** built by traccc. So by default Acts should get it from the same place.

If you want to use another version of Eigen than what traccc is building, you can change all of these defaults. It will take modifying multiple flags, which I admit can get a little tiring. But all the flags are now there to allow any setup that you can think of.

Remember, all of these are cache variables. They can all be overridden from the command line if necessary.

extern/algebra-plugins/CMakeLists.txt
krasznaa: I guess so. But in my mind that would just unnecessarily complicate the CMake code.

These should be set up as cache variables in order to allow us to override them if/when necessary. So if we had a text file called `externals.txt` that we would read all those values out of, we would have to write some non-trivial CMake code that did this during configuration, and then still set up a cache variable with the value coming from that text file. Since I'd still want the value to be overridable.

extern/detray/CMakeLists.txt
krasznaa: Again, explanation in the description of the PR. :stuck_out_tongue: I'll update it shortly...

extern/detray/CMakeLists.txt
krasznaa: I do not agree. Tests should be enabled when we build a project as a top-level project.

Many projects that I've seen test whether they are the top level projects, and then disable their tests by default if they are not. We could potentially go for something like that with these projects.

But I myself don't see a huge issue with how these flags are set at the moment. The same type of complexity was there before my changes as well. The traccc build system was just oblivious to them. (And was not able to affect the behaviour of the upstream projects, as it can with the latest versions of those.)

cmake/traccc-compiler-options.cmake
stephenswat: I'm not sure I really follow along here. I suppose `CXXGPU30_ARCHITECTURES` is a typo? Anyway, I don't think silently failing on CMake <3.18 is a good idea at all! That seems dangerous to me. Also, The problem with NVIDIA architectures is that they are not necessarily backwards compatible. Even a conservative back-up configuration may still fail to compile on a more recent GPU. I think it would be safest to rely on the default provided by the NVIDIA compiler (which is the default behaviour, I think), because it can make a far more informed decision.

extern/acts/CMakeLists.txt
stephenswat: Sure, I didn't really catch that from the PR description, but that makes sense (upgrading should be on the to-do list!).

extern/dfelibs/CMakeLists.txt
krasznaa: The CSV reading code is all templated. And it uses dfelibs. Since we install the CSV reading headers, we must also install the dfelibs ones.

I had to do the same in Detray. The CSV reading code of that also depends on this external. If it didn't install dfelibs with itself (or found it from some other location), then `find_package(detray)` could not provide functional targets to the clients.

extern/detray/CMakeLists.txt
stephenswat: You can download a tarball for any revision in Github, that was my main confusion. :D

CMakeLists.txt
krasznaa: It is all set through `ALGEBRA_PLUGINS_CUSTOM_SCALARTYPE` now. Which is a cache variable in algebra-plugins. Since it's a cache variable, we don't have to define it here. We can override it with `-DALGEBRA_PLUGINS_CUSTOM_SCALARTYPE=float` from the command line, if we absolutely need to.

But as I've been advertising to @niermann999 and @beomki-yeo for the last weeks, I believe making this choice during the CMake configuration is a bad choice to begin with. I will want to re-design the algebra-plugins and detray projects to allow the user deciding about the scalar type as a template parameter. So that they could try out "detector objects" with different scalar types side-by-side, in the same source file if they wanted to.

So removing the setting from here is a precursor for things to come...

CMakeLists.txt
krasznaa: Are you absolutely sure that traccc will only ever be a top-level project? We have the same sort of flag for both algebra-plugins and detray. Because I knew that I will need to turn both of those off once we build them as part of this project.

The flag is here to allow us to potentially use this as a sub-project of even higher level projects.

CMakeLists.txt
krasznaa: I absolutely agree. But since the purpose of this PR was only to upgrade how the externals are handled, I didn't want to perturb the logic in other places too much. But we should absolutely make a lot more changes after this PR goes in.

extern/acts/CMakeLists.txt
stephenswat: These being cache variables is all well and good, but doesn't this mean that you can quite easily break the build just by selecting a non-default way of building Eigen? I would think that the way Acts builds Eigen should be informed by how traccc does it - rather than just disabling the build in Acts all together. Sure, this is unlikely to cause problems in real-world cases and I'm fine to accept this as is, but it feels to me like we're putting together a Jenga tower here, where one (potentially sane) configuration change can break the entire build. :fearful:

extern/algebra-plugins/CMakeLists.txt
stephenswat: Sure, let's skip this for now.

extern/detray/CMakeLists.txt
stephenswat: I think checking whether something is a top-level project is also an acceptable solution. The problem I see with defaulting all these things on be enabled is that you will need to update the traccc configuration whenever one of the dependencies adds any additional flags. Sounds like technical debt that we might want to avoid.

extern/dfelibs/CMakeLists.txt
stephenswat: I suppose this depends on how we plan to use traccc. I was always under the impression that it was a standalone demonstrator that nothing else would depend on, but I suppose it is more flexible if we install dfelibs...

extern/vecmem/CMakeLists.txt
stephenswat: I would be in favour of that. However, that is not really relevant right now. We can take that discussion to the relevant projects.

CMakeLists.txt
stephenswat: Fair enough. :smile:

CMakeLists.txt
stephenswat: That's not really what I mean - if you turn this off then CMake will never find vecmem at all, right? So then what will it know what to link against?

CMakeLists.txt
stephenswat: Fully agree on making this a template parameter. :+1:

cmake/traccc-compiler-options.cmake
krasznaa: :confused: I guess it will help if I give you a bit of a "matrix" of how CMake and CUDA interact with each other these days. Let's use the following example project:

```
[bash][Legolas]:demo-for-stephen > ls
CMakeLists.txt  demo.cu
[bash][Legolas]:demo-for-stephen > more CMakeLists.txt 

cmake_minimum_required( VERSION 3.10 )
project( demo-for-stephen VERSION 0.0.1 LANGUAGES CUDA )

set( CMAKE_CUDA_ARCHITECTURES "52" CACHE STRING "CUDA architecture(s) to use" )

add_executable( demo demo.cu )
[bash][Legolas]:demo-for-stephen > more demo.cu 

__global__
void demo() {

}

int main() {

   demo<<<1, 100>>>();
   return 0;
}
[bash][Legolas]:demo-for-stephen >
```

Now, let's see what happens with a new CMake and CUDA version out of the box.

```
[bash][Legolas]:demo-for-stephen > source ~/software/cuda/11.4.2/x86_64-ubuntu2004/setup.sh 
Configured CUDA from: /home/krasznaa/software/cuda/11.4.2/x86_64-ubuntu2004
[bash][Legolas]:demo-for-stephen > ~/software/cmake/3.21.3/x86_64-ubuntu2004-gcc9-opt/bin/cmake -S . -B build
-- The CUDA compiler identification is NVIDIA 11.4.120
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Check for working CUDA compiler: /home/krasznaa/software/cuda/11.4.2/x86_64-ubuntu2004/bin/nvcc - skipped
-- Detecting CUDA compile features
-- Detecting CUDA compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /data/ssd-1tb/projects/traccc/demo-for-stephen/build
[bash][Legolas]:demo-for-stephen >
```

Yepp, everybody is happy. Let's see what happens when `CMAKE_CUDA_ARCHITECTURES` is not set:

```
[bash][Legolas]:demo-for-stephen > ~/software/cmake/3.21.3/x86_64-ubuntu2004-gcc9-opt/bin/cmake -DCMAKE_CUDA_ARCHITECTURES= -S . -B build
-- The CUDA compiler identification is NVIDIA 11.4.120
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Check for working CUDA compiler: /home/krasznaa/software/cuda/11.4.2/x86_64-ubuntu2004/bin/nvcc - skipped
-- Detecting CUDA compile features
-- Detecting CUDA compile features - done
-- Configuring done
CMake Warning (dev) in CMakeLists.txt:
  Policy CMP0104 is not set: CMAKE_CUDA_ARCHITECTURES now detected for NVCC,
  empty CUDA_ARCHITECTURES not allowed.  Run "cmake --help-policy CMP0104"
  for policy details.  Use the cmake_policy command to set the policy and
  suppress this warning.

  CUDA_ARCHITECTURES is empty for target "demo".
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) in CMakeLists.txt:
  Policy CMP0104 is not set: CMAKE_CUDA_ARCHITECTURES now detected for NVCC,
  empty CUDA_ARCHITECTURES not allowed.  Run "cmake --help-policy CMP0104"
  for policy details.  Use the cmake_policy command to set the policy and
  suppress this warning.

  CUDA_ARCHITECTURES is empty for target "demo".
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Generating done
-- Build files have been written to: /data/ssd-1tb/projects/traccc/demo-for-stephen/build
[bash][Legolas]:demo-for-stephen >
```

CMake gets upset that we didn't set it to anything...

Now let's see what happens with an older version of CMake, but still a new CUDA version.

```
[bash][Legolas]:demo-for-stephen > ~/software/cmake/3.14.7/x86_64-ubuntu2004-gcc9-opt/bin/cmake -S . -B build1
-- The CUDA compiler identification is NVIDIA 11.4.120
-- Check for working CUDA compiler: /home/krasznaa/software/cuda/11.4.2/x86_64-ubuntu2004/bin/nvcc
-- Check for working CUDA compiler: /home/krasznaa/software/cuda/11.4.2/x86_64-ubuntu2004/bin/nvcc -- works
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Configuring done
-- Generating done
-- Build files have been written to: /data/ssd-1tb/projects/traccc/demo-for-stephen/build1
[bash][Legolas]:demo-for-stephen > ~/software/cmake/3.14.7/x86_64-ubuntu2004-gcc9-opt/bin/cmake -DCMAKE_CUDA_ARCHITECTURES= -S . -B build2
-- The CUDA compiler identification is NVIDIA 11.4.120
-- Check for working CUDA compiler: /home/krasznaa/software/cuda/11.4.2/x86_64-ubuntu2004/bin/nvcc
-- Check for working CUDA compiler: /home/krasznaa/software/cuda/11.4.2/x86_64-ubuntu2004/bin/nvcc -- works
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Configuring done
-- Generating done
-- Build files have been written to: /data/ssd-1tb/projects/traccc/demo-for-stephen/build2
[bash][Legolas]:demo-for-stephen >
```

That older version of CMake just does not care about this variable **at all**. It is not possible to influence in this easy way which platforms the build would target.

Now let's try your other point about 5.2 being "too new" of a requirement...

```
[bash][Legolas]:demo-for-stephen > ~/software/cmake/3.21.3/x86_64-ubuntu2004-gcc9-opt/bin/cmake -DCMAKE_CUDA_ARCHITECTURES="35" -S . -B build
-- The CUDA compiler identification is NVIDIA 11.4.120
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Check for working CUDA compiler: /home/krasznaa/software/cuda/11.4.2/x86_64-ubuntu2004/bin/nvcc - skipped
-- Detecting CUDA compile features
-- Detecting CUDA compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /data/ssd-1tb/projects/traccc/demo-for-stephen/build
[bash][Legolas]:demo-for-stephen > ~/software/cmake/3.21.3/x86_64-ubuntu2004-gcc9-opt/bin/cmake --build build
[ 50%] Building CUDA object CMakeFiles/demo.dir/demo.cu.o
nvcc warning : The 'compute_35', 'compute_37', 'compute_50', 'sm_35', 'sm_37' and 'sm_50' architectures are deprecated, and may be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
[100%] Linking CUDA executable demo
[100%] Built target demo
[bash][Legolas]:demo-for-stephen >
```

5.2 is the **oldest** standard that you can still target with CUDA 11 without getting a warning. It is the most conservative option that we can go with at this very moment. That's why I use it in all the Acts projects right now.

And the reason that I do set a default value is because I hate seeing CMake warnings when I forget to set a value explicitly...

extern/acts/CMakeLists.txt
krasznaa: Right now, without https://github.com/acts-project/acts/pull/1071, Acts sets up Eigen with this code:

```cmake
if (ACTS_USE_SYSTEM_EIGEN3)
  find_package(Eigen3 ${_acts_eigen3_version} REQUIRED CONFIG)
else()
  add_subdirectory(thirdparty/eigen3)
endif()
```

https://github.com/acts-project/acts/blob/main/CMakeLists.txt#L154-L158

If I didn't make any changes in Acts, this would force us to always build Eigen before building traccc. The build of Eigen could not be done at the same time as this project. (Since Acts does not use a new enough version of Eigen for our use case.)

Yes. With this many cache variables defined, you can very much come up with non-functional setups. Maybe we can improve on that in the future. But I'm getting the sense that you don't understand what my thinking is with all these cache variables. I guess we should talk on Monday over Skype about this...

cmake/traccc-compiler-options.cmake
stephenswat: Thanks for the demo, but I really feel like we are talking past each other. My point is not that 5.2 is "too new" of a requirement - my point is that NVIDIA architectures are not ordered in the sense that you can run code compiled for an older version of CUDA on a newer device. Case in point: you can design code for CC 5.2 that simply will not run on a CC 5.3 device. Even within the same major version there is _no guarantee of compatibility_ between CUDA versions. I agree that this is confusing, but this is precisely why it makes no sense at all to try to find any kind of default value for this - it's just not how CUDA works. The user should ALWAYS specify the architecture for which they want to build, that is the only way to make sure that it actually works. Frankly, this flows from your own example as well: the compiler will complain at you if you don't set a value. Setting a default at configuration time which does not guarantee that the code will work and _which will obscure this fact from the user by silencing an error_ is, in my opinion, a really bad idea.

CMakeLists.txt
krasznaa: If this project is added with [add_subdirectory](https://cmake.org/cmake/help/latest/command/add_subdirectory.html) to some parent project, then that parent project could've set up the build of VecMem itself. At which point either using `find_package(...)` or trying to set up the VecMem build itself, would be a bad choice in this place.

If we want to allow the usage of this project as a sub-project of something even higher level (like the repository of one of you guys that you use to make measurements for your thesis, against a fixed version of this repository), then it must be possible to tell it not to set up any of its externals. And just expect "the higher level project" to set those up outside of this code.

stephenswat: Hold on are you saying that we are inheriting dependencies from the parent project??? :fearful:

cmake/traccc-compiler-options.cmake
krasznaa: This is not true...

As I wrote, we've been building VecMem with this same setting since forever. And it's been working just fine on my RTX 3080.

stephenswat (approved):

stephenswat (approved):

PR #106, opened 2021-11-08
SYCL Example, main branch (2021.11.08.)

krasznaa:
Added a [SYCL](https://www.khronos.org/sycl/) example application to the project. With the goal of helping the development of @konradkusiak97 and @SylvainJoube,

The ability to use VecMem's "language code" depends on https://github.com/acts-project/vecmem/pull/144. Until that is in, this will stay a draft.

### DIFF
--- a/examples/run/sycl/sycl_language_example.sycl
+++ b/examples/run/sycl/sycl_language_example.sycl
@@ -1,0 +1,54 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+// SYCL include(s).
+#include <CL/sycl.hpp>
+
+// System include(s).
+#include <iostream>
+
+int main() {
+
+    // Loop over all available SYCL platforms.
+    for (const cl::sycl::platform& platform :
+         cl::sycl::platform::get_platforms()) {
+
+        // Print some information about the platform.
+        std::cout << "============ Platform ============" << std::endl;
+        std::cout << " Name   : "
+                  << platform.get_info<cl::sycl::info::platform::name>()
+                  << std::endl;
+        std::cout << " Vendor : "
+                  << platform.get_info<cl::sycl::info::platform::vendor>()
+                  << std::endl;
+        std::cout << " Version: "
+                  << platform.get_info<cl::sycl::info::platform::version>()
+                  << std::endl;
+
+        // Loop over all devices available from this platform.
+        for (const cl::sycl::device& device : platform.get_devices()) {
+
+            // Print some information about the device.
+            std::cout << "------------- Device -------------" << std::endl;
+            std::cout << " Name   : "
+                      << device.get_info<cl::sycl::info::device::name>()
+                      << std::endl;
+            std::cout << " Vendor : "
+                      << device.get_info<cl::sycl::info::device::vendor>()
+                      << std::endl;
+            std::cout << " Version: "
+                      << device.get_info<cl::sycl::info::device::version>()
+                      << std::endl;
+            std::cout << "----------------------------------" << std::endl;
+        }
+
+        std::cout << "==================================" << std::endl;
+    }
+
+    // Retrun gracefully.
+    return 0;
+}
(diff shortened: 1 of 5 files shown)


krasznaa:
Pinging @stephenswat. :wink:

SylvainJoube:
Dear Attila,

Thank you very much for your reply. When following your example I get exactly the same result, and every test succeeds:

```log
[==========] 18 tests from 4 test suites ran. (3293 ms total)
[  PASSED  ] 18 tests.
```

But I wasn't trying to build vecmem, sorry that was not very clear on my part. I was only trying to build traccc with the line `enable_language(SYCL)` written on the main `CMakeLists.txt`.

Last time I tried to build traccc, CMake told me it needed the `CMAKE_SYCL_COMPILER` and `CMAKE_SYCL_COMPILER_ENV_VAR` variables to be set, but I ran cmake again I only needed to write `CMAKE_CXX_COMPILER=dpcpp`, but still got the error mentioned above.

I am very confused as I'm able to build vecmem effortlessly but the not when it is included as a dependency of traccc. From what I understand, your pull request https://github.com/acts-project/traccc/pull/106 will solve the issue (when https://github.com/acts-project/vecmem/pull/144 will me merged too), is that right ?

krasznaa:
Basically we just need this to be merge in by now. VecMem is already updated. :wink:

Once this is in, you should modify your code to use `<foo>.sycl` files for everything that needs to be built with the oneAPI compiler, and "hide all of your code" behind `TRACCC_BUILD_SYCL`.

SylvainJoube:
Thanks ! :)

Should I use the `.sycl` also for header files ? I think we discussed this in a previous parallelization meeting but I don't really remember the outcome of the debate.

So I'll wait and will add to traccc my SparseCCL sycl implementation as soon as possible, following the structure of the current algorithms. I will most likely write two examples, one having exactly the same structure as the usual algorithms but running many kernels, and one more efficient running only one kernel per event but not as close to the current algorithm structure. I guess maybe I will present this at a future ACTS Parallelization Meeting in a few weeks (not until December though), if possible :)

krasznaa:
You should still use `.hpp` for the headers. You should definitely not use `.sycl`, just like how you shouldn't use `.cpp` for header files either. :stuck_out_tongue: (As CMake will actively try to compile files that have the `.sycl` extension. While it largely ignores files that have a `.hpp` extension.)

Currently we "didn't come up" with specific header extensions for HIP or SYCL code. For HIP the `.hip` extension was introduced into CMake itself a while ago. But no extension was introduced for "HIP headers". We also didn't invent anything for SYCL in VecMem. But that's also because we didn't have to.

You see, in my mind `.cuh` should only ever be used with CUDA code when you have to use CUDA keywords in a header. So that that header must only be read by a CUDA compiler. And with a good code design, we can very often avoid having to use CUDA keywords in headers all together. (I don't mean things like`__device__` and `__host__` here, but rather kernel launches, and `__global__` function implementations.)

So as long as you use a function like:

```c++
namespace traccc::sycl {
  void some_clever_algorithm(vecmem::vector_view<...> ..., vecmem::jagged_vector_view<...> ...) {
    ...
  }
}
```

, you should absolutely just use a `some_clever_function.hpp` header for declaring it, and `some_clever_function.sycl` for implementing it. As the declaration could be read by any (C\+\+) compiler, not just a SYCL-capable one.

krasznaa:
Ping again...? @stephenswat? :stuck_out_tongue:

SylvainJoube:
Understood, thank you for this explanation, I will follow these guidelines !

We will be able to ask Stephen directly on Friday during our ACTS Parallelization Meeting, anyways :)

stephenswat (approved):
Who, what, when?! Yes, looks good! :sweat_smile:

examples/run/sycl/sycl_language_example.sycl
```diff
@@ -0,0 +1,54 @@
...
+            std::cout << " Version: "
+                      << device.get_info<cl::sycl::info::device::version>()
+                      << std::endl;
+            std::cout << "----------------------------------" << std::endl;
+        }
+
+        std::cout << "==================================" << std::endl;
+    }
+
+    // Retrun gracefully.
```
stephenswat, line 52: I will let this one slip past for the sake of getting things done quickly. :smile:

PR #108, opened 2021-12-10
Add `link_type` in container for indexing and its application to seeding

beomki-yeo:
### 1. The motivation of the PR
- The `internal_spacepoint` object takes the copy of `spacepoint` as a member variable, which is a bit waste of memory and takes more time in global memory writing during spacepoint_binning

### 2. Major changes
- `container` class defines the `link_type` which is the pair of size types of header and item vector
-  `internal_spacepoint` takes the index of `spacepoint` in `spacepoint_container` with `link_type`

##### Before

https://github.com/acts-project/traccc/blob/b4b8562f219b520b9d854f09d42baaa7c56ce6fa/core/include/edm/internal_spacepoint.hpp#L30-L39

##### After 

```c++
template <typename spacepoint_t>
struct internal_spacepoint {

    using spacepoint_container_type = host_container<geometry_id, spacepoint_t>;
    using link_type = typename spacepoint_container_type::link_type;

    link_type m_link;

    scalar m_x;
    scalar m_y;
    scalar m_z;
    scalar m_r;
    scalar m_varianceR;
    scalar m_varianceZ;
```

### 3. Issue 
`std::pair` seems not working in device code with the following error:
```
/home/beomki/projects/traccc/traccc/core/include/edm/internal_spacepoint.hpp(108): error: calling a __host__ function("std::pair<unsigned long, unsigned long> ::operator =(const ::std::pair<unsigned long, unsigned long> &)") from a __device__ function("traccc::internal_spacepoint< ::traccc::spacepoint> ::operator =") is not allowed

/home/beomki/projects/traccc/traccc/core/include/edm/internal_spacepoint.hpp(108): error: identifier "std::pair<unsigned long, unsigned long> ::operator =" is undefined in device code
```
So, for the moment, I am utilizing `thrust::pair` as default argument for `pair_type`.

### DIFF
--- a/core/include/edm/container.hpp
+++ b/core/include/edm/container.hpp
@@ -10,6 +10,8 @@
 #include <type_traits>
 
 // VecMem include(s).
+#include <thrust/pair.h>
+
 #include <vecmem/containers/data/jagged_vector_buffer.hpp>
 #include <vecmem/containers/data/vector_buffer.hpp>
 #include <vecmem/containers/device_vector.hpp>
@@ -64,7 +66,8 @@ class container_element {
 /// header and item type.
 template <typename header_t, typename item_t,
           template <typename> class vector_t,
-          template <typename> class jagged_vector_t>
+          template <typename> class jagged_vector_t,
+          template <typename...> class pair_type = thrust::pair>
 class container {
     public:
     /// @name Type definitions
@@ -84,6 +87,16 @@ class container {
     /// The item vector type
     using item_vector = jagged_vector_type<item_t>;
 
+    /**
+     * @brief The size type of this container, which is the type by which
+     * its elements are indexed.
+     */
+    using size_type = typename header_vector::size_type;
+
+    /// The element link type
+    using link_type = pair_type<typename header_vector::size_type,
+                                typename item_vector::size_type>;
+
     /// @}
 
     /**
@@ -100,12 +113,6 @@ class container {
     using const_element_view =
         container_element<const header_t,
                           const typename item_vector::value_type>;
-
-    /**
-     * @brief The size type of this container, which is the type by which
-     * its elements are indexed.
-     */
-    using size_type = typename header_vector::size_type;
 
     /**
      * We need to assert that the header vector and the outer layer of the
@@ -172,7 +179,6 @@ class container {
         if (i >= size()) {
             throw std::out_of_range("Index out of range.");
         }
-
         return operator[](i);
     }
 
@@ -184,8 +190,21 @@ class container {
         if (i >= size()) {
             throw std::out_of_range("Index out of range.");
         }
-
         return operator[](i);
+    }
+
+    /**
+     * @brief Bounds-checking mutable item vector element accessor.
+     */
+    TRACCC_HOST_DEVICE
+    item_t at(const link_type& link) { return items[link.first][link.second]; }
+
+    /**
+     * @brief Bounds-checking immutable item vector element accessor.
+     */
+    TRACCC_HOST_DEVICE
+    const item_t at(const link_type& link) const {
+        return items[link.first][link.second];
     }
 
     /**
--- a/core/include/edm/container.hpp
+++ b/core/include/edm/container.hpp
@@ -67,7 +67,7 @@ class container_element {
 template <typename header_t, typename item_t,
           template <typename> class vector_t,
           template <typename> class jagged_vector_t,
-          template <typename...> class pair_type = std::pair>
+          template <typename...> class pair_type = thrust::pair>
 class container {
     public:
     /// @name Type definitions
--- a/core/include/edm/container.hpp
+++ b/core/include/edm/container.hpp
@@ -67,7 +67,7 @@ class container_element {
 template <typename header_t, typename item_t,
           template <typename> class vector_t,
           template <typename> class jagged_vector_t,
-          template <typename...> class pair_type = thrust::pair>
+          template <typename...> class pair_type = std::pair>
 class container {
     public:
     /// @name Type definitions
--- a/core/include/edm/internal_spacepoint.hpp
+++ b/core/include/edm/internal_spacepoint.hpp
@@ -85,6 +85,18 @@ struct internal_spacepoint {
     }
 
     TRACCC_HOST_DEVICE
+    internal_spacepoint& operator=(internal_spacepoint sp) {
+        m_link = sp.m_link;
+        m_x = sp.m_x;
+        m_y = sp.m_y;
+        m_z = sp.m_z;
+        m_r = sp.m_r;
+        m_varianceR = sp.m_varianceR;
+        m_varianceZ = sp.m_varianceZ;
+        return *this;
+    }
+
+    TRACCC_HOST_DEVICE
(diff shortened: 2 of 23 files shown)


stephenswat:
`std::pair` should work with `constexpr` relaxation turned on, you could give that a try.

beomki-yeo:
`--expt-relaxed-constexpr` is already included in compile options. (See `traccc/device/cuda/include/cuda/seeding/CMakeLists.txt`)

I guess it doesn't work because it's not a `constexpr` function. If it is from the missing constexpr relaxation, the error message should have been like this:
```
error: calling a constexpr __host__ function("operator[]") from a __global__ function("doublet_finding_kernel") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.
```

stephenswat:
I don't follow: `std::pair` does not define `operator[]` at all. What exactly is the problem you're facing with `std::pair` here?

beomki-yeo:
No I am not using `operator []` for std::pair. 
```
error: calling a constexpr __host__ function("operator[]") from a __global__ function("doublet_finding_kernel") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.
```
The above error message is just to show what happens if one uses `constexpr` host function without relaxation and has nothing to do with current issue.

The actual error message is the following w/ constexpr relaxation:
```
/home/beomki/projects/traccc/traccc/core/include/edm/internal_spacepoint.hpp(108): error: calling a __host__ function("std::pair<unsigned long, unsigned long> ::operator =(const ::std::pair<unsigned long, unsigned long> &)") from a __device__ function("traccc::internal_spacepoint< ::traccc::spacepoint> ::operator =") is not allowed

/home/beomki/projects/traccc/traccc/core/include/edm/internal_spacepoint.hpp(108): error: identifier "std::pair<unsigned long, unsigned long> ::operator =" is undefined in device code
```

As you can see `operator =` for `std::pair` is not `constexpr` which causes the error.

stephenswat:
I am virtually certain that nvcc should accept `std::pair` without any issue. I would - from what you're showing me here - suspect that this header file is being included in some other translation unit which is being compiled with `constexpr` relaxation. Could you check that?

beomki-yeo:
Hmm I doubt that it is from missing `constexpr` relaxation compile option because `operator =` of `std::pair` doesn't seem `constexpr`
I checked again and the culprit was at populating_grid.cu:
```c++
auto isp = internal_spacepoint<spacepoint>(
     spacepoints_device, {header_idx, sp_idx}, config.beamPos);
```
which calls `operator =` of `std::pair`.

I am not certain how it is possible, but things work after manually writing the assign operator for `internal_spacepoint`

```c++
    TRACCC_HOST_DEVICE
    internal_spacepoint& operator=(internal_spacepoint sp) {
        m_link = sp.m_link; // assign std::pair
        m_x = sp.m_x;
        m_y = sp.m_y;
        m_z = sp.m_z;
        m_r = sp.m_r;
        m_varianceR = sp.m_varianceR;
        m_varianceZ = sp.m_varianceZ;
        return *this;
    }
```

beomki-yeo:
Hmm, I found that spacepoint objects w/ `std::pair` are corrupted at some point during seed finding. (I don't know where exactly though). I will come back to this issue later

beomki-yeo:
Ah OK. I should have edited assignment operator not only `internal_spacepoint` but also `seed`. Now it does work well with `std::pair`.

```c++
    TRACCC_HOST_DEVICE
    internal_spacepoint& operator=(
        const internal_spacepoint<spacepoint_t>& sp) {

        m_link.first = sp.m_link.first;
        m_link.second = sp.m_link.second;
        m_x = sp.m_x;
        m_y = sp.m_y;
        m_z = sp.m_z;
        m_r = sp.m_r;
        m_varianceR = sp.m_varianceR;
        m_varianceZ = sp.m_varianceZ;

        return *this;
    }
```
```c++
    TRACCC_HOST_DEVICE
    seed& operator=(const seed& aSeed) {

        spB_link.first = aSeed.spB_link.first;
        spB_link.second = aSeed.spB_link.second;

        spM_link.first = aSeed.spM_link.first;
        spM_link.second = aSeed.spM_link.second;

        spT_link.first = aSeed.spT_link.first;
        spT_link.second = aSeed.spT_link.second;

        weight = aSeed.weight;
        z_vertex = aSeed.z_vertex;
        return *this;
    }
```

beomki-yeo:
@stephenswat Could you look into this again?

core/include/edm/container.hpp
```diff
@@ -10,6 +10,8 @@
 #include <type_traits>
 
 // VecMem include(s).
+#include <thrust/pair.h>
```
krasznaa, line 13: I'm confused. In the current setup where is the build taking this include from?

In any case, since you seem to be back to using [std::pair](https://en.cppreference.com/w/cpp/utility/pair), you should just remove this include.

core/include/edm/container.hpp
```diff
@@ -64,7 +66,8 @@ class container_element {
 /// header and item type.
 template <typename header_t, typename item_t,
           template <typename> class vector_t,
-          template <typename> class jagged_vector_t>
+          template <typename> class jagged_vector_t,
+          template <typename...> class pair_type = std::pair>
```
krasznaa, line 70: Do you foresee using something other than [std::pair](https://en.cppreference.com/w/cpp/utility/pair) here? Like [std::tuple](https://en.cppreference.com/w/cpp/utility/tuple) let's say?

I'm not completely sold even that we need to template this type at all. But as long as we only want to choose between [std::pair](https://en.cppreference.com/w/cpp/utility/pair) and [thrust::pair](https://thrust.github.io/doc/structthrust_1_1pair.html), you should declare this type rather like:

```c++
template <typename, typename> class pair_type = std::pair
```

I'm also not too happy with the naming here. As we discussed before, I myself would prefer an implementation like:

```c++
template <typename header_t, typename item_t,
          template <typename> class vector_t,
          template <typename> class jagged_vector_t,
          template <typename, typename> class pair_t = std::pair>
class container {
    ...
    template <typename T1, typename T2>
    using pair_type = pair_t<T1, T2>;
    ...
```

Just so that the `container::pair_type` symbol could be used outside of the class's body as well.

P.S. If for some reason you do want to consider [std::tuple](https://en.cppreference.com/w/cpp/utility/tuple) then you must replace all uses of `first` and `second`, and rather access the pair's elements with [std::get](https://en.cppreference.com/w/cpp/utility/pair/get) everywhere. But I don't think that should be necessary...

krasznaa (changes requested):
I think this should be fine. Please just remove the Thrust include from the core library header.

core/include/edm/container.hpp
beomki-yeo: Right. I removed it

core/include/edm/container.hpp
beomki-yeo: It's only for pair - I changed to `template <typename, typename> class pair_t = std::pair`

krasznaa:
For some reason GitHub doesn't even allow me to resolve discussions on this repository. :frowning: But I consider the two that I opened resolved by now.

Issue #93, opened 2021-09-25
Remove the doublet/triplet number estimator for cuda seed finding 


beomki-yeo:
I used to pre-define (and overestimate) the size of doublet and triplet container based on the number of middle spacepoints per grid bin
Now it feels quite silly because I can give an exact number to the vector size from doublet and triplet counter objects.
Will fix it once related PRs are settled down

beomki-yeo:
Resolved by #110

PR #110, opened 2022-01-09
Remove multiplet number estimator and use the counter result to resize the multiplet container

beomki-yeo:
This PR addresses #93.
The size of multiplet (doublet, triplet, and seed) container used to be resized based on the empirical values obtained by cpu seed finding. The changes in this PR makes seed finding no longer rely on the empirical values but use the counting result to resize the container.

To make it possible, the counter objects had to be extended to include the more information on the number of multiplets.

Good thing is that the overall speedup gets improved because the size of doublet and triplet container is not over-estimated anymore:

Cuda seeding Elapsed time for 20 events
Before (1.4 sec) -> After (1.0 sec)


core/include/seeding/detail/doublet.hpp
```diff
@@ -12,7 +12,16 @@
...
 
-/// Header: unsigned int for the number of doublets per spacepoint bin
+/// Header: the number of doublets per spacepoint bin
+struct doublet_per_bin {
+    unsigned int n_doublets = 0;
+
+    TRACCC_HOST_DEVICE
+    unsigned int get_ref_num() const { return n_doublets; }
+
+    TRACCC_HOST
```
krasznaa, line 22: Why didn't you use `TRACCC_HOST_DEVICE` here? Even if this function is not used in device code for the moment, it is absolutely fit to be used there. So we might as well allow client code to use it there, if it so desires.

krasznaa (approved):
Formally I think this looks fine.

I was very confused by the `doublet_per_bin` and `triplet_per_bin` structs at first. Since those are really just an unnecessary complication over simply using an `unsigned int` by itself, like the code has done before.

I see that the interface must've been developed for the `dublet_counter_per_bin` struct at first, where a struct is actually meaningful. But... Is this uniform interface for these 3 structs actually important at any point? Did you write some generic code somewhere that would use these 3 types in a uniform way? I can't see that in the code... (Though I might just be missing it.)

As always, [C\+\+20 concepts](https://en.cppreference.com/w/cpp/language/constraints) would fit this type of code beautifully. Hopefully we'll get those in not too long...

core/include/seeding/detail/doublet.hpp
beomki-yeo: Hmm right. It's changed to TRACCC_HOST_DEVICE

beomki-yeo:
> Did you write some generic code somewhere that would use these 3 types in a uniform way?

Yeah. They have uniform structure to use `get_ref_num()` `cuda/utils/cuda_helper.cuh` :p

krasznaa (approved):
Once #108 is merged in, this should still be updated to include those changes first. That should also give the CI a chance to fix the (hopefully temporary) hiccup that occurred in the last run.

But apart from these technicalities, I'm on board with this PR as well.

beomki-yeo:
I want to merge this - but why `review-required: main` is not satisfied although it was already approved? @paulgessinger

paulgessinger:
The review through the bot gets invalidated when you add new commits to the branch that are not just update commits.

paulgessinger (approved):

paulgessinger:
Should be green now

beomki-yeo:
Thanks!

PR #119, opened 2022-01-31
fix bug in seed selecting

beomki-yeo:
Can't believe that I spent a whole day to fix this bug
There was a non-trivial discrepancy (1~3%) between cpu and cuda seeding, especially for ttbar 100 pileups, which I had no clue. 
I finally spotted the culprit XD 
Anyway I feel like to improve comments in cuda seeding codes at some point... hard to figure out what I have done
I attached print outputs below. 

<details>
  <summary>Click to expand!</summary>

### ttbar <100>, float

event 0
 seed matching rate: 1
 track parameters matching rate: 1
event 1
 seed matching rate: 1
 track parameters matching rate: 1
event 2
 seed matching rate: 1
 track parameters matching rate: 1
event 3
 seed matching rate: 1
 track parameters matching rate: 1
event 4
 seed matching rate: 1
 track parameters matching rate: 1
event 5
 seed matching rate: 1
 track parameters matching rate: 1
event 6
 seed matching rate: 1
 track parameters matching rate: 1
event 7
 seed matching rate: 1
 track parameters matching rate: 1
event 8
 seed matching rate: 1
 track parameters matching rate: 1
event 9
 seed matching rate: 1
 track parameters matching rate: 1
event 10
 seed matching rate: 1
 track parameters matching rate: 1
event 11
 seed matching rate: 1
 track parameters matching rate: 1
event 12
 seed matching rate: 1
 track parameters matching rate: 1
event 13
 seed matching rate: 1
 track parameters matching rate: 1
event 14
 seed matching rate: 1
 track parameters matching rate: 1
event 15
 seed matching rate: 1
 track parameters matching rate: 1
event 16
 seed matching rate: 1
 track parameters matching rate: 1
event 17
 seed matching rate: 1
 track parameters matching rate: 1
event 18
 seed matching rate: 1
 track parameters matching rate: 1
event 19
 seed matching rate: 1
 track parameters matching rate: 1
event 20
 seed matching rate: 1
 track parameters matching rate: 1
event 21
 seed matching rate: 0.999608
 track parameters matching rate: 0.999608
event 22
 seed matching rate: 0.999557
 track parameters matching rate: 0.999557
event 23
 seed matching rate: 1
 track parameters matching rate: 1
event 24
 seed matching rate: 0.999735
 track parameters matching rate: 0.999735
event 25
 seed matching rate: 1
 track parameters matching rate: 1
event 26
 seed matching rate: 1
 track parameters matching rate: 1
event 27
 seed matching rate: 1
[...]

PR #120, opened 2022-01-31
SYCL Seed finding, Track param estimation and Spacepoint binning

konradkusiak97:
This PR aims at introducing the SYCL versions of Seed finding, Track param estimation and Spacepoint binning algorithms. 

Most of the code is derived from the corresponding CUDA versions in the [traccc/device/cuda](https://github.com/acts-project/traccc/tree/main/device). 

The current state of the code lets it successfully run the `seeding_example_sycl` and `seq_example_sycl` tests on NVIDIA cards when compiled for CUDA backend. 

## Issues
1. Right now, the code doesn't perform 100% correctly on Intel devices
2. The code has to be built with TRACCC_ALGEBRA_PLUGINS=ARRAY option 

pinging @krasznaa


device/sycl/src/seeding/counting_grid_capacities.sycl
```diff
@@ -0,0 +1,128 @@
...
+    vecmem::memory_resource& resource, ::sycl::queue* q) {
+
+    auto grid_view = get_data(grid, resource);
+    auto spacepoints_view = get_data(spacepoints, &resource);
+    auto sp_container_indices_view = vecmem::get_data(sp_container_indices);
+    auto grid_capacities_view = vecmem::get_data(grid_capacities);
+
+    // The local number of threads per work-group (block in cuda) (number taken
+    // from cuda implementation, warp size *2)
+    unsigned int localSize = 32 * 4;
```
krasznaa, line 111: We will need to find a better way of establishing these values using the information provided by SYCL about the device that we are running on... :thinking:

examples/algorithms/sycl/include/sycl/track_finding/seeding_algorithm.hpp
```diff
@@ -0,0 +1,72 @@
...
+        sf = std::make_shared<traccc::sycl::seed_finding>(
+            traccc::sycl::seed_finding(m_config, sb->nbins(), mr, q));
+    }
+
+    output_type operator()(
+        host_spacepoint_container&& spacepoints) const override {
+        output_type seeds({host_seed_container(1, &m_mr.get())});
+        auto internal_sp_g2 = sb->operator()(std::move(spacepoints));
+        seeds =
+            sf->operator()(std::move(spacepoints), std::move(internal_sp_g2));
```
krasznaa, line 58: ? You `std::move` the `spacepoints` variable twice. This is **very** undefined behaviour. :frowning: And it nicely shows the pitfalls of being way too enthusiastic with moving objects around... (Khm... @stephenswat... khm...) Please resolve this.

And at the same time you should simplify the incantations here a bit.

```c++
auto internal_sp_g2 = sb(spacepoints);
seeds = sf(spacepoints, internal_sp_g2);
```

should probably work, no?

examples/algorithms/sycl/include/sycl/track_finding/seeding_algorithm.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+        sb = std::make_shared<traccc::sycl::spacepoint_binning>(
+            traccc::sycl::spacepoint_binning(m_config, m_grid_config, mr, q));
+        sf = std::make_shared<traccc::sycl::seed_finding>(
+            traccc::sycl::seed_finding(m_config, sb->nbins(), mr, q));
+    }
+
+    output_type operator()(
+        host_spacepoint_container&& spacepoints) const override {
+        output_type seeds({host_seed_container(1, &m_mr.get())});
```
krasznaa, line 55: :confused: This should just be

```c++
output_type seeds{1, &m_mr.get()};
```

, no?

examples/run/sycl/seeding_example_sycl.sycl
```diff
@@ -0,0 +1,261 @@
...
+
+// Selector to pass to sycl::queue object for running on NVIDIA device
+class CUDADeviceSelector : public ::sycl::device_selector {
+    public:
+    int operator()(const ::sycl::device& device) const override {
+        return device.get_platform().get_backend() == ::sycl::backend::cuda
+                   ? 1
+                   : -1;
+    }
+};
```
krasznaa, line 37: Please remove this custom selector from the code. There's no need to use any custom selector here. We can influence what device the application should run on, using the `SYCL_DEVICE_FILTER` environment variable.

examples/run/sycl/seq_example_sycl.sycl
```diff
@@ -0,0 +1,294 @@
...
+
+// Selector to pass to sycl::queue object for running on NVIDIA device
+class CUDADeviceSelector : public ::sycl::device_selector {
+    public:
+    int operator()(const ::sycl::device& device) const override {
+        return device.get_platform().get_backend() == ::sycl::backend::cuda
+                   ? 1
+                   : -1;
+    }
+};
```
krasznaa, line 38: Please remove this custom device selector as well.

krasznaa (changes requested):
There are unfortunately a good number of things that I'm not super happy about. But since I'm keen on getting the code into the repository in some form at least, let's come back to those in further PRs. (For instance the public headers of `traccc::sycl` will really need to be readable by the host compiler as well.)

We definitely still have a lot of work ahead of us, since the performance of the code at the moment, even under ideal circumstances, is pretty terrible.

```
[bash][Legolas]:build > ./bin/seeding_example_sycl tml_detector/trackml-detector.csv tml_hits/  1 0
Running ./seeding_example tml_detector/trackml-detector.csv tml_hits/  1 0
event 0
 seed matching rate: 0.996299
 track parameters matching rate: 0.999026
==> Statistics ... 
- read    48109 spacepoints from 0 modules
- created        0 internal spacepoints
- created (cpu)  5134 seeds
- created (sycl) 5134 seeds
==> Elpased time ... 
wall time           3.33076   
hit reading (cpu)   0.568527  
seeding_time (cpu)  0.599433  
seeding_time (sycl) 1.76794   
tr_par_esti_time (cpu)    0.00162376
tr_par_esti_time (sycl)   0.0838475 
[bash][Legolas]:build >
```

But again, let's come back to that in subsequent PRs. For now let's just fix the technical issues that I pointed out.

device/sycl/include/sycl/seeding/detail/sycl_helper.hpp
```diff
@@ -0,0 +1,141 @@
...
+            ::sycl::shift_group_left(sg, array[workItemIdx], 1);
+
+        ::sycl::group_barrier(workGroup);
+
+        if (workItemIdx == 0) {
+            for (int i = 1; i < groupDim / 32; i++) {
+                array[workItemIdx] += array[i * 32];
+            }
+        }
+    }
```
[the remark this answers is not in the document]
konradkusiak97: I will play with this function a little more and try making it work on the Intel devices, but this might have to be replaced by some different reduction techniques. Although, other ones that I've implemented so far don't fully work on different platforms either :confused:

device/sycl/src/seeding/counting_grid_capacities.sycl
```diff
@@ -0,0 +1,128 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#define EIGEN_DONT_VECTORIZE
```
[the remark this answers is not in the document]
konradkusiak97: Alright, I will remove those.

examples/algorithms/sycl/include/sycl/track_finding/seeding_algorithm.hpp
konradkusiak97: So here, both operators in `seed_finding` and `spacepoint_binning` accept an R-value reference, so that's why there is `std::move` applied on them, and both the `sb` and `sf` are shared pointers to the `spacepoint_binning` and `seed_finding` functors, so directly calling `sb(spacepoints` and `sf(spacepoints, internal_sp_g2)` does not really work on them - this was taken directly from the CUDA code :stuck_out_tongue:

device/sycl/include/sycl/seeding/detail/sycl_helper.hpp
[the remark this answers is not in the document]
krasznaa: Once this version of the code is in, it should become easier for others (like me :wink:) to play with this as well. We'll just do that...

examples/algorithms/sycl/include/sycl/track_finding/seeding_algorithm.hpp
krasznaa: Both of those functors/algorithms are yours. If the CUDA code defined them with rvalue references, that should be a problem for that code. But in the SYCL version let's just ingest const lvalue references in those algorithms instead. It will make the memory management of the code a whole lot simpler.

For the smart pointers: Yeah, I forgot about that. Though I'm also not entirely sure why we need to use smart pointers to instantiate these functors, and can't just do it by value? :confused:

In any case, as long as these are smart pointer variables, you could write:

```c++
auto internal_sp_g2 = (*sb)(spacepoints);
seeds = (*sf)(spacepoints, internal_sp_g2);
```

konradkusiak97: I'm having some troubles :stuck_out_tongue: I changed passing the `spacepoints` to the `const &` but then inside the `seed_finding.hpp`, inside  the functions that have the kernel calls inside them (such as [doublet_counting](https://github.com/konradkusiak97/traccc/blob/9f005a70144b3024337336b811e227ab7b05d8e4/device/sycl/src/seeding/doublet_counting.sycl)), I'm doing:
```cpp
void doublet_counting(const seedfinder_config& config, const sp_grid& internal_sp,
                      host_doublet_counter_container& doublet_counter_container,
                      vecmem::memory_resource& resource, ::sycl::queue* q) {
...
    auto internal_sp_view = get_data(internal_sp, resource);
```
And it doesn't allow me to call `get_data()` on the `const sp_grid&` type
```shell
[ 94%] Building SYCL object device/sycl/CMakeFiles/traccc_sycl.dir/src/seeding/populating_grid.sycl.o
/home/kkusiak/traccc-test/traccc/device/sycl/src/seeding/doublet_counting.sycl:150:29: error: no matching function for call to 'get_data'
    auto internal_sp_view = get_data(internal_sp, resource);
                            ^~~~~~~~
/home/kkusiak/traccc-test/traccc/core/include/edm/container.hpp:369:41: note: candidate template ignored: could not match 'container' against 'grid2'
inline container_data<header_t, item_t> get_data(
                                        ^
/home/kkusiak/traccc-test/build_nvidia/_deps/detray-src/core/include/detray/grids/grid2.hpp:477:1: note: candidate template ignored: deduced type 'grid2<...>' of 1st parameter does not match adjusted type 'const grid2<...>' of argument [with populator_type = detray::attach_populator, axis_p0_type = detray::axis::circular, axis_p1_type = detray::axis::regular, serializer_type = detray::serializer2, vector_type = detray::dvector, jagged_vector_type = detray::djagged_vector, array_type = detray::darray, tuple_type = detray::dtuple, value_type = traccc::internal_spacepoint<traccc::spacepoint>, kSORT = false, kDIM = 1]
get_data(grid2<populator_type, axis_p0_type, axis_p1_type, serializer_type,
^
/home/kkusiak/traccc-test/build_nvidia/_deps/detray-src/core/include/detray/grids/axis.hpp:656:54: note: candidate function template not viable: requires single argument 'axis', but 2 arguments were provided
inline axis_data<axis_type<array_type, vector_type>> get_data(
```
which is defined in [traccc/core/include/edm/container.hpp](https://github.com/konradkusiak97/traccc/blob/9f005a70144b3024337336b811e227ab7b05d8e4/core/include/edm/container.hpp)
I have the recent changes in the latest [commit](https://github.com/acts-project/traccc/pull/120/commits/9f005a70144b3024337336b811e227ab7b05d8e4)

krasznaa: Hmm... Unfortunately this highlights a shortcoming in the current code. :frowning: It comes from a pretty nuanced issue in how vector views work in [vecmem](https://github.com/acts-project/vecmem).

The "problem" is sort of summarised in this code:

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/vector.hpp#L30-L37

```c++
/// Helper function creating a @c vecmem::data::vector_view object
template <typename TYPE, typename ALLOC>
VECMEM_HOST data::vector_view<TYPE> get_data(std::vector<TYPE, ALLOC>& vec);

/// Helper function creating a @c vecmem::data::vector_view object
template <typename TYPE, typename ALLOC>
VECMEM_HOST data::vector_view<const TYPE> get_data(
    const std::vector<TYPE, ALLOC>& vec);
```

You see that we defined 2 `get_data(...)` functions for `vecmem::vector`. The first one is simple, it translates a non-const `vecmem::vector<T>` object into a `vecmem::data::vector_view<T>` one. But for a `const vecmem::vector<T>` object we can't just make a `const vecmem::data::vector_view<T>` object. Since we pass the view objects around mostly by value, that const declaration would be lost very quickly.

For this reason a constant vector is represented by a `vecmem::data::vector_view<const T>` data type. Hardcoding in the type that we are not allowed to modify the vector's elements through this view.

Once you switched to passing the spacepoint grid as a const reference to the function, the currently defined `get_data(...)` function (coming [from detray](https://github.com/acts-project/detray/blob/main/core/include/detray/grids/grid2.hpp#L457-L482)) became unable to return a proper view type. :frowning: @beomki-yeo and @niermann999, we will need to sort this out in the detray code.

For now let's just do the easy/dirty thing in this PR. Remove `const` from the lvalue reference, and just give a non-const reference to this function. It's a pretty dirty setup for the code, but let's be pragmatic. Once the "infrastructure code" becomes smarter, we'll be able to make the code more const-correct.

konradkusiak97: Hm, but can I actually do that? :thinking: Because the `seed_finding` and `spacepoint_binning` algorithms are inheriting from the `algorithm` class defined in [traccc/core/include/utils/algorithm.hpp](https://github.com/acts-project/traccc/blob/main/core/include/utils/algorithm.hpp) where it ensures that the `rvalue reference` or the ` const lvalue reference` is being used. I'm getting failures of `static_assert` when I'm trying to just use the `lvalue reference` :confused:

krasznaa: Then let's keep the interface with the const reference, and do another type of ugly hack. `const_cast` away the constness of the reference inside of that function.

A safer option would be to make a copy of the object inside of this function. But the code is slow enough as is...

core/include/seeding/spacepoint_binning_helper.hpp
```diff
@@ -67,14 +72,17 @@ inline TRACCC_HOST_DEVICE size_t is_valid_sp(const seedfinder_config& config,
     if (sp.z() > config.zMax || sp.z() < config.zMin) {
         return detray::invalid_value<size_t>();
     }
-    scalar spPhi = std::atan2(sp.y(), sp.x());
+    scalar spPhi = algebra::math::atan2(sp.y(), sp.x());
     if (spPhi > config.phiMax || spPhi < config.phiMin) {
         return detray::invalid_value<size_t>();
     }
-
+#if defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
```
beomki-yeo, line 79: maybe we can use algebra::math::perp here and avoid if macro?

core/include/seeding/track_params_estimation_helper.hpp
```diff
@@ -78,7 +90,7 @@ inline TRACCC_HOST_DEVICE bound_vector seed_to_bound_vector(
     scalar B = uv2[1] - A * uv2[0];
 
     // Curvature (with a sign) estimate
-    scalar rho = -2.0 * B / std::hypot(1., A);
+    scalar rho = -2.0 * B / math_ns::hypot(float(1.), A);
```
beomki-yeo, line 93: As mentioned above, we can consider `algebra::math::perp`

krasznaa: Std::hypot and std::perp are not the same things though. We would have to teach algebra-plugins about std::hypot first.

examples/algorithms/sycl/include/sycl/track_finding/seeding_algorithm.hpp
beomki-yeo: > Though I'm also not entirely sure why we need to use smart pointers to instantiate these functors

 When I tried to use value, it required copy assignment operator for functors. For example;
` sb = traccc::spacepoint_binning(m_config, m_grid_config, mr)); // need copy assignment operator`
 
 and yeah i was too lazy to add that

> where it ensures that the rvalue reference or the const lvalue reference is being used

Yeah, that's why I used rvalue in CUDA code. It seems not so desirable pattern according to @krasznaa 's explanation

beomki-yeo:
It doesn't have to be sorted out in this PR - but some files like `doublet_counter.hpp` can be shared between cuda and sycl implementation to avoid having same copies

core/include/seeding/track_params_estimation_helper.hpp
beomki-yeo: They don't make a same output?

device/sycl/src/seeding/seed_selecting.sycl
```diff
@@ -0,0 +1,313 @@
...
+
+    // -- localSize
+    // The dimension of workGroup (block) is the integer multiple of WARP_SIZE
+    // (=32)
+    unsigned int localSize = 64;
+    // Calculate the global number of threads to run in kernel
+    unsigned int num_groups = 0;
+    for (unsigned int i = 0; i < internal_sp.nbins(); ++i) {
+        num_groups +=
+            triplet_counter_container.get_headers()[i].n_mid_bot / localSize +
```
beomki-yeo, line 289: @konradkusiak97 you may want to check #119 to fix a notorious bug that I had... :)

core/include/seeding/track_params_estimation_helper.hpp
krasznaa: :facepalm: I was mistaking `algebra::getter::perp` for one of the functions I added recently in:

https://github.com/acts-project/algebra-plugins/blob/main/math/common/include/algebra/math/common.hpp

Since I did check that none of those was `std::hypot`.

You're right, the `perp(...)` function from algebra-plugins does calculate this. But that function always works on the `vector2` (or `vector3`) type of whatever storage we are using. Still... maybe you're right. Maybe a more elegant thing would be to replace this line with:

```c++
scalar rho = -2.0 * B / algebra::getter::perp({1., A});
```

:thinking: Could you give that a try Konrad?

device/sycl/src/seeding/seed_selecting.sycl
konradkusiak97: Right, I will fix that... thanks! :wink:

krasznaa:
> It doesn't have to be sorted out in this PR - but some files like `doublet_counter.hpp` can be shared between cuda and sycl implementation to avoid having same copies

Since we should keep the two implementations in sync, I very much agree. Let's not do it in this PR. But later on we should investigate how we could share as much code as possible between the CUDA and SYCL implementations.

core/include/seeding/track_params_estimation_helper.hpp
konradkusiak97: Unfortunately it doesn't work out of the box: 
```shell
In file included from /home/kkusiak/traccc-test/traccc/core/include/seeding/track_params_estimation.hpp:10,
                 from /home/kkusiak/traccc-test/traccc/examples/run/cpu/seq_example.cpp:18:
/home/kkusiak/traccc-test/traccc/core/include/seeding/track_params_estimation_helper.hpp: In function ‘traccc::bound_vector traccc::seed_to_bound_vector(const spacepoint_container_t&, const traccc::seed&, const vector3&, traccc::scalar)’:
/home/kkusiak/traccc-test/traccc/core/include/seeding/track_params_estimation_helper.hpp:93:58: error: no matching function for call to ‘perp(<brace-enclosed initializer list>)’
     scalar rho = -2.0 * B / algebra::getter::perp({1., A});
                                                          ^
In file included from /home/kkusiak/traccc-test/build_nvidia/_deps/algebraplugins-src/math/cmath/include/algebra/math/impl/cmath_cylindrical2.hpp:11,
                 from /home/kkusiak/traccc-test/build_nvidia/_deps/algebraplugins-src/math/cmath/include/algebra/math/cmath.hpp:12,
                 from /home/kkusiak/traccc-test/build_nvidia/_deps/algebraplugins-src/frontend/array_cmath/include/algebra/array_cmath.hpp:11,
                 from /home/kkusiak/traccc-test/traccc/plugins/algebra/include/traccc/array/include/plugins/algebra/array_definitions.hpp:13,
                 from /home/kkusiak/traccc-test/traccc/core/include/definitions/primitives.hpp:11,
                 from /home/kkusiak/traccc-test/traccc/io/include/io/csv.hpp:19,
                 from /home/kkusiak/traccc-test/traccc/examples/run/cpu/seq_example.cpp:11:
/home/kkusiak/traccc-test/build_nvidia/_deps/algebraplugins-src/math/cmath/include/algebra/math/impl/cmath_getter.hpp:51:37: note: candidate: ‘template<class size_type, template<class, size_type <anonymous> > class array_t, class scalar_t, size_type N, typename std::enable_if<(N >= 2), bool>::type <anonymous> > scalar_t algebra::cmath::perp(const array_t<scalar_t, N>&)’
 ALGEBRA_HOST_DEVICE inline scalar_t perp(
                                     ^~~~
/home/kkusiak/traccc-test/build_nvidia/_deps/algebraplugins-src/math/cmath/include/algebra/math/impl/cmath_getter.hpp:51:37: note:   template argument deduction/substitution failed:
In file included from /home/kkusiak/traccc-test/traccc/core/include/seeding/track_params_estimation.hpp:10,
                 from /home/kkusiak/traccc-test/traccc/examples/run/cpu/seq_example.cpp:18:
/home/kkusiak/traccc-test/traccc/core/include/seeding/track_params_estimation_helper.hpp:93:58: note:   couldn't deduce template parameter ‘size_type’
     scalar rho = -2.0 * B / algebra::getter::perp({1., A});
```

beomki-yeo: Lookes like `brace-enclosed initializer list` is not deduced. Have you tried sth like `vector2{1.,A}` instead?

konradkusiak97: That indeed works :+1:

examples/run/sycl/seq_example_sycl.sycl
konradkusiak97: That's done

examples/run/sycl/seeding_example_sycl.sycl
konradkusiak97: Done.

examples/algorithms/sycl/include/sycl/track_finding/seeding_algorithm.hpp
konradkusiak97: Done.

examples/algorithms/sycl/include/sycl/track_finding/seeding_algorithm.hpp
konradkusiak97: That's also done.

device/sycl/src/seeding/counting_grid_capacities.sycl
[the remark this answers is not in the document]
konradkusiak97: Done.

krasznaa (approved):
As long as the CI is happy (note that we will need to add CI tests for the SYCL code!), let's go ahead like this. We'll fix things in subsequent PRs.

PR #121, opened 2022-02-01
Change project status

beomki-yeo:


krasznaa (commented):
I don't think we should change the status of these things yet. :thinking: The SYCL code is very far from working reliably at the moment.

For now I'd leave it as "under constructions" for a little longer.

beomki-yeo:
OK. I demoted them now

krasznaa (approved):
Sure, this we can do...

PR #122, opened 2022-02-01
Code Reorganisation, main branch (2022.02.01.)

krasznaa:
This is one... painful... PR...

I spent some time today re-organising the code of the repository a little. Making sure that all headers of the project would live under the `traccc/` prefix.

As in the case of the other repositories, I did this by introducing a helper function. (`traccc_add_library(...)`) That should later on allow us to conveniently export the project's configuration with CMake as well, but I didn't bother with that in this PR.

Most of the changes were moving files around, and adjusting include paths. I did not do a really careful cleanup of the includes. While I noticed many bad patterns in how files are included, those should be fixed in future PRs. Here I just wanted to move files to "their correct locations".

The noteworthy changes are:
  - A slightly larger re-organisation of the algebra plugin headers. Now they are laid out much like in [detray](https://github.com/acts-project/detray);
  - I simplified the setup of `traccc::cuda` a bit. Merging it into a single library, and  removing the necessity for using `CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES` with it.
    * A "more robust fix" for the setup of that library will come in a future PR.

As scary as this PR is, it would be good to push it through quickly, as any other PR will conflict with it. (And vice versa...)

Pinging @konradkusiak97.


beomki-yeo (approved):

beomki-yeo:
Let's go with this. We can do further cleanup in the next PRs as @krasznaa mentioned

PR #124, opened 2022-02-02
Use `vecmem::vector` for seed container

beomki-yeo:
In this PR, the container type of seed container is changed to `vecmem::vector` from `vecmem::jagged_vector`.
This PR also updates vecmem version to v0.9.0 due to the issue brought up in acts-project/vecmem#159 

The reason for using `vector_buffer` as a kernel input is that I always overestimate the number of seeds (there is no seed_counting kernel)

@konradkusiak97 Once this PR is merged, could you fix SYCL version as well?


konradkusiak97:
Sure! I will update the SYCL code once this is in :+1:

krasznaa:
Let's wait with this for just a moment... I'm about to open a PR that would introduce CI testing for the SYCL code. I wanted to keep this PR back until that is in. To make sure that it would, for sure, not break the SYCL code.

Since it's okay to update aspects of the CUDA and SYCL codes separately for now. (Eventually we shouldn't really even do that...) But it's definitely not okay for one update to break the other...

beomki-yeo:
No problem!

krasznaa:
Unfortunately now it needs to be adapted to all of the other changes. But hopefully that shouldn't be too difficult.

beomki-yeo:
No worries. Installing sycl on my laptop would be only difficult part XD

krasznaa:
You should probably make use of the Docker image that the CI uses.

```
docker run -it --rm -v $PWD:$PWD -w $PWD --user "$(id -u):$(id -g)" ghcr.io/acts-project/ubuntu2004_oneapi:v16.1
```

That's how I debugged the 2021.2.0 issues as well, not with a local installation...

beomki-yeo:
OKay thanks for suggestion

beomki-yeo:
I decided to run it on my laptop rather than docker.
@konradkusiak97 @krasznaa I have some questions 

- How can I make sycl find the device?
```
[beomki@localhost traccc_build]$ ./bin/traccc_seq_example_sycl tml_detector/trackml-detector.csv ../../acts/acts_data/data/digi_generic/ttbar_mu200/ 10 1
Running ./seq_example tml_detector/trackml-detector.csv ../../acts/acts_data/data/digi_generic/ttbar_mu200/  10 1 
terminate called after throwing an instance of 'cl::sycl::runtime_error'
  what():  No device of requested type available. Please check https://software.intel.com/content/www/us/en/develop/articles/intel-oneapi-dpcpp-system-requirements.html -1 (CL_DEVICE_NOT_FOUND)
```  

- Why is the compilation with oneAPI too slow in my laptop? Is it only for me?

beomki-yeo:
Hmm here is the env setup that I am using. any recommendation regarding my question?
```
source /opt/intel/oneapi/setvars.sh
export CXX=`which clang++`
export SYCLCXX=`which dpcpp`
export SYCL_DEVICE_FILTER=cuda:gpu:1
export LDFLAGS="-L${CMPLR_ROOT}/linux/compiler/lib/intel64_lin -lirc"
```

krasznaa:
You cannot use oneAPI (dpcpp) to build code for your NVIDIA card. You need a hand-built version of intel/llvm for that.

You could either follow my instructions from mattermost for that, or the significantly simpler ones from: https://codeplay.com/solutions/oneapi/for-cuda/

Once you managed to compile the code with

SYCLCXX="clang++ -fsycl -fsycl-targets=nvptx64-nvidia-cuda"

, you'll be able to use

SYCL_DEVICE_FILTER=cuda

to force the usage of your card. (You only need your elaborate formalism if you have multiple NVIDIA cards in your system.)

Attila

beomki-yeo:
Okay. traccc_seq_example_sycl runs very well

@konradkusiak97 @krasznaa The performance is much better than you described earlier:

./bin/traccc_seq_example_sycl tml_detector/trackml-detector.csv tml_pixels/ 10 1

krasznaa (approved):
I like the updates. Since today I will want to re-arrange the SYCL source files a bit (to make `traccc_sycl` only privately depend on SYCL), let's just go forward with it as-is.

As for your performance numbers Beomki: Konrad did correct me privately that it **can** indeed happen that on a "cold run" the SYCL code performs quite badly, but on a "warm run" (on let's say a 2nd run of the test in quick succession after the first) the performance would be much better.

Interestingly enough the CUDA executables can behave the same. It seems that the NVIDIA driver has some peculiarities like this. That in some cases it may take it a moment to get the GPU ready for work.

Long story short, the performance of the SYCL code is indeed not as bad as I first feared. As for why the CPU code would perform differently: We'll have to see whether there may be an issue with memory resource usage on the host side in these examples, similar to what we discussed in https://github.com/acts-project/detray/pull/189. But it may well be just a statistical fluctuation as well. We'll have to see...

PR #125, opened 2022-02-02
Executable Reorganisation, main branch (2022.02.02.)

krasznaa:
This is a follow-up from #122, cleaning up the CMake configuration / build of the executables in the project as well.

The PR (re-)introduces 2 functions:
  - `traccc_add_test(...)` is moved to `cmake/traccc-functions.cmake` for a bit more consistency, and is set up to use [gtest_discover_tests(...)](https://cmake.org/cmake/help/git-stage/module/GoogleTest.html#command:gtest_discover_tests). (Much like we did in [vecmem](https://github.com/acts-project/vecmem).
  - `traccc_add_executable(...)` is meant to help with setting up test/benchmark executables in a convenient way.
    * Note that I do not consider executables added with it as unit tests. So I removed some of the [add_test(...)](https://cmake.org/cmake/help/latest/command/add_test.html) calls from the code.
    * This function should allow us to export the configuration of the executables together with the project later on. In case that would ever be necessary.

I chose to re-categorise `seq_single_module` into a unit test, and added a very basic GTest facade around the existing code. Later on it should be made a bit better though, with the code actually testing whether the results are what we would expect.

I re-organised the setup of the "common test headers" a little. The setup was just a little over-complicated for me previously.

Finally, I switched from referring to the GoogleTest libraries with `gtest_main` to `GTest::gtest_main`, so that the build would work also when GoogleTest is "found on the system".

P.S. The names of the executables and tests were also harmonised to have common prefixes. Like we do in the other projects as well. :smile:


stephenswat (approved):
I can't really say I understand the rationale for moving the `add_test` function into the global CMake definitions space when it's only ever needed in the testing subdirectory. This seems to complicate matters unnecessarily, and it makes it harder to build the tests as a stand-alone project.

Also I don't really see why the sequential example needs to be a test? It's just there to provide an example of how algorithms work together in our project, putting it in as a test defeats that purpose, especially if we're not actually testing anything in there? :confused: 

Anyhow, there is also no harm done and we can always iterate if we need to do things differently in the future.

krasznaa:
> it makes it harder to build the tests as a stand-alone project.

I'm really not sure what you mean by that. :confused: Since the `<foo>_add_test` functions are declared in the project's main `cmake/` directory in the other projects (https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-functions.cmake#L71-L95, https://github.com/acts-project/algebra-plugins/blob/main/cmake/algebra-plugins-functions.cmake#L43-L64, https://github.com/acts-project/detray/blob/main/cmake/detray-functions.cmake#L37-L58), I thought this harmonisation would help the developers.

> Also I don't really see why the sequential example needs to be a test? It's just there to provide an example of how algorithms work together in our project, putting it in as a test defeats that purpose, especially if we're not actually testing anything in there? :confused:

That executable seems like the easiest to turn into an actual test from all of them. Note that the executable is defined as a test already now. Along with a bunch of the other executables, which I stopped from being tests in this PR. It was just this one executable that I thought could in principle work as a test. Since the input data is defined inside of the source code itself.

But if you don't want to use it as a test, I'm happy to move it back under `examples/run`, and just declare it as a non-test executable.

stephenswat:
No, I'm not really arguing for that. We can always copy it back to examples later, this is not a problem as far as I am concerned. Since the CI passes, let's just move ahead.

PR #112, opened 2022-01-19
BUG : target_link_libraries for a .sycl not working

SylvainJoube:
Hi,

#106 solved my SYCL compilation issue, I can now compile the `traccc_sycl_language_example` program from the `.sycl` code and run it. But once I try to link any internal library, I got a compilation error.

Maybe I'm missing something again, but to reproduce this error I only need to change the `examples/run/sycl/CMakeLists.txt` and add after the `add_executable( traccc_sycl_language_example "sycl_language_example.sycl" )` line something like `target_link_libraries( traccc_sycl_language_example LINK_PUBLIC traccc::algorithms )` or any other library really, to have the same error. Not the same list of included files, but exactly the same `fatal error: 'stdlib.h' file not found`.

Here is the output I got every time :

```log
[ 94%] Linking CXX executable ../../../bin/seq_single_module
[ 94%] Built target seq_single_module
Scanning dependencies of target traccc_sycl_language_example
[ 94%] Building SYCL object examples/run/sycl/CMakeFiles/traccc_sycl_language_example.dir/sycl_language_example.sycl.o
In file included from /home/data_sync/academique/These/Traccc/traccc_sycl_link_bug_with_link_updated/examples/run/sycl/sycl_language_example.sycl:9:
In file included from /home/sylvain/intel/oneapi/compiler/2021.4.0/linux/bin/../include/sycl/CL/sycl.hpp:11:
In file included from /home/sylvain/intel/oneapi/compiler/2021.4.0/linux/bin/../include/sycl/CL/sycl/accessor.hpp:12:
In file included from /home/sylvain/intel/oneapi/compiler/2021.4.0/linux/bin/../include/sycl/CL/sycl/atomic.hpp:11:
In file included from /home/sylvain/intel/oneapi/compiler/2021.4.0/linux/bin/../include/sycl/CL/__spirv/spirv_ops.hpp:13:
In file included from /home/sylvain/intel/oneapi/compiler/2021.4.0/linux/bin/../include/sycl/CL/sycl/detail/stl_type_traits.hpp:12:
In file included from /usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/iterator:66:
In file included from /usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/streambuf_iterator.h:35:
In file included from /usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/streambuf:41:
[...]

PR #126, opened 2022-02-02
Add SYCL CI, main branch (2022.02.02.)

krasznaa:
This turned out to be quite a bit more painful than I first realised. But finally, building @konradkusiak97's SYCL code is more or less working as part of the CI.

I had to do the following things:
  - Update to the latest version of [vecmem](https://github.com/acts-project/vecmem), to include https://github.com/acts-project/vecmem/pull/163, which also properly fixes the issue discussed in #112;
  - Switch to `TRACCC_ALGEBRA_PLUGINS=ARRAY` by default;
    * The SYCL code only works with this setting, so to simplify the CI, this seemed reasonable for the host and CUDA code as well as a default;
    * I also noticed along the way that not every plugin works even with the host code at the moment. So a larger cleanup will be needed later on.
  - I had to force the oneAPI compiler to link everything against `libirc.so`;
    * This is to work around an issue with Acts using `--coverage` in Debug builds. An issue that is present with oneAPI 2021.4.0 as well. (I also tried to use 2022.1.0, but that one ran into other issues. So at that point I just gave up for now.)
  - I had to modify parts of the SYCL code to make it compatible with oneAPI 2021.2.0;
    * Switching the `::sycl::group_barrier(workGroup);` calls to `item.barrier();` will undoubtedly make things less efficient, but probably not by much;
    * The partial summing written by Konrad does not work in oneAPI 2021.2.0. So for now I switched to the simpler / slower code. We will need to re-think that part of the code anyway...

We **should** update to the latest version of oneAPI in a future PR. But for now this seemed to be the fastest way to get the CI working.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,7 +147,7 @@ if( TRACCC_SETUP_GOOGLETEST )
 endif()
 
 # option for algebra plugins (ARRAY EIGEN SMATRIX VC VECMEM)
-set(TRACCC_ALGEBRA_PLUGINS EIGEN CACHE STRING "Algebra plugin to use in the build")
+set(TRACCC_ALGEBRA_PLUGINS ARRAY CACHE STRING "Algebra plugin to use in the build")
 message(STATUS "Building with plugin type: " ${TRACCC_ALGEBRA_PLUGINS})
 add_definitions(-DALGEBRA_PLUGINS_INCLUDE_${TRACCC_ALGEBRA_PLUGINS})
 
--- a/device/sycl/src/seeding/triplet_counting.sycl
+++ b/device/sycl/src/seeding/triplet_counting.sycl
@@ -135,7 +135,8 @@ class TripletCount {
         }
 
         if (mt_end_idx >= mid_top_doublets_per_bin.size()) {
-            mt_end_idx = fmin(mid_top_doublets_per_bin.size(), mt_end_idx);
+            mt_end_idx =
+                ::sycl::min(mid_top_doublets_per_bin.size(), mt_end_idx);
         }
 
         if (mt_start_idx >= mid_top_doublets_per_bin.size()) {
--- a/device/sycl/src/seeding/triplet_finding.sycl
+++ b/device/sycl/src/seeding/triplet_finding.sycl
@@ -164,7 +164,8 @@ class TripletFind {
         }
 
         if (mt_end_idx >= mid_top_doublets_per_bin.size()) {
-            mt_end_idx = fmin(mid_top_doublets_per_bin.size(), mt_end_idx);
+            mt_end_idx =
+                ::sycl::min(mid_top_doublets_per_bin.size(), mt_end_idx);
         }
 
         if (mt_start_idx >= mid_top_doublets_per_bin.size()) {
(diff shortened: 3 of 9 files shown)


device/sycl/src/seeding/triplet_counting.sycl
```diff
@@ -135,7 +135,8 @@ class TripletCount {
         }
 
         if (mt_end_idx >= mid_top_doublets_per_bin.size()) {
-            mt_end_idx = fmin(mid_top_doublets_per_bin.size(), mt_end_idx);
+            mt_end_idx =
+                ::sycl::min(mid_top_doublets_per_bin.size(), mt_end_idx);
```
krasznaa, line 139: Ahh, I actually forgot about these... @beomki-yeo, I fear that @konradkusiak97 copied these `fmin(...)` calls from the CUDA code directly.

But the code compares two integers here. :confused: So `fmin(...)` is not appropriate. Even more weird, why do this behind an `if(...)` statement? Once we're in there, we **know** that the value has to be replaced, no? :confused:

krasznaa:
Forgot to ping @SylvainJoube on this as well. :wink:

device/sycl/src/seeding/triplet_finding.sycl
```diff
@@ -217,9 +218,10 @@ class TripletFind {
         }
         // Calculate the number of triplets per "block" with reducing sum
         // technique
-        ::sycl::group_barrier(workGroup);
+        // ::sycl::group_barrier(workGroup);
+        item.barrier();
```
konradkusiak97, line 222: Maybe, adding this would make the synchronization pretty much the same as before?
```cpp
item.barrier(::sycl::access::fence_space::local_space);
```

krasznaa: I had the same idea, but after looking at the oneAPI headers for a bit, I wasn't entirely sure that this would be a 1-to-1 replacement. So I went with this safer-looking global barrier in this PR...

device/sycl/src/seeding/triplet_counting.sycl
beomki-yeo: Indeed. That's probably one of dumb codes that I've written

CMakeLists.txt
```diff
@@ -147,7 +147,7 @@ if( TRACCC_SETUP_GOOGLETEST )
 endif()
 
 # option for algebra plugins (ARRAY EIGEN SMATRIX VC VECMEM)
-set(TRACCC_ALGEBRA_PLUGINS EIGEN CACHE STRING "Algebra plugin to use in the build")
+set(TRACCC_ALGEBRA_PLUGINS ARRAY CACHE STRING "Algebra plugin to use in the build")
```
beomki-yeo, line 150: This should be OK for now. But It's gonna be tricky when we implement KF which has bunch of matrix operations that doesn't exist in  `ARRAY` algebra-plugins. We need them ASAP regardless of the default plugin though

krasznaa: The CUDA code can still be compiled with the EIGEN plugin.

And I intend to make this possible for the SYCL code later on as well. :thinking: That will just need some changes in the CUDA code as well. Because the issue with Eigen at the moment is that with that plugin `traccc::vector2` is not trivially copyable. And as such, it can't be given to a SYCL kernel as an argument. (It's actually interesting that CUDA doesn't complain about this at all. Maybe there's some CUDA specific code in Eigen?)

This should all be possible to work around on our side. But the CUDA code will need to be adjusted as well after we change some of the core types.

beomki-yeo (approved):
If @konradkusiak97 is happy with the changes in sycl code, I am good as well

konradkusiak97:
Sure, I'm fine with that too

PR #128, opened 2022-02-04
Bug fix - initializing local memory in doublet finding

konradkusiak97:
This bug was spotted after trying to run the SYCL seeding algorithm in a loop:
```
Running ./seeding_example tml_detector/trackml-detector.csv tml_hits/  9 0
Running on device: NVIDIA GeForce RTX 3050 Ti Laptop GPU
event 0
 seed matching rate: 0.996104
 track parameters matching rate: 0.999026

PI CUDA ERROR:
	Value:           700
	Name:            CUDA_ERROR_ILLEGAL_ADDRESS
	Description:     an illegal memory access was encountered
	Function:        wait
	Source Location: /home/konradd/software/llvm/sycl/plugins/cuda/pi_cuda.cpp:449

terminate called after throwing an instance of 'cl::sycl::runtime_error'
  what():  Native API failed. Native API returns: -999 (Unknown OpenCL error code) -999 (Unknown OpenCL error code)
Aborted (core dumped)
```
The error here comes from `triplet_counting` kernel which is trying to be invoked with an arbitrarily high number of threads per block. That's because, on the second try, the `doublet_finding` gives out a strangely big number of doublets found. And it seems like this was caused by not initializing the arrays in local memory before accessing them. 

By fixing that, everything seems to work fine on the CUDA backend...for now :smiling_face_with_tear: 

pinging @krasznaa

### DIFF
--- a/device/sycl/src/seeding/doublet_finding.sycl
+++ b/device/sycl/src/seeding/doublet_finding.sycl
@@ -79,8 +79,10 @@ class DupletFind {
         auto mid_top_doublets_per_bin =
             mid_top_doublet_device.get_items().at(bin_idx);
 
-        auto num_mid_bot_doublets_per_thread = m_localMemBot.get_pointer();
-        auto num_mid_top_doublets_per_thread = m_localMemTop.get_pointer();
+        auto num_mid_bot_doublets_per_thread = m_localMemBot;
+        auto num_mid_top_doublets_per_thread = m_localMemTop;
+        num_mid_bot_doublets_per_thread[workItemIdx] = 0;
+        num_mid_top_doublets_per_thread[workItemIdx] = 0; 
         // Convenient alias for the number of doublets per thread
         auto& n_mid_bot_per_spM = num_mid_bot_doublets_per_thread[workItemIdx];
         auto& n_mid_top_per_spM = num_mid_top_doublets_per_thread[workItemIdx];
--- a/device/sycl/src/seeding/doublet_finding.sycl
+++ b/device/sycl/src/seeding/doublet_finding.sycl
@@ -82,7 +82,7 @@ class DupletFind {
         auto num_mid_bot_doublets_per_thread = m_localMemBot;
         auto num_mid_top_doublets_per_thread = m_localMemTop;
         num_mid_bot_doublets_per_thread[workItemIdx] = 0;
-        num_mid_top_doublets_per_thread[workItemIdx] = 0; 
+        num_mid_top_doublets_per_thread[workItemIdx] = 0;
         // Convenient alias for the number of doublets per thread
         auto& n_mid_bot_per_spM = num_mid_bot_doublets_per_thread[workItemIdx];
         auto& n_mid_top_per_spM = num_mid_top_doublets_per_thread[workItemIdx];


device/sycl/src/seeding/doublet_finding.sycl
```diff
@@ -79,8 +79,10 @@ class DupletFind {
         auto mid_top_doublets_per_bin =
             mid_top_doublet_device.get_items().at(bin_idx);
 
-        auto num_mid_bot_doublets_per_thread = m_localMemBot.get_pointer();
-        auto num_mid_top_doublets_per_thread = m_localMemTop.get_pointer();
+        auto num_mid_bot_doublets_per_thread = m_localMemBot;
+        auto num_mid_top_doublets_per_thread = m_localMemTop;
+        num_mid_bot_doublets_per_thread[workItemIdx] = 0;
+        num_mid_top_doublets_per_thread[workItemIdx] = 0;
```
krasznaa, line 85: Curious. Notice that you also changed the type of the `num_mid_bot_doublets_per_thread` and `num_mid_top_doublets_per_thread` variables for the rest of the function. :thinking: Which I'm happy with, because using the "SYCL objects" as smart pointers should be safer than to use the raw pointer given back by them.

This does bring up two things though:
  - Is it possible that the initialisation may not even be needed if the manipulation of this local memory is consistently done through the smart pointers?  Though initialising the memory explicitly to `0` certainly seems good.
  - Do you need to make a copy of these objects? I guess you do, since the operator on this functor is `const`. :thinking: I just don't like it all that much that we have to do that copy... But maybe this is indeed the best way to do this.

krasznaa (approved):
With all my bickering, I think this is good. We may just want to keep this part of the code in mind during later optimisations.

device/sycl/src/seeding/doublet_finding.sycl
konradkusiak97: Hmm, yes, the change from calling `get_pointer` sneaked in here. And just removing that didn't solve the problem, the `num_mid_bot_doublets_per_thread` and `num_mid_top_doublets_per_thread` had to be initialized to `0`. 
But that means that right now, the types of those arrays are:
```c++
 sycl::accessor<int, 1, sycl::access::mode::read_write,::sycl::access::target::local>;
```
whereas before (after calling `get_pointer`), they were of the type
```c++
sycl::multi_ptr<int, :sycl::access::target::local>
```
So is `sycl::accessor` equivalent to a smart pointer in this case and `sycl::multi_ptr` to the raw pointer?

It seems like I do make copies of those objects since as you've said, the `operator ()` is marked `const`, but I'm not really sure why the compiler doesn't allow me to change that :thinking:

krasznaa: :smile: It's part of the SYCL standard that the "executable" that we send to the device must be immutable. If it wasn't, if it was allowed to modify values held by the class, you could get the idea that threads would be allowed to communicate with each other through those values.

The member variables of these functors are basically behaving like arguments to a CUDA kernel. But while a CUDA compiler needs to invent its own language to handle global functions with a variadic list of arguments, SYCL kernels pass a variadic list of arguments to the kernel through such member variables using bog standard C\+\+.

But let's not delay it any longer. The code looks good like this.

PR #130, opened 2022-02-09
Add Seeding Performance Writer

beomki-yeo:
This PR is made on top of #133 

---------------------------------------------------------------------------------------------------------------------
I am opening the PR to collect opinions or comments beforehand.

The PR aims for adding performance writer with ROOT to make seeding optimization easier
One issue with ROOT is that it doesn't seem to support `FetchContent` as other externals do.
So I added `TRACCC_BUILD_EXAMPLES` with the default value of `FALSE` and made `examples` directory require ROOT dependency.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -25,7 +25,7 @@ jobs:
           - name: CUDA
             container: ghcr.io/acts-project/ubuntu2004_cuda:v16.1
           - name: SYCL
-            container: ghcr.io/acts-project/ubuntu2004_oneapi:v19
+            container: ghcr.io/acts-project/ubuntu2004_oneapi:v20
         build:
           - Release
           - Debug
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -25,7 +25,7 @@ jobs:
           - name: CUDA
             container: ghcr.io/acts-project/ubuntu2004_cuda:v16.1
           - name: SYCL
-            container: ghcr.io/acts-project/ubuntu2004_oneapi:v16.1
+            container: ghcr.io/acts-project/ubuntu2004_oneapi:v19
         build:
           - Release
           - Debug
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,6 +172,8 @@ add_subdirectory( plugins )
 if ( TRACCC_BUILD_EXAMPLES )
    # Find Boost.
    find_package( Boost REQUIRED COMPONENTS program_options)
+   # Find ROOT.
+   find_package( ROOT COMPONENTS RIO Tree Hist )
    add_subdirectory( examples )
 endif()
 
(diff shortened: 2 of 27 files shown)


stephenswat:
Not to be a party pooper, but do we really want to lock this project into ROOT? I understand that it is popular in HEP but it gives us a massive new dependency that can be pretty difficult to install, and it makes the output of our project way harder to use for anyone not intimately familiar with ROOT compared to, say, CSV or even something like Parquet.

beomki-yeo:
Bro, but installing ROOT is much easier than installing Boost...
And I am adding the ROOT dependency for plotting, but not for I/O. 
if people are not a big fan of root format file, I can make performance writer output to be pdf format

stephenswat:
Eh, it's fine. We can always add additional output writers for other output formats later. :smiling_face_with_tear:

beomki-yeo:
Hmm I managed to finalize `seeding_performance_writer` which (I hope) mimics the ACTS's one.
Meanwhile, seeding performance is very poor in many aspects.

1. The seeding performance of `seq_example` which follows hit clusterization does not give any reasonable result. The efficiency dropped to <1%. I guess it is related to seeding configuration - well I don't know TBH

2. The seeding performance of `seeding_example` without clusterization gives more reasonable result but it is still poor. Following is the result from only-pions events and ttbar event.

**4000 Pions**

image

**ttbat <200>**

image

@stephenswat @paulgessinger Before I investigate ACTS seeding configuration, do you guys have any comments or advice on it?

stephenswat:
Hi Beomki, your efficiency graphs look somewhat similar to [my graphs](https://github.com/acts-project/acts/pull/904#issuecomment-896318460) when I first did some seeding implementation. Back then it was a configuration failure. What detector geometry are you using? In any case, you may want to try increasing `rMax` to about 250, increase `deltaRMax` to about 200, and increase `maxSeedsPerSpM` to 20. See if that helps!

beomki-yeo:
Thanks for the answer. It's tml detector. I will try the values today

beomki-yeo:
@paulgessinger  @krasznaa Could you add ROOT dependency into SYCL?

.github/workflows/builds.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           - name: CUDA
             container: ghcr.io/acts-project/ubuntu2004_cuda:v16.1
           - name: SYCL
-            container: ghcr.io/acts-project/ubuntu2004_oneapi:v16.1
+            container: ghcr.io/acts-project/ubuntu2004_oneapi:root
```
krasznaa, line 28: @beomki-yeo, could you update the PR to `v19`? I think that should work for your changes, and would allow us to finally merge this thing in. :wink:

beomki-yeo: Alright. BTW, #133 should get merged first.

beomki-yeo: hmm it doesn't work...

stephenswat:
Is this ready to go in?

beomki-yeo:
Yes

stephenswat (approved):
Let's get this in then!

PR #132, opened 2022-02-10
Warning Fixes, main branch (2022.02.10.)

krasznaa:
This is meant as a replacement for #80, and is a follow-up from https://github.com/acts-project/algebra-plugins/pull/46 and https://github.com/acts-project/detray/pull/193. (Updated both externals in this PR to include those fixes.)

I re-organised the compiler setup code in the same way in which all the other projects have it set up by now. Separating the settings of the different languages into their own files, and including just the necessary ones in the appropriate directories. This way avoiding to pollute the builds of the externals with flags coming from this project.

Unfortunately I was not able to turn on `-Werror all-warnings` for CUDA just yet. :frowning: As I only found the following issue while working on this PR:

```
/data/ssd-1tb/projects/traccc/build/_deps/detray-src/core/include/detray/grids/grid2.hpp(183): warning #20011-D: calling a __host__ function("std::vector< ::traccc::internal_spacepoint< ::traccc::spacepoint> ,    ::std::pmr::polymorphic_allocator< ::traccc::internal_spacepoint< ::traccc::spacepoint> > > ::vector(const ::std::vector< ::traccc::internal_spacepoint< ::traccc::spacepoint> ,    ::std::pmr::polymorphic_allocator< ::traccc::internal_spacepoint< ::traccc::spacepoint> > > &)") from a __host__ __device__ function("detray::grid2< ::detray::attach_populator,  ::detray::axis::circular,  ::detray::axis::regular,  ::detray::serializer2,  ::detray::dvector,  ::detray::djagged_vector,  ::detray::darray,  ::detray::dtuple,  ::traccc::internal_spacepoint< ::traccc::spacepoint> , (bool)0, (unsigned int)1u> ::bin") is not allowed

/data/ssd-1tb/projects/traccc/build/_deps/detray-src/core/include/detray/grids/grid2.hpp(183): warning #20011-D: calling a __host__ function("std::vector<::std::vector< ::traccc::internal_spacepoint< ::traccc::spacepoint> ,    ::std::pmr::polymorphic_allocator< ::traccc::internal_spacepoint< ::traccc::spacepoint> > > ,    ::std::pmr::polymorphic_allocator<::std::vector< ::traccc::internal_spacepoint< ::traccc::spacepoint> ,    ::std::pmr::polymorphic_allocator< ::traccc::internal_spacepoint< ::traccc::spacepoint> > > > > ::at(unsigned long)") from a __host__ __device__ function("detray::grid2< ::detray::attach_populator,  ::detray::axis::circular,  ::detray::axis::regular,  ::detray::serializer2,  ::detray::dvector,  ::detray::djagged_vector,  ::detray::darray,  ::detray::dtuple,  ::traccc::internal_spacepoint< ::traccc::spacepoint> , (bool)0, (unsigned int)1u> ::bin") is not allowed
```

That one will need a bit of thinking about Beomki... :thinking: (I'll describe the issue in more detail in an issue in [detray](https://github.com/acts-project/detray) once I get to it.)

The fixes to get rid of the warnings are in general all fairly standard stuff. Getting rid of unused variables, unnecessary semicolons, and harmonising integer types between variables that are used together in some way.


examples/run/cpu/ccl_example.cpp
```diff
@@ -42,10 +42,10 @@ void print_statistics(const traccc::host_cell_container& data) {
     for (std::size_t i = 0; i < data.size(); ++i) {
         std::size_t count = data.at(i).items.size();
 
-        for (std::size_t i = 0; i < bins_edges.size(); ++i) {
-            if (count >= bins_edges[i] &&
-                (i + 1 >= bins_edges.size() || count < bins_edges[i + 1])) {
-                ++bins[i];
+        for (std::size_t j = 0; i < bins_edges.size(); ++j) {
```
stephenswat, line 45: This doesn't look right.

stephenswat (changes requested):
Ah, the build system grows ever more complicated. There's one part where I am not convinced a correct code transformation was made, other than that it looks fine.

beomki-yeo:
Other than @stephenswat 's comment, it looks good

examples/run/cpu/ccl_example.cpp
krasznaa: Good catch! Even though I looked at that part so many times...

krasznaa:
> For the warning you mentioned, I need to investigate in detray side. maybe there is a code that is using `std::vector` instead of `vecmem::vector`

Ha! ...but `vecmem::vector` **is** `std::vector`. The problem is that when we specialise `detray::grid2` with `vecmem::vector`, that function can not be marked with `__device__`. As `vecmem::vector` only works on the host... So we'll need some clever formalism that would select the `__host__` or `__host__ __device__` version of a function based on what vector type is being used with that type. :thinking: Not sure yet how to best tackle that, but clever templating should allow us to do it. I think...

krasznaa:
Ping! I believe I fixed the issue that Stephen found...

beomki-yeo (approved):

PR #133, opened 2022-02-11
Add a mapper for validation purpose

beomki-yeo:
This PR is for #130 which will need a map between `measurement` and `particle` to calculate the seeding performance.
Technically, `generate_measurement_particle_map` in `mapper.hpp` generates `std::map<measurement, std::vector<particle>>`.

~~The unit tests are also added for most of upstream functions (but not for `generate_measurement_particle_map` which accompanies clusterization algorithm)~~

The unit tests are added for mock data which I made up


examples/run/cpu/seq_example.cpp
```diff
@@ -7,6 +7,7 @@
 
 // io
 #include "traccc/io/csv.hpp"
+#include "traccc/io/mapper.hpp"
```
stephenswat, line 10: Is there a reason to include this here?

stephenswat (commented):
Overall it looks good, free of obvious flaws at least. I am, however, a little bit uncertain about the way we're adding algorithm execution to the IO part of the project. To me, these parts are strictly separate (the IO part _only_ does IO), and I feel uneasy combining them like this. I'll let the others weigh in to see what they think.

beomki-yeo:
> I feel uneasy combining them like this.

I agree. How about moving it to `core/utils` then?

beomki-yeo:
Ah `mapper` cannot go to `core` which is not linked to `traccc::io`. Hmm

examples/run/cpu/seq_example.cpp
beomki-yeo: I removed it

stephenswat (approved):
Let's get this in. :+1:

PR #136, opened 2022-02-15
Add boost program options

beomki-yeo:
This PR adds boost::program_options for example files
As #130 does, `TRACCC_BUILD_EXAMPLES` is added as a cmake build option which requires boost dependency


examples/run/cuda/seeding_example_cuda.cpp
```diff
@@ -221,21 +223,27 @@ int seq_run(const std::string& detector_file, const std::string& hits_dir,
...
+    po::variables_map vm;
+    po::store(po::parse_command_line(argc, argv, desc), vm);
+    po::notify(vm);
+
+    auto detector_file = vm["detector_file"].as<std::string>();
+    auto hit_directory = vm["hit_directory"].as<std::string>();
+    auto events = vm["events"].as<int>();
+    auto run_cpu = vm["run_cpu"].as<bool>();
+
+    std::cout << "Running ./traccc_seeding_example_cuda " << detector_file
```
krasznaa, line 245: As long as we're doing this update... Could this printout make use of `argv[0]` instead? For a long time now we've been having these applications print wrong names for themselves...

examples/run/cuda/seeding_example_cuda.cpp
```diff
@@ -25,16 +25,20 @@
 #include <iomanip>
 #include <iostream>
 
+// Boost
+#include <boost/program_options.hpp>
```
krasznaa, line 29: I would prefer to put the Boost includes before the standard library ones. As in general I want the includes to go "from specific to generic". As it is in the ATLAS coding standards as well...

tests/cpu/compare_with_acts_seeding.cpp
```diff
@@ -12,9 +12,9 @@
 #include "traccc/io/writer.hpp"
 
 // algorithms
-#include "traccc/clusterization/clusterization_algorithm.hpp"
+#include "traccc/seeding/seed_finding.hpp"
+#include "traccc/seeding/spacepoint_binning.hpp"
```
krasznaa, line 16: I was always a bit confused by `seeding_algorithm` to be honest. Are we moving away from it? :thinking:

CMakeLists.txt
```diff
@@ -159,10 +160,15 @@ endif()
 if( TRACCC_BUILD_SYCL )
    add_subdirectory( device/sycl )
 endif()
-add_subdirectory( examples )
 add_subdirectory( io )
 add_subdirectory( plugins )
 
+if ( TRACCC_BUILD_EXAMPLES )
+   # Set up Boost.
+   find_package( Boost 1.75.0 REQUIRED COMPONENTS program_options)
```
krasznaa, line 168: Why did you impose this 1.75 version minimum? I don't see anything in your code that would not have been part of this part of Boost ever since I've been using it.

CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ option( TRACCC_BUILD_CUDA "Build the CUDA sources included in traccc"
    ${TRACCC_BUILD_CUDA_DEFAULT} )
 option( TRACCC_BUILD_SYCL "Build the SYCL sources included in traccc" FALSE )
 option( TRACCC_BUILD_TESTING "Build the (unit) tests of traccc" TRUE )
+option( TRACCC_BUILD_EXAMPLES "Build the examples of traccc" FALSE )
```
krasznaa, line 45: I think the default should be `TRUE` for this option. Since this is our "top project" for the time being, I like to be able to build it with a minimal set of options specified on the command line.

krasznaa (changes requested):
I'm in favour of using `Boost::program_options` in the applications. But of course the build issues will need to be sorted out first. (On top of my other minor requests...)

tests/cpu/compare_with_acts_seeding.cpp
beomki-yeo: I removed `traccc::algorithm` dependency from `test` directory, so I could not use `seeding_algorithm` here anymoew

CMakeLists.txt
beomki-yeo: I dumbly put the version number that I was using. What is the best way figuring out this?

krasznaa: You don't need to specify anything. You can just write:

```cmake
find_package( Boost REQUIRED COMPONENTS program_options )
```

If you're really curious, you can check what version was actually found using the `${Boost_VERSION}` variable set up by this call. But I doubt that our code needs to know it.

The version number specified for `find_package(...)` is treated as a minimum version that the code needs. Or an exact one if you use the `EXACT` keyword, like:

```cmake
find_package( Boost 1.75.0 EXACT REQUIRED )
```

Long story short, just remove the version number. :wink:

examples/run/cuda/seeding_example_cuda.cpp
beomki-yeo: Right. I changed them

CMakeLists.txt
beomki-yeo: Ok it does work

examples/run/cpu/seq_example.cpp
```diff
@@ -96,27 +100,31 @@ int seq_run(const std::string& detector_file, const std::string& cells_dir,
...
+
+    po::variables_map vm;
+    po::store(po::parse_command_line(argc, argv, desc), vm);
+    po::notify(vm);
+
+    auto detector_file = vm["detector_file"].as<std::string>();
+    auto cell_directory = vm["cell_directory"].as<std::string>();
+    auto events = vm["events"].as<int>();
+
+    std::cout << "Running ./traccc_seq_example " << detector_file << " "
```
krasznaa, line 126: You missed this printout. :stuck_out_tongue: (I see that you did change every other one to use `argv[0]`.)

krasznaa (changes requested):
I still have one request...

examples/run/cpu/seq_example.cpp
beomki-yeo: I changed that definitely! I just didn't hit ctrl+s...

krasznaa (approved):
Let's go ahead with it finally...

krasznaa:
:frowning: I'm used to gitlab.cern.ch having issues, but github.com is having a bit too many issues lately for my liking...

@paulgessinger do you know why the "review-required" rule would not be fulfilled by now? :confused:

krasznaa:
:angry: Okay, so github is also having back-end issues then. I just needed to wait a few minutes...

paulgessinger:
image
 
@krasznaa looks greent to me?

PR #137, opened 2022-02-22
Sycl seeding intel fix 20220222

konradkusiak97:
With this fix in place, the code can be run on the Intel backend without changing the workGroup sizes to `8` in the `doublet_finding` and `triplet_finding` kernels. 

It seems like on the Intel backend, the group synchronization call: `item.barrier()` needs to be visible by all work items - without any exception. This implies that we need to be careful with returning from a kernel call if the current item index is not in the desireable range e.g. 
```cpp
if (idx >= totalRange) return;
```
If after such statement, the call to `item.barrier()` will be invoked, some of the previously returned work items won't see it and so the result will be that the rest of the work group items will wait for them to join the `barrier` call - forever...

pinging @krasznaa :wink:

### DIFF
--- a/device/sycl/src/seeding/doublet_finding.sycl
+++ b/device/sycl/src/seeding/doublet_finding.sycl
@@ -93,9 +93,9 @@ class DupletFind {
             auto& isp = internal_sp_per_bin[sp_idx];
 
             // find the reference (start) index of the doublet container item
-            // vector, where the doublets are recorded The start index is calculated
-            // by accumulating the number of doublets of all previous compatible
-            // middle spacepoints
+            // vector, where the doublets are recorded The start index is
+            // calculated by accumulating the number of doublets of all previous
+            // compatible middle spacepoints
             unsigned int mid_bot_start_idx = 0;
             unsigned int mid_top_start_idx = 0;
             for (unsigned int i = 0; i < item_idx; i++) {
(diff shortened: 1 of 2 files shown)


HadrienG2:
@konradkusiak97 This is also true on some NVidia GPUs, but not all (likely microarchitecture-dependent). In general, calling a collective group operation after early exiting from some threads is undefined behavior from both CUDA and SYCL.

krasznaa (approved):
I think this should be fine as-is. Though it definitely raises some uncomfortable questions...

If what @HadrienG2 says is true, then how can we ever confidently use synchronisation in CUDA code after the common

```c++
    const std::size_t i = blockIdx.x * blockDim.x + threadIdx.x;
    if (i >= input.size()) {
        return;
    }
```

style beginning to a kernel? :confused: Would a `__syncthreads()` call really halt such kernels on older GPUs? :thinking:

HadrienG2:
> I think this should be fine as-is. Though it definitely raises some uncomfortable questions...
> 
> If what @HadrienG2 says is true, then how can we ever confidently use synchronisation in CUDA code after the common
> 
> ```c++
>     const std::size_t i = blockIdx.x * blockDim.x + threadIdx.x;
>     if (i >= input.size()) {
>         return;
>     }
> ```
> 
> style beginning to a kernel? confused Would a `__syncthreads()` call really halt such kernels on older GPUs? thinking

I don't remember in which way it misbehaved, but I've definitely seen it misbehave in CUDA and OpenCL code on some hardware. This is what prompted me to read the docs and figure out that this intuitive pattern is unfortunately UB...

stephenswat:
Just for posterity, the PTX standard guarantees that `bar.sync`, which `__syncthreads()` _should_ compile to, only requires non-exited threads to pass it. So this is a peculiar situation where according to the CUDA standard this is unsafe to do, but according to the PTX standard this should not be a problem.

PR #138, opened 2022-02-22
Program Options Improvements, main branch (2022.02.22.)

krasznaa:
Added `--help` options to all of the (example) executables.

At the same time making sure that when an error is made on the command line, the printout would be a bit more helpful than it is now. Like:

```
[bash][atspot01]:build > ./bin/traccc_seq_example
Couldn't interpret command line options because of:

the option '--cell_directory' is required but missing

Allowed options:
  -h [ --help ]         Give some help with the program's options
  --detector_file arg   specify detector file
  --cell_directory arg  specify the directory of cell files
  --events arg          number of events

[bash][atspot01]:build >
```

Note that my overall views of clang-format have not improved too much in this exercise. :frowning: I was trying to format these calls like this:

https://gitlab.cern.ch/atlas/athena/-/blob/master/PhysicsAnalysis/D3PDMaker/D3PDMakerReader/src/apps/d3pdReaderMaker.cxx#L108-121

But clang-format made that really unreadable. :frowning:

Even weirder, note how in the exception handling `&` is put in a different place in the SYCL source files compared to the C\+\+ ones. :confused:

So yeah, not the biggest fan of clang-format still...

### DIFF
--- a/examples/run/cpu/seq_example.cpp
+++ b/examples/run/cpu/seq_example.cpp
@@ -20,6 +20,7 @@
 #include <boost/program_options.hpp>
 
 // System include(s).
+#include <exception>
 #include <iostream>
 
 namespace po = boost::program_options;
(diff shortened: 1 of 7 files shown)


krasznaa:
Wait, I misspoke... It's the OpenMP files in which `&` is placed differently. :confused:

We could also introduce some shared code between the executables to reduce the amount of boilerplate later on. But didn't want to invent anything elaborate for this PR.

beomki-yeo (approved):
Good!

krasznaa:
Could somebody re-bless this PR? Unfortunately GH does not let me merge it. :frowning:

stephenswat (approved):
I hereby bless this pull request!

PR #139, opened 2022-02-25
SYCL Code Cleanup, main branch (2022.02.25.)

krasznaa:
Both the description and the changelog are pretty long, sorry about that... :frowning:

I cleaned up the SYCL code of the project a little.

Made it possible to include every public header of the `traccc_sycl` library by a regular C\+\+ compiler, and generally cleaned up a few things about the implementation.

These are the commit messages recorded during the development process:
  - Starting with the cleanup.
  - Capturing the broken state of the code.
  - Saving yet another snapshot.
  - Finished updating the `traccc::sycl::spacepoint_binning` algorithm.
  - Moved the implementation of `traccc::sycl::seed_finding` into compiled code.
  - Updated the duplet finding part of the seed finding.
  - Modernised the triplet_counting and triplet_finding kernels.
  - Finished with the cleanup of the `traccc::sycl::seed_finding` algorithm.
  - Cleaned up the `traccc::sycl::track_params_estimation` algorithm.
  - Hid the implementation specific header files from the users of `traccc_sycl`.

At first I wanted to fix all const-correctness issues in this PR already, but as I had to find (already last week), that was a bit too ambitious. I still made it possible to use `detray::get_data(...)` on constant containers at least, but didn't end up exercising it too much in the code. That will come with another (set of) PR(s)...

The re-write also allowed me to turn a couple of `.sycl` files into `.cpp` ones, which should help code robustness a little in my mind.

Finally: This re-write did slow down the code a little. :frowning: That is because I removed the mutex from `traccc::sycl::seed_finding`. That class at the moment (in both the CUDA and SYCL implementations) has a persistent state, and can only be used by a single thread at a time. This way it sets up the memory that it needs **during initialisation**. Meaning that the time measurement does not see that!

In this implementation the data objects are created "on the stack" instead, so there's no need to lock the algorithm. Which will allow us to use it in parallel from multiple threads at one point.

This means that I go from:

```
[bash][Legolas]:build > SYCL_DEVICE_FILTER=cuda ./bin/traccc_seeding_example_sycl --detector_file=tml_detector/trackml-detector.csv --hit_directory=tml_hits/ --events=1 --run_cpu=true
Running ./bin/traccc_seeding_example_sycl tml_detector/trackml-detector.csv tml_hits/ 1
event 0
 seed matching rate: 0.993484
 track parameters matching rate: 0.997917
==> Statistics ... 
- read    48109 spacepoints from 0 modules
- created (cpu)  18722 seeds
- created (sycl) 18943 seeds
==> Elpased time ... 
wall time           4.3619    
hit reading (cpu)   0.568056  
seeding_time (cpu)  2.87935   
seeding_time (sycl) 0.124653  
tr_par_esti_time (cpu)    0.00639176
tr_par_esti_time (sycl)   0.00388667
[bash][Legolas]:build >
```

To:

```
[bash][Legolas]:build > SYCL_DEVICE_FILTER=cuda ./bin/traccc_seeding_example_sycl --detector_file=tml_detector/trackml-detector.csv --hit_directory=tml_hits/ --events=1 --run_cpu=true
Running ./bin/traccc_seeding_example_sycl tml_detector/trackml-detector.csv tml_hits/ 1
event 0
 seed matching rate: 0.992629
 track parameters matching rate: 0.997062
==> Statistics ... 
- read    48109 spacepoints from 0 modules
- created (cpu)  18722 seeds
- created (sycl) 18942 seeds
==> Elpased time ... 
wall time           4.37163   
hit reading (cpu)   0.567533  
seeding_time (cpu)  2.88422   
seeding_time (sycl) 0.144498  
tr_par_esti_time (cpu)    0.00633689
tr_par_esti_time (sycl)   0.00303242
[bash][Legolas]:build >
```

This is not great, but I believe that it more correctly represents what is going on. And in any case, the overhead from this initial memory allocation should go away once we process multiple events, and use some smart (thread safe) caching method for the device memory.

Pinging @konradkusiak97.


konradkusiak97:
This makes the code much cleaner indeed, I can only say thanks for this :smile:

device/sycl/src/seeding/doublet_counting.sycl
```diff
@@ -151,12 +161,13 @@ void doublet_counting(const seedfinder_config& config, sp_grid& internal_sp,
...
-         DupletCount kernel(config, internal_sp_view,
-                            doublet_counter_container_view);
-         h.parallel_for<class doublet_count_kernel>(doubletCountNdRange,
-                                                    kernel);
-     }).wait_and_throw();
+    details::get_queue(queue)
+        .submit([&](::sycl::handler& h) {
+            DupletCount kernel(config, internal_sp_view,
+                               doublet_counter_container_view);
+            h.parallel_for<DupletCount>(doubletCountNdRange, kernel);
```
konradkusiak97, line 168: Just, since it's a clean up, I realized later on that `parallel_for` doesn't need to be templated with the name of the kernel if they are created as functors...but I can remove that in some other PR.

krasznaa: It's a oneAPI extension that you can forego specifying a template parameter to `parallel_for`. I think we should leave these be.

The change I made here is just to make use of the functor types that we have there already, instead of declaring types just inside of the function call itself. Which turns into a warning with [ComputeCpp](https://developer.codeplay.com/products/computecpp/ce/home/). (We discussed this either on Mattermost or on a ticket with @HadrienG2 a while ago.)

All in all, I think this part of the code should just stay like this. I'll have some other ideas for you to clean up, don't worry. :wink:

konradkusiak97 (approved):
Let's speed this along. :smile:

krasznaa:
@paulgessinger could you have a look? I was trying to take a shortcut with Konrad approving this PR, so his development could go along. But the CI does not seem to accept his approval. :frowning:

paulgessinger:
@krasznaa yeah so far he wasn't on the "Reviewer" list. I invited him now, but I'm not sure the review bot will get triggered again. You can either push another commit, or I can override the check and merge manually.

paulgessinger:
It did get updated, so this should be good to go!

PR #143, opened 2022-03-11
update traccc data

beomki-yeo:
This traccc-data update adds ttbar samples from 10 to 300 pileups in `tml_full` directory.

After changing files in `compare_with_acts_seeding`, few track parameters could not be reconstructed in Acts seeding due to the error in global -> local transformation:

```c++
Acts::Result<Acts::Vector2> Acts::PlaneSurface::globalToLocal(
    const GeometryContext& gctx, const Vector3& position,
    const Vector3& /*unused*/, double tolerance) const {
  Vector3 loc3Dframe = transform(gctx).inverse() * position;
  if (loc3Dframe.z() * loc3Dframe.z() > tolerance * tolerance) {
    return Result<Vector2>::failure(SurfaceError::GlobalPositionNotOnSurface);
  }
  return Result<Vector2>::success({loc3Dframe.x(), loc3Dframe.y()});
}
```

Technically, `loc3Dframe.z()` was evaluated slightly higher than the tolerance which is given as `constexpr` value

Thus, I lower the huddle a little. :p

### DIFF
--- a/tests/cpu/compare_with_acts_seeding.cpp
+++ b/tests/cpu/compare_with_acts_seeding.cpp
@@ -297,7 +297,7 @@ TEST_P(CompareWithActsSeedingTests, Run) {
 
     // Ensure that ACTS and traccc give the same result
     EXPECT_EQ(seeds.size(), seedVector.size());
-    EXPECT_FLOAT_EQ(seed_match_ratio, 1);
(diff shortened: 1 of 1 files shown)


stephenswat (approved):
Looks great! :smile:

PR #145, opened 2022-03-14
Fix bugs in options

beomki-yeo:
This is a hotfix of stupid bugs in option... I will merge it once the CI is done

PR #146, opened 2022-03-15
Organize options

beomki-yeo:
This PR add `seeding_input_options` and `seq_input_options` to clean up messy codes in `main()` body.

This PR also fixes a bug raised in #145 where `check_performance` always is `true`.

`skip` option is also added to the options, which indicates how many first events will be skipped during the executables. This is going to be useful when we do multi-processing


examples/options/include/traccc/throw_exception.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace traccc {
+
+enum exception : int {
+    print_help = 0,
+    input_error = 1,
+    no_exception = 2,
+};
+
+int throw_exception(const po::options_description& desc,
```
paulgessinger, line 21: I'm not sure this function (and file) have a good name. Maybe `handle_argument_errors`?

paulgessinger (commented):
This is kind of what we were doing in core as well. At some point, we then start factorizing the common options into the helper functions (in this case `events` and `skip` for examples). 

I strongly think we should keep the number of options as low as feasible.

This here seems fine to me though.

examples/options/include/traccc/throw_exception.hpp
beomki-yeo: Makes sense

krasznaa (changes requested):
I would do it differently... :thinking:

While we shouldn't over-engineer things, I think a better idea would be to make this a bit more object oriented. I would've rather introduced `traccc::full_tracking_input_config` like:

```c++
struct full_tracking_input_config {
    std::string detector_file;
    std::string cell_directory;
    std::string hit_directory;
    std::string particle_directory;
    bool check_seeding_performance;
    unsigned int events;
    int skip;

    full_tracking_input_config(po::options_description& desc);
    void read(const po::variables_map& vm);
};
```

Similarly for `seeding_input_options` of course. Though, and this is where this should be made more "object oriented", the options that they share, should be extracted into another struct. :thinking: (And they should call each others' functions, like types in any proper inheritance structure do. :stuck_out_tongue:)

For handling `--help` I'm a bit less sure. :frowning: I don't like this `traccc::handle_argument_errors` function too much, but I'm not sure how to make it much better. Maybe the function should just use [std::exit](https://en.cppreference.com/w/cpp/utility/program/exit) itself, and not leave it up to its caller to do the right thing. :thinking: That could simplify the function a bit, as it wouldn't need to return anything.

But the thing that really made me write a review... We should stop just adding header-only stuff all the time. There is no reason why this code should all live in headers. Please introduce the new library as a shared library, and move the implementation of the functions into `.cpp` files.

paulgessinger:
@krasznaa this is then precisely what we ended up doing in ACTS core, which I referred to as "factorizing" in my previous comment. This is cleaner, but also comes with increased complexity. 

Doesn't mean it should be done per se, but it's interesting that we're retracing also these steps in this project here.

stephenswat:
Saving this conversation into my "evidence that everyone implicitly knows that composition is the best way to control complexity in software engineering but they don't want to admit it" folder.

beomki-yeo:
PR updated with the given comments. I think the `handle_argument_errors` should be polished more - but let's leave it as it is for now...

beomki-yeo:
Can this get an approval unless there is follow-up comments?

stephenswat (approved):
I'll sign off on this. :smile:

PR #147, opened 2022-03-17
changed Cluster EDM

konradkusiak97:
This change of cluster EDM aims at making the clusterization algorithm more parallelizable. 

The idea for those changes was triggered by my attempt to port the `measurement_creation` to SYCL. This is still a draft though and definitely some feedback would be very much appreciated :smiley:

pinging @krasznaa :wink:

### DIFF
--- a/examples/run/cpu/ccl_example.cpp
+++ b/examples/run/cpu/ccl_example.cpp
@@ -80,7 +80,7 @@ void print_statistics(const traccc::host_cell_container& data) {
 void run_on_event(traccc::component_connection& cc,
                   traccc::host_cell_container& data) {
     for (std::size_t i = 0; i < data.size(); ++i) {
-        traccc::cluster_collection clusters_per_module =
+        traccc::host_cluster_container clusters_per_module =
             cc(data.at(i).items, data.at(i).header);
     }
 }
--- a/examples/run/cpu/ccl_example.cpp
+++ b/examples/run/cpu/ccl_example.cpp
@@ -80,7 +80,7 @@ void print_statistics(const traccc::host_cell_container& data) {
 void run_on_event(traccc::component_connection& cc,
                   traccc::host_cell_container& data) {
     for (std::size_t i = 0; i < data.size(); ++i) {
-        traccc::host_cluster_container clusters_per_module =
+        traccc::host_cluster_container clusters =
             cc(data.at(i).items, data.at(i).header);
     }
 }
--- a/examples/run/openmp/io_dec_par_example.cpp
+++ b/examples/run/openmp/io_dec_par_example.cpp
@@ -65,7 +65,7 @@ traccc::demonstrator_result run(traccc::demonstrator_input input_data,
             traccc::host_cluster_container clusters_per_module =
                 cc(cells_per_event.get_items()[i],
                    cells_per_event.get_headers()[i]);
-            for (auto& cl_id : clusters_per_module.get_headers())
+            for (auto &cl_id : clusters_per_module.get_headers())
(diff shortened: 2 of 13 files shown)


examples/algorithms/cpu/include/traccc/clusterization/clusterization_algorithm.hpp
```diff
@@ -66,9 +66,10 @@ class clusterization_algorithm
             auto module = cells_per_event.at(i).header;
 
             // The algorithmic code part: start
-            traccc::cluster_collection clusters_per_module = cc->operator()(
+            traccc::host_cluster_container clusters_per_module = cc->operator()(
```
krasznaa, line 69: You should rename the variable as well.

In the "new EDM" we don't store the clusters separately for the modules. So the variable name should not try to confuse people about it. :wink:

examples/run/cpu/ccl_example.cpp
```diff
@@ -80,7 +80,7 @@ void print_statistics(const traccc::host_cell_container& data) {
 void run_on_event(traccc::component_connection& cc,
                   traccc::host_cell_container& data) {
     for (std::size_t i = 0; i < data.size(); ++i) {
-        traccc::cluster_collection clusters_per_module =
+        traccc::host_cluster_container clusters_per_module =
```
krasznaa, line 83: The same here. Let's just go with a simple `clusters` variable name.

examples/run/openmp/io_dec_par_example.cpp
```diff
@@ -62,10 +62,11 @@ traccc::demonstrator_result run(traccc::demonstrator_input input_data,
                 traccc::pixel_segmentation{-8.425, -36.025, 0.05, 0.05};
 
             // The algorithmic code part: start
-            traccc::cluster_collection clusters_per_module =
+            traccc::host_cluster_container clusters_per_module =
```
krasznaa, line 65: And again. :stuck_out_tongue:

examples/run/openmp/par_example.cpp
```diff
@@ -84,9 +84,10 @@ int par_run(const std::string &detector_file, const std::string &cells_dir,
                 traccc::pixel_segmentation{-8.425, -36.025, 0.05, 0.05};
 
             // The algorithmic code part: start
-            traccc::cluster_collection clusters_per_module =
+            traccc::host_cluster_container clusters_per_module =
```
krasznaa, line 87: And once more.

io/include/traccc/io/mapper.hpp
```diff
@@ -185,18 +185,19 @@ measurement_cell_map generate_measurement_cell_map(
         auto module = cells_per_event.at(i).header;
 
         // The algorithmic code part: start
-        cluster_collection clusters_per_module =
+        host_cluster_container clusters_per_module =
```
krasznaa, line 188: You know the drill by now...

krasznaa (changes requested):
Apart from the superficial variable name changes, I'm completely happy with this myself.

Of course we'll need to discuss about it in today's meeting, and you'll also need to resolve the conflict(s) that it collected since yesterday. But other than that, I think this is the way to go...

examples/algorithms/cpu/include/traccc/clusterization/clusterization_algorithm.hpp
konradkusiak97: alright, done.

krasznaa:
Ahh... turns out I was able to do that myself... :smile:

PR #148, opened 2022-03-24
Remove dependency on ACTS unit test headers

stephenswat:
Right now, our tests are relying on some ACTS-internal headers which are defined for the ACTS unit tests. This does not work, because those headers are not internal, and they are not exported or installed, leading to compiler errors when building traccc. This commit is a stop-gap solution that copies over the necessary files into the domain of traccc such that the code can compile.


beomki-yeo (approved):
Should be good to go

PR #149, opened 2022-03-24
Fix range-loop-construct warning in module map

stephenswat:
Recent versions of gcc introduce a new `range-loop-construct` warning which detects non-explicit copies or references in for loops. This started throwing warnings for a particular loop in our module map, which then got escalated to an error. This commit fixes the problem and allows the code to compile with gcc 11.2.0.

### DIFF
--- a/core/include/traccc/geometry/module_map.hpp
+++ b/core/include/traccc/geometry/module_map.hpp
@@ -265,7 +265,7 @@ class module_map {
          */
         std::vector<K> keys;
         keys.reserve(input.size());
-        for (const std::pair<K, V>& i : input) {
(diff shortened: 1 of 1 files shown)


HadrienG2 (approved):

stephenswat:
Hmm, why is @HadrienG2's review not counting towards the `review-required` bot?

beomki-yeo (approved):

beomki-yeo:
I think I saw that once.
My approval somehow resolved that

stephenswat:
Thanks :smile:

beomki-yeo (approved):

PR #150, opened 2022-03-25
Run tests only for CPU-based builds

stephenswat:
As we start to add CUDA and SYCL tests, we are running into issues where the CI nodes are trying to run code for devices they don't have. This is a bad idea, because it causes those tests to fail. This commit modifies the CI setup such that the tests are only run for the CPU-based builds.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -49,6 +49,7 @@ jobs:
           source ${GITHUB_WORKSPACE}/.github/ci_setup.sh ${{ matrix.platform.name }}
           cmake --build build
       - name: Test
+        if: "matrix.platform.name != 'CUDA'"
         run: |
           cd build
           source ${GITHUB_WORKSPACE}/.github/ci_setup.sh ${{ matrix.platform.name }}


.github/workflows/builds.yml
```diff
@@ -49,6 +49,7 @@ jobs:
           source ${GITHUB_WORKSPACE}/.github/ci_setup.sh ${{ matrix.platform.name }}
           cmake --build build
       - name: Test
+        if: "matrix.platform.name == 'CPU'"
```
krasznaa, line 52: :thinking: Not quite... Most of the SYCL tests should function on the CI nodes as well with `SYCL_DEVICE_FILTER=host`. So (as long as we don't add a HIP flavour for now) a

```
"matrix.platform.name != 'CUDA'"
```

would seem more appropriate.

krasznaa (changes requested):
Let's just exclude CUDA for now. Running both native/CPU and SYCL tests in the CI would seem better than not running any device code at all.

.github/workflows/builds.yml
stephenswat: Fixed. :smile:

PR #153, opened 2022-03-28
Add infrastructure for CUDA testing

stephenswat:
This commit adds all the necessary infrastructure for testing CUDA code. In particular, this means adding the necessary build system features, adding a CUDA-appropriate Google Test main function, and adding some basic tests to make sure the CUDA device and the CUDA runtime are functional.

### DIFF
--- a/tests/cuda/CMakeLists.txt
+++ b/tests/cuda/CMakeLists.txt
@@ -1,0 +1,35 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2021-2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+include(traccc-compiler-options-cuda)
+
+enable_language(CUDA)
+
+find_package(CUDAToolkit REQUIRED)
+
+add_library(
+    traccc_tests_cuda_main
+    STATIC
+    cuda_main.cpp
+)
+
+target_link_libraries(
+    traccc_tests_cuda_main
+    PRIVATE
+    CUDA::cudart
+    GTest::gtest
+)
+
+traccc_add_test(
+    cuda
+
+    # Define the sources for the test.
+    test_basic.cu
+
+    LINK_LIBRARIES
+    GTest::gtest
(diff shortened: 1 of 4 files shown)


tests/cuda/test_basic.cu
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include <gtest/gtest.h>
+
+TEST(CUDABasic, DeviceCount) {
+    int nDev;
```
krasznaa, line 12: Please set default values to every variable. Having just been through 2 weeks of programming tutorials, we should not ignore one of the rules that we teach to all newcomers...

tests/cuda/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+)
+
+traccc_add_test(
+    cuda
+
+    # Define the sources for the test.
+    test_basic.cu
+
+    LINK_LIBRARIES
+    traccc::cuda
```
krasznaa, line 31: Please remove this dependency. It's not needed.

tests/cuda/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2021-2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+include(traccc-compiler-options-cuda)
+
```
krasznaa, line 8: Please add

```cmake
find_package( CUDAToolkit )
```

into this file. It gets called upstream, making `CUDA::cudart` visible at this point. But we should not rely on that. (I don't even fully understand right now why that imported target is actually visible here... :confused:)

krasznaa (changes requested):
Just some technical requests...

krasznaa (approved):
As I wrote in #155, as long as the CMake 3.18 requirement is made correctly there, this PR should be okay as-is. :smile:

PR #154, opened 2022-03-28
Implement and test a CUDA version of the module map

stephenswat:
The module map was originally designed to be GPU friendly, and we now reap the fruits of that decision. This implementation of a CUDA version of that code is quite short and simple to understand. This pull request adds an owning CUDA version of the module map, as well as a light-weight non-owning view object which is designed to be passed to kernels. It also tests the class.

This pull request depends on #153.

### DIFF
--- a/core/include/traccc/geometry/module_map.hpp
+++ b/core/include/traccc/geometry/module_map.hpp
@@ -15,6 +15,11 @@
 #include <iostream>
 
 namespace traccc {
+namespace cuda {
+template <typename, typename>
+class module_map;
+}
+
 /**
  * @brief A fast and GPU-friendly map for consecutive module IDs to values.
  *
@@ -397,5 +402,7 @@ class module_map {
      * keep indices in this array instead of pointers.
      */
     std::vector<V> m_values;
+
+    friend class cuda::module_map<K, V>;
 };
 }  // namespace traccc
--- a/device/cuda/CMakeLists.txt
+++ b/device/cuda/CMakeLists.txt
@@ -11,6 +11,8 @@ enable_language( CUDA )
 include( traccc-compiler-options-cpp )
 include( traccc-compiler-options-cuda )
 
+find_package(CUDAToolkit)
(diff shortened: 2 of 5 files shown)


stephenswat:
This PR also depends on https://github.com/acts-project/vecmem/pull/172.

stephenswat:
This is now ready to be merged.

core/include/traccc/geometry/module_map.hpp
```diff
@@ -15,6 +15,11 @@
 #include <iostream>
```
beomki-yeo, line 15: I think you can remove it as well

beomki-yeo, line 15: I think you can remove it as well

beomki-yeo (commented):
Just very trivial comments...

stephenswat:
Thanks! You have a keen eye for this! :smile:

stephenswat:
Closing this as there seems to be little interest in it.

beomki-yeo:
Actually I need it for GPU. 
I was not event aware of that this PR has existed so I made a (very simple) module map for CKF study recently.
So I guess it is worth keeping this alive

stephenswat:
Mind you this is a map from module identifiers to transformation matrices, although we can easily repurpose this for spacepoint identifier to module identifier mappings, which I guess is what you need for the CKF.

beomki-yeo:
> although we can easily repurpose this for spacepoint identifier to module identifier mappings,

Exactly. I hope we can generalize it for multiple purposes.

krasznaa:
Right now you could still argue that we don't have the correct thing in place yet. But as I wrote in https://github.com/acts-project/traccc/pull/592#discussion_r1618846697, I'll have some EDM updates in the foreseeable future, which will absolutely replace this PR.

So yeah, I agree with closing it. Even if I was only going to suggest it a little later, once I pushed my EDM changes through. :wink:

PR #156, opened 2022-03-29
Remove variance from spacepoint EDM

stephenswat:
As it stands, the spacepoint EDM contains a measure of variance which is underdefined and unused. I've discussed with Andi, and we've concluded that this serves no purpose at all but to slow down our development, and that we'd be better served by trimming the fat out of our EDM. This commit does just that; it removes the variance from the spacepoint, and amends any code that uses it (which no code did meaningfully).

### DIFF
--- a/core/include/traccc/edm/internal_spacepoint.hpp
+++ b/core/include/traccc/edm/internal_spacepoint.hpp
@@ -40,8 +40,6 @@ struct internal_spacepoint {
     scalar m_y;
     scalar m_z;
     scalar m_r;
-    scalar m_varianceR;
-    scalar m_varianceZ;
 
     internal_spacepoint() = default;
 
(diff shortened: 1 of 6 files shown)


core/include/traccc/edm/internal_spacepoint.hpp
```diff
@@ -56,9 +56,8 @@ struct internal_spacepoint {
         m_z = sp.global[2];
         m_r = std::sqrt(m_x * m_x + m_y * m_y);
 
-        // Need to fix this part
-        m_varianceR = sp.variance[0];
-        m_varianceZ = sp.variance[1];
+        m_varianceR = 0.;
+        m_varianceZ = 0.;
```
krasznaa, line 60: Do we have a use for these 2 variables? Shouldn't we remove these as well?

stephenswat:
Depends on #154 for now.

core/include/traccc/edm/internal_spacepoint.hpp
beomki-yeo: I think it's OK to remove them for now. We can revive them anytime if it turns out to be useful later

stephenswat:
This is now ready to go in!

beomki-yeo (approved):

stephenswat:
Thanks! :smile:

PR #157, opened 2022-03-29
Return internal spacepoint members by value

stephenswat:
Currently, we are returning a lot of the members of the internal spacepoint class by constant reference, which we really should not be doing. This commit changes all of these getter methods to use return-by-value semantics instead, which will be more maintainable and more performant.

### DIFF
--- a/core/include/traccc/edm/internal_spacepoint.hpp
+++ b/core/include/traccc/edm/internal_spacepoint.hpp
@@ -112,22 +112,22 @@ struct internal_spacepoint {
     TRACCC_HOST_DEVICE const scalar& x() const { return m_x; }
 
     TRACCC_HOST_DEVICE
-    const scalar& y() const { return m_y; }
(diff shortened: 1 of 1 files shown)


paulgessinger:
This is because for the scalars the copies are cheaper than the pointer copies (64bit?)?

stephenswat:
That's exactly right: returning the value directly not only gets rid of the overhead of returning 32 extra bits (although, in practice, register allocation will make this a _very_ muddy topic.

But the bigger issue is that I discussed with Andi that we will want to get rid of the variance in the spacepoint objects, which means that the variance methods in the internal spacepoints will compute the variance on demand. That means we need to return an rvalue, which you can't do through const references; that only works through values.

krasznaa (approved):
I thought it would've gone in already. :stuck_out_tongue: Of course we should make this change! :smile:

PR #158, opened 2022-03-29
Implement CUDA spacepoint formation with flat EDM

stephenswat:
This commit contains an implementation of spacepoint formation on the CUDA platform, employing a flat style of EDM. This is one of two algorithms I would like to implement for spacepoint formation.

This depends on #154.


stephenswat:
This depends on #154.

stephenswat:
As far as I understand this has been sorted.

PR #159, opened 2022-03-31
Update SYCL Seedfinding Status, main branch (2022.03.31.)

krasznaa:
Marked the SYCL seed-finding as "existing". That code works reasonably reliably by now, so no need to say that work has merely started on it.

Pinging @konradkusiak97.


krasznaa:
I believe you should be able to write to my branch. You should try. If it doesn't work, I'll apply that patch myself.

stephenswat:
Okay, I'll try. :smile:

stephenswat (approved):
Applied, now let's get this in.

stephenswat (approved):

PR #160, opened 2022-04-05
Use buffer instead of container in CUDA seeding

beomki-yeo:
The motivation is that cuda & sycl seeding speedups become only 2-3x after the seeding parameter optimization in #130 , which was reported by @konradkusiak97 

This PR recovers cuda speedup upto x8 but what's fixed here is totally independent of #130 

One of the major bottlenecks in seeding was from bunch of resize() function call in `seed_finding.hpp` - for example:
https://github.com/acts-project/traccc/blob/90e34f17e72b51e5d7ff943d20cfcf7a2e5a2c1c/device/cuda/include/traccc/cuda/seeding/seed_finding.hpp#L71-L74

This is quite inefficient becuase the cudaMalloc will be called as many as the size of jagged vector.

Therefore, I changed all internal containers to buffer type and allocated the memory in one shot.

You can check how the kernel functions get cluttered with the changes

Main:
image
seeding_time (cuda) 0.576947  for 10 events (ttbar<200>)

This PR:
image
seeding_time (cuda) 0.198063 for 10 events (ttbar<200>)

### DIFF
--- a/device/cuda/include/traccc/cuda/seeding/seed_finding.hpp
+++ b/device/cuda/include/traccc/cuda/seeding/seed_finding.hpp
@@ -48,25 +48,26 @@ struct seed_finding : public algorithm<host_seed_collection(
         vecmem::cuda::copy copy;
         unsigned int nbins = g2_buffer._buffer.m_size;
 
-        // resize the doublet counter container with the number of middle
-        // spacepoint
+        // Fill the size vector for double counter container
         std::vector<size_t> n_spm_per_bin;
         for (unsigned int i = 0; i < nbins; ++i) {
             n_spm_per_bin.push_back(g2_buffer._buffer.m_ptr[i].size());
         }
 
+        // Create doublet counter container buffer
         doublet_counter_container_buffer dcc_buffer{
             {nbins, m_mr.get()}, {n_spm_per_bin, m_mr.get()}};
         copy.setup(dcc_buffer.headers);
 
-        // doublet counting
+        // Run doublet counting
         traccc::cuda::doublet_counting(m_seedfinder_config, g2_buffer,
                                        dcc_buffer, m_mr.get());
 
-        // resize the doublet container with the number of doublets
+        // Take header of the doublet counter container into host
         vecmem::vector<doublet_counter_per_bin> dcc_headers(&m_mr.get());
         copy(dcc_buffer.headers, dcc_headers);
 
+        // Fill the size vectors for doublet containers
         std::vector<size_t> n_mid_bot_per_bin;
         std::vector<size_t> n_mid_top_per_bin;
         for (const auto& h : dcc_headers) {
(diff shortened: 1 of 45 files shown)


beomki-yeo:
Ah btw this is only for the CUDA...

device/cuda/include/traccc/cuda/seeding/seed_finding.hpp
```diff
@@ -24,112 +24,129 @@
...
-#include <iostream>
-#include <mutex>
 
 namespace traccc {
 namespace cuda {
 
 /// Seed finding for cuda
 struct seed_finding : public algorithm<host_seed_collection(
-                          host_spacepoint_container&&, sp_grid&&)> {
+                          host_spacepoint_container&&, sp_grid_buffer&&)> {
```
krasznaa, line 33: This algorithm should take a view object (`sp_grid_view`) like the other functions do. Functions/classes should basically never receive data or buffer objects. They should always receive view objects. The caller can then decide whether they want to give the function/class a view on top of a host object, a view on top of a buffer, or whatever else...

device/cuda/include/traccc/cuda/seeding/seed_finding.hpp
```diff
@@ -24,112 +24,129 @@
...
-                 vecmem::memory_resource& mr)
-        : m_seedfinder_config(config),
-          m_mr(mr),
-          doublet_counter_container(nbins, &m_mr.get()),
-          mid_bot_container(nbins, &m_mr.get()),
-          mid_top_container(nbins, &m_mr.get()),
-          triplet_counter_container(nbins, &m_mr.get()),
-          triplet_container(nbins, &m_mr.get()),
-          seed_collection(&m_mr.get()) {}
+    seed_finding(seedfinder_config& config, vecmem::memory_resource& mr)
```
krasznaa, line 40: It should really receive a const reference to the configuration object. Requiring a non-const reference is just asking for trouble.

device/cuda/src/seeding/doublet_counting.cu
```diff
@@ -12,6 +12,12 @@
 namespace traccc {
 namespace cuda {
 
+__global__ void set_zero_kernel(doublet_counter_container_view dcc_view) {
+    device_doublet_counter_container dcc_device(
+        {dcc_view.headers, dcc_view.items});
+    dcc_device.get_headers().at(threadIdx.x).zeros();
```
krasznaa, line 18: You should add a

```c++
if (threadIdx.x >= dcc_device.get_headers().size()) {
  return;
}
```

check. Or better, the full

```c++
const std::size_t i = blockIdx.x * blockSize.x + threadIdx.x;
if (i >= >= dcc_device.get_headers().size()) {
  return;
}
```

, as you'll see in my next comment...

device/cuda/src/seeding/doublet_counting.cu
```diff
@@ -24,13 +30,18 @@ __global__ void doublet_counting_kernel(
...
+                      doublet_counter_container_view dcc_view,
                       vecmem::memory_resource& resource) {
 
-    auto doublet_counter_container_view =
-        get_data(doublet_counter_container, &resource);
-    auto internal_sp_view = get_data(internal_sp, resource);
+    unsigned int nbins = internal_sp_view._data_view.m_size;
+
+    // zero initialization
+    set_zero_kernel<<<1, nbins>>>(dcc_view);
```
krasznaa, line 41: I don't like it that we're maximising `nbins` at 1024. It would be worth to generalise this kernel launch to any bin count.

device/cuda/src/seeding/doublet_finding.cu
```diff
@@ -12,6 +12,15 @@
 namespace traccc {
 namespace cuda {
 
+__global__ void set_zero_kernel(doublet_container_view mbc_view,
+                                doublet_container_view mtc_view) {
+    device_doublet_container mbc_device({mbc_view.headers, mbc_view.items});
```
krasznaa, line 17: Same comment as for the doublet counting.

device/cuda/src/seeding/doublet_finding.cu
```diff
@@ -28,16 +37,21 @@ __global__ void doublet_finding_kernel(
...
                      vecmem::memory_resource& resource) {
 
-    auto doublet_counter_view = get_data(doublet_counter_container, &resource);
-    auto mid_bot_doublet_view = get_data(mid_bot_doublet_container, &resource);
-    auto mid_top_doublet_view = get_data(mid_top_doublet_container, &resource);
-    auto internal_sp_view = get_data(internal_sp, resource);
+    unsigned int nbins = internal_sp_view._data_view.m_size;
+
+    // zero initialization
+    set_zero_kernel<<<1, nbins>>>(mbc_view, mtc_view);
```
krasznaa, line 51: Same comment as for the doublet counting.

krasznaa (changes requested):
I fear that all of the additional copy operations in `traccc::cuda::seed_finding` are still giving us more overhead than we should have.

But generally I'm happy with this update. We should try out different ways of making the code faster. I just hope that we'll even be able to go further than this. :wink:

For now I "just" have some technical comments.

device/cuda/src/seeding/doublet_counting.cu
beomki-yeo: Resolved

device/cuda/src/seeding/doublet_counting.cu
beomki-yeo: Indeed. Resolved.

device/cuda/include/traccc/cuda/seeding/seed_finding.hpp
beomki-yeo: Right

device/cuda/include/traccc/cuda/seeding/seed_finding.hpp
beomki-yeo: Resolved

krasznaa (approved):
I was just about to open an issue about the CUDA and SYCL executables not running the CPU seed finding... We really need to improve the output of those executables. When the CPU code is not enabled, we should not print that 0 seeds were reconstructed...

Other than that, let's merge this in now...

PR #164, opened 2022-04-06
Changed seed matching comparison

konradkusiak97:
It turns out that so far we've been comparing found seeds from CPU vs GPU by checking if the indices of bottom/top/middle spacepoints are the same. So for instance:
```cpp
...
if (lhs.spB_link == rhs.spB_link && 
    lhs.spM_link == rhs.spM_link &&
    lhs.spT_link == rhs.spT_link) return true;
...
```
This of course doesn't work if both seed collection from CPU and seed collection from GPU refer with their indices to two different spacepoint containers (which could potentially have the spacepoints in different order). 

To make this work, I changed it so we compare if the actual spacepoints are the same.
pinging @krasznaa @beomki-yeo

### DIFF
--- a/core/include/traccc/edm/seed.hpp
+++ b/core/include/traccc/edm/seed.hpp
@@ -62,6 +62,41 @@ inline bool operator==(const seed& lhs, const seed& rhs) {
             lhs.spT_link == rhs.spT_link);
 }
 
+template <typename spacepoint_container_t>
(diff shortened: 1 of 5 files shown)


konradkusiak97:
I made the changes, is this what you had in mind or should I change something else? :stuck_out_tongue:

krasznaa (approved):
Yes, this is what I had in mind. :smile:

PR #167, opened 2022-04-07
Reorganize seeding structure

beomki-yeo:
This PR reorganize some files for seeding in core directory and remove some annoying struct in helper files

Detailed changes:
1.  Move files commonly used by HOST, CUDA and SYCL into `seeding/common` directory
2.  Remove helper struct (e.g. `doublet_finding_helper`, `cuda_helper` and `sycl_helper`) and define functions only

It could be just my imagination but CUDA seeding seems get faster with ~5% after removing helper struct


beomki-yeo:
Actually not sure if the change is really better than the original. Comments will be appreciated..

krasznaa (commented):
I don't know...

I was myself planning on creating a proposal today/tomorrow for how I want to share a **lot** of code between the CUDA and SYCL implementations. A lot more than what this PR proposes.

Moving some of the functions into simple stand-alone functions may very well have a good effect on the generated CUDA code. I could well imagine that changes like this could indeed help.

But let me leave this one on hold for a bit. I'll see today if I could propose something a bit more drastic, and then in the light of that come back to this PR...

beomki-yeo:
Um No problem

PR #168, opened 2022-04-08
Container Improvements, main branch (2022.04.08.)

krasznaa:
Following up from #165, I ended up working with the container code a bit more before getting to the algorithmic code tweaks for #167...

After @stephenswat's nagging (re-)Introduced some move constructors and assignment operators, in case we end up making use of them.

Introduced a specific constructor for `traccc::device_container` that would allow constructing such objects directly from "view objects". Then updated all CUDA and SYCL code to make use of this more convenient way of creating the device containers.

Finally updated `traccc::container_element` to make it usable in device code. Since as @konradkusiak97 found, such things couldn't actually be constructed in device code so far. This is because [vecmem::jagged_device_vector](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/jagged_device_vector.hpp) doesn't actually return references to [vecmem::device_vector](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/device_vector.hpp) for technical reasons...

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/jagged_device_vector.hpp#L60-L63

### DIFF
--- a/core/include/traccc/edm/details/container_element.hpp
+++ b/core/include/traccc/edm/details/container_element.hpp
@@ -21,11 +21,17 @@ namespace traccc {
  * low-overhead access to the struct-of-arrays container class to emulate an
  * array-of-structs architecture.
  *
- * @tparam header_t The type of the header object.
- * @tparam vector_t The fully qualified vector type.
+ * @tparam header_reference_t The type of reference of the header object.
+ * @tparam vector_reference_t The fully qualified vector reference type.
  */
-template <typename header_t, typename vector_t>
+template <typename header_reference_t, typename vector_reference_t>
 struct container_element {
+
+    /// Header reference type
+    using header_reference = header_reference_t;
+    /// Vector reference type
+    using vector_reference = vector_reference_t;
+
     /**
      * @brief Construct a new container element view.
      *
@@ -37,10 +43,11 @@ struct container_element {
      * @param[in] v The vector object reference.
      */
     TRACCC_HOST_DEVICE
-    container_element(header_t& h, vector_t& v) : header(h), items(v) {}
+    container_element(header_reference h, vector_reference v)
+        : header(h), items(v) {}
 
-    header_t& header;
-    vector_t& items;
+    header_reference header;
+    vector_reference items;
 };  // struct container_element
 
 }  // namespace traccc
--- a/core/include/traccc/edm/details/device_container.hpp
+++ b/core/include/traccc/edm/details/device_container.hpp
@@ -34,6 +34,15 @@ class device_container
     /// Inherit all of the base class's constructors
     using base_type::base_type;
 
+    /// Constructor from a "view object"
+    ///
+    /// Note that it may also be a "data" or "buffer" object. It just needs to
+    /// look like a "view object".
+    ///
+    template <template <typename, typename> class view_t>
+    TRACCC_HOST_DEVICE device_container(const view_t<header_t, item_t>& view)
+        : base_type(view.headers, view.items) {}
+
     /**
      * @brief Bounds-checking mutable element accessor.
      */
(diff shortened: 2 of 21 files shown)


core/include/traccc/edm/details/device_container.hpp
```diff
@@ -34,6 +34,15 @@ class device_container
...
     using base_type::base_type;
 
+    /// Constructor from a "view object"
+    ///
+    /// Note that it may also be a "data" or "buffer" object. It just needs to
+    /// look like a "view object".
+    ///
+    template <template <typename, typename> class view_t>
+    TRACCC_HOST_DEVICE device_container(const view_t<header_t, item_t>& view)
+        : base_type(view.headers, view.items) {}
```
stephenswat, line 44: This is undefined behaviour if it is called on the host with host-inaccessible memory.

core/include/traccc/edm/details/container_element.hpp
```diff
@@ -21,11 +21,17 @@ namespace traccc {
...
  */
-template <typename header_t, typename vector_t>
+template <typename header_reference_t, typename vector_reference_t>
 struct container_element {
+
+    /// Header reference type
+    using header_reference = header_reference_t;
+    /// Vector reference type
+    using vector_reference = vector_reference_t;
+
```
stephenswat, line 34: Please add a static assertion to assert that these types are either primitives, references, or pointers, because this is a massive amount of (performance) bug surface area.

stephenswat (commented):
Thanks for putting the move constructors back.

core/include/traccc/edm/details/device_container.hpp
krasznaa: :confused: What?

Yes, it's also bad if you give null pointers to this constructor. But I really don't understand what's you're getting to here. This is the exact same kind of constructor that we have for all of the device types in vecmem. Are those constructors somehow fundamentally flawed as well because they make it possible to trigger undefined behaviour with them?

Note that CUDA is pretty good with making your code crash if you try to access GPU-only memory in host code. So I'm not even convinced that UB is the right term here...

core/include/traccc/edm/details/container_element.hpp
krasznaa: But they aren't. This is my whole point. When you access an "inner vector" in a `vecmem::jagged_device_vector`, you access a `vecmem::device_vector` that was created on the fly on the stack. Don't you remember?

The "device version" of this will have `header_reference == header_t&` and `vector_reference == vecmem::device_vector<item_t>`. By (current) design...

core/include/traccc/edm/details/device_container.hpp
stephenswat: Well, this was the reason why these types were so conservatively equipped with const-reference constructors in the first place, because they can be pretty dangerous in heterogeneous code. What I'm trying to say is: shouldn't this be a `TRACCC_DEVICE` function, rather than `TRACCC_HOST_DEVICE`? In what case would you actually want to use this on the host? :thinking:

core/include/traccc/edm/details/container_element.hpp
stephenswat: Ah, I see. I haven't used jagged vectors much, so I was not aware. Still, is there a reason why we want these types to be so general? I don't necessarily see why a more specific template type (from which we can then construct the device vector types) wouldn't suffice?

core/include/traccc/edm/details/device_container.hpp
krasznaa: For any number of testing purposes. :confused: In general I see very little value in making a function `TRACCC_DEVICE` only. Not in this project...

Note that by definition when we use this constructor in SYCL code, it will be compiled into "host code", whatever CUDA qualifiers we put on it. So only thinking about the CUDA qualifiers here is pretty flawed to start with.

stephenswat (approved):

PR #169, opened 2022-04-08
Add measurement creation implementation in SYCL

konradkusiak97:
This PR adds `SYCL clusterization algorithm` implementation to traccc - from which only `measurement creation` was so far ported to SYCL kernel. This opens a door for including SYCL component connection and spacepoint formation kernels - as the next steps. 

- Currently, the SYCL clusterization uses the CPU version of component connection algorithm from `traccc/core` and this step is done in a "per module" manner. 
-  Then, the SYCL measurement creation kernel performs the computation across clusters from all the modules combined. 
-  At last, the CPU spacepoint formation algorithm from `traccc/core` is used for each module to create the final `spacepoint_container`

The `Cluster EDM` was modified again as I realized that `std::function` cannot be used inside the kernel code. To work around that, I created a separate function: `signal_cell_modelling()` and the `position_estimation` member was replaced with `pixel_segmentation` functor similarly as in the `cell_module`.

pinging @krasznaa

### DIFF
--- a/core/include/traccc/clusterization/measurement_creation_helper.hpp
+++ b/core/include/traccc/clusterization/measurement_creation_helper.hpp
@@ -1,0 +1,57 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/definitions/common.hpp"
+#include "traccc/definitions/primitives.hpp"
+#include "traccc/definitions/qualifiers.hpp"
+#include "traccc/edm/cluster.hpp"
+
+namespace traccc {
+
+TRACCC_HOST_DEVICE
+inline scalar signal_cell_modelling(scalar signal_in,
+                                    const cluster_id& /*cl_id*/) {
+    // Retrieve the signal based on the cl_id.module_idx
+    return signal_in;
+}
+
+/// Function for pixel segmentation
+TRACCC_HOST_DEVICE
+inline vector2 position_from_cell(cell c, cluster_id cl_id) {
+    // Retrieve the specific values based on module idx
+    return {cl_id.pixel.min_center_x + c.channel0 * cl_id.pixel.pitch_x,
+            cl_id.pixel.min_center_y + c.channel1 * cl_id.pixel.pitch_y};
+}
+
+TRACCC_HOST_DEVICE
+inline vector2 get_pitch(cluster_id cl_id) {
+    // return the values based on the module idx
+    return {cl_id.pixel.pitch_x, cl_id.pixel.pitch_y};
+}
+
+TRACCC_HOST_DEVICE
+inline void calc_cluster_properties(const cell& c, const cluster_id& cl_id,
+                                    point2& mean, point2& var,
+                                    scalar& totalWeight) {
+    scalar weight = signal_cell_modelling(c.activation, cl_id);
+    if (weight > cl_id.threshold) {
+        totalWeight += c.activation;
+        const point2 cell_position = position_from_cell(c, cl_id);
+        const point2 prev = mean;
+        const point2 diff = cell_position - prev;
+
+        mean = prev + (weight / totalWeight) * diff;
+        for (std::size_t i = 0; i < 2; ++i) {
+            var[i] = var[i] + weight * (diff[i]) * (cell_position[i] - mean[i]);
+        }
+    }
+}
+
+}  // namespace traccc
--- a/core/include/traccc/clusterization/measurement_creation_helper.hpp
+++ b/core/include/traccc/clusterization/measurement_creation_helper.hpp
@@ -38,10 +38,10 @@ inline vector2 get_pitch(cluster_id cl_id) {
 
 /// Function used for calculating the properties of the cluster inside
 /// measurement creation
-template <template <typename> class vector_type>
+template <template <typename> class vector_type, typename cell_t>
 TRACCC_HOST_DEVICE inline void calc_cluster_properties(
-    const cell_collection<vector_type>& cluster, const cluster_id& cl_id,
-    point2& mean, point2& var, scalar& totalWeight) {
+    const vector_type<cell_t>& cluster, const cluster_id& cl_id, point2& mean,
+    point2& var, scalar& totalWeight) {
     for (const auto& cell : cluster) {
         scalar weight = signal_cell_modelling(cell.activation, cl_id);
         if (weight > cl_id.threshold) {
--- a/core/include/traccc/clusterization/measurement_creation_helper.hpp
+++ b/core/include/traccc/clusterization/measurement_creation_helper.hpp
@@ -15,10 +15,10 @@
 
 namespace traccc {
 
+/// Function used for retrieving the cell signal based on the module id
 TRACCC_HOST_DEVICE
 inline scalar signal_cell_modelling(scalar signal_in,
                                     const cluster_id& /*cl_id*/) {
-    // Retrieve the signal based on the cl_id.module_idx
     return signal_in;
 }
 
@@ -36,20 +36,25 @@ inline vector2 get_pitch(cluster_id cl_id) {
     return {cl_id.pixel.pitch_x, cl_id.pixel.pitch_y};
 }
 
-TRACCC_HOST_DEVICE
-inline void calc_cluster_properties(const cell& c, const cluster_id& cl_id,
-                                    point2& mean, point2& var,
-                                    scalar& totalWeight) {
-    scalar weight = signal_cell_modelling(c.activation, cl_id);
-    if (weight > cl_id.threshold) {
-        totalWeight += c.activation;
-        const point2 cell_position = position_from_cell(c, cl_id);
-        const point2 prev = mean;
-        const point2 diff = cell_position - prev;
+/// Function used for calculating the properties of the cluster inside
+/// measurement creation
+template <template <typename> class vector_type>
+TRACCC_HOST_DEVICE inline void calc_cluster_properties(
+    const cell_collection<vector_type>& cluster, const cluster_id& cl_id,
+    point2& mean, point2& var, scalar& totalWeight) {
+    for (const auto& cell : cluster) {
+        scalar weight = signal_cell_modelling(cell.activation, cl_id);
+        if (weight > cl_id.threshold) {
+            totalWeight += cell.activation;
+            const point2 cell_position = position_from_cell(cell, cl_id);
+            const point2 prev = mean;
+            const point2 diff = cell_position - prev;
 
-        mean = prev + (weight / totalWeight) * diff;
-        for (std::size_t i = 0; i < 2; ++i) {
-            var[i] = var[i] + weight * (diff[i]) * (cell_position[i] - mean[i]);
+            mean = prev + (weight / totalWeight) * diff;
+            for (std::size_t i = 0; i < 2; ++i) {
+                var[i] =
+                    var[i] + weight * (diff[i]) * (cell_position[i] - mean[i]);
+            }
         }
     }
 }
--- a/core/include/traccc/edm/cluster.hpp
+++ b/core/include/traccc/edm/cluster.hpp
@@ -60,23 +60,4 @@ using cluster_container_view = container_view<cluster_id, cell>;
 using cluster_container_const_view =
     container_view<const cluster_id, const cell>;
 
-/// Function for signal modelling
-inline scalar signal_cell_modelling(scalar signal_in,
-                                    const cluster_id& /*cl_id*/) {
-    // Retrieve the signal based on the cl_id.module_idx
-    return signal_in;
-}
-
-/// Function for pixel segmentation
-inline vector2 position_from_cell(cell c, cluster_id cl_id) {
-    // Retrieve the specific values based on module idx
-    return {cl_id.pixel.min_center_x + c.channel0 * cl_id.pixel.pitch_x,
-            cl_id.pixel.min_center_y + c.channel1 * cl_id.pixel.pitch_y};
-}
-
-inline vector2 get_pitch(cluster_id cl_id) {
-    // return the values based on the module idx
-    return {cl_id.pixel.pitch_x, cl_id.pixel.pitch_y};
-}
-
 }  // namespace traccc
--- a/core/include/traccc/edm/cluster.hpp
+++ b/core/include/traccc/edm/cluster.hpp
@@ -17,32 +17,10 @@
 
 namespace traccc {
 
-/// Function for signal modelling
-inline scalar signal_cell_modelling(scalar signal_in) {
-    return signal_in;
-}
-
-/// Function for pixel segmentation
-inline vector2 position_from_cell(cell c, pixel_data pixel, bool is_default) {
-    if (is_default)
-        return {static_cast<scalar>(c.channel0),
-                static_cast<scalar>(c.channel1)};
-    else
-        // Retrieve the specific values based on module idx (for now hard-coded)
-        return {pixel.min_center_x + c.channel0 * pixel.pitch_x,
-                pixel.min_center_y + c.channel1 * pixel.pitch_y};
-}
-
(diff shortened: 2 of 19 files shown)


konradkusiak97:
> and the `position_estimation` member was replaced with `pixel_segmentation` functor similarly as in the `cell_module`.

That was clearly not the best idea - the `test_cca` doesn't pass because of this. I will need to rethink my approach here... :thinking:

stephenswat:
Interesting, we can have a chat next week when I'll be back from my holiday, if you want.

konradkusiak97:
Now we're talkin. 

I changed the `pixel_segmentation` functor to a standalone function, which would in the future produce the result based on a module id. This function behaves in a twofold way, based on a `is_default` parameter that now is part od the `cluster_id`. I'm not sure if this is a way to go though... :thinking:

core/include/traccc/edm/cluster.hpp
```diff
@@ -9,31 +9,46 @@
...
-        return {static_cast<scalar>(ch0), static_cast<scalar>(ch1)};
-    };
-
     scalar threshold = 0.;
-    signal_modeling signal = [](scalar signal_in) -> scalar {
-        return signal_in;
-    };
+
+    pixel_data pixel;
+    bool is_default = true;
```
krasznaa, line 51: Could you explain a bit what this `is_default` bit is for? :thinking: I see that you use it, but do we really need a separate variable for it?

core/include/traccc/edm/cluster.hpp
```diff
@@ -9,31 +9,46 @@
...
+inline vector2 get_pitch(pixel_data pixel) {
+    // return the values based on the module idx (for now hard-coded)
+    return {pixel.pitch_x, pixel.pitch_y};
+}
+
+using position_estimation = std::function<vector2(channel_id, channel_id)>;
 struct cluster_id {
 
     event_id event = 0;
+    scalar module_idx = 0;
```
krasznaa, line 45: Please change this to `std::size_t`. A bit confusingly `scalar` refers to a floating point number. Either `double` or `float` in our current builds.

core/include/traccc/edm/cluster.hpp
```diff
@@ -9,31 +9,46 @@
...
 // System include(s).
 #include <functional>
 #include <vector>
 
 namespace traccc {
 
-using position_estimation = std::function<vector2(channel_id, channel_id)>;
-using signal_modeling = std::function<scalar(scalar)>;
+/// Function for signal modelling
+inline scalar signal_cell_modelling(scalar signal_in) {
```
krasznaa, line 21: This function will need to receive a cluster/module ID as well. Right now the "signal" from the detectors is not processed in any shape or form. But in principle there was a reason why this function was part of the "ID object" before. Because it could very much depend on the detector module, or even the specific cluster, how significant the same "digitised counts" from the detector would be treated as.

core/include/traccc/edm/cluster.hpp
```diff
@@ -9,31 +9,46 @@
...
+    return signal_in;
+}
+
+/// Function for pixel segmentation
+inline vector2 position_from_cell(cell c, pixel_data pixel, bool is_default) {
```
krasznaa, line 26: You should just pass the "cluster ID" object to this function. No need to pass the `pixel_data` and `bool` parameters individually. The function should be allowed to use any identification characteristic of the detector module to figure out how to calculate X-Y coordinates for a cell.

device/sycl/include/traccc/sycl/clusterization/measurement_creation.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include "traccc/edm/cluster.hpp"
+#include "traccc/edm/measurement.hpp"
+#include "traccc/sycl/utils/queue_wrapper.hpp"
+#include "traccc/utils/algorithm.hpp"
+
+namespace traccc::sycl {
+
+struct measurement_creation
+    : public algorithm<host_measurement_container(
+          const host_cluster_container &, const std::vector<std::size_t> &,
```
krasznaa, line 22: It seems redundant to pass both the cluster container and a vector describing the sizes of the cluster container as input to this algorithm. The information stored in the `std::vector<std::size_t>` variable is available from the `traccc::host_cluster_container` object as well. Passing it separately just gives a chance for passing incorrect values to the algorithm.

device/sycl/src/clusterization/measurement_creation.sycl
```diff
@@ -0,0 +1,138 @@
...
+                    return;
+
+                // items: cluster of cells at current idx
+                // header: cluster_id object with the information about the cell
+                // module
+                const auto &cluster = clusters_device.get_items().at(idx);
+                const auto &cluster_id = clusters_device.get_headers().at(idx);
+
+                const auto pitch = get_pitch(cluster_id.pixel);
+                const auto module_idx = cluster_id.module_idx;
```
krasznaa, line 77: There is way too much `auto` being used here for my taste. :frowning: Please use explicit types, otherwise the lay reader will have no idea what is going on in this code...

krasznaa (changes requested):
I'm on board with all of this of course, I just have the mandatory technical comments...

core/include/traccc/edm/cluster.hpp
konradkusiak97: So as far as I understand, if the "whole" clusterization algorithm is called (with component connection -> measurement creation - > spacepoint formation), the `position_estimation()` function returns the `vector2` based on the pixel_data. So my understanding before was that the "default" behaviour of this function which was previously:
 ```cpp
position_estimation position_from_cell = [](channel_id ch0,
                                                channel_id ch1) -> vector2 {
        return {static_cast<scalar>(ch0), static_cast<scalar>(ch1)};
    };
```
is basically never called.
But I saw that this is not true and it is called in some cases - in some of the tests. I suppose that those tests are running just measurement creation on its own and that's when the default behaviour is called. 

So this bit is basically always switched to false if the whole clusterization algorithm is called but otherwise it's true - and this guides the behaviour of the `position_estimation` function. 

Not sure if that makes sense... :stuck_out_tongue: I'm aware that this is not ideal since from what I understand, in the end this would be removed and position estimation would be just: 
```cpp
inline vector2 position_from_cell(cell c, pixel_data pixel) {
    return {pixel.min_center_x + c.channel0 * pixel.pitch_x,
            pixel.min_center_y + c.channel1 * pixel.pitch_y};
```
instead of:
```cpp
inline vector2 position_from_cell(cell c, pixel_data pixel, bool is_default) {
 if (is_default)
        return {static_cast<scalar>(c.channel0),
                static_cast<scalar>(c.channel1)};
    else
        // Retrieve the specific values based on module idx (for now hard-coded)
        return {pixel.min_center_x + c.channel0 * pixel.pitch_x,
                pixel.min_center_y + c.channel1 * pixel.pitch_y};
}
```
But I haven't found any other way to work around it

core/include/traccc/edm/cluster.hpp
konradkusiak97: An unused module ID parameter won't compile in a full debug mode. Is there a way to work around it?

konradkusiak97: Done.

core/include/traccc/edm/cluster.hpp
konradkusiak97: done

device/sycl/include/traccc/sycl/clusterization/measurement_creation.hpp
konradkusiak97: Alright, this is done too.

core/include/traccc/edm/cluster.hpp
konradkusiak97: oh of course, that's true...  :stuck_out_tongue: I changed it now.

device/sycl/src/clusterization/measurement_creation.sycl
```diff
@@ -0,0 +1,150 @@
...
+                //     edition, chapter 4.2.2.
+                point2 mean = {0., 0.}, var = {0., 0.};
+
+                // Should not happen
+                if (cluster.empty()) {
+                    return;
+                }
+
+                for (const auto &cell : cluster) {
+                    scalar weight =
```
beomki-yeo, line 110: Let's write a helper function which can be shared by host and device algorithm like we did for the seeding algorithm. We don't have to repeat the same codes

device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
```diff
@@ -0,0 +1,116 @@
...
+#include "traccc/edm/spacepoint.hpp"
+#include "traccc/geometry/pixel_data.hpp"
+
+// clusterization
+#include "traccc/clusterization/component_connection.hpp"
+#include "traccc/clusterization/spacepoint_formation.hpp"
+#include "traccc/sycl/clusterization/measurement_creation.hpp"
+
+#ifdef _OPENMP
+#include "omp.h"
```
beomki-yeo, line 23: Do we need openMP for this?

device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
```diff
@@ -0,0 +1,116 @@
...
+ * (c) 2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include "traccc/edm/cell.hpp"
+#include "traccc/edm/cluster.hpp"
+#include "traccc/edm/internal_spacepoint.hpp"
```
beomki-yeo, line 12: We might not need this header

beomki-yeo (commented):
Looks good in general. It would be a lot better if host and device can share sub-algorithmic codes.

device/sycl/src/clusterization/measurement_creation.sycl
krasznaa: It's a good idea. Let's move the code from this for-loop into a separate helper function. Which you should also update the CPU code to use.

I'm thinking of something like:

```c++
TRACCC_HOST_DEVICE
void calc_cluster_properties(const cluster& cl, point2& mean, point2& variance,
                             scalar& totalWeight);
```

konradkusiak97: I made a `measurement_creation_helper.hpp` inside core/clusterization where I put the functions that are reused for CPU and SYCL.

I was thinking at first of making a `traccc::common` library, similarly as you've done in `device/common` in #170. Would that be something to consider for the future?

device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
konradkusiak97: You're right, we don't

device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
konradkusiak97: Correct, thanks! :wink:

device/sycl/src/clusterization/measurement_creation.sycl
krasznaa: Not exactly. I should have been a little clearer... :frowning:

Please move the for-loop into the new function as well. Don't call that function from the same sort of for-loop in both the device and the host code. This is why I wrote that the function should receive an entire cluster, not just one cell.

konradkusiak97: I templated the function to receive either `host_cell_collection` or `device_cell_collection`. For the device code, it would actually need to receive a `device_cell_const_collection`... but for now I "casted away" the const in SYCL code to make it work.

core/include/traccc/clusterization/measurement_creation_helper.hpp
```diff
@@ -0,0 +1,62 @@
...
+            const point2 diff = cell_position - prev;
+
+            mean = prev + (weight / totalWeight) * diff;
+            for (std::size_t i = 0; i < 2; ++i) {
+                var[i] =
+                    var[i] + weight * (diff[i]) * (cell_position[i] - mean[i]);
+            }
+        }
+    }
+}
```
krasznaa, line 60: No, let's not introduce any new code that would have to use `const_cast` just to work...

Please try to template the function like:

```c++
template <template <typename> class vector_t, typename cell_t>
TRACCC_HOST_DEVICE
void calc_cluster_properties(const vector_t<cell_t>& cluster, const cluster_id& cl_id,
                              point2& mean, point2& car, scalar& totalWeight) {...}
```

This type of function should be able to receive both const and non-const clusters.

You could also try to make this work using `traccc::container_element`, but I also don't see off the bat how to make that work nicely. So let's just stay with templating both the vector, and the exact cell type.

konradkusiak97: Alright, thanks!

device/sycl/src/clusterization/measurement_creation.sycl
beomki-yeo: > I was thinking at first of making a traccc::common library
IMAO, It should be enough to make `traccc/clusterization/common` directory and move helper files into it.

krasznaa (approved):
Some things are still a bit suspicious. When I run the current main branch's code on my system, I get:

```
[bash][Legolas]:build > SYCL_DEVICE_FILTER=cuda ./bin/traccc_seq_example_sycl --detector_file=tml_detector/trackml-detector.csv --cell_directory=tml_full/ttbar_mu140/ --events=1 --run_cpu=1
Running ./bin/traccc_seq_example_sycl tml_detector/trackml-detector.csv tml_full/ttbar_mu140/ 1
Module libc not found.
event 0
 number of seeds (cpu): 2654
 number of seeds (sycl): 2654
 seed matching rate: 1
 track parameters matching rate: 1
==> Statistics ... 
- read    67281 spacepoints from 12486 modules
- created        234275 cells           
- created        67281 meaurements     
- created        67281 spacepoints     
- created (cpu)  2654 seeds
- created (sycl) 2654 seeds
==> Elpased time ... 
wall time           29.5874   
file reading (cpu)        0.498042  
clusterization_time (cpu) 1.51321   
seeding_time (cpu)        0.074182  
seeding_time (sycl)       0.0674212 
tr_par_esti_time (cpu)    0.00134518
tr_par_esti_time (sycl)   0.00210228
[bash][Legolas]:build >
```

While this PR's code gives me:

```
[bash][Legolas]:build > SYCL_DEVICE_FILTER=cuda ./bin/traccc_seq_example_sycl --detector_file=tml_detector/trackml-detector.csv --cell_directory=tml_full/ttbar_mu140/ --events=1 --run_cpu=1
Running ./bin/traccc_seq_example_sycl tml_detector/trackml-detector.csv tml_full/ttbar_mu140/ 1
Running Seeding on device: NVIDIA GeForce RTX 3080
Module libc not found.
event 0
 number of seeds (cpu): 2654
 number of seeds (sycl): 2654
 measurements matching rate: 1
 seed matching rate: 0.963451
 track parameters matching rate: 1
==> Statistics ... 
- read    67281 spacepoints from 12486 modules
- created        234275 cells           
- created        67281 meaurements     
- created (sycl) 67281 measurements     
- created        67281 spacepoints     
- created (cpu)  2654 seeds
- created (sycl) 2654 seeds
==> Elpased time ... 
wall time           23.6961   
file reading (cpu)        0.501552  
clusterization_time (cpu) 0.0110778 
clusterization_time (sycl) 1.70964   
seeding_time (cpu)        0.0733167 
seeding_time (sycl)       0.0635474 
tr_par_esti_time (cpu)    0.000555182
tr_par_esti_time (sycl)   0.00140205
[bash][Legolas]:build >
```

I.e. the clusterization time measurement seems very suspicious. :confused:

But let's still go ahead finally, we'll deal with further code harmonisation and performance issues in further PRs.

konradkusiak97:
> 

This cpu clusterization being much faster is from changing it to work on host_memory_resource. In the version in main (before this PR being merged) it was using shared_memory_resource.

beomki-yeo:
I think you have to run multiple events to warm up the GPU device

beomki-yeo:
NVM it doesn't make big difference with 10 events..

PR #170, opened 2022-04-08
CUDA/SYCL Code Sharing, main branch (2022.04.08.)

krasznaa:
Shared the spacepoint binning code between CUDA and SYCL. Moved the CUDA algorithms into functions that could be used by both the CUDA and SYCL code, avoiding code duplication.

Unfortunately I realised only now that I'll need to re-write the SYCL code to use a spacepoint grid buffer as well for it to be able to use the newly introduced `traccc::device::populate_grid` function. And I'm just not willing to start with that this evening anymore...

Note that I bumped into some interesting stuff along the way. For instance how `detday::grid2`'s "device constructor" is only usable in device code with CUDA. I was not able to use that constructor from a `TRACCC_HOST_DEVICE` function. It's sort of fitting that we had a long exchange with @stephenswat about exactly this just today...

This is very much work in progress at the moment, even though the CUDA code does work already. But I'll only consider it being ready once the SYCL code works correctly with the shared code as well. At the same time, I only aim to share the code of the spacepoint binning in this PR. This should then serve as a template for how we would synchronize all other parts of the seedfinding in subsequent PRs as well.


device/cuda/src/seeding/spacepoint_binning.cu
```diff
@@ -0,0 +1,92 @@
...
+    kernels::count_grid_capacities<<<num_blocks, num_threads>>>(
+        m_config, m_axes.first, m_axes.second, traccc::get_data(spacepoints),
+        vecmem::get_data(sp_prefix_sum), vecmem::get_data(grid_capacities));
+    CUDA_ERROR_CHECK(cudaGetLastError());
+    CUDA_ERROR_CHECK(cudaDeviceSynchronize());
+
+    // Create the grid buffer.
+    sp_grid_buffer grid_buffer(m_axes.first, m_axes.second,
+                               std::vector<std::size_t>(grid_bins, 0),
+                               std::vector<std::size_t>(grid_capacities.begin(),
```
beomki-yeo, line 77: Hmm you removed the minimum capacity of 60 that I added in #160. 
I don't know if it's a bug in vecmem, the small pileup ttbar sample (e.g. ttbar<40>) made run time error.

krasznaa: Kudos for noticing!

Yes, I removed that hack, hoping to bump into the issue that made you add it. I think you're right, there may very well be an issue with the VecMem code. @konradkusiak97 saw something very similar in his code, where he ended up adding +1 to every size to avoid a crash.

Thanks for the hint about the mu40 samples. I'll play around with those a bit.

device/cuda/src/seeding/spacepoint_binning.cu
```diff
@@ -0,0 +1,97 @@
...
+                                                        grid_capacities.end()),
+                               m_mr.get());
+    vecmem::cuda::copy copy;
```
beomki-yeo, line 83: `vecmem::copy` ?

device/common/include/traccc/seeding/device/impl/count_grid_capacities.ipp
```diff
@@ -0,0 +1,58 @@
...
+        sp_prefix_sum(sp_prefix_sum_view);
+    if (globalIndex >= sp_prefix_sum.size()) {
+        return;
+    }
+
+    // Get the spacepoint that we need to look at.
+    const get_spacepoint_prefix_sum_element_t sp_idx =
+        sp_prefix_sum[globalIndex];
+    device_spacepoint_const_container spacepoints(spacepoints_view);
+    const spacepoint sp = spacepoints.at(sp_idx);
```
beomki-yeo, line 38: shouldn't const reference be better here?

beomki-yeo, line 38: shouldn't const reference be better here?

device/common/include/traccc/seeding/device/impl/populate_grid.ipp
```diff
@@ -0,0 +1,55 @@
...
+        sp_prefix_sum(sp_prefix_sum_view);
+    if (globalIndex >= sp_prefix_sum.size()) {
+        return;
+    }
+
+    // Get the spacepoint that we need to look at.
+    const get_spacepoint_prefix_sum_element_t sp_idx =
+        sp_prefix_sum[globalIndex];
+    device_spacepoint_const_container spacepoints(spacepoints_view);
+    const spacepoint sp = spacepoints.at(sp_idx);
```
beomki-yeo, line 34: Same here

beomki-yeo, line 34: Same here

device/common/include/traccc/seeding/device/impl/count_grid_capacities.ipp
beomki-yeo, line 38: shouldn't const reference be better here?

device/common/include/traccc/seeding/device/impl/populate_grid.ipp
beomki-yeo, line 34: Same here

beomki-yeo (approved):

device/common/include/traccc/seeding/device/impl/count_grid_capacities.ipp
krasznaa: It is very good that you asked this question.

I don't have any measurements in these particular cases. But in the Acts code (https://github.com/acts-project/acts/tree/main/Plugins/Cuda/src/Seeding2) I did experiment with this just a little bit some time ago. I saw that pulling small objects into local (register) memory was usually a better idea compared to accessing them multiple times from global memory.

Since remember, when you use a (const) reference, it's only the pointer/reference that is in the thread's register. As soon as you try to access something from the referenced object, the thread has to go to global memory to access the info. Of course caching has improved a lot on GPUs in recent years, so in this case none of this might be important at all.

But I'm definitely standing on shaky ground with this... And unfortunately I just couldn't find the usernames of Nicolin on Xola, so let me pull them into this discussion in some other way... :thinking:

nicolin: for scalar variables const really does not make much of a difference. For CUDA we can have a look at the generated PTX code as NVCC will put it in a register anyway. PS: unless you need the extra range offered by 32 bit int you can do int16 which will save registers on Ampere and above GPUs as half precision ints will be packed into 32bit registers

krasznaa: :thinking: Let's just go ahead like this for now. Probably it doesn't even make a difference whether we use references here or not, as the compiler may be smart enough to generate "the optimal" code in either case.

Doing this type of profiling of how well our code behaves is the exact sort of thing that I want Xola and Nicolin to help us with. So if later on we find that we should tweak this code, we will of course do it. :wink:

PR #173, opened 2022-04-14
Rewrite seed collection comparator

beomki-yeo:
It took like a year to compare seeding result between HOST and DEVICE (@konradkusiak97 Didn't it happen to you?)
To avoid massive copy, I changed the container member values to const references


beomki-yeo:
It still takes a lot for SYCL...

konradkusiak97:
Yes, I noticed this later, that it is really slow while comparing those...

beomki-yeo:
PR is ready now. I rewrote the seed comparator by flattening seed collection into `vector<vector<spacepoint>>`

core/include/traccc/edm/seed.hpp
```diff
@@ -46,56 +46,60 @@ struct seed {
...
+        result.reserve(3);
         result.push_back(spacepoints.at(spB_link).meas);
         result.push_back(spacepoints.at(spM_link).meas);
         result.push_back(spacepoints.at(spT_link).meas);
         return result;
     }
+
+    TRACCC_HOST
+    std::vector<spacepoint> get_spacepoints(
+        const host_spacepoint_container& spacepoints) const {
```
krasznaa, line 62: :thinking: How would you feel about using `std::array<measurement, 3>` and `std::array<spacepoint, 3>` objects instead?

core/include/traccc/edm/seed.hpp
```diff
@@ -46,56 +46,60 @@ struct seed {
...
 inline bool operator==(const seed& lhs, const seed& rhs) {
     return (lhs.spB_link == rhs.spB_link && lhs.spM_link == rhs.spM_link &&
             lhs.spT_link == rhs.spT_link);
 }
 
-template <typename spacepoint_container_t>
-bool is_same(const seed& seed1, const spacepoint_container_t& container1,
-             const seed& seed2, const spacepoint_container_t& container2) {
+inline bool operator==(const std::vector<spacepoint>& lhs,
+                       const std::vector<spacepoint>& rhs) {
```
krasznaa, line 78: If we used arrays, we could avoid the runtime for-loop from the check. It may even be possible to declare the comparison in a variadic template way where the comparison operator would not require arrays of size 3. :thinking: Though I don't know off the bat how exactly that would have to be written.

krasznaa (commented):
I'm still on board with this, we may just want to consider the vector -> array switch...

core/include/traccc/edm/seed.hpp
beomki-yeo: Makes sense.
I was taking account of the case where we have different size of seeds, but I don't think we will do such crazy thing.

core/include/traccc/edm/seed.hpp
stephenswat: Hm? Arrays (of arbitrary size) have a well-defined `operator==` which you could use for the comparison. Or is there something I am missing there?

beomki-yeo: This equality operation is removed now

beomki-yeo:
Any approval unless there are more comments?

stephenswat (approved):

PR #175, opened 2022-04-15
General Jagged Vector Prefix Sums, main branch (2022.04.15.)

krasznaa:
Introduced generic prefix sum making functions, replacing the one that was specific for the spacepoint container.

While starting to work on the next set of kernels to make them share code between the CUDA and SYCL implementations, I had to realise that it will make sense to create prefix sums for some of those as well. Even though currently they are not using this setup. (They rather rely on the kernels figuring out which jagged vector elements they should look at using for-loops in the kernel.)

Long story short, this is just a generalisation of the existing code...


stephenswat (approved):
It's a little too specific for my liking in the sense that it only works on these size vectors, but I suppose this is acceptable for the time being. Admittedly I'm a little confused why we're doing these prefix sums on the host. :confused:

stephenswat:
Also, I'm guessing the title of this PR is a type (and that it should be host), might be worth fixing it for clarity.

krasznaa:
What would be an efficient way to build up such prefix sums in device code? :confused: I'm absolutely open to doing it that way, but I thought that this was something that would be much faster to do on the host...

The "device" in the title is because the function is called `traccc::device::get_prefix_sum(...)`. Which I called like this because we'll only want to produce such prefix sums for execution on a device. In single-threaded code we can happily loop over the jagged vectors in "double for loops".

But when it comes to the title of the PR, I don't really care...

stephenswat:
There's a _lot_ of literature on how to do parallel prefix sums:

1. https://developer.download.nvidia.com/compute/cuda/2_2/sdk/website/projects/scan/doc/scan.pdf
2. https://www.sciencedirect.com/science/article/pii/0890540189900369
3. https://escholarship.org/content/qt6j57h5zw/qt6j57h5zw.pdf

Etc. Could be worth looking into. :smile:

beomki-yeo:
Actually it's not a bad idea to do it on the device code as @stephenswat suggested. We can use only `vector_buffer` for prefix sum rather than creating `vector`, unless we use it in the host side. I will take a look on it later...

PR #177, opened 2022-04-18
Add binary file I/O

beomki-yeo:
This PR adds binary I/O which makes file reading faster.  Fast file reading will make multi-process study (e.g. CUDA-MPS or MIG) much easier. 

------------------
### 1. How to use

User can run the existing examples with binary input file. :
```bash
# data_format options has default value of csv
./bin/traccc_seq_example --detector_file=tml_detector/trackml-detector.csv --cell_directory=tml_full/ttbar_mu200/  --events=1 --input-binary
```

Even though I am not going to add binary files into `traccc-data` repository, users can generate binary files by themselves using `traccc_create_binaries` example

```bash
# This will generate one binary file from event000000000-cells.csv
./bin/traccc_create_binaries --detector_file=tml_detector/trackml-detector.csv --cell_directory=tml_full/ttbar_mu200/  --events=1 

# This will generate one binary file from event000000000-hits.csv
./bin/traccc_create_binaries --detector_file=tml_detector/trackml-detector.csv --hit_directory=tml_full/ttbar_mu200/  --events=1 
```

------------------
### 2. Benchmark

Here is the cell file reading benchmark with one 200 pileup event.

  | host resource | cuda managed resource
-- | -- | --
csv file reading [sec] | 0.66 | 1.15
binary file reading [sec] | 0.007 | 0.14

Note that binary file reading with cuda managed memory takes 0.14 second, which is still much longer than cuda seeding (~0.02 second).

More details on profiling will be provided later. 

Update:
It's obvious most of time is taken by the bunch of cudaMalloc call for jagged vector :/ But I believe that we can minimize the effect with downstream memory caching later...

### DIFF
--- a/examples/run/cpu/seq_example.cpp
+++ b/examples/run/cpu/seq_example.cpp
@@ -56,14 +56,14 @@ int seq_run(const traccc::full_tracking_input_config& i_cfg,
     sd_performance_writer.add_cache("CPU");
 
     // Loop over events
-    for (unsigned int event = common_opts.skip; event < common_opts.events + common_opts.skip;
-         ++event) {
+    for (unsigned int event = common_opts.skip;
+         event < common_opts.events + common_opts.skip; ++event) {
 
         // Read the cells from the relevant event file
         traccc::host_cell_container cells_per_event =
             traccc::read_cells_from_event(event, i_cfg.cell_directory,
-                                          common_opts.data_format, surface_transforms,
-                                          host_mr);
+                                          common_opts.input_data_format,
+                                          surface_transforms, host_mr);
 
         /*-------------------
             Clusterization
--- a/examples/run/cpu/seq_example.cpp
+++ b/examples/run/cpu/seq_example.cpp
@@ -20,6 +20,7 @@
 #include "traccc/efficiency/seeding_performance_writer.hpp"
 
 // options
+#include "traccc/options/common_options.hpp"
 #include "traccc/options/full_tracking_input_options.hpp"
 #include "traccc/options/handle_argument_errors.hpp"
 
@@ -29,7 +30,8 @@
 
 namespace po = boost::program_options;
 
-int seq_run(const traccc::full_tracking_input_config& i_cfg) {
+int seq_run(const traccc::full_tracking_input_config& i_cfg,
+            const traccc::common_options& common_opts) {
 
     // Read the surface transforms
     auto surface_transforms = traccc::read_geometry(i_cfg.detector_file);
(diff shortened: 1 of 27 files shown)


examples/run/cpu/seq_example.cpp
```diff
@@ -105,11 +105,6 @@ int seq_run(const traccc::full_tracking_input_config& i_cfg) {
             sd_performance_writer.write("CPU", seeds, spacepoints_per_event,
                                         evt_map);
         }
-
-        traccc::write_measurements(event, measurements_per_event);
-        traccc::write_spacepoints(event, spacepoints_per_event);
-        traccc::write_seeds(event, spacepoints_per_event, seeds);
-        traccc::write_estimated_track_parameters(event, params);
```
stephenswat, line 112: Is there a reason why we are no longer writing the output of these executables to disk? I imagine this might still be useful, right?

io/include/traccc/io/binary.hpp
```diff
@@ -0,0 +1,108 @@
...
+/// Function for binary file writing
+///
+/// @param in_name is the input filename which includes the path
+/// @param copy is the vecem copy helper object
+/// @param resource is the vecmem memory resource
+template <typename container_t>
+container_t read_binary(const std::string& in_name, vecmem::copy& copy,
+                        vecmem::memory_resource& resource) {
+    std::ifstream in_file(in_name, std::ios::binary);
+
```
stephenswat, line 68: I would add the same static checks here.

io/include/traccc/io/binary.hpp
```diff
@@ -0,0 +1,108 @@
...
+namespace traccc {
+
+/// Function for binary file writing
+///
+/// @param out_name is the output filename which includes the path
+/// @param container is the input traccc container
+template <typename container_t>
+void write_binary(const std::string& out_name, const container_t& container) {
+    std::ofstream out_file(out_name, std::ios::out | std::ios::binary);
+
```
stephenswat, line 29: Sorry, apparently Github did not like one of my comments. I am a little bit worried about marshaling data to and from disk byte-wise like this, especially because we compile this code with so many different compilers. The Itanium ABI is relatively stable and should enable us to do this in a somewhat stable fashion, but only if our data (both header and item types) are PODs. Thus I would like to suggest we add the following; 

```suggestion
    static_assert(std::is_trivial_v<container_t::header_t>, "Container header type must be trivial.");
    static_assert(std::is_standard_layout_v<container_t::header_t>, "Container header type must be standard layout.");
    static_assert(std::is_trivial_v<container_t::item_t>, "Container item type must be trivial.");
    static_assert(std::is_standard_layout_v<container_t::item_t>, "Container item type must be standard layout.");
```

examples/run/cpu/seq_example.cpp
beomki-yeo: I guess nobody has been using this feature. Even though we might need to write final track fit result to disk at some point, I believe we need to implement it with a better interface

io/include/traccc/io/reader.hpp
```diff
@@ -34,39 +37,84 @@ inline traccc::geometry read_geometry(const std::string &detector_file) {
...
+    vecmem::memory_resource &resource) {
+
     // Read the cells from the relevant event file
-    std::string io_cells_file =
-        data_directory() + cells_dir + get_event_filename(event, "-cells.csv");
-    traccc::cell_reader creader(
-        io_cells_file,
-        {"geometry_id", "hit_id", "cannel0", "channel1", "activation", "time"});
-    return traccc::read_cells(creader, resource, &surface_transforms);
+    if (data_format == "csv") {
```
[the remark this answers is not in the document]
beomki-yeo: That's a good comment so I switched to use enumeration.

beomki-yeo:
@stephenswat Could you review again?

stephenswat (approved):
Looks good to me.

PR #178, opened 2022-04-26
Detray Update, main branch (2022.04.26.)

krasznaa:
Updated the project to [Detray v0.6.0](https://github.com/acts-project/detray/releases/tag/v0.6.0).

To make this happen, I updated the include paths of the headers that were moved in the new version. Which was the simple part of the whole thing.

What was more tricky is that Detray modified the definition of `detray::darray` in the new tag. Which made me realise that `detray::darray` was also declared in this project. Instead of updating this project's definition of `detray::darray` (and `traccc::darray`), I decided to re-write how [algebra-plugins](https://github.com/acts-project/algebra-plugins) would be set up in this project.

@beomki-yeo, of course things should become a **lot** cleaner once the plugins will be used as template arguments in Detray. But for now this was the best that I could come up with...

Note that this also revealed an interesting bug in the code, with a missing `std::move(...)` call in the CUDA code. Since the seed-finding algorithm needs an rvalue reference as input, it makes sense that we need to use `std::move(...)` there. But I was still surprised that this was only revealed by the compiler now, after the latest round on "`detray::grid2` cleanups".

### DIFF
--- a/core/include/traccc/definitions/primitives.hpp
+++ b/core/include/traccc/definitions/primitives.hpp
@@ -46,24 +46,3 @@ template <unsigned int kSize>
 using traccc_sym_matrix = Eigen::Matrix<scalar, kSize, kSize>;
 
 }  // namespace traccc
-
-namespace detray {
-
-using scalar = traccc::scalar;
-
-template <typename value_type, unsigned int kDIM>
-using darray = traccc::darray<value_type, kDIM>;
-
-template <typename value_type>
-using dvector = traccc::dvector<value_type>;
-
-template <typename value_type>
-using djagged_vector = traccc::djagged_vector<value_type>;
-
-template <typename key_type, typename value_type>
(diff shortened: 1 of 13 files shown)


core/include/traccc/definitions/primitives.hpp
```diff
@@ -8,12 +8,16 @@
 #pragma once
 
 #if ALGEBRA_PLUGINS_INCLUDE_ARRAY
+#include "detray/plugins/algebra/array_definitions.hpp"
```
beomki-yeo, line 11: I wonder if you can include detray definitions headers in traccc definition headers rather than here?

krasznaa: That indeed sounds like a better setup... :thinking:

beomki-yeo (approved):

PR #179, opened 2022-04-27
Add Component connection in SYCL

konradkusiak97:
As the title says, this PR adds Component Connection implementation in SYCL. With this, we'll have a full clusterization algorithm ported to SYCL. Unfortunately, the performance is not the best... but there might be some paths to search for improvement :crossed_fingers: 

#### Some notable changes: 
- Sparce_ccl was rewritten to accept references of objects subjected to change rather to return the results as a `std::pair`. It was also templated on the vector type (host/device). With this we can reuse this code for the CPU and SYCL implementations.
- The SYCL algorithm is parallelized across modules and it outputs the `host_cluster_container` which is then passed to SYCL measurement creation.

#### Initial performance results
- When comparing the performance with the CPU version, there seems to be a big difference, depending if we use shared_memory_resource for the CPU algorithm or host_memory resource: 

Running on the HOST - cpu(host_mr) sycl(shared_mr)
```
clusterization_time (cpu) 0.0065893 
clusterization_time (sycl) 1.12543   
```
Running on the HOST - cpu(shared_mr) sycl(shared_mr)
```
clusterization_time (cpu) 0.763698  
clusterization_time (sycl) 1.12048   
```
Running on Intel Integrated Graphics 630 - cpu(shared_mr) sycl(shared_mr)
```
clusterization_time (cpu) 0.754215  
clusterization_time (sycl) 12.8749   
```
Running on NVIDIA RTX 2060 - cpu(shared_mr) sycl(shared_mr)
```
clusterization_time (cpu) 1.02629   
clusterization_time (sycl) 0.758348
```

So one's hope might be that when we switch to using device_memory_resource for the SYCL implementations, things would improve :thinking: 

There are also some next steps in optimizing this algorithm. Right now, the output cluster container has too much memory allocated in itself. The way to get rid of that would be to go with the similar approach as in the seeding code, where we have a separate kernel at the beginning that counts the number of clusters that will be created. 

P.S.
(I know that we talked about it with @krasznaa but in the end, I commented out the operator for a device vector of core component connection. This right now wouldn't work with the changed sparse_ccl - in order to keep it in, I would need to template sparse_ccl on yet another vector type :thinking: ?)


core/include/traccc/clusterization/detail/sparse_ccl.hpp
```diff
@@ -56,7 +60,8 @@ inline unsigned int make_union(std::vector<unsigned int>& L, unsigned int e1,
 /// @param b the second cell
 ///
 /// @return boolan to indicate 8-cell connectivity
-inline bool is_adjacent(cell a, cell b) {
+template <typename cell_t>
+TRACCC_HOST_DEVICE inline bool is_adjacent(cell_t a, cell_t b) {
```
krasznaa, line 64: :thinking: Why did you need to template the cell type? Don't we have just a single cell type in both host and device code? :confused:

I also wonder if passing the cells to this function with const references may be a better idea. Cells are small, but are they small enough for passing around by value? (I'm honestly not sure...)

core/include/traccc/edm/cell.hpp
```diff
@@ -123,9 +123,22 @@ using host_cell_container = host_container<cell_module, cell>;
...
 /// Convenience declaration for the cell container data type to use in host code
 using cell_container_data = container_data<cell_module, cell>;
 
+/// Convenience declaration for the cell container data type to use in host code
+/// (const)
+using cell_const_container_data = container_data<const cell_module, const cell>;
+
+/// Convenience declaration for the cell container view type to use in host code
+/// (const)
+using cell_const_container_view = container_view<const cell_module, const cell>;
```
krasznaa, line 140: Let's call these `cell_container_const_data` and `cell_container_const_view` instead. Just to keep with the naming used for the spacepoint types.

device/sycl/include/traccc/sycl/clusterization/component_connection.hpp
```diff
@@ -0,0 +1,49 @@
...
+    /// @param cells_per_event are the input cells grouped by module
+    host_cluster_container operator()(
+        const host_cell_container &c) const override;
+
+    /// Callable operator for component connection for all the cells per event
+    ///
+    /// @param cells_per_event are the input cells grouped by module
+    /// @return a container of clusters from all the modules
+    void operator()(const host_cell_container &cells_per_event,
+                    output_type &clusters) const;
```
krasznaa, line 42: It's fine to implement a function with this interface as well. But in general, this was not needed. It would be enough to implement the "standard" algorithm function.

The reason that the host version of the algorithm outsourced the execution into a function like this was to make it possible to use the same code with both host- and device containers. That is not a requirement here.

And if/when we'll start to re-think how containers are passed around, we'll need to touch this code anyway. So we might as well keep it as simple as possible for now.

krasznaa (changes requested):
Apart from some technicalities, I really like the code!

core/include/traccc/clusterization/detail/sparse_ccl.hpp
konradkusiak97: I templated them because we're passing just `cell` in the CPU code but `const cell` in the device. I can change it to receive `cell_t&` then? :thinking:

krasznaa: As we discussed before lunch, let's just go back to passing `cell` objects by value to these functions. You just need to add `TRACCC_HOST_DEVICE` on them in this PR.

device/sycl/include/traccc/sycl/clusterization/component_connection.hpp
konradkusiak97: I see. I removed this extra `operator()` step.

core/include/traccc/edm/cell.hpp
konradkusiak97: done

krasznaa (approved):
Let's go ahead like this. We'll probably add tweaks later on anyway.

PR #181, opened 2022-05-02
 Rewrite SYCL clusterization to reduce copies 

konradkusiak97:
The SYCL component connection and measurement creation algorithms were rewritten to reduce the copies (Device to Device) that were slowing down the execution significantly. The design of the execution of the clusterization chain was modified to mimic the seedfinder code. As a result we now have something like this (the arrow is a replacement of word "calls" :smile: )
```
clusterization_algorithm -> cluster_finding -> (component connection kernel, measurement creation kernel) -> result (measurements)
```
With this, I modified slightly the component connection kernel, so that no copies need to be made to measurement creation kernel. This increased the performance roughly 8-10 times on all platforms (You can compare with numbers from #179).

#### Some simple benchmarks
NVIDIA RTX 2060 (shared_mr), CPU (host_mr)
```
clusterization_time (cpu) 0.00841581
clusterization_time (sycl) 0.0802062
```
Intel Integrated Graphics 630 (shared_mr), CPU (host_mr)
```
clusterization_time (cpu) 0.00609147
clusterization_time (sycl) 1.66646
```
HOST (shared_mr), CPU (host_mr)
```
clusterization_time (cpu) 0.00640465
clusterization_time (sycl) 0.381006
```
pinging @krasznaa


device/sycl/src/clusterization/cluster_finding.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    traccc::sycl::measurement_creation(measurement_buffer, clusters_buffer,
+                                       range, m_queue);
+
+    host_measurement_container measurements(&m_mr.get());
+    copy(measurement_buffer.items, measurements.get_items());
+
+    for (std::size_t i = 0; i < num_modules; ++i) {
+        auto module = cells_per_event.get_headers()[i];
+        measurements.get_headers().push_back(std::move(module));
```
krasznaa, line 92: I would've simply written:

```c++
measurements.get_headers().push_back(cells_per_event.get_headers()[i]);
```

You could've also replaced the whole for-loop with:

```c++
for (const cell_module& cm : cells_per_event.get_headers()) {
    measurements.get_headers().push_back(cm);
}
```

But in this case it would be worth adding an

```c++
assert(cells_per_event.get_headers().size() == measurements.get_items().size())
```

check, just in case...

device/sycl/src/clusterization/component_connection.hpp
```diff
@@ -0,0 +1,32 @@
...
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// SYCL library include(s).
+#include "traccc/sycl/utils/queue_wrapper.hpp"
+
+// Project include(s).
+#include "traccc/clusterization/detail/sparse_ccl.hpp"
```
krasznaa, line 14: This include is not needed in the header.

device/sycl/src/clusterization/component_connection.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include "traccc/edm/cell.hpp"
+#include "traccc/edm/cluster.hpp"
+
+// Vecmem include(s).
+#include <vecmem/memory/memory_resource.hpp>
```
krasznaa, line 19: You should add an explicit include for `vecmem::data::vector_view`.

device/sycl/src/clusterization/component_connection.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+namespace traccc::sycl {
+
+/// Forward declaration of component connection function
+///
+void component_connection(
+    cluster_container_view clusters_view, cell_container_const_view cells_view,
+    const host_cell_container& cells_per_event,
```
krasznaa, line 27: The function should either receive a `cell_container_const_view` or a `host_cell_container` parameter. The two describe the same container, so only one is necessary. (You can create a `cell_container_const_view` inside of the function from `host_cell_container` if you choose to pass the latter to the function.)

device/sycl/src/clusterization/component_connection.sycl
```diff
@@ -5,86 +5,54 @@
...
-    output_type clusters(&m_mr.get());
+void component_connection(
+    cluster_container_view clusters_view, cell_container_const_view cells_view,
+    const host_cell_container& cells_per_event,
+    vecmem::data::vector_view<unsigned int> clusters_count_view,
+    vecmem::data::vector_view<unsigned int> total_clusters_view,
+    vecmem::memory_resource& resource, queue_wrapper queue) {
 
     // Execution size of the algorithm
     unsigned int N = cells_per_event.size();
```
krasznaa, line 27: Please use `std::size_t` for size variables instead of `unsigned int` or `unsigned long`.

Generally `unsigned int` should only be used when we want to handle that variable atomically on the device.

device/sycl/src/clusterization/component_connection.sycl
```diff
@@ -5,86 +5,54 @@
...
-    cluster_container_view clusters_view(clusters_buffer);
-
-    // Cells view to pass to the device
-    auto cells_data = get_data(cells_per_event, &m_mr.get());
-    cell_container_const_view cells_view(cells_data);
-
     // Get the sizes of the cells in each module
     std::vector<std::size_t> cell_sizes(N, 0);
-    for (unsigned long j = 0; j < cell_sizes.size(); ++j) {
+    for (unsigned long j = 0; j < N; ++j) {
```
krasznaa, line 31: Same comment about `std::size_t`.

krasznaa (changes requested):
A bunch of technical comments, but of course I'm quite happy about the update itself. :wink:

device/sycl/src/clusterization/cluster_finding.cpp
```diff
@@ -0,0 +1,98 @@
...
+        {std::vector<std::size_t>(cells_total_size, 0),
+         std::vector<std::size_t>(cells_total_size, max_cluster_size),
+         m_mr.get()}};
+
+    // Vector for counts of clusters per each module
+    vecmem::vector<unsigned int> cluster_sizes(num_modules, 1, &m_mr.get());
+
+    // Atomic count of all clusters (needed inside component connection kernel
+    // but also here)
+    vecmem::vector<unsigned int> total_clusters(1, 0, &m_mr.get());
```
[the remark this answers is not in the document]
konradkusiak97: So the `total_clusters` is needed anyway inside the component connection kernel (its intermediate atomic sums are used for correct indexing). 

For the `cluster_sizes`, using `copy.get_sizes()` doesn't really work here because what I need is the vector of "number of clusters per each module", where the get_sizes gives me numbers of cells in each cluster - it's the jagged vector of `vector<vector<cell>`.

device/sycl/src/clusterization/cluster_finding.cpp
```diff
@@ -0,0 +1,98 @@
...
+            ->size();
+
+    // Cluster container buffer for the clusters and headers (cluster ids)
+    // assuming the worst case where 1 cell = 1 cluster
+    // and max. number of cells per module is max. number of cells per cluster
+    cluster_container_buffer clusters_buffer{
+        {cells_total_size, m_mr.get()},
+        {std::vector<std::size_t>(cells_total_size, 0),
+         std::vector<std::size_t>(cells_total_size, max_cluster_size),
+         m_mr.get()}};
```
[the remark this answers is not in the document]
konradkusiak97: So the idea here is that the data from this buffer is not needed on the host side at all. It stays on the device for the next kernel to use. I thought that would also work when we use host/device resources? :thinking:

device/sycl/src/clusterization/component_connection.hpp
konradkusiak97: Oh wow, I don't know how did that happen, of course, those are the same.

device/sycl/src/clusterization/component_connection.sycl
konradkusiak97: done

device/sycl/src/clusterization/component_connection.sycl
konradkusiak97: done

device/sycl/src/clusterization/component_connection.hpp
konradkusiak97: done

device/sycl/src/clusterization/component_connection.hpp
konradkusiak97: done

device/sycl/src/clusterization/cluster_finding.cpp
konradkusiak97: done

device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
```diff
@@ -7,12 +7,8 @@
 
 #pragma once
 
-#include "traccc/edm/measurement.hpp"
-#include "traccc/edm/spacepoint.hpp"
-
 // clusterization
-#include "traccc/sycl/clusterization/component_connection.hpp"
-#include "traccc/sycl/clusterization/measurement_creation.hpp"
+#include "traccc/sycl/clusterization/cluster_finding.hpp"
```
krasznaa, line 11: We should completely remove this class (`traccc::sycl::clusterization_algorithm`) in a future PR. Not necessarily now. But by now this algorithm is just a useless wrapper around `traccc::sycl::cluster_finding`.

krasznaa (approved):
Let's finally go ahead with it... :+1:

PR #184, opened 2022-05-05
Const Correctness Updates, main branch (2022.05.05.)

krasznaa:
As the final bite-sized update before the monster that will unify the CUDA and SYCL doublet finding implementations, this PR is simply making it possible to access constant and non-constant host objects through constant views in device code.

@beomki-yeo, you may recognise `traccc::details::is_same_nc` from https://github.com/acts-project/detray/pull/252. :wink: Although this version is just a tiny bit different from `detray::details::is_same_nc` (and `vecmem::details::is_same_nc`).

Since hopefully this is not too controversial, if it could be approved quickly, I could still open that "monster PR" still today. :wink: (Pinging @stephenswat and @paulgessinger for good measure...)

### DIFF
--- a/core/include/traccc/utils/type_traits.hpp
+++ b/core/include/traccc/utils/type_traits.hpp
@@ -1,0 +1,32 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+namespace traccc::details {
+
+/// Helper trait for detecting when a type is a non-const version of another
+///
+/// This comes into play multiple times to enable certain constructors
+/// conditionally through SFINAE.
+///
+template <typename CTYPE, typename NCTYPE>
+struct is_same_nc {
+    static constexpr bool value = false;
+};
+
+template <typename TYPE>
+struct is_same_nc<TYPE, TYPE> {
+    static constexpr bool value = true;
+};
+
+template <typename TYPE>
+struct is_same_nc<const TYPE, TYPE> {
+    static constexpr bool value = true;
+};
+
+}  // namespace traccc::details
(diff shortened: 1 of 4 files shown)


core/include/traccc/utils/type_traits.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+template <typename TYPE>
+struct is_same_nc<TYPE, TYPE> {
+    static constexpr bool value = true;
+};
+
+template <typename TYPE>
+struct is_same_nc<const TYPE, TYPE> {
+    static constexpr bool value = true;
+};
```
stephenswat, line 30: I would be in favour of getting rid of this entire file and just writing `std::is_same_v<std::decay_t<T1>, std::std::decay_t<T2>>` instead.

stephenswat, line 30: Even if we keep it this, there are a few problems with it that we might want to consider:

1. This does not properly handle the case where the first type is non-const and the second type is const.
2. This breaks down if we are talking about references or pointers, in which case the reference or pointer type constructor is applied after the const type constructor, and must be stripped from it. Although that may be intended.

stephenswat (approved):
This PR really represents another worrying step in the direction of having to many typedefs that nobody really understands what's happening in traccc anymore. :worried:

core/include/traccc/utils/type_traits.hpp
krasznaa: That would not give the same result... There is a good reason why we have this same setup in vecmem as well. As you may remember, we couldn't find a more standard way for expressing that a type should either be the same as another one, or the constant version of it.

I.e. your proposal would also allow a non-const view to be constructed from a const object.

krasznaa: This was the intention... The trait should "only work in one direction".

stephenswat: Then I would argue it is very confusingly named. :fearful:

krasznaa:
After some separate back-and-forth, let's just go with this for now...

PR #185, opened 2022-05-05
Common Device Doublet Finding, main branch (2022.05.05.)

krasznaa:
After a **lot** of preparations, this is the 1000lb gorilla that I was talking about since a while...

I created a common implementation for the doublet finding between the CUDA and SYCL libraries. There are a lot of updates here, so let me try to just highlight some stuff:
  - I added a common type of doublet counter, which required steps after the doublet finding to be updated as well. Which is one reason why I ended up touching so many files.
  - Instead of letting the device threads figure out which elements of the input jagged vectors they should work with (through the `traccc::cuda::cuda_helper::find_idx_on_jagged_vector`, `traccc::cuda::cuda_helper::find_idx_on_container`, `traccc::sycl::sycl_helper::find_idx_on_jagged_vector` and `traccc::sycl::sycl_helper::find_idx_on_container` functions), the doublet counting and doublet finding kernels now use host-created prefix sums for this purpose instead.
    * Making use of the previously introduced `traccc::device::get_prefix_sum` helper function.
  - I removed the thread-synchronization/collaboration from the doublet finding kernels. The way that shared memory is used there at the moment for slightly reducing the number of atomic updates into global memory just didn't seem to make a lot of sense/difference for me. But @beomki-yeo may very well disaggree.
    * I further delayed the introduction of dealing with thread synchronisation in a uniform way in these shared functions like this. But I do have ideas for how to implement thread synchronisation uniformly between CUDA and SYCL if/when the time comes.

Finally, just to show that the updates don't make things worse. In fact, they make the seed finding a little faster. (The absolute values are not too relevant, but all measurements between the old/current and new codes were done on the same hardware. So the relative values are meaningful.)

| &#965; | Current CUDA | New CUDA | Current SYCL | New SYCL |
|-|-|-|-|-|
| 10 | 0.0456415 | 0.0353137 | 0.160323 | 0.0521146 |
| 40 | 0.0552883 | 0.0429054 | 0.258301 | 0.0871926 |
| 80 | 0.0665741 | 0.0558218 | 0.31336 | 0.12144 |
| 140 | 0.0904985 | 0.0809856 | 0.383457 | 0.169536 |
| 200 | 0.117286 | 0.107773 | 0.443463 | 0.217447 |

As you can see, one of my main motivations for all this harmonisation is that the SYCL code is a lot slower by now. Not having received a number of updates that Beomki made in the CUDA code. Hopefully once all kernels are harmonised in this way, the performance numbers will be a lot closer to each other...

I'm curious what everybody will think. :wink:


beomki-yeo (approved):
Lovely. Let's merge this in

> I removed the thread-synchronization/collaboration from the doublet finding kernels. 

I think it is a good change. I have realized that it doesn't improve the performance actually.

beomki-yeo:
I want to point out a few things for your future work.

- `weight_updating` is not sharing the code from CPU seed filtering algorithm. I tried to do that when I initially implemented it, but couldn't find a nice way. Maybe you can take a look.
- `seed_selecting` is the kernel that I am not pretty satisfied with. It's quite unreadable and the performance is worst among the seed finding kernels. The performance degradation (I guess) is from the bad indexing strategy. You can also think about how to improve it...

Of course, those points do not have to be solved in your next PRs for CUDA-SYCL code sharing, but just as a note :)

PR #186, opened 2022-05-06
Container Cleanup, main branch (2022.05.06.)

krasznaa:
This is probably going to be a little controversial. Especially since I myself am also not 100% sure that this is the best way to go. But here it goes...

When I started setting up the `traccc::device::triplet_counter` type today as the next step after #185. I really got fed up with copy-pasting a bunch of standard type declarations. Which we have a **lot** of. And they are even a bit inconsistent here and there.

So I decided to set up the collection / container types through helper macros instead. I introduced `TRACCC_DECLARE_COLLECTION_TYPES(...)` and `TRACCC_DECLARE_CONTAINER_TYPES(...)`. Which one could use like:

```c++
// Declare all doublet collection/container types
TRACCC_DECLARE_COLLECTION_TYPES(doublet);
TRACCC_DECLARE_CONTAINER_TYPES(doublet, doublet_per_bin, doublet);
```

As the in-line documentation lists, `TRACCC_DECLARE_COLLECTION_TYPES` declares all of the following type names:
  - `foo::bar_collection<T>`
  - `foo::bar_const_collection<T>`
  - `foo::host_bar_collection`
  - `foo::device_bar_collection`
  - `foo::device_bar_const_collection`

While `TRACCC_DECLARE_CONTAINER_TYPES` sets up all the following:
  - `foo::host_bar_container`
  - `foo::device_bar_container`
  - `foo::device_bar_const_container`
  - `foo::bar_container_view`
  - `foo::bar_container_const_view`
  - `foo::bar_container_data`
  - `foo::bar_container_const_data`
  - `foo::bar_container_buffer`

All this ensures that all const and non-const types would always be declared, in a consistent way.

While at it, I also made some of the EDM functions a little nicer.

Now... Using macros in EDM declarations is quite evil. There is a reason why we've forbidden macro usage in the xAOD headers. So I'm not super happy about what this does to the readability of our code. At the same time, the code repetition is just going out of hand by now. :frowning: If not like this, I would go with trait classes for declaring all these types. Which may be a better idea in the end to be honest. (The reason I didn't start with that implementation is that I only thought of it right now... :stuck_out_tongue:)

So depending on what everyone thinks, we may or may not want to go with this macro-heavy implementation...

### DIFF
--- a/core/include/traccc/edm/cell.hpp
+++ b/core/include/traccc/edm/cell.hpp
@@ -7,31 +7,23 @@
 
 #pragma once
 
-// traccc include(s).
+// Library include(s).
+#include "traccc/definitions/common.hpp"
 #include "traccc/definitions/primitives.hpp"
 #include "traccc/edm/container.hpp"
 #include "traccc/geometry/pixel_data.hpp"
 
-// VecMem include(s).
-#include <vecmem/containers/data/jagged_vector_buffer.hpp>
-#include <vecmem/containers/data/vector_buffer.hpp>
-#include <vecmem/containers/device_vector.hpp>
-#include <vecmem/containers/jagged_device_vector.hpp>
-#include <vecmem/containers/jagged_vector.hpp>
-#include <vecmem/containers/vector.hpp>
-
 // System include(s).
+#include <cmath>
 #include <limits>
 
 namespace traccc {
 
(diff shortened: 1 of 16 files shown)


krasznaa:
Closing in favour of #188.

PR #187, opened 2022-05-09
Optimized SYCL Clusterization

konradkusiak97:
Here are some further optimizations for the SYCL clusterization algorithm. By now, I think this code is as optimized as it can be, but don't hesitate to free me from this illusion. 

#### Right now the algorithmic structure is the following:
- `Clusters_sum` is the first kernel that counts the number of clusters that will be found. In order to do that, it runs sparse_ccl. The cluster indices for each cell that are found  are saved in a jagged vector buffer - for the next kernels to use. In this setup, sparse_ccl is performed only once, in this reduction kernel. The output that goes back to the host is the single number - the `total_clusters`. (The indices from sparse_ccl stay on the device for the next kernel to read from).

- `Cluster_counting` kernel is the next and it fills the `cluster_sizes` vector with the sizes of each cluster. The trick here is to make sure that the next kernel - component connection, fills each cluster with cells at the appropriate index. For that, we have the following logic in the cluster_counting:
```cpp
// get the number of already written clusters from other threads
auto prefix_sum = atomicAdd(&all_clusters, num_clusters)

for (i = 0; i < cells_size; ++i) {

     // get the cluster index for the current cell
     cindex = sparse_ccl_indices[i]

     // Increment the size at this position
     cluster_sizes[prefix_sum + cindex]++ 
}

// save the prefix sum for this thread idx
prefix_sums_vector[idx] = prefix_sum;
``` 
- Then, coming to the last kernel: `Component_connection`, it retrieves the prefix sum for each idx and pushes the cells at the appropriate position.

This approach ensures that we don't allocate more memory than actually needed. And it is faster for about `30-40%` than before. Current numbers are a little bit better than this because the `vecmem::sycl::copy` was changed to just `vecmem::copy` which should be used with shared memory for the CUDA backend, as  @krasznaa made me aware of.
So here it goes (for comparison, the previous numbers are in #181): 
- Intel UHD Graphics 630
```
clusterization_time (cpu) 0.00593337
clusterization_time (sycl) 0.44185
```
- Nvidia GeForce RTX 2060
```
clusterization_time (cpu) 0.00697169
clusterization_time (sycl) 0.0220094 
```


konradkusiak97:
UPDATE: The above logic of prefix sums is moved to the cluster_sums kernel. There, the prefix sums vector is filled out and the next kernels can access it. Cluster_counting and Component_connection kernels were made 2-dimensional with each thread corresponding to 1 cell in a module. In this setup, the x dimension is the number of modules and the y dimension is the maximum number of cells per module. 

This made the total time of kernel execution 2x faster. This is not very visible with those new numbers below but it can be seen with nsight compute :wink: 
- Intel UHD Graphics 630
```
clusterization_time (cpu) 0.00609065
clusterization_time (sycl) 0.415542  
```
- Nvidia GeForce RTX 2060
```
clusterization_time (cpu) 0.00701469
clusterization_time (sycl) 0.0199222 
```

device/sycl/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -0,0 +1,126 @@
...
+         m_mr.get()}};
+    copy.setup(measurement_buffer.items);
+
+    // Measurement creation kernel
+    traccc::sycl::measurement_creation(measurement_buffer, clusters_buffer,
+                                       *total_clusters, m_queue);
+
+    // Copy the results back to the host
+    host_measurement_container measurements(&m_mr.get());
+    copy(measurement_buffer.items, measurements.get_items());
```
beomki-yeo, line 114: I have a quick question; how long does this line take?

device/sycl/src/clusterization/clusters_sum.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <vecmem/memory/unique_ptr.hpp>
+
+namespace traccc::sycl {
+
+/// Forward declaration of clusters sum function
+///
+void clusters_sum(
+    const cell_container_types::host& cells_per_event,
+    vecmem::data::jagged_vector_view<unsigned int> sparse_ccl_indices_view,
+    vecmem::unique_alloc_ptr<unsigned int>& total_clusters,
```
krasznaa, line 26: This is not a good way for passing this variable to the function. :thinking: A general rule of thumb is that unique pointers should only ever be passed around by value. When you pass a unique pointer by (const) reference, you should instead just pass the underlying object by (const) reference. Since you don't want to pass ownership to the function. And at that point the function shouldn't have to know how the lifetime of the object in question is managed in the caller site.

Long story short, just pass this as `unsigned int&`. :wink:

device/sycl/src/clusterization/clusterization_algorithm.cpp
konradkusiak97: That's around 0.012 s

device/sycl/src/clusterization/component_connection.hpp
```diff
@@ -24,8 +23,9 @@ namespace traccc::sycl {
 void component_connection(
     cluster_container_view clusters_view,
     const cell_container_types::host& cells_per_event,
-    vecmem::data::vector_view<unsigned int> clusters_count_view,
-    vecmem::unique_alloc_ptr<unsigned int>& total_clusters,
-    vecmem::memory_resource& resource, queue_wrapper queue);
+    vecmem::data::jagged_vector_view<unsigned int> sparse_ccl_indices_view,
+    vecmem::data::vector_view<std::size_t> cluster_prefix_sum_view,
+    const std::size_t& cells_max, vecmem::memory_resource& resource,
```
krasznaa, line 28: Just pass `cells_max` by value. Primitive types should in 99% of the cases just be passed by value.

device/sycl/src/clusterization/measurement_creation.hpp
```diff
@@ -12,15 +12,12 @@
...
-#include <vecmem/memory/memory_resource.hpp>
-
 namespace traccc::sycl {
 
 /// Forward decleration of measurement creation kernel
 ///
 void measurement_creation(measurement_container_view measurements_view,
                           cluster_container_view clusters_view,
-                          const unsigned int& range, queue_wrapper queue);
+                          const std::size_t& n_clusters, queue_wrapper queue);
```
krasznaa, line 21: Why do you need to pass this size as a separate variable? :thinking: Isn't this the same as `clusters_view.m_size`?

krasznaa (changes requested):
I know, I know... I just keep giving you work with my conflicting updates... :frowning:

Please resolve the conflicts, and make the technical changes that I pointed out. After that we should go ahead with this, before I add any more conflicts for you. :stuck_out_tongue:

device/sycl/src/clusterization/clusters_sum.hpp
konradkusiak97: Hm, I see. So I actually need the pointer inside this kernel so I will just pass the raw pointer then.

krasznaa: Inside the function, when starting the kernel, yes. You should pass a pointer there.

But in the signature of this function you should have a reference. That way the caller of this function is forced to provide a valid object. (Or at least they have to go out of their way to pass a non-valid reference.) Whenever a function receives a pointer, you have to add a check that it would be a valid pointer. So generally we only pass things by pointer in modern C\+\+ when the argument is allowed to be a null pointer as well. (For optional arguments in general.)

konradkusiak97: I don't believe I follow... (I understand why we don't usually pass raw pointers in modern C++) but here, I need this pointer for the atomic addition inside the kernel. And the results of this addition I need back in the main clusterization algorithm. 

So do you mean that I should pass total clusters by `unsigned int&`, then inside the clusters_sum function I create the vecmem unique ptr, pass it to the kernel, do atomics on it and then after the kernel finishes, I assign the results from the pointer to the `unsigned int& total_clusters`?

krasznaa: I mean that you should implement the code as:

```c++
void clusters_sum(..., unsigned int& total_clusters, ...) {
   ...
   queue.parallel_for<...>(range, [..., total_clusters = &total_clusters, ...] {...});
   ...
}
```

I.e. the function should receive a const reference. But you will need to pass a pointer to the kernel inside the function. Because non-const references can't easily be passed to a kernel. (If at all...)

konradkusiak97: Gotch ya, done.

device/sycl/src/clusterization/measurement_creation.hpp
konradkusiak97: Alright, I changed it to `clusters_view.headers.size()`

device/sycl/src/clusterization/component_connection.hpp
konradkusiak97: Done.

krasznaa (approved):
Let's just go forward with this. We'll anyway need to do some serious profiling on the CUDA/SYCL code still.

PR #188, opened 2022-05-09
Container Cleanup, main branch (2022.05.09.)

krasznaa:
This is meant as a replacement for #186.

Instead of declaring "top level" types through macros, I introduced the `traccc::collection_types` and `traccc::container_types` trait structs for declaring all the types that one may want to use in (device) code.

I then updated `traccc::device::doublet_counter` to make use of these traits. Updating all of the code that uses this recently introduced type to make use of the trait types.

Note that replacing all EDM container types like this will be more work than what #186 would've required. But we can do those updates one-by-one, which should hopefully help with making the PRs palatable... And this way we can avoid using macros. Which, as I wrote in #186 as well, I didn't really like to begin with.

I'm curious what everybody will think about this. :smile:

### DIFF
--- a/core/include/traccc/edm/container.hpp
+++ b/core/include/traccc/edm/container.hpp
@@ -19,6 +19,8 @@
 #include <vecmem/containers/data/jagged_vector_view.hpp>
 #include <vecmem/containers/data/vector_buffer.hpp>
 #include <vecmem/containers/data/vector_view.hpp>
+#include <vecmem/containers/device_vector.hpp>
+#include <vecmem/containers/vector.hpp>
 
 // System include(s).
 #include <type_traits>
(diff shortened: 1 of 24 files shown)


core/include/traccc/edm/container.hpp
```diff
@@ -122,4 +124,62 @@ inline container_data<const header_t, const item_t> get_data(
             {vecmem::get_data(cc.get_items(), resource)}};
 }
 
+/// Type trait defining all "collection types" for an EDM class
+template <typename item_t>
+struct collection_types {
+
+    /// @c item_t must not be a constant type
+    static_assert(std::is_const<item_t>::value == false,
+                  "The template parameter must not be a constant type");
```
stephenswat, line 133: We may also want to make sure that the type is not a reference.

PR #189, opened 2022-05-10
Cell EDM Update, main branch (2022.05.10.)

krasznaa:
Getting drunk on the success of #188, I went ahead with simplifying `cell.hpp`.

This required touching a good number of files, most of which were simple type name replacements.

But I decided to use this occasion to do a bit of a bigger re-write on the CCL code as part of the EDM changes. The `traccc::component_connection` algorithm was just not a nice thing in its latest implementation. :frowning:

The big update is that I removed **all** templating from the sparse CCL code. Since that code can really just be implemented on top of "constant views" of the cell EDM. So that's what I did, I made it use `traccc::cell_collection_types::const_device` in all cases. :smile: Even when using device code. Since setting up "device containers" in the host implementation is really no biggie. (Note that I also just switched to providing the "output array" to the CCL code through a bare pointer. Since all access to that memory was already done through non-protected array accesses anyway.)

As before, I ran some performance checks, just to be sure that nothing really bad happened as a result of the updates... These are the reported "clusterization times" from `traccc_seq_example_sycl`:

| &#965; | Current CPU | New CPU | Current SYCL | New SYCL |
|-|-|-|-|-|
| 40 | 0.0211379 | 0.0229111 | 0.397941 | 0.43047 |
| 80 | 0.0358985 | 0.0361159 | 0.570159 | 0.571931 |
| 200 | 0.0746189 | 0.0753304 | 1.06429 | 0.976579 |

Note that I would not expect these to change at all as a result of this update. And while one may see a systematic slowdown for the CPU times if looking at the numbers, I think that's dominantly a statistical uncertainty...

If you guys don't disagree, I'll go on to clusters next. :wink:

### DIFF
--- a/core/include/traccc/clusterization/detail/sparse_ccl.hpp
+++ b/core/include/traccc/clusterization/detail/sparse_ccl.hpp
@@ -7,9 +7,15 @@
 
 #pragma once
 
+// Library include(s).
 #include "traccc/definitions/qualifiers.hpp"
 #include "traccc/edm/cell.hpp"
-#include "traccc/edm/cluster.hpp"
+
+// VecMem include(s).
+#include <vecmem/containers/device_vector.hpp>
+
+// System include(s).
+#include <cassert>
 
 namespace traccc {
 
@@ -24,12 +30,14 @@ namespace detail {
 /// @param L an equivalance table
 ///
 /// @return the root of @param e
-template <template <typename> class vector_t>
 TRACCC_HOST_DEVICE inline unsigned int find_root(
-    const vector_t<unsigned int>& L, const unsigned int& e) {
+    const vecmem::device_vector<unsigned int>& L, unsigned int e) {
+
     unsigned int r = e;
+    assert(r < L.size());
     while (L[r] != r) {
         r = L[r];
+        assert(r < L.size());
     }
     return r;
 }
@@ -39,16 +47,17 @@ TRACCC_HOST_DEVICE inline unsigned int find_root(
 /// @param L an equivalance table
 ///
 /// @return the rleast common ancestor of the entries
-template <template <typename> class vector_t>
-TRACCC_HOST_DEVICE inline unsigned int make_union(vector_t<unsigned int>& L,
-                                                  const unsigned int& e1,
-                                                  const unsigned int& e2) {
+TRACCC_HOST_DEVICE inline unsigned int make_union(
+    vecmem::device_vector<unsigned int>& L, unsigned int e1, unsigned int e2) {
+
     int e;
     if (e1 < e2) {
         e = e1;
+        assert(e2 < L.size());
         L[e2] = e;
     } else {
         e = e2;
+        assert(e1 < L.size());
         L[e1] = e;
     }
     return e;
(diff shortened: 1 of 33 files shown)


core/include/traccc/clusterization/detail/sparse_ccl.hpp
```diff
@@ -24,12 +27,18 @@ namespace detail {
 /// @param L an equivalance table
 ///
 /// @return the root of @param e
-template <template <typename> class vector_t>
-TRACCC_HOST_DEVICE inline unsigned int find_root(
-    const vector_t<unsigned int>& L, const unsigned int& e) {
+TRACCC_HOST_DEVICE inline unsigned int find_root(const unsigned int* L,
+                                                 unsigned int e,
+                                                 unsigned int size) {
```
stephenswat, line 32: This I don't like - we should not be introducing additional parameters simply for debugging assertions. Instead, we should turn `L` back into a vector reference, which will also give us the size without passing it separately.

core/include/traccc/clusterization/detail/sparse_ccl.hpp
```diff
@@ -24,12 +27,18 @@ namespace detail {
...
+
+    // Don't complain about an unused variable in optimised builds.
+    (void)size;
+
     unsigned int r = e;
+    assert(r < size);
```
stephenswat, line 38:
```suggestion
    assert(r < L.size());
```

core/include/traccc/clusterization/detail/sparse_ccl.hpp
```diff
@@ -24,12 +27,18 @@ namespace detail {
...
     while (L[r] != r) {
         r = L[r];
+        assert(r < size);
```
stephenswat, line 41:
```suggestion
        assert(r < L.size());
```

core/include/traccc/clusterization/detail/sparse_ccl.hpp
```diff
@@ -39,16 +48,22 @@ TRACCC_HOST_DEVICE inline unsigned int find_root(
...
 ///
 /// @return the rleast common ancestor of the entries
-template <template <typename> class vector_t>
-TRACCC_HOST_DEVICE inline unsigned int make_union(vector_t<unsigned int>& L,
-                                                  const unsigned int& e1,
-                                                  const unsigned int& e2) {
+TRACCC_HOST_DEVICE inline unsigned int make_union(unsigned int* L,
+                                                  unsigned int e1,
+                                                  unsigned int e2,
+                                                  unsigned int size) {
```
stephenswat, line 54: Same here, `L` should be a vector.

core/include/traccc/clusterization/detail/sparse_ccl.hpp
```diff
@@ -83,20 +98,22 @@ TRACCC_HOST_DEVICE inline bool is_far_enough(traccc::cell a, traccc::cell b) {
 /// @param L is the vector of the output indices (to which cluster a cell
 /// belongs to)
 /// @param labels is the number of clusters found
-template <template <typename> class vector_t, typename cell_t>
-TRACCC_HOST_DEVICE inline void sparse_ccl(const vector_t<cell_t>& cells,
-                                          vector_t<unsigned int>& L,
-                                          unsigned int& labels) {
+TRACCC_HOST_DEVICE inline void sparse_ccl(
+    const cell_collection_types::const_device& cells, unsigned int* L,
```
stephenswat, line 102:
```suggestion
    const cell_collection_types::const_device& cells, const std::vector<unsigned int> & L,
```

core/src/clusterization/component_connection.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    // Set up a device collection on top of the host collection.
+    const cell_collection_types::const_view cells_view =
+        vecmem::get_data(cells);
+    const cell_collection_types::const_device cells_device(cells_view);
+
+    // Run the algorithm
+    unsigned int num_clusters = 0;
+    std::vector<unsigned int> connected_cells(cells.size());
+    detail::sparse_ccl(cells_device, connected_cells.data(), num_clusters);
```
stephenswat, line 32:
```suggestion
    detail::sparse_ccl(cells_device, connected_cells, num_clusters);
```

stephenswat (approved):

core/include/traccc/clusterization/detail/sparse_ccl.hpp
krasznaa: Moved the code to using `vecmem::device_vector<unsigned int>`...

core/include/traccc/clusterization/detail/sparse_ccl.hpp
krasznaa: Moved the code to using `vecmem::device_vector<unsigned int>`...

core/include/traccc/clusterization/detail/sparse_ccl.hpp
```diff
@@ -83,20 +98,22 @@ TRACCC_HOST_DEVICE inline bool is_far_enough(traccc::cell a, traccc::cell b) {
...
     unsigned int start_j = 0;
-    for (unsigned int i = 0; i < cells.size(); ++i) {
+    for (unsigned int i = 0; i < n_cells; ++i) {
         L[i] = i;
         int ai = i;
         if (i > 0) {
             for (unsigned int j = start_j; j < i; ++j) {
                 if (is_adjacent(cells[i], cells[j])) {
-                    ai = make_union<vector_t>(L, ai, find_root<vector_t>(L, j));
+                    ai = make_union(L, ai, find_root(L, j, n_cells), n_cells);
```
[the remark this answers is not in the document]
krasznaa: Went with the `vecmem::device_vector` route instead...

core/include/traccc/clusterization/detail/sparse_ccl.hpp
krasznaa: Moved the code to using `vecmem::device_vector<unsigned int>`...

stephenswat (approved):

PR #190, opened 2022-05-11
Cluster EDM Update, main branch (2022.05.11.)

krasznaa:
Continuing on from #188 and #189, here I update the definition of the cluster EDM container(s).

As before, this EDM update also triggered some algorithmic code update. I now hid the implementation of `traccc::measurement_creation` into `measurement_creation.cpp`, moved the helper functions used by measurement creation into the `traccc::detail` namespace, and then updated all the project for the name changes.

P.S. This might very well conflict with #187, so we'll need to be thoughtful about the order of their merge...


stephenswat (approved):
Looks like harmless code shuffling. :+1:

PR #192, opened 2022-05-16
Add SYCL Spacepoint formation

konradkusiak97:
This PR adds the Spacepoint formation in SYCL. With this in place we will have all the pieces to  start playing with using just device memory all the way from cells to seeding :metal:  :cowboy_hat_face: 

I didn't separate this implementation from the clusterization algorithm (similarly to the CPU version) because there was some information needed from the "clusterization chain", to construct the buffers for the spacepoints. 

I also rewrote the current SYCL clusterization algorithm to make use of the code that constructs [prefix sums](https://github.com/acts-project/traccc/blob/main/device/common/include/traccc/device/get_prefix_sum.hpp) out of our 2-dimensional containers that @krasznaa  recently added. Getting rid of 2-dim kernels as a result. This logic is also implemented for the spacepoint formation.

#### Some benchmarks:
- Intel UHD Graphics 630
```
clusterization_time (cpu) 0.00602939
spacepoint_formation_time (cpu) 0.0152623 
clusterization_time (sycl) 0.370286  
```
NVIDIA RTX 2060 (shared_mr), CPU (host_mr)
```
clusterization_time (cpu) 0.00562569
spacepoint_formation_time (cpu) 0.0138134 
clusterization_time (sycl) 0.0251253 
```
Keep in mind that sycl clusterization here means clusterization + spacepoint formation

### DIFF
--- a/device/sycl/CMakeLists.txt
+++ b/device/sycl/CMakeLists.txt
@@ -30,6 +30,8 @@ traccc_add_library( traccc_sycl sycl TYPE SHARED
   "src/clusterization/measurement_creation.hpp"
   "src/clusterization/component_connection.hpp"
   "src/clusterization/component_connection.sycl"
+  "src/clusterization/spacepoint_formation.hpp"
+  "src/clusterization/spacepoint_formation.sycl"
   "src/seeding/seed_finding.sycl"
   "src/seeding/seed_selecting.hpp"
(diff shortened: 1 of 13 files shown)


krasznaa (approved):
I didn't find anything screamingly incorrect in it, so let's just go ahead like this. We anyway will be tweaking the code quite a bit for better performance as the next step.

PR #193, opened 2022-05-17
Measurement EDM Update, main branch (2022.05.17.)

krasznaa:
Updated the measurement containers to the "new declaration style", following in the footsteps of #188, #189 and #190.

Updated all of the code to use the new type names, and moved the implementation of `traccc::spacepoint_formation` into a `.cpp` file.

The only not-completely-trivial update is that I removed the `traccc::measurement::invalid_value()` and `traccc::spacepoint::invalid_value()` functions. These were just ill-defined, and we never used them anywhere luckily.


beomki-yeo (approved):
Looks good to me!

PR #194, opened 2022-05-17
Fix a bug in thread block configuration in `seed_finding.cu`

beomki-yeo:
The values for block and thread were switched


krasznaa (approved):
:facepalm: Amazing that it even worked. Goes to show that we just don't use all that many blocks at the moment...

beomki-yeo:
Yeah it gave error only for 300 pileups.

PR #195, opened 2022-05-18
Rewrote SYCL algorithms to accept view objects

konradkusiak97:
In the current implementation, there are good amounts of memory copies between launching the algorithms (e.i. clusterization -> seed finding -> track param estimation). This is because at the end of each algorithmic execution, we copy the results back to the host and return a host object. 

In many cases, this is not really necessary and the data can stay on the device for the next algorithm to use. To do that, I changed the SYCL algorithms to receive view objects and return buffers. 

I also included in this PR some updates to the SYCL seedfinding that @beomki-yeo made some time ago in #160 to the CUDA version.

P. S. @krasznaa let me know if we want to split this PR in the end :wink:

### DIFF
--- a/device/sycl/src/clusterization/cluster_counting.sycl
+++ b/device/sycl/src/clusterization/cluster_counting.sycl
@@ -34,18 +34,22 @@ void cluster_counting(
                                        ::sycl::range<1>(wGroupSize)};
 
     // Set zero kernel
-    details::get_queue(queue).submit([&ndrange, &cluster_sizes_view](::sycl::handler& h){
-        h.parallel_for<class SetZeroClusterCounting>(ndrange, [=](::sycl::nd_item<1> item){
-            auto idx = item.get_global_linear_id();
+    details::get_queue(queue)
+        .submit([&ndrange, &cluster_sizes_view](::sycl::handler& h) {
+            h.parallel_for<class SetZeroClusterCounting>(
+                ndrange, [=](::sycl::nd_item<1> item) {
+                    auto idx = item.get_global_linear_id();
 
-            vecmem::device_vector<std::size_t> device_cluster_sizes(cluster_sizes_view);
+                    vecmem::device_vector<std::size_t> device_cluster_sizes(
+                        cluster_sizes_view);
 
-            if (idx >= device_cluster_sizes.size())
-                return;
-            
-            device_cluster_sizes[idx] = 0;
-        });
-    }).wait_and_throw();
+                    if (idx >= device_cluster_sizes.size())
+                        return;
+
+                    device_cluster_sizes[idx] = 0;
+                });
+        })
+        .wait_and_throw();
 
     // Cluster Counting kernel
     details::get_queue(queue)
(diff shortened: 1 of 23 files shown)


device/sycl/src/clusterization/cluster_counting.sycl
```diff
@@ -33,6 +33,25 @@ void cluster_counting(
     auto ndrange = ::sycl::nd_range<1>{::sycl::range<1>(num * wGroupSize),
                                        ::sycl::range<1>(wGroupSize)};
 
+    // Set zero kernel
+    details::get_queue(queue)
```
beomki-yeo, line 37: Can we use `copy.memset` here?

beomki-yeo: Maybe it's OK. We might be able to add it later when we implement CUDA twin

device/sycl/src/clusterization/cluster_counting.sycl
```diff
@@ -73,13 +92,13 @@ void cluster_counting(
                         device_cluster_prefix_sum[module_idx];
 
                     // Vector to fill in with the sizes of each cluster
-                    vecmem::device_vector<unsigned int> device_cluster_sizes(
+                    vecmem::device_vector<std::size_t> device_cluster_sizes(
```
krasznaa, line 95: Wait... does this really work? :confused:

While I prefer to use `std::size_t` for size-type variables as much as possible, `unsigned int` should rather be used for any device-side atomic operations. Since those operations definitely exist on all modern GPUs for `unsigned int`. But not necessarily for `unsigned long` (i.e. `std::size_t`).

device/sycl/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -139,11 +143,11 @@ host_spacepoint_container clusterization_algorithm::operator()(
         vecmem::get_data(measurements_prefix_sum), m_queue);
 
     // Copy the results back to the host
-    host_spacepoint_container spacepoints_per_event(&m_mr.get());
-    copy(spacepoints_buffer.headers, spacepoints_per_event.get_headers());
-    copy(spacepoints_buffer.items, spacepoints_per_event.get_items());
+    // host_spacepoint_container spacepoints_per_event(&m_mr.get());
+    // copy(spacepoints_buffer.headers, spacepoints_per_event.get_headers());
+    // copy(spacepoints_buffer.items, spacepoints_per_event.get_items());
```
krasznaa, line 148: You should just remove these lines completely.

krasznaa (approved):
As Beomki wrote, we'll need to switch to memsets from from the "zeroing kernels". But we can do that in a later step.

The `std::size_t` <-> `unsigned int` changes you should still have another look at, just to make sure that you would use the correct types everywhere. But other than that, this seems good to go in.

device/sycl/src/clusterization/clusterization_algorithm.cpp
konradkusiak97: thanks, I forgot about this bit

device/sycl/src/clusterization/cluster_counting.sycl
konradkusiak97: Alright, I changed it back to unsigned int

krasznaa (approved):
Let's go ahead with this...

PR #197, opened 2022-05-22
Make Clusterization Result Correct

beomki-yeo:
The clusterization result has been wrong which makes the seeding efficiency of `seq_example` drop to nearly zero. 
And I am glad that I finally fixed the problems (as far as I can detect)

There are ~~two~~ three bugs that I fixed:

#### Bug 1. Wrong binning setup for clusterization
The data files that I updated in `traccc-data` were generated from the digitization configuration file (`acts/Examples/Algorithms/Digitization/share/default-geometric-config-generic.json`), which sets different binning for each module. 

Meanwhile, the traccc `cell_module` always have the same binning scheme: 
```c++
// in cell.hpp
pixel_data pixel{-8.425, -36.025, 0.05, 0.05};
```
Thus, I took out binning information using `Acts::GeometryHierarchyMap` and apply to `pixel_data` in `csv.hpp`

#### Bug 2. Cell module was shifted during reading in `csv.hpp`
There was still a mismatch between header and item of `cell_container`. It was because the cell module information of the next cell group was recorded for the current cell group. It could be fixed by replacing `geometry_id` to `reference_id` in `read_cells` function.

```c++
// in csv.hpp
    while (creader.read(iocell)) {

        if (first_line_read and iocell.geometry_id != reference_id) {
            if (tfmap != nullptr) {
               // reference_id should be used instead of geometry_id
                if (tfmap->contains(iocell.geometry_id)) { 
                    module.placement = (*tfmap)[iocell.geometry_id];
                }
            }
            ...
```
#### Bug 3. Use measurement-simhit-map for correct indexing
See https://github.com/acts-project/traccc/pull/197#issuecomment-1136478051

### Physics performance after fix

I will compare two cases:

1) `seq_example` with ttbar<140> cell file (clusterization+seeding)

2) `seeding_example` with ttbar<140> hit file (seeding)

### DIFF
--- a/tests/common/tests/cca_test.hpp
+++ b/tests/common/tests/cca_test.hpp
@@ -107,7 +107,7 @@ class ConnectedComponentAnalysisTests
                 result.at(io_truth.geometry_id);
 
             traccc::scalar tol = std::max(
-                0.01, 0.0001 * std::max(io_truth.channel0, io_truth.channel1));
+                0.1, 0.0001 * std::max(io_truth.channel0, io_truth.channel1));
 
             auto match = std::find_if(
                 meas.begin(), meas.end(),
--- a/tests/cpu/test_clusterization_resolution.cpp
+++ b/tests/cpu/test_clusterization_resolution.cpp
@@ -1,0 +1,125 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+// Project include(s).
+#include "traccc/clusterization/clusterization_algorithm.hpp"
+#include "traccc/clusterization/spacepoint_formation.hpp"
+#include "traccc/io/reader.hpp"
+
+// VecMem include(s).
+#include <vecmem/memory/host_memory_resource.hpp>
+
+// GTest include(s).
+#include <gtest/gtest.h>
+
+class SurfaceBinningTests
+    : public ::testing::TestWithParam<
+          std::tuple<std::string, std::string, std::string, unsigned int>> {};
+
+// This defines the local frame test suite
+TEST_P(SurfaceBinningTests, Run) {
+
+    std::string detector_file = std::get<0>(GetParam());
+    std::string digi_config_file = std::get<1>(GetParam());
+    std::string data_dir = std::get<2>(GetParam());
+    unsigned int event = std::get<3>(GetParam());
+
+    // Read the surface transforms
+    auto surface_transforms = traccc::read_geometry(detector_file);
+
+    // Read the digitization configuration file
+    auto digi_cfg = traccc::read_digitization_config(digi_config_file);
+
+    // Memory resource used by the EDM.
+    vecmem::host_memory_resource host_mr;
+
+    // Algorithms
+    traccc::clusterization_algorithm ca(host_mr);
+    traccc::spacepoint_formation sf(host_mr);
+
+    // Read the cells from the relevant event file
+    traccc::cell_container_types::host cells_truth =
+        traccc::read_cells_from_event(event, data_dir, traccc::data_format::csv,
+                                      surface_transforms, digi_cfg, host_mr);
+
+    // Get Reconstructed Spacepoints
+    auto measurements_recon = ca(cells_truth);
+    auto spacepoints_recon = sf(measurements_recon);
+
+    // Read the hits from the relevant event file
+    traccc::spacepoint_container_types::host spacepoints_truth =
+        traccc::read_spacepoints_from_event(event, data_dir,
+                                            traccc::data_format::csv,
+                                            surface_transforms, host_mr);
+
+    // Check the size of spacepoints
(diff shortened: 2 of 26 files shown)


beomki-yeo:
One thing that I still don't understand is some events of `ttbar<200>` and `ttbar<300>` give me terrible efficiency. 
Precisely, 5th event of ttbar<200> and many of ttbat<300> have the problem
All ttbar<20 - 140> samples  look fine.

./bin/traccc_seq_example --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --cell_directory=tml_full/ttbar_mu200/ --hit_directory=tml_full/ttbar_mu200/  --particle_directory=tml_full/ttbar_mu200/  --events=1 --skip=4

beomki-yeo:
Um finally, I found the reason.
I was overlooking the existence of `measurement-simhit-map.csv`. In the past, I simply ignored it because I thought measurement id and hit id are always equal, which is wrong. That's why the seeding efficiency dropped after a random eta value
Now it is fixed.

10 ttbar<200> events:

beomki-yeo:
Anybody review please :p

tests/common/tests/cca_test.hpp
```diff
@@ -107,7 +107,7 @@ class ConnectedComponentAnalysisTests
                 result.at(io_truth.geometry_id);
 
             traccc::scalar tol = std::max(
-                0.01, 0.0001 * std::max(io_truth.channel0, io_truth.channel1));
+                0.1, 0.0001 * std::max(io_truth.channel0, io_truth.channel1));
```
krasznaa, line 110: 😕  Did the tolerance worsen as a result of all of this? I thought if anything, this would make the test more accurate...

tests/cpu/test_clusterization_resolution.cpp
```diff
@@ -0,0 +1,125 @@
...
+        for (std::size_t j = 0; j < sp_collection_recon.size(); j++) {
+            const auto& sp_recon = sp_collection_recon[j];
+            const auto& sp_truth = sp_collection_truth[j];
+
+            // Make sure that the difference in spacepoint position is less than
+            // 1%
+            EXPECT_TRUE(
+                traccc::getter::norm(sp_recon.global - sp_truth.global) /
+                    traccc::getter::norm(sp_recon.global) <
+                0.01);
```
krasznaa, line 88: 🤔  Interesting. Is a percentage figure really the most adequate here? In reality I would hope that our reconstruction accuracy is some absolute value on the surface of the detectors. And not 0 in one corner, and some larger value in the opposite corner. To me the most physically accurate test here would be to say that the reconstructed position has to be within X um from the true position.

Which I would then express like:

```c++
EXPECT_LT(traccc::getter::norm(sp_recon.global - sp_truth.global), 0.01);
```

Assuming that we want to go for 10 um for instance. Though admittedly at this point we should start using the units from Acts to express these magic numbers more clearly.

krasznaa (approved):
Apart from questions about how tolerances are handled in the tests, I think this looks good.

I was very pleased to see that the SYCL code was propagating this info everywhere correctly already. Kudos to @konradkusiak97. I was absolutely expecting that we would've forgotten about at least 1-2 places. 😛

Depending on how you feel about setting up the tests, you could just go ahead and merge the PR like this. :wink:

tests/cpu/test_clusterization_resolution.cpp
beomki-yeo: You are right but let's go with percentage for the moment. Once we add lorentz shift to clusterization algorithm, we can test with expected spatial resolution.

tests/common/tests/cca_test.hpp
beomki-yeo: Well. the test didn't pass the variance test after changes.
```
            EXPECT_NEAR(match->variance[0], io_truth.variance0, tol);
            EXPECT_NEAR(match->variance[1], io_truth.variance1, tol);
```
TBH I don't know how this test works and data was created :/ But it's not surprising at all as out variance estimation is not that robust

beomki-yeo:
@krasznaa Hmm I think it needs the approval again after force pushing

tests/common/tests/cca_test.hpp
beomki-yeo: I think I figured out the reason.
The default pitch size of  `pixel_data` is 1 which makes the reconstructed variance (`match->variance[0]` and `match->variance[1]` ) 1/12 or ~0.08.
And the `io_truth.variance0` (which is truth input from synthetic data) is 0.
That's why `0.01` tolerance doesn't work here.

Then someone will ask why it has been OK so far.
Before this PR, `pixel_data` of `cell_module` gets the default pitch size of 0.05:
```
pixel_data pixel{-8.425, -36.025, 0.05, 0.05};
```
This makes the reconstructed variance `0.05/12`, which is fine with 0.01 tolerance. 
This PR removes the default value of `pixel_data` and causes the unit test fails with 0.08 variance.

The conclusion is `cca_test` or synthetic data needs some modification :p...

krasznaa (approved):
Let's go ahead like this then.

PR #198, opened 2022-05-23
Comment cuda cca fast sv 2

ugGit:
This PR adds comments to the two fast_sv algorithms and the variables `adjc` and `adjv` are mentioned to be abbreviations for `adjecency count` respectively `adjecency vector`. 

Further, the variables `f` and `f_next` are introduced to be more coherent with the vocabulary of [the paper describing the Fast SV algorithm](https://epubs.siam.org/doi/pdf/10.1137/1.9781611976137.5). 
Concretely:
* in fast_sv_1: `f` (<-`gf`) and `f_next` (<-`f`)
* in fast_sv_2: `f` (<-`f`) and `f_next` (<-`gf`)

### DIFF
--- a/device/cuda/src/cca/component_connection.cu
+++ b/device/cuda/src/cca/component_connection.cu
@@ -130,17 +130,17 @@ __device__ void reduce_problem_cell(cell_container& cells, index_t tid,
 }
 
 /*
- * Implementation of a SV algorithm with the following steps:
- *   1) stochastic hooking to parent cell in new cluster
- *   2) aggressive hooking to grand parent cell in new cluster
- *   3) shortcutting to grand parent cell in same cluster
+ * Implementation of a FastSV algorithm with the following steps:
+ *   1) mix of stochastic and aggressive hooking
+ *   2) shortcutting
  *
  * The implementation corresponds to an adapted versiion of Algorithm 3 of
  * the following paper:
  * https://www.sciencedirect.com/science/article/pii/S0743731520302689
  *
  * f      = array holding the parent cell ID for the current iteration.
- * gf     = buffer array holding updated information for the next iteration.
+ * gf     = array holding grandparent cell ID from the previous iteration. 
+            This array only gets updated at the end of the iteration to prevent race conditions.
  */
 __device__ void fast_sv_1(index_t* f, index_t* gf, unsigned char adjc[],
                           index_t adjv[][8], unsigned int size) {
@@ -154,15 +154,15 @@ __device__ void fast_sv_1(index_t* f, index_t* gf, unsigned char adjc[],
     do {
         /*
          * Reset the end-parameter to false, so we can set it to true if we
-         * make a change to the f array.
+         * make a change to the gf array.
          */
         gf_changed = false;
 
         /*
          * The algorithm executes in a loop of three distinct parallel
-         * stages. In this first one, stochastic hooking, we examine
-         * adjacent cells and copy their parents cluster ID if it is 
-         * lower than our, essentially merging the two together.
+         * stages. In this first one, a mix of stochastic and aggressive hooking, we examine
+         * adjacent cells and copy their grand parents cluster ID if it is 
+         * lower than ours, essentially merging the two together.
          */
         for (index_t tst = 0, tid;
              (tid = tst * blockDim.x + threadIdx.x) < size; ++tst) {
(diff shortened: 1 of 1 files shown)


stephenswat:
Thank you for this follow-up PR, Uchendu. You are indeed correct that the implementation of `fast_sv_1` does not quite match the paper, and I am confused why. This was not intended. I do, however, vaguely remember coming across two versions of the same paper, and I was confused back then why there were discrepancies between them. The problem is, I can't find the other version of the paper anymore.

If you look at this paper: https://ir.cwi.nl/pub/31053/31053.pdf, you will notice that a lot of the naming of variables there is similar to what I have in the traccc implementation, so I suspect that I am not misremembering and that there was, at some point, a different version of the FastSV paper with this different implementation. I realise that I am starting to sound like a conspiracy nut here, but I really can't explain it any other way.

I would suggest we proceed as follows. We keep the implementation of `fast_sv_1`, because it seems to work. Then we add an implementation of Algorithm 1 (Shiloach-Vishkin). Finally, we should validate that all three implementations produce correct results through the use of the test suite, and we should see which one performs best.

I'm really sorry for confusing you and wasting your time by creating this issue (and by not keeping adequate documentation).

ugGit:
Alright. It feels like the pieces are finally coming together. I'll shortly summarize in chronological order the different papers:
1. [FastSV: A Distributed-Memory Connected Component Algorithm with Fast Convergence](https://epubs.siam.org/doi/epdf/10.1137/1.9781611976137.5): 
Proposing an optimized Shiloach-Vishkin (FastSV) algorithm to build connected components. They start off with a simplified version, and then present 4 improvements for faster convergence (hooking to grandparent, stochastic hooking, aggressive hooking, and early termination). The Algorithm 4 in the paper, embodying all improvements, corresponds to `fast_sv_2`. The paper includes a benchmark on the number of iterations needed to converge with each improvement added.
The algorithms use two arrays. The first array represents the current state child-parent relation (array `f[]`), and the second array (`f_next[]`) holds the updates of the current iteration to prevent read-write race conditions for comparisons with the current state.
Further, the paper suggests a GraphBLAS based solution, and a corresponding benchmark with a reference algorithm.
2. [Parallel algorithms for finding connected components using linear algebra](https://www.sciencedirect.com/science/article/am/pii/S0743731520302689): 
Proposes another variation of the FastSV algorithm, as well a summary of the Awerbuch–Shiloach (AS) algorithm. The algorithm `fast_sv_1` implemented currently in traccc corresponds to Algorithm 3 of this paper, and uses a mix of stochastic and aggressive hooking, in addition to shortcutting.
As proposed in the preceding paper, the FastSV algorithm presented here, uses two arrays, but in a different manner. The first array stores again the child-parent information (`f[]`). The second array (`gf[]`), however, stores information about the grandparents (somewhat similar to duplicating information, if ever only `f[]` is updated).
The rest of the paper presents implementations of the same algorithms using GraphBLAS. The benchmarks, AFAIU, are based on the later implementations using CombBLAS.
3. [LAGraph: Linear Algebra, Network Analysis Libraries, and the Study of Graph Algorithms](https://ir.cwi.nl/pub/31053/31053.pdf): 
Presenting a library to solve various graph-based problems, which includes the linear algebra based version of the FastSV algorithm.

Having sorted these things out, I finally reply to your comment:
> I would suggest we proceed as follows. We keep the implementation of fast_sv_1, because it seems to work. Then we add an implementation of Algorithm 1 (Shiloach-Vishkin). Finally, we should validate that all three implementations produce correct results through the use of the test suite, and we should see which one performs best.

This sounds reasonable. I could work on this issue during the next week. I think, the modification from `fast_sv_2` to the `simplified _sv`, Algorithm 1 (SV), should be fairly easy. Further, I will fix the comments.
Both already implemented algorithms pass all the tests. Thus, the work should be fairly quickly done.

ugGit:
@stephenswat, the PR should now be alright. It contains:
* commented implementations and corresponding references of the already implemented FastSV algorithms
* a new implementation of the simplified SV algorithm

All algorithms pass the CCA tests. 

So far, I haven't run the code through a profiler or done any measurements. I plan to do so throughout the next few weeks, and will then share the results with you.

stephenswat (approved):
Truly excellent PR, looks great to me. Thanks! :smile:

PR #201, opened 2022-05-30
Add choice of host/device mem resource to SYCL clusterization

konradkusiak97:
This PR is the beginning of the work to make SYCL algorithms run with either shared or host/device memory resources.

I introduced `traccc::memory_resource` struct, which contains a reference to `host_mr` and a pointer to `device_mr`. This setup assumes that the `host_mr` will be either shared memory resource or host memory resource. An object of this struct is then passed to algorithms.

For now, I implemented the `if statement` logic into SYCL Clusterization to decide how the algorithm should behave depending if the `device_mr` exists or not. This turned out not to be super messy for now (might be worse with seedfinding). Another approach, as we discussed with @krasznaa, would be to create separate versions of algorithms that behave according to the memory resources they will receive. But we'll see in the future what would be the best choice. 

Note that although the SYCL clusterization is able to work on shared or host&device memory resources, we can't use host and device out of the box in `seq_example_sycl`, because seedfinding code is not yet made to receive a buffer with data from device memory resource.

### DIFF
--- a/device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
+++ b/device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
@@ -19,6 +19,7 @@
 
 // VecMem include(s).
 #include <vecmem/memory/memory_resource.hpp>
+#include <vecmem/utils/copy.hpp>
 
 // traccc library include(s).
 #include "traccc/utils/memory_resource.hpp"
@@ -34,7 +35,8 @@ class clusterization_algorithm
     ///
     /// @param mr is a struct of memory resources (shared or host & device)
     /// @param queue is the sycl queue for kernel invocation
-    clusterization_algorithm(traccc::memory_resource& mr, queue_wrapper queue);
+    clusterization_algorithm(const traccc::memory_resource& mr,
+                             queue_wrapper queue);
 
     /// Callable operator for clusterization algorithm
     ///
@@ -48,6 +50,7 @@ class clusterization_algorithm
     private:
     traccc::memory_resource m_mr;
     mutable queue_wrapper m_queue;
+    std::unique_ptr<vecmem::copy> m_copy;
 };
 
 }  // namespace traccc::sycl
--- a/device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
+++ b/device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
@@ -8,7 +8,6 @@
 #pragma once
 
 // SYCL library include(s).
-#include "traccc/sycl/utils/memory_resource.hpp"
 #include "traccc/sycl/utils/queue_wrapper.hpp"
 
 // Project include(s).
@@ -21,6 +20,9 @@
 // VecMem include(s).
 #include <vecmem/memory/memory_resource.hpp>
 
+// traccc library include(s).
+#include "traccc/utils/memory_resource.hpp"
+
 namespace traccc::sycl {
 
 class clusterization_algorithm
@@ -30,10 +32,9 @@ class clusterization_algorithm
     public:
     /// Constructor for clusterization algorithm
     ///
-    /// @param mr is the memory resource
+    /// @param mr is a struct of memory resources (shared or host & device)
     /// @param queue is the sycl queue for kernel invocation
-    clusterization_algorithm(traccc::sycl::memory_resource& mr,
-                             queue_wrapper queue);
+    clusterization_algorithm(traccc::memory_resource& mr, queue_wrapper queue);
 
     /// Callable operator for clusterization algorithm
     ///
@@ -45,7 +46,7 @@ class clusterization_algorithm
         const cell_container_types::host& cells_per_event) const override;
 
     private:
-    std::reference_wrapper<traccc::sycl::memory_resource> m_mr;
+    traccc::memory_resource m_mr;
     mutable queue_wrapper m_queue;
 };
 
--- a/device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
+++ b/device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
@@ -8,6 +8,7 @@
 #pragma once
 
 // SYCL library include(s).
+#include "traccc/sycl/utils/memory_resource.hpp"
 #include "traccc/sycl/utils/queue_wrapper.hpp"
 
 // Project include(s).
@@ -31,8 +32,7 @@ class clusterization_algorithm
     ///
     /// @param mr is the memory resource
     /// @param queue is the sycl queue for kernel invocation
-    clusterization_algorithm(vecmem::memory_resource& mr,
-                             vecmem::memory_resource& device_mr,
+    clusterization_algorithm(traccc::sycl::memory_resource& mr,
                              queue_wrapper queue);
 
     /// Callable operator for clusterization algorithm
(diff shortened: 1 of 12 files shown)


konradkusiak97:
I also now got rid of atomics from clusters sum kernel. The necessary prefix sum for proper indexing is now being constructed on the host. It actually increased performance a little for higher pile up. It also will make my life easier for the future development (moving all kernels to common functions).

device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
```diff
@@ -29,9 +32,9 @@ class clusterization_algorithm
     public:
     /// Constructor for clusterization algorithm
     ///
-    /// @param mr is the memory resource
+    /// @param mr is a struct of memory resources (shared or host & device)
     /// @param queue is the sycl queue for kernel invocation
-    clusterization_algorithm(vecmem::memory_resource& mr, queue_wrapper queue);
+    clusterization_algorithm(traccc::memory_resource& mr, queue_wrapper queue);
```
krasznaa, line 37: This is where things become more complicated with const-correctness... :stuck_out_tongue:

We shouldn't pass the object like this. Because this formalism requires us to give an lvalue object to the algorithm. But in some situations it may be nice to be able to create the `traccc::memory_resource` object (using lvalue `vecmem::memory_resource` objects) "on the fly". Like:

```c++
vecmem::memory_resource& host_mr = ...;
vecmem::memory_resource& device_mr = ...;
sycl::queue& queue = ...;
...
traccc::clusterization_algorithm alg({host_mr, &device_mr}, &queue);
```

For the `vecmem::memory_resource` argument we actually wanted to require the user passing in an lvalue. :wink: But here we don't.

And now comes the weird const-correctness aspect of it all: You should just receive the `traccc::memory_resource` object as a const reference. After you initialise the `m_mr` member variable from it, it will not matter whether the argument to the constructor was const or not. (You could also just make the constructor receive `traccc::memory_resource` by value like we do for `traccc::sycl::queue_wrapper`. But a const reference is probably just ever slightly better to use.)

device/sycl/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -18,32 +18,31 @@
...
 clusterization_algorithm::output_type clusterization_algorithm::operator()(
-    const cell_container_types::host &cells_per_event) const {
+    const cell_container_types::host& cells_per_event) const {
 
     // Number of modules
     unsigned int num_modules = cells_per_event.size();
 
     // Vecmem copy object for moving the data between host and device
-    vecmem::copy copy;
+    vecmem::sycl::copy copy(m_queue.queue());
```
krasznaa, line 39: :thinking: When using `vecmem::sycl::shared_memory_resource` I believe we need to stick to `vecmem::copy`.

You should add an `std::unique_ptr<vecmem::copy>` member to the algorithm, and initialise it with either `vecmem::copy` or `vecmem::sycl::copy` based on the values in the `traccc::memory_resource` given to the algorithm constructor. (Remember that `vecmem::sycl::copy` inherits from `vecmem::copy`.)

device/sycl/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -55,93 +54,117 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
 
     // Helper container for sparse CCL calculations
     vecmem::data::jagged_vector_buffer<unsigned int> sparse_ccl_indices(
-        cell_sizes_plus, m_mr.get());
+        cell_sizes_plus, (m_mr.device ? *(m_mr.device) : m_mr.host),
+        (m_mr.device ? &(m_mr.host) : nullptr));
```
krasznaa, line 58: :thinking: Right... I forgot about this...

The logic in vecmem is just the opposite to how I made you set up `traccc::memory_resource`. Jagged buffers (amongst other things) assume that the "main" memory resource may or may not be host accessible. And if its not, then the optional `vecmem::memory_resource` pointer would provide the host-accessible resource.

Instead of implementing this reverse logic everywhere, I think we should rather make `traccc::memory_resource` follow vecmem's logic. I.e. implement it like:

```c++
namespace traccc {
    struct memory_resource {
        vecmem::memory_resource& main;
        vecmem::memory_resource* host = nullptr;
    };
}
```

This way we could feed those variables directly into the buffer constructors... :thinking:

krasznaa (changes requested):
Sorry for the delay...

device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
konradkusiak97: I see, makes sense :+1:

device/sycl/src/clusterization/clusterization_algorithm.cpp
konradkusiak97: I thought so too but I tested it on CUDA backend and it seemed to work, that's why I kept only `vecmem::sycl::copy`. But I also don't know why now it would work :stuck_out_tongue: so you're right, I will change make this distinction for `shared_mr`.

device/sycl/src/clusterization/clusterization_algorithm.cpp
konradkusiak97: done

device/sycl/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -55,93 +60,113 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
...
+
+    // Perform the exclusive scan operation. It will results with a vector of
+    // prefix sums, starting with 0, until the second to last sum. The total
+    // number of clusters in this case will be the last prefix sum + the
+    // clusters in the last idx
+    unsigned int total_clusters = clusters_per_module_host.back();
+    std::exclusive_scan(clusters_per_module_host.begin(),
+                        clusters_per_module_host.end(),
+                        clusters_per_module_host.begin(), 0);
+    total_clusters += clusters_per_module_host.back();
```
krasznaa, line 120: Wait... :confused: Do I understand correctly that you're doing the following in a much more complicated way here?

```c++
const std::size_t total_clusters = std::accumulate(clusters_per_module_host.begin(), clusters_per_module_host.end(), 0);
```

I had to look at the description of [std::exclusive_scan](https://en.cppreference.com/w/cpp/algorithm/exclusive_scan) for a bit to see what you're doing here...

krasznaa (changes requested):
I'm generally happy with the code, I'm only suspicious about that `std::exclusive_scan`... :thinking:

device/sycl/src/clusterization/clusterization_algorithm.cpp
konradkusiak97: Partially. I need the `total_clusters` but the main reason for this operation is to get a vector of prefix sums, starting from 0 to up to second-to-last prefix sum. So after `exclusive scan`, the `clusters_per_module_host` is this vector. It is then copied to the device, into `cl_per_module_prefix_buff`

stephenswat: He uses this prefix sum later as well, so in that case it's a pretty clever way to combine the prefix sum and accumulation operations.

konradkusiak97: Yes, exactly, I could use std accumulate alongside the exclusive scan but that's adding some more work which is avoided in this way.

krasznaa: Okay, let's go with this then.

krasznaa (approved):
Let's finally put this in.

PR #203, opened 2022-06-08
Add choice of host/device resource to SYCL seeding

konradkusiak97:
Following the work from #201, I added the choice of `host/device` memory resource to SYCL spacepoint binning, seeding and track parameter estimation. 

#### Notable changes 
- The modified algorithms have now a dual interface, one that accepts views and the other that accepts buffers. This was due to the fact that we can't access the `size_ptr` from a view that is created from a buffer, which is in the device memory.
The underlying functionality of the algorithm is implemented in a private function. This avoids code duplication. 

- Apart from that, the main change is that now we can make a choice of using `shared` or `host/device` memory resources in all SYCL algorithms.

This PR depends on my recent addition to acts-project/detray#262.


beomki-yeo:
I made a new release for detray update

beomki-yeo (approved):
Looks good to me

PR #204, opened 2022-06-09
Updated detray version

konradkusiak97:
Updated detray version to `v0.8.0`. This includes the following updates:
- Add host accessible memory resource to grid2 buffer
- Add Generic container for tuple of vector and tuple of array
- Add scalar template parameter for some classes


krasznaa (approved):
All good. The CI just does not want to hurry itself...

PR #207, opened 2022-06-16
Change options

beomki-yeo:
There are few changes in this PR:
1. Users need to pass `input_directory` rather than `cell_directory ` or `hit_directory`
2. The root performance file is not generated anymore unless `check_performace` is passed


examples/performance/include/traccc/efficiency/seeding_performance_writer.hpp
```diff
@@ -70,6 +61,16 @@ class seeding_performance_writer {
         }
     }
 
+    void init() {
+        auto path = m_cfg.file_path;
```
krasznaa, line 65: Is there any reason for making this copy? :thinking: Why not just refer to `m_cfg.file_path` in the 2 places in this function?

Even if you'd like to make the variable name shorter, please declare the type explicitly as

```c++
const std::string&
```

instead of using `auto`.

examples/performance/include/traccc/efficiency/seeding_performance_writer.hpp
```diff
@@ -70,6 +61,16 @@ class seeding_performance_writer {
...
+
+        m_output_file = TFile::Open(path.c_str(), m_cfg.file_mode.c_str());
```
krasznaa, line 67: I only now realised that the memory management of this file is far from great. :frowning:

Please declare the file variable like:

```c++
std::unique_ptr<TFile> m_output_file;
```

That will make sure that we wouldn't leak memory. Which we currently do. (Calling `m_output_file->Close()` is not the same as deleting the object from memory.)

It doesn't necessarily have to happen in this PR, but it might as well.

examples/performance/include/traccc/efficiency/seeding_performance_writer.hpp
```diff
@@ -70,6 +61,16 @@ class seeding_performance_writer {
         }
     }
 
+    void init() {
```
krasznaa, line 64: You added a bunch of `init()` calls in the clients, where they had none before.

Would it not be possible to make this `init()` function protected/private and make the object call it whenever necessary. (When the user calls a function that requires the object to be initialised, while it's not initialised yet.) Then the clients would only need to put their `add_cache(...)` calls behind `if` statements.

krasznaa (changes requested):
I'm generally in favour of this, modulo my comments.

Plus: Some of the tests at least were able to receive concrete files as input instead of an input directory. I'm not sure what will happen with those. (Though I'm just wondering. I'm not too attached to being able to run test applications with just a single file/event.)

beomki-yeo:
> Some of the tests at least were able to receive concrete files as input instead of an input directory. I'm not sure what will happen with those.

Not sure of what you meant with "receive concrete files" exactly...
If I understand your comment correctly, we had to put arguments for cell, particle, and hit directory separately. I thought there is no point in separating them, and this PR requires all input files to be in the same directory (as `input_directory`).

examples/performance/include/traccc/efficiency/seeding_performance_writer.hpp
beomki-yeo: I will change to `const std::string&`.

examples/performance/include/traccc/efficiency/seeding_performance_writer.hpp
beomki-yeo: Yeah thanks for the comments. I used the `unique_ptr` for `TFile` now

examples/performance/include/traccc/efficiency/seeding_performance_writer.hpp
beomki-yeo: Ok. I changed to call `init()` in `add_cache(...)` if the file is not initiailized.

PR #208, opened 2022-06-28
Single Precision Fixes, main branch (2022.06.28.)

krasznaa:
Eliminated some double-precision calculations when `scalar` is set to `float`.

In order to allow using these code pieces on hardware that does not support double-precision calculations, we need to be very explicit about the types that we would use in certain places. On my shiny new laptop, with its Intel Xe GPU, I was running into runtime errors like:

```
...
error: double type is not supported on this platform
in file: track_params_estimation_helper.hpp:95

in kernel: 'typeinfo name for traccc::sycl::TrackParamsEstimation'
error: backend compiler failed build.
 -11 (PI_ERROR_BUILD_PROGRAM_FAILURE)
Aborted (core dumped)
```

(The line number is not too meaningful in the example. It came from an "intermediate" state of the code while I was trying things out.)

With these changes I am able to run `traccc_seq_example_sycl` (once again) on hardware that dislikes double-precision calculations. (My laptop...)

### DIFF
--- a/core/include/traccc/seeding/track_params_estimation_helper.hpp
+++ b/core/include/traccc/seeding/track_params_estimation_helper.hpp
@@ -7,9 +7,13 @@
 
 #pragma once
 
+// Library include(s).
 #include "traccc/edm/seed.hpp"
 #include "traccc/edm/spacepoint.hpp"
 #include "traccc/edm/track_parameters.hpp"
+
+// System include(s).
+#include <cmath>
 
 namespace traccc {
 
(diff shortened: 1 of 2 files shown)


stephenswat:
Hmm, good idea, but wouldn't it be even better to template these functions on the scalar type they are to use?

krasznaa:
Yes, we should go to templating in one way or another... eventually... For now we have the `traccc::scalar` type.

https://github.com/acts-project/traccc/blob/main/plugins/algebra/array/include/traccc/plugins/algebra/array_definitions.hpp#L30

I **was** considering using that type more explicitly in the code instead of hardcoding `float` types for the number literals. But I thought it would not be worth the extra characters. But if you think it is, let me know. Then I'll need to introduce named variables for all constants. (I remember one of the compilers not liking `static_cast<float>(...)` for some reason. But I don't remember anymore why... :confused:)

stephenswat (approved):
Let's go with this for now then, definitely a step in the right direction.

PR #212, opened 2022-07-13
Added a common implementation for triplet counting

guilhermeAlmeida1:
This PR follows the work done for seed_finding by creating an implementation for triplet counting that can be used from both CUDA and SYCL, similarly to what was done with clusterization, doublets counting & doublets finding by @krasznaa & @konradkusiak97 .

Similar work to triplets finding & seed selecting to come afterwards.


device/common/include/traccc/seeding/device/impl/count_triplets.ipp
```diff
@@ -0,0 +1,163 @@
...
+
+    // if the number of triplets per mb is larger than 0, write the triplet
+    // counter into the container
+    if (num_triplets_per_mb > 0) {
+        vecmem::device_atomic_ref<uint32_t> obj(num_compat_mb_per_bin);
+        const int pos = obj.fetch_add(1);
+        vecmem::device_atomic_ref<uint32_t> objT(num_triplets);
+        objT.fetch_add(num_triplets_per_mb);
+        triplet_counter_at_bin[pos] = {mid_bot.sp1.sp_idx,
+                                        num_triplets_per_mb};
```
krasznaa, line 157: :thinking: It should be possible to do this a little more elegantly...

First off, you should not use `uint32_t` to refer to `unsigned int` variables. They are the same in practice, but let's use consistent type names throughout the code.

But the more interesting thing is how the triplet counter container is being filled. :thinking: We should try to do that with a `push_back(...)` function. You see, right now the `num_compat_mb_per_bin` variable is just there to help choose the correct bin in the `triplet_counter_at_bin` vector. But we could make that jagged vector "resizable", and just use the `vecmem::device_vector<T>::push_back(...)` function on it to add a new element into it.

But since this will require a bit more tweaks in the code, for now let's just replace `uint32_t` with `unsigned int`. That should be enough for this PR.

device/common/src/seeding/make_triplet_counter_buffer.cpp
```diff
@@ -0,0 +1,37 @@
...
+    
+    // Calculate the capacities for the buffer.
+    const device::triplet_counter_container_types::buffer::header_vector::
+        size_type buffer_size = mb_doublet_sizes.size();
+
+    // Create the buffer object.
+    device::triplet_counter_container_types::buffer buffer{
+        {buffer_size, mr},
+        {std::vector<std::size_t>(buffer_size, 0),
+         std::vector<std::size_t>(mb_doublet_sizes.begin(), mb_doublet_sizes.end()), mr,
```
krasznaa, line 25: :thinking: The triplet counter buffer **is** set up to be resizable. So we should double-check how its filling is working exactly after all...

device/sycl/src/seeding/seed_finding.sycl
```diff
@@ -172,17 +177,59 @@ vecmem::data::vector_buffer<seed> seed_finding::operator()(
         doublet_counts.begin(), doublet_counts.end(), mb_buffer_sizes.begin(),
         [](const device::doublet_counter_header& dc) { return dc.m_nMidBot; });
 
-    // create the triplet_counter container with the number of doublets
+    // Get the prefix sum for the midBot doublets using buffer.
+    const device::prefix_sum_t mb_prefix_sum = device::get_prefix_sum(
+        mb_buffer_sizes, (m_mr.host ? *(m_mr.host) : m_mr.main));
```
krasznaa, line 182: Instead of introducing a whole new overload for `traccc::device::get_prefix_sum`, let's use the existing function like:

```c++
const device::prefix_sum_t mb_prefix_sum = device::get_prefix_sum(
    std::vector<unsigned int>(mb_buffer_sizes.begin(), mb_buffer_sizes.end()),
    (m_mr.host ? *(m_mr.host) : m_mr.main));
```

Or something like that... I didn't try exactly what the formalism should be. The main point is that you can convert between different integer type vectors with a single expression.

device/sycl/src/seeding/triplet_counting.sycl
```diff
@@ -56,56 +56,57 @@ class TripletCount {
                                            mb_idx, item);
 
         // get internal spacepoints for current bin
-        auto internal_sp_per_bin = internal_sp_device.bin(bin_idx);
-        auto& num_compat_spM_per_bin =
+        vecmem::device_vector internal_sp_per_bin = internal_sp_device.bin(bin_idx); 
+        const unsigned int& num_compat_spM_per_bin =
```
krasznaa, line 60: This whole file should just be removed by this PR. Let's not make it nicer, just so that we could remove it right after...

krasznaa (changes requested):
A couple of general comments:
  - I fear that the filling of the triplet counter buffer may not be correct just yet. Since you didn't switch the downstream kernels to using the output of your new setup, we don't actually know if it's output exactly the right info. :frowning:
  - You should be a bit more liberal with `const`. :wink: If something can be declared `const`, it should be declared `const`.
    * This makes absolutely zero difference for the performance of the code that is generated. It's purely a code readability/maintainability question.

But to make it clear, I'm pretty pleased with how things look on this very first step. :smile:

device/common/CMakeLists.txt
```diff
@@ -39,6 +39,11 @@ traccc_add_library( traccc_device_common device_common TYPE SHARED
    "include/traccc/seeding/device/make_doublet_counter_buffer.hpp"
    "src/seeding/make_doublet_counter_buffer.cpp"
    "include/traccc/seeding/device/make_doublet_buffers.hpp"
-   "src/seeding/make_doublet_buffers.cpp" )
+   "src/seeding/make_doublet_buffers.cpp"
+   "include/traccc/seeding/device/count_triplets.hpp"
+   "include/traccc/seeding/device/impl/count_triplets.ipp"
+   "include/traccc/edm/device/triplet_counter.hpp"
```
krasznaa, line 45: Just to silence my OCD, please move the `triplet_counter.hpp` line up a bit, just after the `doublet_counter.hpp` line. :wink:

device/common/include/traccc/edm/device/triplet_counter.hpp
```diff
@@ -0,0 +1,60 @@
...
+    /// The total number of Triplets in a given geometric bin
+    unsigned int m_nTriplets = 0;
+
+    TRACCC_HOST_DEVICE
+    unsigned int get_ref_num() const { return m_nMidBot; }
+
+    void zeros() {
+        m_nMidBot = 0;
+        m_nTriplets = 0;
+    }
```
krasznaa, line 36: You don't need either of these functions on this class. In the "old" implementation @beomki-yeo was generalising some of the code with the help of these functions. But with the "new" implementation we don't need these anymore.

device/common/include/traccc/edm/device/triplet_counter.hpp
```diff
@@ -0,0 +1,60 @@
...
+};  // struct triplet_counter_header
+
+/// Item type for the "triplet counter container"
+///
+/// It stores the number of triplets for one specific Mid Bottom Doublet.
+///
+struct triplet_counter {
+
+    /// indices of two spacepoints of midbot doublet
+    doublet mid_bot_doublet;
```
krasznaa, line 47: As long as we're going for a "more ATLAS-like" naming, let's call this variable `m_minBotDoublet`.

device/common/include/traccc/seeding/device/impl/count_triplets.ipp
```diff
@@ -0,0 +1,164 @@
...
+    const device_doublet_container mid_top_doublet_device(mid_top_doublet_view);
+
+    // Create device copy of output parameter
+    triplet_counter_container_types::device triplet_counter(triplet_view);
+
+    // Get all spacepoints
+    const const_sp_grid_device internal_sp_device(sp_view);
+
+    // Get internal spacepoints for current bin
+    const unsigned int& num_compat_spM_per_bin =
```
krasznaa, line 62: Instead of using a reference here, let's just get this variable into a "const value". So that the value can be handled in a register, and the code would not think about loading it from global memory on every access. (I.e. just remove the `&` from the declaration.)

device/common/include/traccc/seeding/device/impl/count_triplets.ipp
```diff
@@ -0,0 +1,164 @@
...
+        doublet_counter_device.get_headers().at(bin_idx).m_nSpM;
+
+    // Header of doublet counter : number of compatible middle sp per bin
+    // Item of doublet counter : doublet counter objects per bin
+    const vecmem::device_vector<const doublet_counter> doublet_counter_per_bin =
```
krasznaa, line 67: Let's use

```c++
const doublet_counter_collection_types::const_device
```

here instead. (Hopefully I got the type correct... But the point is, there should be a typedef that we could use instead of using the VecMem type directly.)

device/common/include/traccc/seeding/device/impl/count_triplets.ipp
```diff
@@ -0,0 +1,164 @@
...
+    const unsigned int& spM_idx = mid_bot.sp1.sp_idx;
+    // middle spacepoint
+    const traccc::internal_spacepoint<traccc::spacepoint>& spM =
+        internal_sp_device.bin(spM_bin)[spM_idx];
+    // bottom spacepoint indexes
+    const unsigned int& spB_bin = mid_bot.sp2.bin_idx;
+    const unsigned int& spB_idx = mid_bot.sp2.sp_idx;
+    // bottom spacepoint
+    const traccc::internal_spacepoint<traccc::spacepoint>& spB =
+        internal_sp_device.bin(spB_bin)[spB_idx];
```
krasznaa, line 86: We should get all of these into local variables as well. Definitely at least for all of the `unsigned int` variables. (The internal spacepoint variable is maybe more debatable.)

device/common/include/traccc/seeding/device/impl/count_triplets.ipp
```diff
@@ -0,0 +1,164 @@
...
+    unsigned int num_triplets_per_mb = 0;
+
+    // iterate over mid-top doublets
+    for (unsigned int i = mt_start_idx; i < mt_end_idx; ++i) {
+        const traccc::doublet& mid_top_doublet = mid_top_doublets_per_bin[i];
+
+        const unsigned int& spT_bin = mid_top_doublet.sp2.bin_idx;
+        const unsigned int& spT_idx = mid_top_doublet.sp2.sp_idx;
+        const traccc::internal_spacepoint<traccc::spacepoint>& spT =
+            internal_sp_device.bin(spT_bin)[spT_idx];
```
krasznaa, line 135: Again, let's just lose the references.

device/common/include/traccc/seeding/device/impl/count_triplets.ipp
```diff
@@ -0,0 +1,164 @@
...
+    // Apply the conformal transformation to middle-bot doublet
+    traccc::lin_circle lb =
+        doublet_finding_helper::transform_coordinates(spM, spB, true);
+
+    // Calculate some physical quantities required for triplet compatibility
+    // check
+    scalar iSinTheta2 = 1 + lb.cotTheta() * lb.cotTheta();
+    scalar scatteringInRegion2 = config.maxScatteringAngle2 * iSinTheta2;
+    scatteringInRegion2 *= config.sigmaScattering * config.sigmaScattering;
+    scalar curvature, impact_parameter;
```
krasznaa, line 97: Maybe add a comment to these variables, why they are needed. It took me a bit to find why you didn't want to declare them closer to their usage.

It's good that you didn't declare them inside of the for-loop. But a line saying what these are (and why we don't care about their value) could be useful.

device/common/include/traccc/seeding/device/make_triplet_counter_buffer.hpp
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+#pragma once
+
+// Library include(s).
+#include "traccc/edm/device/triplet_counter.hpp"
+
+// VecMem include(s).
+#include <vecmem/memory/memory_resource.hpp>
+#include <vecmem/utils/copy.hpp>
```
krasznaa, line 15: At this point I'm **super** pedantic, but let's add an include for both `<vector>` and `<cstddef>` to this header. Just to make sure that the `std::vector<std::size_t>` type would be understood. Even if upstream headers change their own includes.

device/common/include/traccc/seeding/device/make_triplet_counter_buffer.hpp
```diff
@@ -0,0 +1,35 @@
...
+///
+/// @param mb_doublet_sizes vector of MidBot doublet sizes
+/// @param copy The object to use for accessing @c spacepoint_grid
+/// @param mr The memory resource for the buffer
+/// @param mr_host The (optional) host-accessible memory resource for the
+///                buffer, in case @c mr is not host-accessible
+/// @return A buffer usable by @c traccc::device::count_triplets
+///
+triplet_counter_container_types::buffer make_triplet_counter_buffer(
+    const std::vector<size_t>& mb_doublet_sizes, vecmem::copy& copy,
```
krasznaa, line 32: Here I'm even more pedantic, but let's use `std::size_t` instead of just `size_t`. It really doesn't make a practical difference, but the former is "more C\+\+".

krasznaa (changes requested):
As I said in person as well, I'm **very** pleased with the PR. These are really all just pedantic requests on my part...

device/common/include/traccc/edm/device/triplet_counter.hpp
guilhermeAlmeida1: I cannot yet remove the get_ref_num() function as this header is still being used in triplet_finding.sycl. When I make that part with the "new" implementation as well, I'll remove it

krasznaa: :+1:

krasznaa (approved):
Once the CI comes back successfully, let's merge it in!

PR #217, opened 2022-08-18
Cuda seed_finding uses common triplet counting

guilhermeAlmeida1:
Cuda seed_finding now makes use of the common implementation of triplet counting created in #212


krasznaa (approved):
After some soul-searching, let's just go ahead with this as-is. Even though the prefix sum for the triplet counting is surprisingly large, let's just use it for now.

If later on we decide that we should solve the issue without a prefix sum, or with a "sort of" prefix sum that requires some post-processing in the kernel (but is much smaller in memory), we'll do that in sync between CUDA and SYCL.

PR #218, opened 2022-08-18
Changed prefix sum element type and creation of prefix sum vector function

guilhermeAlmeida1:
@krasznaa 

Performance improvement on get_prefix_sum function.

Sample output in same conditions old / new:

**Old:**
wall time           120.212   
file reading (cpu)        63.9631   
clusterization_time (cpu) 0.65169   
spacepoint_formation_time (cpu) 0.129127  
clusterization and sp formation (cuda) 0.241384  
seeding_time (cpu)        2.79764   
seeding_time (cuda)       0.128792  
tr_par_esti_time (cpu)    0.311556  
tr_par_esti_time (cuda)   0.00514919

**New:**
wall time           119.376   
file reading (cpu)        63.3151   
clusterization_time (cpu) 0.65213   
spacepoint_formation_time (cpu) 0.129153  
clusterization and sp formation (cuda) 0.147325  
seeding_time (cpu)        2.80033   
seeding_time (cuda)       0.0913527 
tr_par_esti_time (cpu)    0.312316  
tr_par_esti_time (cuda)   0.00516295

### DIFF
--- a/device/common/src/get_prefix_sum.cpp
+++ b/device/common/src/get_prefix_sum.cpp
@@ -10,6 +10,7 @@
 
 // System include(s).
 #include <algorithm>
+#include <cassert>
 
 namespace traccc::device {
 
@@ -19,15 +20,15 @@ prefix_sum_t get_prefix_sum(
     vecmem::memory_resource& mr) {
 
     // Create the result object.
-    prefix_sum_t result(&mr);
     const std::size_t nelements = static_cast<std::size_t>(
         std::accumulate(sizes.begin(), sizes.end(), 0));
-    result.reserve(nelements);
+    prefix_sum_t result(nelements, &mr);
 
     // Fill the result object.
-    for (std::size_t i = 0; i < sizes.size(); ++i) {
-        for (std::size_t j = 0; j < sizes[i]; ++j) {
-            result.push_back({i, j});
+    for (std::size_t i = 0, k = 0; i < sizes.size(); ++i) {
+        for (std::size_t j = 0; j < sizes[i]; ++j, ++k) {
+            assert(k < result.size());
+            result[k] = {i, j};
         }
     }
 
--- a/core/include/traccc/edm/internal_spacepoint.hpp
+++ b/core/include/traccc/edm/internal_spacepoint.hpp
@@ -48,7 +48,7 @@ struct internal_spacepoint {
         const spacepoint_container_t& sp_container, const link_type& sp_link,
         const vector2& offsetXY)
         : m_link(std::move(sp_link)) {
-        const spacepoint_t& sp = sp_container.at({sp_idx.first, sp_idx.second});
+        const spacepoint_t& sp = sp_container.at({sp_link.first, sp_link.second});
         m_x = sp.global[0] - offsetXY[0];
         m_y = sp.global[1] - offsetXY[1];
         m_z = sp.global[2];
--- a/core/include/traccc/edm/internal_spacepoint.hpp
+++ b/core/include/traccc/edm/internal_spacepoint.hpp
@@ -48,7 +48,7 @@ struct internal_spacepoint {
         const spacepoint_container_t& sp_container, const link_type& sp_link,
         const vector2& offsetXY)
         : m_link(std::move(sp_link)) {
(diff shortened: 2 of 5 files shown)


device/common/src/get_prefix_sum.cpp
```diff
@@ -19,15 +19,14 @@ prefix_sum_t get_prefix_sum(
...
-    result.reserve(nelements);
+    prefix_sum_t result(nelements, &mr);
 
     // Fill the result object.
-    for (std::size_t i = 0; i < sizes.size(); ++i) {
-        for (std::size_t j = 0; j < sizes[i]; ++j) {
-            result.push_back({i, j});
+    for (std::size_t i = 0, k = 0; i < sizes.size(); ++i) {
+        for (std::size_t j = 0; j < sizes[i]; ++j, ++k) {
+            result[k] = {i, j};
```
krasznaa, line 29: Let's do something fancy. Let's add:

```c++
assert(k < nelements);
```

before this line. That way assuring that debug builds would test the assumption that we're making in the code. (You will also need to include `<cassert>` for this.)

krasznaa (commented):
As @stephenswat found, after I told him about this issue, we apparently stumbled upon a performance issue in the C\+\+ standard library that was found by others as well.
  - https://cplusplus.github.io/LWG/issue2969
  - https://cplusplus.github.io/LWG/issue2975
  - https://cplusplus.github.io/LWG/issue3525

And should be fixed in C\+\+20...

But that's really just here for me to justify that thinking that the previous implementation would be okay performance-wise was not such a silly idea. :stuck_out_tongue:

device/common/src/get_prefix_sum.cpp
krasznaa: You know what, let's make that instead:

```c++
assert(k < result.size());
```

Since the whole thing is not there in an optimised build, we should not even assume what the size of the `result` object is. :wink:

krasznaa (approved):
Looks good! Once the CI finishes, it should go in.

krasznaa:
:frowning: The CI is not in a good shape right now...

https://www.githubstatus.com/

Still, let's wait at least a little longer...

krasznaa:
GitHub's infrastructure seems to be working again.

But in the meantime I had an idea. Since the performance of this function still does not seem to be amazing, even after your changes, please try something even a bit more drastic... Try to replace the current definition of `traccc::device::prefix_sum_element_t` with something like:

```c++
struct prefix_sum_element {
    std::size_t first;
    std::size_t second;
};
typedef prefix_sum_element prefix_sum_element_t;
```

(I chose the `first` and `second` names so that this could replace the current definition without having to change any other parts of the code. At least I hope that's the case.)

With that we should be avoiding the above mentioned C\+\+17 issues.

Please give this a try, and see how this affects the performance of your test job.

guilhermeAlmeida1:
Changing the prefix sum element type from std::pair to a equivalent struct significantly improved performance of the get_prefix_sum function.

Sample output:
./bin/traccc_seq_example_cuda --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --input_directory=tml_full/ttbar_mu100/ --run_cpu=0 --events=10

**OLD:**
==> Elpased time ... 
wall time           33.4155   
file reading (cpu)        33.1363   
clusterization_time (cpu) 0         
spacepoint_formation_time (cpu) 0         
clusterization and sp formation (cuda) 0.147958  
seeding_time (cpu)        0         
seeding_time (cuda)       0.126959  
tr_par_esti_time (cpu)    0         
tr_par_esti_time (cuda)   0.00427735

**NEW:**
==> Elpased time ... 
wall time           33.0237   
file reading (cpu)        32.7963   
clusterization_time (cpu) 0         
spacepoint_formation_time (cpu) 0         
clusterization and sp formation (cuda) 0.121422  
seeding_time (cpu)        0         
seeding_time (cuda)       0.101692  
tr_par_esti_time (cpu)    0         
tr_par_esti_time (cuda)   0.00431525

krasznaa (approved):
The documentation of the new code is lacking a bit, and we'll definitely need to re-think the usage of `std::pair` in general in the code, but for now this should be good.

PR #222, opened 2022-08-30
Add a Futhark implementation of CCA

stephenswat:
Futhark is a purely functional programming language for writing high-performance code. It compiles to C, OpenMP, CUDA, Python, OpenCL, and more. I plan to write Futhark algorithms for more of the reconstruction chain, but this commit adds it for clusterization at least.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,3 +183,7 @@ include( CTest )
 if( BUILD_TESTING AND TRACCC_BUILD_TESTING )
    add_subdirectory( tests )
 endif()
+
+if(TRACCC_BUILD_FUTHARK)
+   add_subdirectory(device/futhark)
+endif()
--- a/device/futhark/CMakeLists.txt
+++ b/device/futhark/CMakeLists.txt
@@ -1,0 +1,140 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2021-2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# The C language is not enabled in our project by default, but we need it to
+# compile Futhark's code.
+enable_language(C)
+
+# Try to find our dependencies, which are of course Futhark itself (required)
+# and the CUDA toolkit for compiling CUDA projects. Also notify the user of
+# what we find.
+find_program(Futhark futhark REQUIRED)
+find_package(CUDAToolkit)
+
+message(STATUS "Found Futhark compiler at ${Futhark}")
+
+# This function describes how to turn a Futhark source into a CMake library.
+# The function creates a target called `TARGET_NAME` from a source file
+# `_SOURCE_NAME`. The Futhark target platform (CUDA or C) is given
+# (case-insensitively) by `_LANGUAGE_TARGET`
+function(add_futhark_library TARGET_NAME _LANGUAGE_TARGET _SOURCE_NAME)
+    # Set a few useful variables.
+    set(FUTHARK_LANGUAGES c cuda)
+    set(SOURCE_NAME ${CMAKE_CURRENT_SOURCE_DIR}/${_SOURCE_NAME})
+    string(TOLOWER ${_LANGUAGE_TARGET} LANGUAGE_TARGET)
+    string(TOUPPER ${_LANGUAGE_TARGET} LANGUAGE_TARGET_FANCY)
+
+    # Notify the user of what we are doing.
+    message(STATUS "Building ${LANGUAGE_TARGET_FANCY} Futhark target ${TARGET_NAME} from ${SOURCE_NAME}")
+
+    # Make sure that the source file actually exists.
+    if(NOT EXISTS ${SOURCE_NAME})
+        message(FATAL_ERROR "File ${SOURCE_NAME} does not exist!")
+    endif()
+
+    # Ensure that the requested language is actually a supported one, namely C
+    # or CUDA.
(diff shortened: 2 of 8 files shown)


CMakeLists.txt
```diff
@@ -183,3 +183,7 @@ include( CTest )
 if( BUILD_TESTING AND TRACCC_BUILD_TESTING )
    add_subdirectory( tests )
 endif()
+
+if(TRACCC_BUILD_FUTHARK)
```
krasznaa, line 187: Let's declare a proper cache variable with this. Like:

```cmake
option( TRACCC_BUILD_FUTHARK "Build the Futhark code included in traccc" FALSE )
```

You should put this around line 50, with the other similar "options".

device/futhark/CMakeLists.txt
```diff
@@ -0,0 +1,140 @@
...
+# compile Futhark's code.
+enable_language(C)
+
+# Try to find our dependencies, which are of course Futhark itself (required)
+# and the CUDA toolkit for compiling CUDA projects. Also notify the user of
+# what we find.
+find_program(Futhark futhark REQUIRED)
+find_package(CUDAToolkit)
+
+message(STATUS "Found Futhark compiler at ${Futhark}")
```
krasznaa, line 17: :thinking: This may be a bit too much effort for this, but here we go...

I would prefer to package up most of the code in this `CMakeLists.txt` file into a `FindFuthark.cmake` module. So that it could be used conveniently throughout this project for generating code. With the familiar

```cmake
find_package(Futhark REQUIRED)
```

device/futhark/CMakeLists.txt
```diff
@@ -0,0 +1,140 @@
...
+
+# This function describes how to turn a Futhark source into a CMake library.
+# The function creates a target called `TARGET_NAME` from a source file
+# `_SOURCE_NAME`. The Futhark target platform (CUDA or C) is given
+# (case-insensitively) by `_LANGUAGE_TARGET`
+function(add_futhark_library TARGET_NAME _LANGUAGE_TARGET _SOURCE_NAME)
```
krasznaa, line 23: I don't like this interface. :frowning: I don't like that we're forced into building a separate **shared** library out of every Futhark file.

I can be on board with having a dedicated "Futhark library" creating function, which can have regular C\+\+ code as well along with the specific "Futhark code".

But I would much rather have a function like `add_futhark_sources_to_target(...)` that would add the generation of sources for a specific, already existing target.

Would that be possible to do? Of course the goal would be to allow to add multiple "Futhark sources", in multiple steps for this. Like:

```cmake
traccc_add_library( Foo ... )
add_futhark_source_to_target( Foo bar1.fut ... )
add_futhark_source_to_target( Foo bar2.fut ... )
```

device/futhark/include/traccc/futhark/cca.hpp
```diff
@@ -0,0 +1,31 @@
...
+                                  const cell_container_types::host& cells)> {
+    component_connection();
+
+    ~component_connection();
+
+    output_type operator()(const cell_container_types::host& cells) const;
+
+    struct futhark_context_config* cfg;
+    struct futhark_context* ctx;
+};
```
krasznaa, line 29: Some (Doxygen) comments would go a very long way...

device/futhark/src/cca.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+    for (std::size_t i = 0; i < data.size(); ++i) {
+        total_cells += data.at(i).items.size();
+    }
+
+    uint64_t* host_event = new uint64_t[total_cells];
+    uint64_t* host_geometry = new uint64_t[total_cells];
+    int64_t* host_channel0 = new int64_t[total_cells];
+    int64_t* host_channel1 = new int64_t[total_cells];
+    float* host_activation = new float[total_cells];
```
krasznaa, line 36: I would love to see some smart pointers being used here.

device/futhark/src/cca.cpp
```diff
@@ -0,0 +1,149 @@
...
+    futhark_free_f32_1d(ctx, out_position0);
+    futhark_free_f32_1d(ctx, out_position1);
+    futhark_free_f32_1d(ctx, out_variance0);
+    futhark_free_f32_1d(ctx, out_variance1);
+
+    output_type out;
+    vecmem::host_memory_resource mem;
+
+    for (std::size_t i = 0; i < data.size(); ++i) {
+        vecmem::vector<measurement> v(&mem);
```
krasznaa, line 124: You should reserve an appropriate size for the vector. I think over-reserving is still a bit better performance-wise than just letting `std::vector` reserve the memory in multiple iterations.

device/futhark/src/cca.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+        for (std::size_t j = 0; j < total_measurements; ++j) {
+            if (host_out_geometry[j] == data.at(i).header.module) {
+                measurement m;
+
+                m.local = {host_out_position0[j], host_out_position1[j]};
+                m.variance = {host_out_variance0[j], host_out_variance1[j]};
+
+                v.push_back(m);
```
krasznaa, line 133: It's a super stylistic question, I agree, but I would've written this as:

```c++
v.emplace_back({host_out_position0[j], host_out_position1[j]},
               {host_out_variance0[j], host_out_variance1[j]});l
```

(Or whatever the exact formalism is... I didn't check...)

krasznaa (changes requested):
I'm pretty curious to see how this language does. So despite the comments, I'm pretty positive about all of this. :wink:

PR #220, opened 2022-08-26
Cuda seed finding uses common triplet finding and weight updating

guilhermeAlmeida1:
This PR depends on #219 , ~~yet to be merged~~

Cuda seed_finding now makes use of the common implementation of triplet finding and weight updated created in #219 

No significant performance changes, which was expected due to the poor performance of the get_prefix_sum function already observed in #217 #218.

Sample performance output:
./bin/traccc_seq_example_cuda --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --input_directory=tml_full/ttbar_mu100/ --run_cpu=0 --events=10

**NEW:**
wall time           33.975    
file reading (cpu)        33.7376   
clusterization_time (cpu) 0         
spacepoint_formation_time (cpu) 0         
clusterization and sp formation (cuda) 0.125502  
seeding_time (cpu)        0         
seeding_time (cuda)       0.107332  
tr_par_esti_time (cpu)    0         
tr_par_esti_time (cuda)   0.00454855

**OLD:**
wall time           35.4842   
file reading (cpu)        35.2446   
clusterization_time (cpu) 0         
spacepoint_formation_time (cpu) 0         
clusterization and sp formation (cuda) 0.130167  
seeding_time (cpu)        0         
seeding_time (cuda)       0.105077  
tr_par_esti_time (cpu)    0         
tr_par_esti_time (cuda)   0.00432057

PR #219, opened 2022-08-24
Added common implementation of triplet finding and weight updating

guilhermeAlmeida1:
Following #212 , a common implementation was now created for the 2 next steps in seed finding.

No significant difference in performance was observed for now, which was expected due to the poor performance of the get_prefix_sum function already observed in #217 #218 .

Sample performance output:
./bin/traccc_seq_example_sycl --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --input_directory=tml_full/ttbar_mu100/ --run_cpu=0 --events=10

**NEW:**
==> Elpased time ... 
wall time           34.8908   
file reading (cpu)        30.805    
clusterization_time (cpu) 0         
spacepoint_formation_time (cpu) 0         
clusterization_time (sycl) 0.566461  
seeding_time (cpu)        0         
seeding_time (sycl)       3.31363   
tr_par_esti_time (cpu)    0         
tr_par_esti_time (sycl)   0.205617

**OLD:**
wall time           34.6785   
file reading (cpu)        30.5384   
clusterization_time (cpu) 0         
spacepoint_formation_time (cpu) 0         
clusterization_time (sycl) 0.570876  
seeding_time (cpu)        0         
seeding_time (sycl)       3.36168   
tr_par_esti_time (cpu)    0         
tr_par_esti_time (sycl)   0.207476

PR #223, opened 2022-09-01
Changed get_prefix_sum to device kernel

guilhermeAlmeida1:
This PR depends on #219 #220 #221 , ~~yet to be merged~~

Following the poor performance observed in the get_prefix_sum function (see #217 #218), this function was changed to run in parallel on the device rather than the cpu.

### DIFF
--- a/device/common/include/traccc/device/fill_prefix_sum.hpp
+++ b/device/common/include/traccc/device/fill_prefix_sum.hpp
@@ -1,0 +1,43 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/definitions/qualifiers.hpp"
+#include "traccc/device/get_prefix_sum.hpp"
+#include "traccc/edm/container.hpp"
+
+// VecMem include(s).
+#include <vecmem/memory/memory_resource.hpp>
+#include <vecmem/utils/copy.hpp>
+
+namespace traccc::device {
+
+using prefix_sum_size_t = vecmem::data::vector_view<int>::size_type;
+
+/// Function filling the prefix sum for a "size vector"
+///
+/// This simple function is just meant to translate the received "size vector"
+/// into a "prefix sum". I.e. into a list of index pairs that would allow
+/// visiting all elements of the jagged vector described by this "size vector".
+///
+/// @param[in] globalIndex The index of the current thread
+/// @param[in] sizes_view View on the sizes of the "inner vectors" of a jagged
+/// vector
+/// @param[out] ps_view   View on the result vector of index pairs
+///
+TRACCC_HOST_DEVICE
+void fill_prefix_sum(
+    std::size_t globalIndex,
+    const vecmem::data::vector_view<const prefix_sum_size_t>& sizes_view,
+    vecmem::data::vector_view<prefix_sum_element_t> ps_view);
+
+}  // namespace traccc::device
+
+// Include the implementation.
+#include "traccc/device/impl/fill_prefix_sum.ipp"
--- a/device/common/include/traccc/device/make_prefix_sum_buffer.hpp
+++ b/device/common/include/traccc/device/make_prefix_sum_buffer.hpp
@@ -1,0 +1,57 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/definitions/qualifiers.hpp"
+#include "traccc/device/fill_prefix_sum.hpp"
+#include "traccc/device/get_prefix_sum.hpp"
+#include "traccc/utils/memory_resource.hpp"
+
+// VecMem include(s).
+#include <vecmem/utils/copy.hpp>
+
+// System include(s).
+#include <variant>
+
+namespace traccc::device {
+
+/** @struct make_result_t
+ * @brief Helper class for return type of make_sum_buffer function
+ * @param view View on the result vector
+ * @param variant Object posessing memory of the result vector
+ * @param totalSize Last member of result vector
+ */
+struct prefix_sum_buffer {
+    vecmem::data::vector_view<prefix_sum_size_t> view;
+    std::variant<vecmem::vector<prefix_sum_size_t>,
+                 vecmem::data::vector_buffer<prefix_sum_size_t>>
+        variant;
+    prefix_sum_size_t totalSize;
+};  // struct prefix_sum_buffer
+
+using prefix_sum_buffer_t = prefix_sum_buffer;
+
+/// Function providing the prefix sum for a "size vector"
+///
+/// This simple function is just meant to translate the received "size vector"
+/// into its prefix sum. I.e. Each element of the output vector equals the
+/// summation of all the elements up to that point of the input vector
+///
+/// @param sizes The sizes of the "inner vectors" of a jagged vector
+/// @param copy A "copy object" capable of dealing with the view
+/// @param mr    The memory resource to use of the result
+/// @return A class containing the prefix sum view, a memory posessing object on
+/// the view elements, and the total summation of the size vector
+///
+TRACCC_HOST
+prefix_sum_buffer_t make_prefix_sum_buffer(
+    const std::vector<prefix_sum_size_t>& sizes, vecmem::copy& copy,
+    const traccc::memory_resource& mr);
+
+}  // namespace traccc::device
--- a/device/cuda/src/seeding/seed_finding.cu
+++ b/device/cuda/src/seeding/seed_finding.cu
@@ -10,7 +10,9 @@
 #include "traccc/cuda/utils/definitions.hpp"
 
 // Project include(s).
+#include "traccc/device/fill_prefix_sum.hpp"
 #include "traccc/device/get_prefix_sum.hpp"
+#include "traccc/device/make_prefix_sum_buffer.hpp"
 #include "traccc/edm/device/doublet_counter.hpp"
 #include "traccc/seeding/device/count_doublets.hpp"
 #include "traccc/seeding/device/count_triplets.hpp"
@@ -32,6 +34,15 @@
 
 namespace traccc::cuda {
 namespace kernels {
+
+/// CUDA kernel for running @c traccc::device::fill_prefix_sum
+__global__ void fill_prefix_sum(
+    vecmem::data::vector_view<const device::prefix_sum_size_t> sizes_view,
+    vecmem::data::vector_view<device::prefix_sum_element_t> ps_view) {
+
+    device::fill_prefix_sum(threadIdx.x + blockIdx.x * blockDim.x, sizes_view,
+                            ps_view);
+}
 
 /// CUDA kernel for running @c traccc::device::count_doublets
 __global__ void count_doublets(
@@ -127,6 +138,30 @@ __global__ void select_seeds(
 
 }  // namespace kernels
 
+vecmem::data::vector_buffer<device::prefix_sum_element_t> make_prefix_sum_buff(
+    const std::vector<device::prefix_sum_size_t>& sizes, vecmem::copy& copy,
+    const traccc::memory_resource& mr) {
+
+    const device::prefix_sum_buffer_t make_sum_result =
+        device::make_prefix_sum_buffer(sizes, copy, mr);
+    const vecmem::data::vector_view<const device::prefix_sum_size_t>
+        sizes_sum_view = make_sum_result.view;
+    const unsigned int totalSize = make_sum_result.totalSize;
+
+    // Create buffer and view objects for prefix sum vector
+    vecmem::data::vector_buffer<device::prefix_sum_element_t> prefix_sum_buff(
+        totalSize, mr.main);
+    copy.setup(prefix_sum_buff);
+
+    // Fill the prefix sum vector
+    kernels::fill_prefix_sum<<<(sizes_sum_view.size() / 32) + 1, 32>>>(
+        sizes_sum_view, prefix_sum_buff);
+    CUDA_ERROR_CHECK(cudaGetLastError());
+    CUDA_ERROR_CHECK(cudaDeviceSynchronize());
+
+    return prefix_sum_buff;
+}
+
 seed_finding::seed_finding(const seedfinder_config& config,
                            const traccc::memory_resource& mr)
     : m_seedfinder_config(config), m_mr(mr) {
@@ -162,32 +197,24 @@ vecmem::data::vector_buffer<seed> seed_finding::operator()(
     const sp_grid_const_view& g2_view,
     const std::vector<unsigned int>& grid_sizes) const {
 
-    // Get the prefix sum for the spacepoint grid using buffer.
-    const device::prefix_sum_t sp_grid_prefix_sum = device::get_prefix_sum(
-        grid_sizes, (m_mr.host ? *(m_mr.host) : m_mr.main));
+    // Create prefix sum buffer and its view
+    vecmem::data::vector_buffer sp_grid_prefix_sum_buff =
+        make_prefix_sum_buff(grid_sizes, *m_copy, m_mr);
 
     // Set up the doublet counter buffer.
     device::doublet_counter_container_types::buffer doublet_counter_buffer =
         device::make_doublet_counter_buffer(grid_sizes, *m_copy, m_mr.main,
                                             m_mr.host);
 
-    // Set up the buffer of the prefix sum and its view
-    vecmem::data::vector_buffer<device::prefix_sum_element_t>
-        sp_grid_prefix_sum_buff(sp_grid_prefix_sum.size(), m_mr.main);
-    m_copy->setup(sp_grid_prefix_sum_buff);
-    (*m_copy)(vecmem::get_data(sp_grid_prefix_sum), sp_grid_prefix_sum_buff);
-    vecmem::data::vector_view<device::prefix_sum_element_t>
-        sp_grid_prefix_sum_view = sp_grid_prefix_sum_buff;
-
     // Calculate the number of threads and thread blocks to run the doublet
     // counting kernel for.
     const unsigned int nDoubletCountThreads = WARP_SIZE * 2;
     const unsigned int nDoubletCountBlocks =
-        sp_grid_prefix_sum.size() / nDoubletCountThreads + 1;
+        sp_grid_prefix_sum_buff.size() / nDoubletCountThreads + 1;
 
     // Count the number of doublets that we need to produce.
     kernels::count_doublets<<<nDoubletCountBlocks, nDoubletCountThreads>>>(
-        m_seedfinder_config, g2_view, sp_grid_prefix_sum_view,
+        m_seedfinder_config, g2_view, sp_grid_prefix_sum_buff,
         doublet_counter_buffer);
     CUDA_ERROR_CHECK(cudaGetLastError());
     CUDA_ERROR_CHECK(cudaDeviceSynchronize());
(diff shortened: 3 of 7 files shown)


guilhermeAlmeida1:
I did some performance tests on using shared memory both before and after the changes done in this PR.

performance tests

I observed a very large increase in kernel execution times if using shared memory (x2.14) yet no significant difference in times between kernels, which one could've expected to decrease. 
Before the changes of get prefix sum to device, I observed an even worse result for shared memory usage. Not only were kernel execution times worse (x2.71) but time between kernels was also now slower for shared memory (1.15x). 
From comparing times before and after implementing get prefix sum on device, using separate memory resources, one sees an improvement of 70% on the total execution time of seed finding.

device/common/include/traccc/device/fill_prefix_sum.hpp
```diff
@@ -0,0 +1,39 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2022 CERN for the benefit of the ACTS project
```
krasznaa, line 3: I was not paying attention to this so far. But on any new file you should just use the current year. We try to show in these file headers when the code was actually developed.

Of course let's not worry about this for any of the other files. But here we may as well set it just to `2022`.

device/common/include/traccc/device/fill_prefix_sum.hpp
```diff
@@ -0,0 +1,39 @@
...
+/// into a "prefix sum". I.e. into a list of index pairs that would allow
+/// visiting all elements of the jagged vector described by this "size vector".
+///
+/// @param[in] globalIndex The index of the current thread
+/// @param[in] sizes_view View on the sizes of the "inner vectors" of a jagged
+/// vector
+/// @param[out] ps_view   View on the result vector of index pairs
+///
+TRACCC_HOST_DEVICE
+void fill_prefix_sum(unsigned int globalIndex,
```
krasznaa, line 32: Let's use `std::size_t` for the global index, like we do for the other functions.

device/common/include/traccc/device/fill_prefix_sum.hpp
```diff
@@ -0,0 +1,39 @@
...
+                     const vecmem::data::vector_view<unsigned int>& sizes_view,
```
krasznaa, line 33: This will be fairly anal, but... Let's not hardcode `unsigned int` here. Let's rather write:

```c++
const vecmem::data::vector_view<vecmem::device_vector<int>::size_type>& sizes_view
```

instead. So that if at any point we may change how `vecmem::device_vector` represents its size, we wouldn't need to change the code here.

To help client code a bit, you could introduce a shorter typedef here. Like:

```c++
namespace traccc::device {
    using prefix_sum_size_t = vecmem::device_vector<int>::size_type;
}
```

:thinking:

Also note that since this is an input variable, the type should actually be `const`. Like:

```c++
const vecmem::data::vector_view<const prefix_sum_size_t>& sizes_view
```

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -33,6 +34,15 @@
 namespace traccc::cuda {
 namespace kernels {
 
+/// CUDA kernel for running @c traccc::device::fill_prefix_sum
+__global__ void fill_prefix_sum(
+    vecmem::data::vector_view<unsigned int> sizes_view,
```
krasznaa, line 39: You could switch to the proposed `traccc::device::prefix_sum_size_t` in all appropriate places in this file and in `seed_finding.sycl`. (Keeping "const correctness" in mind.)

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -162,28 +172,42 @@ vecmem::data::vector_buffer<seed> seed_finding::operator()(
     const sp_grid_const_view& g2_view,
     const std::vector<unsigned int>& grid_sizes) const {
 
-    // Get the prefix sum for the spacepoint grid using buffer.
-    const device::prefix_sum_t sp_grid_prefix_sum = device::get_prefix_sum(
-        grid_sizes, (m_mr.host ? *(m_mr.host) : m_mr.main));
+    // Create prefix sum of grid_sizes vector
+    std::vector<unsigned int> grid_sizes_sum(grid_sizes.size());
```
krasznaa, line 176: `std::vector` works here, but I would still rather use `vecmem::vector`. That way, when we use a single (shared/managed) memory resource, we could skip the explicit copy later on.

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -162,28 +172,42 @@ vecmem::data::vector_buffer<seed> seed_finding::operator()(
...
+    std::partial_sum(grid_sizes.begin(), grid_sizes.end(),
+                     grid_sizes_sum.begin(), std::plus<unsigned int>());
+    const unsigned int& totalGrid = grid_sizes_sum[grid_sizes_sum.size() - 1];
```
krasznaa, line 179: Let's just make it a `const unsigned int`. For primitive types there's no upside to using references. (A reference is 64 bits, and needs to be followed through once indirection in memory. While `unsigned int` is just 32 bits, and is "right there".)

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -162,28 +172,42 @@ vecmem::data::vector_buffer<seed> seed_finding::operator()(
...
-    // Set up the doublet counter buffer.
-    device::doublet_counter_container_types::buffer doublet_counter_buffer =
-        device::make_doublet_counter_buffer(grid_sizes, *m_copy, m_mr.main,
-                                            m_mr.host);
+    // Create buffer and view objects for summation of sizes
+    vecmem::data::vector_buffer<unsigned int> grid_sizes_buff(
+        grid_sizes_sum.size(), m_mr.main);
+    m_copy->setup(grid_sizes_buff);
+    (*m_copy)(vecmem::get_data(grid_sizes_sum), grid_sizes_buff);
+    vecmem::data::vector_view<unsigned int> grid_sizes_view(grid_sizes_buff);
```
krasznaa, line 186: This whole copy should be dependent on whether we use 2 memory resources, or just one. (I still fear that we don't do this correctly in many parts of the code.)

If we use just one memory resource, then the kernel should receive `vecmem::get_data(grid_sizes_sum)` directly.

krasznaa (changes requested):
There is a bit too much copy-pasting in the code for my liking. :frowning: Could you wrap up the creation/filling of the prefix sums into a helper function? :thinking:

device/common/include/traccc/device/make_sum_buffer.hpp
```diff
@@ -0,0 +1,59 @@
...
+/// visiting all elements of the jagged vector described by this "size vector".
+///
+/// @param sizes The sizes of the "inner vectors" of a jagged vector
+/// @param copy A "copy object" capable of dealing with the view
+/// @param mr    The memory resource to use of the result
+/// @return The prefix sum for efficiently iterating over all element of a
+///         jagged vector
+///
+TRACCC_HOST
+make_buffer_result_t<prefix_sum_size_t> make_sum_buffer(
```
krasznaa, line 51: Let's call it `make_prefix_sum_buffer(...)` instead. You used that name in the CUDA and SYCL codes already. But the `traccc::device` namespace will separate this function from the ones in the `traccc::cuda` and `traccc::sycl` namespaces.

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -127,6 +138,33 @@ __global__ void select_seeds(
 
 }  // namespace kernels
 
+vecmem::data::vector_buffer<device::prefix_sum_element_t> make_prefix_sum_buff(
+    const std::vector<device::prefix_sum_size_t> sizes,
```
krasznaa, line 142: Use a const reference here as well.

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -127,6 +138,33 @@ __global__ void select_seeds(
...
+    const std::unique_ptr<vecmem::copy>& copy,
```
krasznaa, line 143: Pass the object as a non-const reference here as well.

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -127,6 +138,33 @@ __global__ void select_seeds(
...
+    const vecmem::data::vector_view<const device::prefix_sum_size_t>
+        sizes_sum_view = make_sum_result.view;
+    const unsigned int totalSize = make_sum_result.totalSize;
+
+    // Create buffer and view objects for prefix sum vector
+    vecmem::data::vector_buffer<device::prefix_sum_element_t> prefix_sum_buff(
+        totalSize, mr.main);
+    copy->setup(prefix_sum_buff);
+    vecmem::data::vector_view<device::prefix_sum_element_t> prefix_sum_view =
+        prefix_sum_buff;
```
krasznaa, line 157: In CUDA code you can skip creating this variable and just pass `prefix_sum_buff` directly to the kernel function. It gets cast into a view automatically. (Since the kernel function explicitly declares that it needs a view, and the buffer can be cast into a view.)

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -162,28 +200,22 @@ vecmem::data::vector_buffer<seed> seed_finding::operator()(
...
     const std::vector<unsigned int>& grid_sizes) const {
 
-    // Get the prefix sum for the spacepoint grid using buffer.
-    const device::prefix_sum_t sp_grid_prefix_sum = device::get_prefix_sum(
-        grid_sizes, (m_mr.host ? *(m_mr.host) : m_mr.main));
+    // Create prefix sum buffer and its view
+    vecmem::data::vector_buffer sp_grid_prefix_sum_buff =
+        make_prefix_sum_buff(grid_sizes, m_copy, m_mr);
+    vecmem::data::vector_view<device::prefix_sum_element_t>
+        sp_grid_prefix_sum_view = sp_grid_prefix_sum_buff;
```
krasznaa, line 207: Same here. You can just use the buffer directly in CUDA code, no need to explicitly cast it to a view.

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -201,24 +233,16 @@ vecmem::data::vector_buffer<seed> seed_finding::operator()(
...
-    vecmem::data::vector_buffer<device::prefix_sum_element_t>
-        doublet_prefix_sum_buff(doublet_prefix_sum.size(), m_mr.main);
-    m_copy->setup(doublet_prefix_sum_buff);
-    (*m_copy)(vecmem::get_data(doublet_prefix_sum), doublet_prefix_sum_buff);
-    vecmem::data::vector_view<device::prefix_sum_element_t>
-        doublet_prefix_sum_view = doublet_prefix_sum_buff;
+    // Create prefix sum buffer and its view
+    vecmem::data::vector_buffer doublet_prefix_sum_buff = make_prefix_sum_buff(
+        m_copy->get_sizes(doublet_counter_buffer.items), m_copy, m_mr);
+    vecmem::data::vector_view doublet_prefix_sum_view = doublet_prefix_sum_buff;
```
krasznaa, line 239: No need for the explicit view here either. Or in any of the next occurrences. But I'll stop pointing these out from here on. :stuck_out_tongue:

device/sycl/src/seeding/seed_finding.sycl
```diff
@@ -5,6 +5,9 @@
  * Mozilla Public License Version 2.0
  */
 
+// System include(s).
+#include <algorithm>
+
```
krasznaa, line 10: It's not harmful, but I guess this was left in by mistake, no?

device/sycl/src/seeding/seed_finding.sycl
```diff
@@ -55,6 +60,40 @@ class select_seeds;
 
 }  // namespace kernels
 
+template <class kernelName>
+vecmem::data::vector_buffer<device::prefix_sum_element_t> make_prefix_sum_buff(
+    const std::vector<device::prefix_sum_size_t> sizes,
+    const std::unique_ptr<vecmem::copy>& copy,
```
krasznaa, line 66: You know by now... :stuck_out_tongue:

device/sycl/src/seeding/seed_finding.sycl
```diff
@@ -55,6 +60,40 @@ class select_seeds;
...
+    const traccc::memory_resource& mr, queue_wrapper& queue) {
```
krasznaa, line 67: The `traccc::sycl::queue_wrapper` type is useful for passing around `::sycl::queue` objects in "pure C\+\+" code. Since here we are defining this function in a source file that has to be built by a SYCL compiler, you could switch to just pass a `::sycl::queue&` non-const reference to the function directly as well.

examples/run/sycl/seq_example_sycl.sycl
```diff
@@ -104,6 +104,7 @@ int seq_run(const traccc::full_tracking_input_config& i_cfg,
 
     // Struct with memory resources to pass to SYCL algorithms
     traccc::memory_resource mr{device_mr, &sycl_host_mr};
+    // traccc::memory_resource mr{shared_mr};
```
krasznaa, line 107: Let's remove this line from this PR. We can come back to a nice setup for switching between memory resource setups in a future PR. :wink:

krasznaa (approved):
I'm still very positive about the whole thing, I just had a **lot** of technical comments...

krasznaa (changes requested):
Approved the request by mistake. :facepalm:

device/common/include/traccc/device/make_sum_buffer.hpp
```diff
@@ -0,0 +1,59 @@
...
+    const std::vector<prefix_sum_size_t> sizes,
+    const std::unique_ptr<vecmem::copy>& copy,
```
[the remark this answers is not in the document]
stephenswat: I mean, passing smart pointers by reference definitely has its uses. But not here. :stuck_out_tongue:

device/sycl/src/seeding/seed_finding.sycl
guilhermeAlmeida1: I intentionally added it because of an std::transform used in this file. The std::transform was already there, not added in this PR, but seemed better to explicitly include it

device/common/include/traccc/device/make_sum_buffer.hpp
```diff
@@ -0,0 +1,59 @@
...
+///
+/// @param sizes The sizes of the "inner vectors" of a jagged vector
+/// @param copy A "copy object" capable of dealing with the view
+/// @param mr    The memory resource to use of the result
+/// @return The prefix sum for efficiently iterating over all element of a
+///         jagged vector
+///
+TRACCC_HOST
+make_buffer_result_t<prefix_sum_size_t> make_sum_buffer(
+    const std::vector<prefix_sum_size_t> sizes,
```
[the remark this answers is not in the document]
guilhermeAlmeida1: I didn't pass this as const reference because in seed_finding I'm calling this function directly with copy->get_sizes(...). If I were to pass by reference I'd need to in each call first create a 

```
auto foo = copy->get_sizes(...)
make_sum_buffer(foo, ...)
```

device/common/src/make_sum_buffer.cpp
```diff
@@ -0,0 +1,48 @@
...
+    const std::vector<prefix_sum_size_t> sizes,
+    const std::unique_ptr<vecmem::copy>& copy,
+    const traccc::memory_resource& mr) {
+
+    // Create vector with summation of sizes
+    vecmem::vector<device::prefix_sum_size_t> sizes_sum(
+        sizes.size(), mr.host ? mr.host : &(mr.main));
+    std::partial_sum(sizes.begin(), sizes.end(), sizes_sum.begin(),
+                     std::plus<prefix_sum_size_t>());
+    const unsigned int totalSize = sizes_sum[sizes_sum.size() - 1];
```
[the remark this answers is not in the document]
guilhermeAlmeida1: I'm not a huge fan of adding this return to the current implementation. I reckon the code would be a bit cluttered as this requires functionality to be added for both cases of shared and separate memory. And if we're getting an empty input here something has already gone very wrong. Would you be unhappy if I just added an `assert(size != 0);` at the beggining?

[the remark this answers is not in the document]
krasznaa: I'm very surprised by this comment. :confused: This particular function can and should work if it's given a jagged vector without any elements. While in our current case we don't expect that to happen, in the future we may very well have a use case for it.

And no, the implementation does not need to be complicated. Since the 1-dimensional buffer that we need to return is empty, we don't need to create anything. We can just do:

```c++
if (sizes.size() == 0) {
    return {{}, {}, 0};
}
```

The default constructors of both the vector view and the variant will be fine for this. (The default constructor of a vector view creates a view of an empty vector. Which is not pointing anywhere in memory, since an empty view does not need to point anywhere in memory.)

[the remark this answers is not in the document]
guilhermeAlmeida1: I see, I was overcomplicating how this could be done. Fixed now

device/sycl/src/seeding/seed_finding.sycl
krasznaa: :+1:

device/common/include/traccc/device/make_sum_buffer.hpp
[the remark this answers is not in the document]
guilhermeAlmeida1: Oh, of course. I was already doing it in fact in the previous version within the traccc::sycl::make_prefix_sum_buff. I just got a compiler error when making this change which led me to think about this the wrong way. It's fixed now

device/sycl/src/seeding/seed_finding.sycl
```diff
@@ -55,6 +60,39 @@ class select_seeds;
 
 }  // namespace kernels
 
+template <class kernelName>
+vecmem::data::vector_buffer<device::prefix_sum_element_t> make_prefix_sum_buff(
+    const std::vector<device::prefix_sum_size_t>& sizes, vecmem::copy& copy,
+    const traccc::memory_resource& mr, ::sycl::queue& queue) {
```
krasznaa, line 66: The function should not be implemented like this, as a template. You shouldn't give different "names" for the different prefix filling kernels.

Just like how with CUDA we can have a single `traccc::cuda::kernels::fill_prefix_sum` global function, you should define a single `traccc::sycl::kernels::fill_prefix_sum` class in this file as well, and then use that to identify every kernel submission from this function. Since the kernel is really the same in all cases, the same code just operates on different inputs. (Note that this has a real performance impact. Since the SYCL runtime separately compiles and loads kernels with different names/identifiers at runtime. So making them share the same identifier will have benefits for the code.)

device/common/include/traccc/device/make_prefix_sum_buffer.hpp
```diff
@@ -0,0 +1,57 @@
...
+///
+/// This simple function is just meant to translate the received "size vector"
+/// into a "prefix sum". I.e. into a list of index pairs that would allow
+/// visiting all elements of the jagged vector described by this "size vector".
+///
+/// @param sizes The sizes of the "inner vectors" of a jagged vector
+/// @param copy A "copy object" capable of dealing with the view
+/// @param mr    The memory resource to use of the result
+/// @return The prefix sum for efficiently iterating over all element of a
+///         jagged vector
```
krasznaa, line 50: Since I have a legitimate complaint as well, let's me point out this documentation here... Please update it to correctly describe what this function is actually doing.

examples/run/sycl/seq_example_sycl.sycl
krasznaa: :confused: You resolved this discussion, but the commented line is still there...

krasznaa (changes requested):
Just one actual request this time around. But please address the other two as well.

device/common/include/traccc/device/make_prefix_sum_buffer.hpp
guilhermeAlmeida1: Is the current version clear on this?

krasznaa (approved):
Good! Let's finally have this in...

device/common/include/traccc/device/make_prefix_sum_buffer.hpp
krasznaa: :+1:

PR #226, opened 2022-09-05
Add support for the Futhark programming language

stephenswat:
This commit provides some of the necessary utilities for adding support for Futhark to the traccc code base. It contains useful wrappers for Futhark C interfaces, as well as the necessary CMake functions to make this work.

### DIFF
--- a/cmake/FindFuthark.cmake
+++ b/cmake/FindFuthark.cmake
@@ -1,0 +1,154 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# Necessary CMake includes
+include(FindPackageHandleStandardArgs)
+include(CheckLanguage)
+
+# The C language is not enabled in our project by default, but we need it to
+# compile Futhark's code.
+check_language(C)
+if(CMAKE_C_COMPILER)
+    enable_language(C)
+endif()
+
+# Try to find our dependencies, which are of course Futhark itself (required)
+# and the CUDA toolkit for compiling CUDA projects. Also notify the user of
+# what we find.
+find_program(Futhark_EXECUTABLE futhark)
+mark_as_advanced(Futhark_EXECUTABLE)
+find_package(CUDAToolkit)
+
+# Report the status of loading Futhark to the user.
+find_package_handle_standard_args(
+    Futhark
+    REQUIRED_VARS
+    Futhark_EXECUTABLE
+    CMAKE_C_COMPILER
+)
+
+# This function describes how to add a Futhark source to a CMake library. The
+# function adds sources generated from a source file `_SOURCE_NAME` to a target
+# called `TARGET_NAME`. The Futhark target platform (CUDA or C) is given
+# (case-insensitively) by `_LANGUAGE_TARGET`
+function(add_futhark_to_library TARGET_NAME _LANGUAGE_TARGET _SOURCE_NAME)
+    # Parse the arguments that we get.
+    cmake_parse_arguments(
+        ARGS
+        ""
+        ""
+        "DEPENDENCIES"
+        ${ARGN}
+    )
+
+    # Set a few useful variables.
+    set(FUTHARK_LANGUAGES c cuda)
+    set(SOURCE_NAME "${CMAKE_CURRENT_SOURCE_DIR}/${_SOURCE_NAME}")
+    string(TOLOWER ${_LANGUAGE_TARGET} LANGUAGE_TARGET)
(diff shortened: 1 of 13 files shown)


cmake/FindFuthark.cmake
```diff
@@ -0,0 +1,125 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2021-2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# The C language is not enabled in our project by default, but we need it to
+# compile Futhark's code.
+enable_language(C)
```
krasznaa, line 9: "Find modules" are not meant to make a CMake configuration fail themselves. They can instruct their caller to bail out because they didn't find what they were looking for, but they shouldn't have the equivalent of `exit(1)` ever.

Long story short, this should be replaced by:

```cmake
include(CheckLanguage)
check_language(C)
if(CMAKE_C_COMPILER)
   enable_language(C)
endif()
```

And then `CMAKE_C_COMPILER` needs to be a "required variable" for this module "to be found". (More on that later on.)

cmake/FindFuthark.cmake
```diff
@@ -0,0 +1,125 @@
...
+
+# Try to find our dependencies, which are of course Futhark itself (required)
+# and the CUDA toolkit for compiling CUDA projects. Also notify the user of
+# what we find.
+find_program(Futhark futhark REQUIRED)
```
krasznaa, line 14: Along the same lines as before, if `find_package(Futhark)` is **not** called with `REQUIRED`, then it must not make the configuration fail just because Futhark was not found.

Change this line to:

```cmake
find_program(Futhark_EXECUTABLE futhark)
mark_as_advanced(Futhark_EXECUTABLE)
```

(Variables declared by the module should have a common prefix. Either `FUTHARK_` or `Futhark_`.)

cmake/FindFuthark.cmake
```diff
@@ -0,0 +1,125 @@
...
+find_package(CUDAToolkit)
```
krasznaa, line 15: Later in the code, before you try to make use of the targets/variables set by this find-module, you should check whether `CUDAToolkit_FOUND` is `TRUE` or not. And give a meaningful error if the user asked for CUDA sources without CUDA being available.

cmake/FindFuthark.cmake
```diff
@@ -0,0 +1,125 @@
...
+
+message(STATUS "Found Futhark compiler at ${Futhark}")
```
krasznaa, line 17: Let's remove this line from here, and put the following at the end of the file instead:

```cmake
include(FindPackageHandleStandardArgs)
find_package_handle_standard_args(Futhark
   REQUIRED_VARS Futhark_EXECUTABLE)
```

This will give the message a nice CMake-y look. And very importantly, it will automatically make the module do the right thing for the `QUIET` and `REQUIRED` arguments.

krasznaa (changes requested):
Possibly further comments to follow after the updates. :stuck_out_tongue:

cmake/FindFuthark.cmake
```diff
@@ -0,0 +1,152 @@
...
+        message(
+            FATAL_ERROR
+            "Futhark language must be one of ${FUTHARK_LANGUAGES}, not ${LANGUAGE_TARGET}"
+        )
+    endif()
+
+    # Define some needed directory and file names.
+    get_filename_component(FILE_ROOT ${SOURCE_NAME} NAME_WE)
+    set(FUTHARK_OUTPUT_SOURCE_DIR ${CMAKE_CURRENT_BINARY_DIR}/src)
+    set(FUTHARK_OUTPUT_HEADER_DIR ${CMAKE_CURRENT_BINARY_DIR}/include/traccc/futhark)
```
krasznaa, line 80: Let's make everything be under `"${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/"`. `CMAKE_FILES_DIRECTORY` is a very badly documented variable that pretty much is just an alias for `"/CMakeFiles.dir"`. (With the leading slash.)

The point is, with an "in source" build such a setup makes sure that all the generated files would be tucked away in the directory anyway created by CMake. And not possibly interfere with the source files in the source directory.

krasznaa (changes requested):
Please put quotes around every filename and path expression in the find-module. To make sure that path- and file names with spaces in them would work correctly.

Other than that, I'm happy with the PR's state.

krasznaa (approved):
Let's finally go ahead with it. :wink:

stephenswat:
Please wait with the merge, there seems to be an issue.

stephenswat:
Okay, _now_ it's ready (it's `CMAKE_COMMAND`, not `CMAKE_EXECUTABLE`).

krasznaa (approved):
Let's go ahead...

PR #227, opened 2022-09-06
Add Futhark implementation of measurement creation

stephenswat:
Implemented using the FastSV algorithm.

Depends on #226.

### DIFF
--- a/device/futhark/CMakeLists.txt
+++ b/device/futhark/CMakeLists.txt
@@ -13,6 +13,11 @@ traccc_add_library(
     futhark
     TYPE SHARED
     "src/context.cpp"
+    "src/component_connection.cpp"
+    "include/traccc/futhark/context.hpp"
+    "include/traccc/futhark/component_connection.hpp"
+    "include/traccc/futhark/utils.hpp"
+    "include/traccc/futhark/wrapper.hpp"
 )
 
 # Link the new outward-facing library against our code librar(y|ies)
@@ -34,4 +39,5 @@ add_futhark_to_library(
     src/edm.fut
     src/maybe.fut
     src/zip.fut
+    src/measurement_creation.fut
 )
--- a/device/futhark/include/traccc/futhark/wrapper.hpp
+++ b/device/futhark/include/traccc/futhark/wrapper.hpp
@@ -12,10 +12,56 @@
 #include <cstring>
 #include <numeric>
 #include <traccc/futhark/context.hpp>
+#include <traccc/futhark/utils.hpp>
 #include <tuple>
 #include <vector>
 
 namespace traccc::futhark {
+struct futhark_u64_1d_wrapper {
+    using cpp_t = uint64_t;
+    using futhark_t = struct futhark_u64_1d;
+    static constexpr futhark_t *(*alloc_f)(struct futhark_context *,
+                                           const cpp_t *,
+                                           int64_t) = &futhark_new_u64_1d;
+    static constexpr int (*free_f)(struct futhark_context *,
+                                   futhark_t *) = &futhark_free_u64_1d;
+    static constexpr const int64_t *(*shape_f)(
+        struct futhark_context *, futhark_t *) = &futhark_shape_u64_1d;
+    static constexpr int (*values_f)(struct futhark_context *, futhark_t *,
+                                     cpp_t *) = &futhark_values_u64_1d;
+    static constexpr std::size_t rank_v = 1;
+};
+
+struct futhark_i64_1d_wrapper {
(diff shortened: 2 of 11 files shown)


device/futhark/src/cca.cpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    cells_to_measurements_wrapper::output_t r =
+        cells_to_measurements_wrapper::run(
+            std::move(host_event), std::move(host_geometry),
+            std::move(host_channel0), std::move(host_channel1),
+            std::move(host_activation));
+
+    output_type out;
+    vecmem::host_memory_resource mem;
```
krasznaa, line 48: I would prefer to make the algorithm receive the memory resource that it should use, from the outside. So that a user could for instance set it up to put its output into SYCL shared memory. Or HIP host memory. Or some cached version of any of those...

In general when I see short-live memory resource objects, that tends to be a bad code smell for me. Even if creating `vecmem::host_memory_resource` on demand is of course very cheap.

tests/futhark/test_cca.cpp
```diff
@@ -0,0 +1,53 @@
...
+namespace {
+traccc::futhark::component_connection cc;
+
+cca_function_t f = [](const traccc::cell_container_types::host &data) {
+    std::map<traccc::geometry_id, vecmem::vector<traccc::measurement>> result;
+
+    traccc::measurement_container_types::host mss = cc(data);
+
+    for (std::size_t i = 0; i < mss.size(); ++i) {
+        vecmem::vector<traccc::measurement> msv;
```
krasznaa, line 31: You should not rely on the default memory resource with any host container. While (thanks to the polyfill we have in VecMem) this works on all platforms at the moment, we should really use explicit memory resources for all containers.

device/futhark/include/traccc/futhark/wrapper.hpp
```diff
@@ -88,4 +152,16 @@ struct wrapper<T, std::tuple<IArgs...>, std::tuple<OArgs...>> {
...
+    : public wrapper<
+          cells_to_measurements_wrapper,
+          std::tuple<futhark_u64_1d_wrapper, futhark_u64_1d_wrapper,
+                     futhark_i64_1d_wrapper, futhark_i64_1d_wrapper,
+                     futhark_f32_1d_wrapper>,
+          std::tuple<futhark_u64_1d_wrapper, futhark_u64_1d_wrapper,
+                     futhark_f32_1d_wrapper, futhark_f32_1d_wrapper,
+                     futhark_f32_1d_wrapper, futhark_f32_1d_wrapper>> {
+    static constexpr auto *entry_f = &futhark_entry_cells_to_measurements;
+};
```
krasznaa, line 166: You should move this definition to some more appropriate place.

krasznaa (changes requested):
Just some organisational comments/requests.

stephenswat:
Thanks, I've resolved the issues. :smile:

device/futhark/include/traccc/futhark/cca.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#pragma once
+
+#include "traccc/edm/cell.hpp"
+#include "traccc/edm/measurement.hpp"
+#include "traccc/utils/algorithm.hpp"
+#include "vecmem/memory/memory_resource.hpp"
+
+struct futhark_context_config;
+struct futhark_context;
```
krasznaa, line 16: There are not actually needed here, right? :confused:

device/futhark/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ traccc_add_library(
     traccc_futhark
     futhark
     TYPE SHARED
+    "src/cca.cpp"
```
krasznaa, line 15: I would prefer if the header files were listed as well, but it's not mandatory...

krasznaa (commented):
I'm not crazy about the `cca.hpp` header name, but I guess it could work... (Though generally in ATLAS we require headers to have the same name as the "main component" that they describe.)

device/futhark/include/traccc/futhark/cca.hpp
stephenswat: Indeed not, removed.

device/futhark/CMakeLists.txt
stephenswat: Headers added.

stephenswat:
I have resolved your comments and renamed the files to `measurement_creation.{h,c}pp`.

krasznaa (approved):
Let's do this finally...

PR #228, opened 2022-09-07
Update Acts dependency to version 20.1.0

stephenswat:
We're currently using a very, very old version of Acts, namely v10. It is also a modified version from Attila's repository. I think it's important that we update the version of Acts so we can get the latest utilities, and keep up with the latest developments in algorithms.

### DIFF
--- a/extern/acts/CMakeLists.txt
+++ b/extern/acts/CMakeLists.txt
@@ -13,7 +13,7 @@ message( STATUS "Building Acts as part of the TRACCC project" )
 
 # Declare where to get Acts from.
 set( TRACCC_ACTS_SOURCE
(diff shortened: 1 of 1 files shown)


beomki-yeo:
As expected, the comparison with Acts seeding and digitization gets an error..

krasznaa:
:thinking: The fact that the build succeeds right away is at least good.

I agree with Stephen, we need to make this update with some priority. We'll need to make this project work with the very latest version of Acts before we could think of merging its worthy parts into that project.

So... who wants to hunt down the differences in seeding between Acts and traccc's CPU code? :stuck_out_tongue:

beomki-yeo:
OK. Meanwhile, I would like to let you know that some algorithms will experience non-trivial changes once I start merging detray into traccc soon. Technically, traccc algorithms are assuming cartesian coordinate in all places which might not be right. I would need some time to find out where to fix. Once I fix the issues, I guess we can start updating ACTS version. Do you think it is reasonable?

PR #229, opened 2022-09-07
Introduce traccc::pair, main branch (2022.09.07.)

krasznaa:
Introduced `traccc::pair`, and updated the container code to use it.

Some functions of the [std::pair](https://en.cppreference.com/w/cpp/utility/pair) type are not available in device code, which required us declaring some awkward functions earlier. With the `traccc::pair` type those functions could be deleted.

However I still introduced more new lines of code than how many I managed to delete. :stuck_out_tongue: So I'm curious what everyone thinks about the merit of this PR... (I'll want to make additional updates to the EDM code after this, I just didn't like how `traccc::seed` looked like at the moment...)

PR #231, opened 2022-09-14
Switch traccc::container to thrust::pair, main branch (2022.09.14.)

krasznaa:
This is a replacement for #229. Instead of writing our own `traccc::pair` type it seemed better to just use `thrust::pair`. Since we already use [Thrust](https://github.com/NVIDIA/thrust) in [detray](https://github.com/acts-project/detray) anyway.

Note that this needs to wait for an update to a version of detray that will include https://github.com/acts-project/detray/pull/298. Until then this will remain a draft.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -70,4 +70,5 @@ traccc_add_library( traccc_core core TYPE SHARED
   "include/traccc/seeding/spacepoint_binning.hpp"
   "src/seeding/spacepoint_binning.cpp" )
 target_link_libraries( traccc_core
-  PUBLIC Eigen3::Eigen vecmem::core detray::core ActsCore ActsPluginJson traccc::algebra )
+  PUBLIC Eigen3::Eigen vecmem::core detray::core ActsCore ActsPluginJson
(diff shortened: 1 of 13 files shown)


core/CMakeLists.txt
```diff
@@ -70,4 +70,5 @@ traccc_add_library( traccc_core core TYPE SHARED
   "include/traccc/seeding/spacepoint_binning.hpp"
   "src/seeding/spacepoint_binning.cpp" )
 target_link_libraries( traccc_core
-  PUBLIC Eigen3::Eigen vecmem::core detray::core ActsCore ActsPluginJson traccc::algebra )
+  PUBLIC Eigen3::Eigen vecmem::core detray::core ActsCore ActsPluginJson
```
stephenswat, line 73: Not for this Pr, but we should move towards one-line-per-target for these sorts of things at some point...

stephenswat (approved):
Looks harmless.

core/CMakeLists.txt
krasznaa: :+1:

stephenswat (approved):

PR #232, opened 2022-09-15
CI Update, main branch (2022.09.15.)

krasznaa:
Switched the CI to the latest Acts Docker images. Mainly for the inclusion of a newer CMake version in the "vanilla" image (needed for #231), but at that point we might as well just use the latest images for all build flavours.


stephenswat (approved):
Let's merge this if the CI passes

PR #234, opened 2022-09-15
Changed configuration parameters' units to match ACTS

guilhermeAlmeida1:
Changed parameters in seeding configs to use explicit units using ACTS::UnitConstants, matching current ACTS formalism. 

OUTDATED: [ This PR will currently **not work** as the version of ACTS in use (10.x , from July 1, 2021) still uses the old formalism of implicit units. ]

EDIT: I changed compare_with_acts.cpp to work with the old version of ACTS currently in use so that we can push these changes without a major rewrite of the whole code, but these changes should be reverted when upgrading to the current version of ACTS

### DIFF
--- a/core/src/seeding/seeding_algorithm.cpp
+++ b/core/src/seeding/seeding_algorithm.cpp
@@ -19,16 +19,16 @@ namespace {
 traccc::seedfinder_config default_seedfinder_config() {
 
     traccc::seedfinder_config config;
-    config.highland = 13.6 * std::sqrt(config.radLengthPerSeed) *
-                      (1 + 0.038 * std::log(config.radLengthPerSeed));
-    float maxScatteringAngle = config.highland / config.minPt;
+    traccc::seedfinder_config config_copy = config.toInternalUnits();
+    config.highland = 13.6 * std::sqrt(config_copy.radLengthPerSeed) *
+                      (1 + 0.038 * std::log(config_copy.radLengthPerSeed));
+    float maxScatteringAngle = config.highland / config_copy.minPt;
     config.maxScatteringAngle2 = maxScatteringAngle * maxScatteringAngle;
     // helix radius in homogeneous magnetic field. Units are Kilotesla, MeV
     // and millimeter
-    // TODO: change using ACTS units
-    config.pTPerHelixRadius = 300. * config.bFieldInZ;
+    config.pTPerHelixRadius = 300. * config_copy.bFieldInZ;
     config.minHelixDiameter2 =
-        std::pow(config.minPt * 2 / config.pTPerHelixRadius, 2);
+        std::pow(config_copy.minPt * 2 / config.pTPerHelixRadius, 2);
     config.pT2perRadius =
         std::pow(config.highland / config.pTPerHelixRadius, 2);
     return config;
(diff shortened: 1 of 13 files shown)


beomki-yeo (commented):
I think the original seeding parameters of traccc was following `1 = 1 * MeV`, `1 = 1 T` and `1 = 1 * mm` convention, which is different from ACTS unit conventions (`1 = 1 * GeV`, `0.000299792458 = 1 T` and `1 = 1 * mm`).

Therefore, I like the changes you made in this PR because we need to follow the ACTS unit. 
Meanwhile, I suppose you added `toInternalUnits` to have the same results after changing the units. 
Instead of having the internal unit conversion, how about correcting the doublet and triplet finding functions directly?

Edit: Maybe you don't have to touch algorithm part :p

beomki-yeo:
For example, you might need to edit `triplet_finding_helper::is_Compatible()` which calculates the radius of curvature:

```c++
    // A and B are evaluated as a function of the circumference parameters
    // x_0 and y_0
    scalar A = (lt.V() - lb.V()) / dU;
    scalar S2 = 1. + A * A;
    scalar B = lb.V() - A * lb.U();
    scalar B2 = B * B;
    // sqrt(S2)/B = 2 * helixradius
    // calculated radius must not be smaller than minimum radius
    if (S2 < B2 * config.minHelixDiameter2) {
        return false;
    }

    // 1/helixradius: (B/sqrt(S2))*2 (we leave everything squared)
    scalar iHelixDiameter2 = B2 / S2;
```

core/src/seeding/seeding_algorithm.cpp
```diff
@@ -19,16 +19,16 @@ namespace {
...
+    traccc::seedfinder_config config_copy = config.toInternalUnits();
+    config.highland = 13.6 * std::sqrt(config_copy.radLengthPerSeed) *
+                      (1 + 0.038 * std::log(config_copy.radLengthPerSeed));
+    float maxScatteringAngle = config.highland / config_copy.minPt;
     config.maxScatteringAngle2 = maxScatteringAngle * maxScatteringAngle;
     // helix radius in homogeneous magnetic field. Units are Kilotesla, MeV
     // and millimeter
-    // TODO: change using ACTS units
-    config.pTPerHelixRadius = 300. * config.bFieldInZ;
+    config.pTPerHelixRadius = 300. * config_copy.bFieldInZ;
```
beomki-yeo, line 29: Maybe changing this to  `config.pTPerHelixRadius [GeV/mm] = 3.33 * 10^3 * config_copy.bFieldInZ [T]` would make sense

guilhermeAlmeida1:
@beomki-yeo Actually Acts also uses the same notation right now of having the config in explicit units and using toInternalUnits on the config when calling the seedfinder (see [line 20](https://github.com/acts-project/acts/blob/8bc896e3067a71f6096a85ad464900e0e2cad373/Core/include/Acts/Seeding/Seedfinder.ipp#L20)) We could probably change the actual functions to deal with explicit units but I did not do to have the same as is in Acts

beomki-yeo:
I think ACTS developers didn't bother using the explicit unit convention for all parts of seeding. While I still think this should be changed, it's OK at the moment

beomki-yeo (approved):

PR #235, opened 2022-09-15
Improve memory performance of CUDA CCA code

stephenswat:
As it stands, the CUDA CCA code uses managed memory because it's the path of least resistance. Sadly, it is also the path of least performance. In this commit, we change the CUDA CCA code to use device memory rather than managed memory, and this speeds up the algorithm significantly. In my preliminary measurements, the difference is roughly a factor five.

Also changes the code to use the new `traccc::memory_resource` type, and switches from 64-bit to 32-bit integers for better on-GPU performance.

### DIFF
--- a/device/cuda/src/cca/component_connection.cu
+++ b/device/cuda/src/cca/component_connection.cu
@@ -12,9 +12,7 @@
 #include "traccc/edm/cluster.hpp"
 #include "traccc/edm/measurement.hpp"
 #include "vecmem/containers/vector.hpp"
-#include "vecmem/memory/allocator.hpp"
-#include "vecmem/memory/binary_page_memory_resource.hpp"
-#include "vecmem/memory/cuda/managed_memory_resource.hpp"
+#include "vecmem/memory/cuda/device_memory_resource.hpp"
 
 namespace {
 static constexpr std::size_t MAX_CELLS_PER_PARTITION = 2048;
@@ -29,8 +27,8 @@ namespace details {
  * Structure that defines the start point of a partition and its size.
  */
 struct ccl_partition {
-    std::size_t start;
-    std::size_t size;
+    unsigned int start;
+    unsigned int size;
 };
 
 /*
@@ -38,7 +36,7 @@ struct ccl_partition {
  * an array/vector of cells.
  */
 struct cell_container {
-    std::size_t size = 0;
+    unsigned int size = 0;
     channel_id* channel0 = nullptr;
     channel_id* channel1 = nullptr;
     scalar* activation = nullptr;
@@ -51,7 +49,6 @@ struct cell_container {
  * an array/vector of measures.
  */
 struct measurement_container {
-    unsigned int size = 0;
     scalar* channel0 = nullptr;
     scalar* channel1 = nullptr;
     scalar* variance0 = nullptr;
@@ -518,7 +515,7 @@ __device__ void aggregate_clusters(const cell_container& cells,
 
 __global__ __launch_bounds__(THREADS_PER_BLOCK) void ccl_kernel(
     const cell_container container, const ccl_partition* partitions,
-    measurement_container& _out_ctnr) {
+    measurement_container _out_ctnr, unsigned int* measurement_count) {
     const ccl_partition& partition = partitions[blockIdx.x];
 
     /*
@@ -646,7 +643,7 @@ __global__ __launch_bounds__(THREADS_PER_BLOCK) void ccl_kernel(
      * amount of threads per block, this has no sever implications.
      */
     if (threadIdx.x == 0) {
-        outi = atomicAdd(&_out_ctnr.size, outi);
+        outi = atomicAdd(measurement_count, outi);
     }
 
     __syncthreads();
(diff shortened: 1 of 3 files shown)


stephenswat:
CI failures seem to be unrelated.

device/cuda/src/cca/component_connection.cu
```diff
@@ -727,14 +724,16 @@ vecmem::vector<details::ccl_partition> partition(
...
 }  // namespace details
 
+component_connection::component_connection(const traccc::memory_resource& mr)
+    : m_mr(mr) {}
+
 component_connection::output_type component_connection::operator()(
     const cell_container_types::host& data) const {
-    vecmem::cuda::managed_memory_resource upstream;
-    vecmem::binary_page_memory_resource mem(upstream);
+    vecmem::cuda::device_memory_resource dmr;
```
krasznaa, line 732: We should really come to an agreement about what to do here. :thinking:

I'm still firmly in the belief that algorithms must not own their own memory resources, and must depend on outside resources to be given to them in their constructors. I'm okay with this algorithm not relying on a single `traccc::memory_resource` if you say that by definition it must not be used with a single shared/managed resource, but then make it require 2 resources all the time. Like:

```c++
component_connection(vecmem::memory_resource& host_mr, vecmem::memory_resource& device_mr);
```

But I can't give up on the exact memory resource "coming from the outside"...

device/cuda/src/cca/component_connection.cu
```diff
@@ -743,19 +742,19 @@ component_connection::output_type component_connection::operator()(
      * more efficiently. This removes the hierarchy level that
      * references to the cell module.
      */
-    vecmem::vector<channel_id> channel0(&mem);
+    std::vector<channel_id> channel0;
```
krasznaa, line 745: With an interface requiring a specific "host memory resource" I'd really like these to return to being `vecmem::vector` instances.

device/cuda/src/cca/component_connection.cu
```diff
@@ -764,16 +763,49 @@ component_connection::output_type component_connection::operator()(
...
+    vecmem::unique_alloc_ptr<channel_id[]> channel0_device =
+        vecmem::make_unique_alloc<channel_id[]>(dmr, channel0.size());
+    vecmem::unique_alloc_ptr<channel_id[]> channel1_device =
+        vecmem::make_unique_alloc<channel_id[]>(dmr, channel1.size());
+    vecmem::unique_alloc_ptr<scalar[]> activation_device =
+        vecmem::make_unique_alloc<scalar[]>(dmr, activation.size());
+    vecmem::unique_alloc_ptr<scalar[]> time_device =
+        vecmem::make_unique_alloc<scalar[]>(dmr, time.size());
+    vecmem::unique_alloc_ptr<geometry_id[]> module_id_device =
+        vecmem::make_unique_alloc<geometry_id[]>(dmr, module_id.size());
```
krasznaa, line 778: Why not `vecmem::data::vector_buffer` objects? Then the code could easily use vector views and device vectors around these things. In sync with how all the other code is written. :thinking:

device/cuda/src/cca/component_connection.cu
```diff
@@ -764,16 +763,49 @@ component_connection::output_type component_connection::operator()(
...
+                                cudaMemcpyHostToDevice));
+    CUDA_ERROR_CHECK(cudaMemcpy(activation_device.get(), activation.data(),
+                                total_cells * sizeof(scalar),
+                                cudaMemcpyHostToDevice));
+    CUDA_ERROR_CHECK(cudaMemcpy(time_device.get(), time.data(),
+                                total_cells * sizeof(scalar),
+                                cudaMemcpyHostToDevice));
+    CUDA_ERROR_CHECK(cudaMemcpy(module_id_device.get(), module_id.data(),
+                                total_cells * sizeof(geometry_id),
+                                cudaMemcpyHostToDevice));
```
krasznaa, line 797: Of course if using vector buffers, then these copies could be made with `vecmem::cuda::copy`.

device/cuda/src/cca/component_connection.cu
```diff
@@ -764,16 +763,49 @@ component_connection::output_type component_connection::operator()(
...
-    container.channel0 = channel0.data();
-    container.channel1 = channel1.data();
-    container.activation = activation.data();
-    container.time = time.data();
-    container.module_id = module_id.data();
+    container.channel0 = channel0_device.get();
+    container.channel1 = channel1_device.get();
+    container.activation = activation_device.get();
+    container.time = time_device.get();
+    container.module_id = module_id_device.get();
```
krasznaa, line 808: I find it **very** interesting that you use a SoA data type here. Since I believe eventually we'll need to transition all our data types to an SoA layout to be compatible with ATLAS's EDM.

Out of curiosity, did you try what happens with this code when using a simple AoS layout? Does the performance suffer in such a setup? I'm genuinely very interested in this.

krasznaa (changes requested):
I hope my standpoint is understandable here. I'd really like to make this algorithm as CUDA-independent as possible. (As I'll want to see if we can make it run under other languages as well.) But this PR is only increasing the algorithm's reliance on CUDA. But I don't see it really benefiting from that more direct reliance.

device/cuda/src/cca/component_connection.cu
stephenswat: To be honest this code was written so long ago I don't really remember whether I did this out of dogma or whether there was actual performance-related evidence for this... Agreed we should re-evaluate the SoA vs AoS approach on this.

device/cuda/src/cca/component_connection.cu
stephenswat: I'm not sure I see the point there, this is strictly host-only code that gains nothing from custom allocation schemes. And as in the other pull request, this is just data pre-processing; not performance-relevant and should be removed for production code. Not sure we should really spend much development time on that kind of code.

stephenswat:
This has fallen way too far behind and I can't be arsed to rebase it.

PR #236, opened 2022-09-16
Add Futhark implementation of spacepoint formation

stephenswat:
This commit adds a Futhark implementation of spacepoint formation, which is extremely brief in this functional style. There are some improvements to be made by using accelerator structures such as search trees, but this will do for a proof of concept.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,10 @@ flowchart LR
     %% Futhark measurement creation
     cell -->|<a href='https://github.com/acts-project/traccc/blob/main/device/futhark/src/measurement_creation.fut'>CCA</a>| meas;
     linkStyle 23 stroke: brown;
+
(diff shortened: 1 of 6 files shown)


krasznaa (commented):
What's the status on this? I didn't realise that this was standing here since so long. :frowning:

But also: Declaring `spacepoint_formation.fut` is missing from the CMake configuration. I don't think this PR would build as-is. Which is yet another point for reviving https://github.com/acts-project/machines/pull/71. So that the Futhark PRs could be properly tested in the CI...

stephenswat:
It would compile, but you are correct it should be in the CMake command so changes in the Futhark code can be properly tracked. I'll make the change; other than that this is good to go.

stephenswat:
Done. :slightly_smiling_face:

PR #237, opened 2022-09-19
ci: Add basic GPU CI

paulgessinger:


Chamodya-ka:
Thanks! Runner with a gpu will be helpful to maintain the MPS scripts.

beomki-yeo:
Hmm do you know why the CI fails in SurfaceBinningTest and IO tests?

paulgessinger:
Yes, it's missing input files. I changed the data repo clone to do a sparse checkout only, because otherwise the job takes too long downloading >1GB of inputs, and I only want to run the CUDA relevant tests. I think I need to disable the unit tests that try to use these files.

stephenswat:
Pinging @guilhermeAlmeida1 because he prodded me about this, Paul and I are going to try to get this back on the rails. :slightly_smiling_face:

paulgessinger:
CI fails with 

```
698 - CUDAArrayWrapper.AoSLayout (Failed)
```

Feel free to push a commit fixing (or disabling) this test, and then let's merge it?

stephenswat:
#387 should make this viable and then we can put it in :parrot:

stephenswat (approved):

PR #238, opened 2022-09-19
Container Copy Algorithms, main branch (2022.09.19.)

krasznaa:
After some discussion with @stephenswat last week, added H->D and D->H container copy algorithms to the project. These are meant to become useful in full algorithm chains a bit later, once the algorithms have been re-worked a little. As "reconstruction algorithms" should not have to deal with memory movement. They should just assume that whatever is given to them as input, is already on the appropriate device.

Added a simple unit test for the code, just to check that the most basic things would work.

### DIFF
--- a/device/common/include/traccc/device/container_d2h_copy_alg.hpp
+++ b/device/common/include/traccc/device/container_d2h_copy_alg.hpp
@@ -1,0 +1,72 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/utils/algorithm.hpp"
+#include "traccc/utils/memory_resource.hpp"
+
+// VecMem include(s).
+#include <vecmem/utils/copy.hpp>
+
+namespace traccc::device {
+
+/// Algorithm to copy a container from a device to the host
+///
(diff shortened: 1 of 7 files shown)


device/common/include/traccc/device/container_d2h_copy_alg.hpp
```diff
@@ -0,0 +1,67 @@
...
+/// differently, so the user should make sure to pass in the correct type.
+///
+/// @tparam CONTAINER_TYPES One of the "container types" traits
+///
+template <typename CONTAINER_TYPES>
+class container_d2h_copy_alg
+    : public algorithm<typename CONTAINER_TYPES::host(
+          const typename CONTAINER_TYPES::const_view&)>,
+      public algorithm<typename CONTAINER_TYPES::host(
+          const typename CONTAINER_TYPES::buffer&)> {
```
stephenswat, line 35: Is it too much to ask to have support for passing in data by rvalue reference? :angel:

stephenswat (approved):
I don't know enough about this to really give an in-depth verdict, but it's certainly a necessary development so let's go ahead with it.

device/common/include/traccc/device/container_d2h_copy_alg.hpp
krasznaa: :confused: Instead of just arguing with you, here you go. Now you explain to me how this version of the code is better... As I truly believe it's worse like this. (Harder to maintain and more error prone for no performance gain.)

PR #241, opened 2022-09-26
Updated readme to match current development status

guilhermeAlmeida1:


stephenswat (approved):

stephenswat:
Actually, before we merge this could you fix the Mermaid graph as well?

stephenswat (changes requested):
Actually, before we merge this could you fix the Mermaid graph as well?

guilhermeAlmeida1:
> Actually, before we merge this could you fix the Mermaid graph as well?

I think this should be good now

stephenswat (approved):
Great, thanks! I would have added the new edges at the end of the graph but this also works of course. :smile:

PR #242, opened 2022-09-28
Allow use of system libraries by default

stephenswat:
This commit adds a new flag to the CMake build system that enables the use of system-installed libraries by default, as an alternative to the current default of fetching libraries on the fly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,9 @@ option( TRACCC_BUILD_FUTHARK "Build the Futhark sources included in traccc" FALS
 option( TRACCC_BUILD_TESTING "Build the (unit) tests of traccc" TRUE )
 option( TRACCC_BUILD_EXAMPLES "Build the examples of traccc" TRUE )
 
+# Flags controlling the meta-build system.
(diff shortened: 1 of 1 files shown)


krasznaa (approved):
I guess... Making the default values of CACHE variables depend on each other can result in some awkward situations. (Updating the value of `TRACCC_USE_SYSTEM_LIBS` in an existing build directory will not do anything to the values of the other flags.)

But there is no better way for achieving this type of behaviour. :thinking: I was thinking of [cmake_dependent_option](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html), but that's also not any more appropriate in this case...

stephenswat:
That's a good point, but I think the `TRACCC_USE_SYSTEM_XXX` flags are not actually cache variables, right? So presumably it would work?

krasznaa:
Every [option(...)](https://cmake.org/cmake/help/latest/command/option.html) variable is a boolean CACHE variable. Otherwise you couldn't override them from the command line. (That's one of the things that makes a cache variable a cache variable.)

Still, as I wrote before, I think this is as good as we can make this logic.

krasznaa (approved):
:+1:

PR #243, opened 2022-09-29
Entirely removed usage of get_prefix_sum to fill_prefix_sum device kernel

guilhermeAlmeida1:
The outdated host function get_prefix_sum which was upgraded to a device kernel was removed from clusterization. 

* Moved make_prefix_sum_buff function from of seed_findinding.sycl(cuda) to separate file make_prefix_sum_buff.sycl(cuda).
* Moved definitions in get_prefix_sum.hpp to fill_prefix_sum.hpp
* Removed get_prefix_sum


device/cuda/src/seeding/spacepoint_binning.cu
```diff
@@ -10,7 +10,8 @@
 #include "traccc/cuda/utils/definitions.hpp"
 
 // Project include(s).
-#include "traccc/device/get_prefix_sum.hpp"
+#include "traccc/cuda/utils/make_prefix_sum_buff.hpp"
+#include "traccc/device/fill_prefix_sum.hpp"
```
krasznaa, line 14: Is it really necessary to include `fill_prefix_sum.hpp` explicitly here? I know that right now that header is included in a **lot** of places. But I thought this file didn't need it explicitly. :thinking:

krasznaa (approved):
I like it. For my OCD we should do some follow-up cleanup later on. (We should really move the typedefs that all clients need into something like `traccc/device/prefix_sum_types.hpp` instead of including `traccc/device/fill_prefix_sum.hpp` everywhere.) But I think this PR could just go ahead as-is.

PR #244, opened 2022-09-29
MPS benchmarking scripts and test workflow

Chamodya-ka:
This PR adds Multi process bench-marking scripts and a workflow to test it. Moreover, this PR is used to check if multi_process_script workflow works as intended with the CI runner with gpu.


krasznaa (commented):
What's the idea with the CI setup exactly? Will this work with the CERN-based CI machine? @paulgessinger, will such a thing "fit into" #237?

I'm absolutely in favour of adding scripts for performance measurements. Though we'll probably need to write better applications than the current ones for proper throughput tests.

Chamodya-ka:
> Though we'll probably need to write better applications than the current ones for proper throughput tests.
This was not clear to me, Do you suggest an alternative approach for benchmarking?

krasznaa:
> This was not clear to me, Do you suggest an alternative approach for benchmarking?

Don't worry about that for now. You should just use these executables in your scripts for now. I just wanted to give you a heads' up that these applications are not able to do a super high throughput yet. (Their "host code" is not super efficient.)

PR #225, opened 2022-09-05
Device/Kokkos

Yhatoh:
Hi!

So here it comes the first pull request for the kokkos implementation. To remember a little of what I did was writing code of some of the seeding directory. I use cuda some cuda code of the functionalities that are not implemented (like seeding/specepoint_binning and seeding/track_params_estimation.cu).

Also, I writed the CMake file for kokkos and add it to the general CMake file, but I couldn't write a CMake code to download automatically Kokkos github and install it alone, the principal factor was Kokkos has a lot of flags to use it, so I didn't know if it was better to put all ON or not. To be known apparently for what i have been testing you need to use g++9.X to make Kokkos work (i know is weird apparently for now g++10.X and g++11.X Kokkos has a bug).

Finally, i don't know if you are interested in some Kokkos testing (like the Cuda ones) to test if Kokkos initialize correctly and that stuff.

I will be watching any commentary you do.

Cheers,
Gabriel

PR #245, opened 2022-10-06
Kokkos Base Setup, main branch (2022.10.06.)

krasznaa:
This is meant as a replacement for #225.

I pretty much just stole / transcribed @Yhatoh's code from that PR into this one. Such that it would be able to work...

Note that I disabled the CUDA support in Kokkos for now. Because I just didn't find a good way in which to enable it.
  - If I don't specify for Kokkos explicitly what CUDA architecture to use, it tries to auto-detect it. And fails in the CI with:

```
CMake Error at build/_deps/kokkos-src/cmake/kokkos_arch.cmake:776 (MESSAGE):
-- Built-in Execution Spaces:
  CUDA enabled but no NVIDIA GPU architecture currently enabled and
--     Device Parallel: Kokkos::Cuda
--     Host Parallel: NoTypeDefined
--       Host Serial: SERIAL
-- 
-- Architectures:
  auto-detection failed.  Please give one -DKokkos_ARCH_{..}=ON' to enable an
  NVIDIA GPU architecture.

  You can yourself try to compile
  /__w/traccc/traccc/build/_deps/kokkos-src/cmake/compile_tests/cuda_compute_capability.cc
  and run the executable.  If you are cross-compiling, you should try to do
  this on a compute node.
  build/_deps/kokkos-src/cmake/kokkos_tribits.cmake:243 (INCLUDE)
  build/_deps/kokkos-src/CMakeLists.txt:188 (KOKKOS_SETUP_BUILD_ENVIRONMENT)
```

  - If I just specify an older architecture manually, one that matches the value set for `CMAKE_CUDA_ARCHITECTURES`, then I got the following error during the build on a machine with a newer GPU:

```
[ 56%] Linking CXX executable ../../bin/traccc_test_kokkos
Kokkos::Cuda::initialize ERROR: running kernels compiled for compute capability 5.2 on device with compute capability 7.5 is not supported by CUDA!

Backtrace:
                                     Kokkos::Impl::save_stacktrace() [0x7f220a5b5082]
                    Kokkos::Impl::traceback_callstack(std::ostream&) [0x7f220a5aceda]
                               Kokkos::Impl::host_abort(char const*) [0x7f220a5acf5b]
                                                                           [0x403836]
     Kokkos::Impl::CudaInternal::initialize(int, CUstream_st*, bool) [0x7f220a5be1ea]
Kokkos::Cuda::impl_initialize(Kokkos::InitializationSettings const&) [0x7f220a5c0048]
                                                                     [0x7f220a5a7eac]
                                    Kokkos::initialize(int&, char**) [0x7f220a5a7e01]
                                                                           [0x402748]
                                                   __libc_start_main [0x7f2208930083]
                                                                           [0x40266e]
CMake Error at /software/cmake/3.24.1/x86_64-ubuntu2004-gcc9-opt/share/cmake-3.24/Modules/GoogleTestAddTests.cmake:112 (message):
  Error running test executable.

    Path: '/home/krasznaa/ATLAS/projects/traccc/build/bin/traccc_test_kokkos'
    Result: Subprocess aborted
    Output:
      

  /software/cmake/3.24.1/x86_64-ubuntu2004-gcc9-opt/share/cmake-3.24/Modules/GoogleTestAddTests.cmake:225 (gtest_discover_tests_impl)
```

Which is really bull@%&*. :frowning: @Yhatoh, is it really not possible to tell Kokkos to include PTX code in the binaries? (For JIT-ing it at runtime.) Please try to look this up. Because otherwise we won't be able to test this code properly.

It's also not a good look that `Kokkos::initialize(...)` seems to require a GPU to be available if CUDA support was turned on. That again will make using Kokkos very difficult in this project. :frowning:

P.S. I added a bunch of

```cmake
if( POLICY CMP0135 )
   cmake_policy( SET CMP0135 NEW )
endif()
```

statements to the PR as well. Just because I was lazy to open a separate PR for these...

### DIFF
--- a/tests/kokkos/CMakeLists.txt
+++ b/tests/kokkos/CMakeLists.txt
@@ -1,0 +1,13 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+traccc_add_test( kokkos
+   kokkos_main.cpp
+   kokkos_basic.cpp
+   LINK_LIBRARIES
+   GTest::gtest
+   Kokkos::kokkos
+)
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -26,6 +26,8 @@ jobs:
             container: ghcr.io/acts-project/ubuntu2004_cuda:v30
           - name: SYCL
             container: ghcr.io/acts-project/ubuntu2004_oneapi:v30
+          - name: KOKKOS
+            container: ghcr.io/acts-project/ubuntu2004:v30
         build:
           - Release
           - Debug
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,10 @@ endif()
 option( TRACCC_BUILD_CUDA "Build the CUDA sources included in traccc"
    ${TRACCC_BUILD_CUDA_DEFAULT} )
 option( TRACCC_BUILD_SYCL "Build the SYCL sources included in traccc" FALSE )
-option( TRACCC_BUILD_FUTHARK "Build the Futhark sources included in traccc" FALSE)
+option( TRACCC_BUILD_FUTHARK "Build the Futhark sources included in traccc"
+   FALSE )
+option( TRACCC_BUILD_KOKKOS "Build the Kokkos sources included in traccc"
+   FALSE )
 option( TRACCC_BUILD_TESTING "Build the (unit) tests of traccc" TRUE )
 option( TRACCC_BUILD_EXAMPLES "Build the examples of traccc" TRUE )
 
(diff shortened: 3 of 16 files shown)


tests/kokkos/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+traccc_add_test( kokkos
+   kokkos_main.cpp
```
stephenswat, line 8: Can we put this into a separate shared/static library like we do for CUDA?

stephenswat (changes requested):

tests/kokkos/CMakeLists.txt
krasznaa: :confused: I don't see much point in doing that. Once we want to build multiple separate executables that test Kokkos, we can move to such a setup. But as long as we only have a single executable, it's an unnecessary complication. (And I imagine that we'll only ever have a single executable that would contain all the Kokkos GoogleTests.)

stephenswat: OK. :+1:

PR #246, opened 2022-10-10
Refactor: Changed triplet counter class to include item_idx

guilhermeAlmeida1:
This PR introduces the item_idx to the triplet counter class. 
Before, the find_triplets kernel started by finding this position given the midBot doublet contained in the class. This was extremely inefficient.
The find_triplets kernel is now ~15x faster than it was before (e.g. for a 17k seed event: 1.886ms -> 0.126ms).

Also a bit of cleanup:
Removed unnecesary creation of view objects on top of buffers in cuda clusterization algorithm, as kernels can be directly called with the buffers with implicit translation to views.
Changed cuda seed_finding calculation of number of blocks to correctly match sycl implementation.

### DIFF
--- a/device/cuda/src/clusterization/clusterization_algorithm.cu
+++ b/device/cuda/src/clusterization/clusterization_algorithm.cu
@@ -258,8 +258,8 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
     m_copy->setup(spacepoints_buffer.items);
 
     // Calculating grid size for measurements creation kernel (block size 64)
-    blocksPerGrid =
-        (clusters_buffer.headers.size() - 1 + threadsPerBlock) / threadsPerBlock;
+    blocksPerGrid = (clusters_buffer.headers.size() - 1 + threadsPerBlock) /
+                    threadsPerBlock;
 
     // Invoke measurements creation will call create measurements kernel
     kernels::create_measurements<<<blocksPerGrid, threadsPerBlock>>>(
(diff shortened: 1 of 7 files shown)


device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -163,18 +163,14 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
 
     // Create views to pass to cluster finding kernel
     const cell_container_types::const_view cells_view(cells_data);
-    vecmem::data::jagged_vector_view<unsigned int> sparse_ccl_indices_view =
```
krasznaa, line 166: I am fully on board with this simplification in the CUDA code. But please put a few sentences into the description of the PR that mention this change/simplification.

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -180,7 +180,8 @@ vecmem::data::vector_buffer<seed> seed_finding::operator()(
     // counting kernel for.
     const unsigned int nDoubletCountThreads = WARP_SIZE * 2;
     const unsigned int nDoubletCountBlocks =
-        sp_grid_prefix_sum_buff.size() / nDoubletCountThreads + 1;
+        (sp_grid_prefix_sum_buff.size() + nDoubletCountThreads - 1) /
+        nDoubletCountThreads;
```
krasznaa, line 184: I am also happy with these changes in the CUDA code. But again, please add 1-2 sentences about these into the PR's description.

krasznaa (approved):
I like it. :+1: As long as you update the PR's description a bit, I believe this is good to go.

device/cuda/src/clusterization/clusterization_algorithm.cu
guilhermeAlmeida1: 👍

PR #247, opened 2022-10-10
VecMem Update, Code Simplification, main branch (2022.10.10.)

krasznaa:
These are the simplifications that I wanted to achieve with https://github.com/acts-project/vecmem/pull/197. With the latest version of VecMem it is no longer necessary to handle `vecmem::data::jagged_vector_buffer` differently from `vecmem::data::jagged_vector_view`. Which meant that a lot of our algorithms could be simplified. :wink:

In my local tests everything still works, so I'm pretty sure that this is a safe update. :crossed_fingers:


beomki-yeo (approved):
Looks good to me!

PR #248, opened 2022-10-10
Seed EDM Update, main branch (2022.10.10.)

krasznaa:
While working on #247 I was reminded that we should really continue with making all the EDM types be defined in the same way.

@guilhermeAlmeida1, once this is in, you should also harmonise the implementations of `traccc::cuda::track_parameter_estimation` and `traccc::sycl::track_parameter_estimation`. (Introducing common code into the `traccc::device` namespace as for all the other algorithms.) That should be a relatively simple exercise.

Note that this PR depends on #247, so it's a Draft until that one gets merged in.

### DIFF
--- a/core/include/traccc/edm/seed.hpp
+++ b/core/include/traccc/edm/seed.hpp
@@ -58,36 +58,7 @@ TRACCC_HOST std::vector<std::array<spacepoint, 3>> get_spacepoint_vector(
     return result;
 }
 
-/// Container of internal_spacepoint for an event
-template <template <typename> class vector_t>
-using seed_collection = vector_t<seed>;
-
-/// Convenience declaration for the seed collection type to use
-/// in host code
-using host_seed_collection = seed_collection<vecmem::vector>;
-
-/// Convenience declaration for the seed collection type to use
-/// in device code
-using device_seed_collection = seed_collection<vecmem::device_vector>;
-
-/// Convenience declaration for the seed container type to use in
-/// host code
-using host_seed_container = host_container<unsigned int, seed>;
-
-/// Convenience declaration for the seed container type to use in
-/// device code
-using device_seed_container = device_container<unsigned int, seed>;
-
-/// Convenience declaration for the seed container data type to
-/// use in host code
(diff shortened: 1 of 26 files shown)


device/common/include/traccc/seeding/device/select_seeds.hpp
```diff
@@ -1,3 +1,10 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
```
guilhermeAlmeida1, line 7: Thank you for this correction. Is it worth adding a unit test for this?

guilhermeAlmeida1 (commented):
Looks good to me. However, I think it's worth adding in the same changes in the seeding specific types.

device/common/include/traccc/seeding/device/select_seeds.hpp
krasznaa: I don't think we managed to figure out how to efficiently test for this in the ATLAS offline code either. :frowning: I wouldn't spend too much time on this, unless you already know how the CI could be taught to test for this.

core/include/traccc/edm/seed.hpp
```diff
@@ -58,36 +58,7 @@ TRACCC_HOST std::vector<std::array<spacepoint, 3>> get_spacepoint_vector(
...
-
-/// Convenience declaration for the seed container buffer type to
-/// use in host code
-using seed_container_buffer = container_buffer<unsigned int, seed>;
-
-/// Convenience declaration for the seed container view type to
-/// use in host code
-using seed_container_view = container_view<unsigned int, seed>;
+/// Declare all seed collection types
+using seed_collection_types = collection_types<seed>;
```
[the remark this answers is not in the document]
krasznaa: We absolutely need to do those updates as well. I just didn't want to do it in one huge PR.

If you're up to it, I'm happy if you can make those changes yourself. :wink:

PR #249, opened 2022-10-11
Seeding details EDM update

guilhermeAlmeida1:
Similarly to #248 , updated the seeding specific data types (listed below) to match others.

- singlet (not actually being used anywhere in the code, might be worth revisiting
- doublet
- triplet
- lin_circle
- bound_track_parameters

### DIFF
--- a/core/include/traccc/edm/track_parameters.hpp
+++ b/core/include/traccc/edm/track_parameters.hpp
@@ -98,19 +98,9 @@ inline bool operator==(const bound_track_parameters& lhs,
     return false;
 }
 
-/// Collection of bound_track_parameters
-template <template <typename> class vector_t>
-using bound_track_parameters_collection = vector_t<bound_track_parameters>;
-
-/// Convenient declaration for the bound track parameters collection type to use
-/// in host code
-using host_bound_track_parameters_collection =
-    bound_track_parameters_collection<vecmem::vector>;
-
-/// Convenient declaration for the bound track parameters collection type to use
-/// in device code
-using device_bound_track_parameters_collection =
-    bound_track_parameters_collection<vecmem::device_vector>;
+/// Declare all track_parameters collection types
+using bound_track_parameters_collection_types =
+    collection_types<bound_track_parameters>;
 
 struct curvilinear_track_parameters {
     using vector_t = bound_vector;
--- a/core/include/traccc/seeding/detail/lin_circle.hpp
+++ b/core/include/traccc/seeding/detail/lin_circle.hpp
@@ -51,7 +51,4 @@ struct lin_circle {
 /// Declare all lin_circle collection types
 using lin_circle_collection_types = collection_types<lin_circle>;
 
-/// Declare all lin_circle container types
-using lin_circle_container_types = container_types<unsigned int, lin_circle>;
-
 }  // namespace traccc
--- a/core/include/traccc/seeding/detail/lin_circle.hpp
+++ b/core/include/traccc/seeding/detail/lin_circle.hpp
@@ -48,37 +48,10 @@ struct lin_circle {
     const scalar& V() const { return m_V; }
 };
 
-/// Container of lin_circle belonging to one detector module
-template <template <typename> class vector_t>
-using lin_circle_collection = vector_t<lin_circle>;
+/// Declare all lin_circle collection types
+using lin_circle_collection_types = collection_types<lin_circle>;
 
-/// Convenience declaration for the lin_circle collection type to use in host
-/// code
-using host_lin_circle_collection = lin_circle_collection<vecmem::vector>;
-
-/// Convenience declaration for the lin_circle collection type to use in device
-/// code
-using device_lin_circle_collection =
-    lin_circle_collection<vecmem::device_vector>;
-
-/// Convenience declaration for the lin_circle container type to use in host
-/// code
-using host_lin_circle_container = host_container<unsigned int, lin_circle>;
-
-/// Convenience declaration for the lin_circle container type to use in device
-/// code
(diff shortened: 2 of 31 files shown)


core/include/traccc/seeding/detail/lin_circle.hpp
```diff
@@ -48,37 +48,10 @@ struct lin_circle {
...
-
-/// Convenience declaration for the lin_circle container buffer type to use in
-/// host code
-using lin_circle_container_buffer = container_buffer<unsigned int, lin_circle>;
-
-/// Convenience declaration for the lin_circle container view type to use in
-/// host code
-using lin_circle_container_view = container_view<unsigned int, lin_circle>;
+/// Declare all lin_circle container types
+using lin_circle_container_types = container_types<unsigned int, lin_circle>;
```
krasznaa, line 55: Let's not even declare this type. It doesn't seem to be used anywhere. (And the first argument of the template seems to be a dummy anyway.)

core/include/traccc/seeding/detail/singlet.hpp
```diff
@@ -30,32 +30,10 @@ inline TRACCC_HOST_DEVICE bool operator!=(const sp_location& lhs,
...
-/// Container of singlet belonging to one detector module
-template <template <typename> class vector_t>
-using singlet_collection = vector_t<sp_location>;
+/// Declare all singlet collection types
+using singlet_collection_types = collection_types<sp_location>;
 
-/// Convenience declaration for the singlet collection type to use in host code
-using host_singlet_collection = singlet_collection<vecmem::vector>;
+/// Declare all singlet container types
+using singlet_container_types = container_types<unsigned int, sp_location>;
```
krasznaa, line 37: Yeah, let's just drop these definitions. If they are not needed, let's not burden the compiler with them.

core/include/traccc/seeding/doublet_finding.hpp
```diff
@@ -18,9 +18,9 @@ namespace traccc {
 
 /// Doublet finding to search the combinations of two compatible spacepoints
 struct doublet_finding
-    : public algorithm<
-          std::pair<host_doublet_collection, host_lin_circle_collection>(
-              const sp_grid&, const sp_location&, const bool&)> {
+    : public algorithm<std::pair<doublet_collection_types::host,
+                                 lin_circle_collection_types::host>(
+          const sp_grid&, const sp_location&, const bool&)> {
```
krasznaa, line 23: Not for this PR, but wow... We really provide a boolean to the algorithm as a constant reference?!? :scream: We should really fix that in a separate PR.

device/common/include/traccc/seeding/device/impl/select_seeds.ipp
```diff
@@ -81,14 +81,14 @@ void select_seeds(
...
-    device_triplet_container triplet_device(triplet_view);
+    triplet_container_types::const_device triplet_device(triplet_view);
     seed_collection_types::device seed_device(seed_view);
 
     // Header of triplet: number of triplets per bin
     // Item of triplet: triplet objects per bin
     const unsigned int num_triplets_per_bin =
         triplet_device.get_headers().at(bin_idx).n_triplets;
-    const vecmem::device_vector<triplet> triplets_per_bin =
+    const vecmem::device_vector<const triplet> triplets_per_bin =
```
krasznaa, line 91: Instead of the explicit type name, which is correct, we should use either

```c++
triplet_collection_types::const_device
```

, or better yet:

```c++
triplet_container_types::item_vector::const_reference
```

(I hope I wrote the correct types. I didn't try to compile this of course.) This way if the underlying types change, as long as the type definitions are correctly defined, this will keep working.

krasznaa (changes requested):
Looks good. Though at one point we should really clean up how we pass arguments to the `traccc::device::` functions. Right now we pass some constant arguments by value, and some by const reference. We should just use one. (I think const reference would be better.)

device/common/include/traccc/seeding/device/impl/select_seeds.ipp
guilhermeAlmeida1: `triplet_container_types::const_device::item_vector::const_reference`

I think this version is a bit overkill and goes against what we're using throughout all our device kernels. Isn't the point of having both the collection & container types declared to also be able to use them together coherently?

guilhermeAlmeida1:
@krasznaa Other than the requested changes, while I was looking at track_params_est I noticed that there we were still using some older version of container classes there, so I changed those as well and added to this PR

device/common/include/traccc/seeding/device/impl/select_seeds.ipp
guilhermeAlmeida1: (so I changed it to `triplet_collection_types::const_device`)

krasznaa: Fine...

core/include/traccc/edm/track_parameters.hpp
```diff
@@ -98,19 +98,25 @@ inline bool operator==(const bound_track_parameters& lhs,
...
+// use
+// /// in host code
+// using host_bound_track_parameters_collection =
+//     bound_track_parameters_collection<vecmem::vector>;
+
+// /// Convenient declaration for the bound track parameters collection type to
+// use
+// /// in device code
+// using device_bound_track_parameters_collection =
+//     bound_track_parameters_collection<vecmem::device_vector>;
```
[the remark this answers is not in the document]
guilhermeAlmeida1: whoops, missed this

krasznaa (approved):
Yeah, let's do it.

PR #250, opened 2022-10-12
Added common implementation of track parameters estimation

guilhermeAlmeida1:
Similarly to #221  , a common implementation of the track parameters estimation was created, now being used by both SYCL and CUDA backends.

No significant differences in performance were observed.

### DIFF
--- a/device/common/include/traccc/seeding/device/impl/estimate_track_params.ipp
+++ b/device/common/include/traccc/seeding/device/impl/estimate_track_params.ipp
@@ -1,0 +1,45 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/seeding/device/estimate_track_params.hpp"
+#include "traccc/seeding/track_params_estimation_helper.hpp"
+
+namespace traccc::device {
+
+TRACCC_HOST_DEVICE
+inline void estimate_track_params(
+    const std::size_t globalIndex,
+    const spacepoint_container_types::const_view& spacepoints_view,
+    const seed_collection_types::const_view& seeds_view,
+    bound_track_parameters_collection_types::view params_view) {
+
+    // Check if anything needs to be done.
+    const seed_collection_types::const_device seeds_device(seeds_view);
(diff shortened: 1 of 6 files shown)


guilhermeAlmeida1:
Currently a draft as it depends on #249

krasznaa:
Note that once an MR is approved, I think you should be able to merge it yourself. :wink: Though I'm not 100% sure. (You should definitely check this the next time.)

guilhermeAlmeida1:
> Note that once an MR is approved, I think you should be able to merge it yourself. 😉 Though I'm not 100% sure. (You should definitely check this the next time.)

I think I need write permissions for this, which I currently don't have

device/cuda/src/seeding/track_params_estimation.cu
```diff
@@ -69,7 +82,7 @@ track_params_estimation::operator()(
     unsigned int num_blocks = seeds_size / num_threads + 1;
 
     // run the kernel
-    track_params_estimating_kernel<<<num_blocks, num_threads>>>(
+    kernels::estimate_track_params<<<num_blocks, num_threads>>>(
```
krasznaa, line 85: Please remove the declaration of `traccc::cuda::track_params_estimating_kernel` from line 37. :stuck_out_tongue:

device/cuda/src/seeding/track_params_estimation.cu
```diff
@@ -69,7 +82,7 @@ track_params_estimation::operator()(
     unsigned int num_blocks = seeds_size / num_threads + 1;
```
krasznaa, line 82: You should fix this calculation. This is not the right way of calculating the number of blocks.

device/common/include/traccc/seeding/device/impl/estimate_track_params.ipp
```diff
@@ -0,0 +1,45 @@
...
+#pragma once
+
+// Project include(s).
+#include "traccc/seeding/device/estimate_track_params.hpp"
+#include "traccc/seeding/track_params_estimation_helper.hpp"
+
+namespace traccc::device {
+
+TRACCC_HOST_DEVICE
+void estimate_track_params(
```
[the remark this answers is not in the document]
guilhermeAlmeida1: I was planning on changing all the kernels to inline in a PR which was going to follow this one, just so they'd all go together, but might as well put this one in now

krasznaa (approved):
Looks good to me.

PR #251, opened 2022-10-12
Made all single source kernels inline

guilhermeAlmeida1:
Our current status of single source files with the header file include the implementation can easily lead to compile-time multiple definition errors if a header file is simultaneously included in different places. For this reason, we changed all of these to inline which prevents such error.


beomki-yeo (approved):
Yeah I can't agree anymore

PR #252, opened 2022-10-14
Changed doublets & triplets data structure to spM based jagged vectors

guilhermeAlmeida1:
Our current implementation of find_doublets (and count/find triplets) is currently based on storing the resulting information in jagged vectors based on each subvector sharing the same geometrical bin for it's spM. 
This, however, results in having poor reutilization of calculations already done in previous kernels. e.g. [the find_doublets kernel starts with finding the position in which the current work item can be positioned in the result object.](https://github.com/acts-project/traccc/blob/7143f0a8710eca0695f39f2899e14d2af54b17cf/device/common/include/traccc/seeding/device/impl/find_doublets.ipp#L81)

This PR implements an alternative structure where jagged vectors are based on each subvector sharing the same spM (so we get a more subvectors which are a smaller, but the same ammount of total items). This helps reusing previously calculated information and reduces the ammount of calculations needed in each device kernel.

The same logic can be applied to the count / find triplets kernels. As these use the same geometrical bin based jagged structure which then once again is inneficient especially when looking further downstream at select_seeds which once again bases it's calculations on triplets which share the same spM. As such, I intend on making similar changes to count/find_triplets in a following PR.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -60,6 +60,8 @@ traccc_add_library( traccc_core core TYPE SHARED
   "include/traccc/seeding/detail/lin_circle.hpp"
   "include/traccc/seeding/detail/doublet.hpp"
   "include/traccc/seeding/detail/triplet.hpp"
+  "include/traccc/seeding/detail/doublet_spM.hpp"
+  "include/traccc/seeding/detail/triplet_spM.hpp"
   "include/traccc/seeding/detail/singlet.hpp"
   "include/traccc/seeding/detail/seeding_config.hpp"
   "include/traccc/seeding/detail/spacepoint_grid.hpp"
--- a/core/include/traccc/seeding/detail/doublet_spM.hpp
+++ b/core/include/traccc/seeding/detail/doublet_spM.hpp
@@ -1,0 +1,44 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2023 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include "traccc/seeding/detail/singlet.hpp"
+
+namespace traccc {
+
+/// Header: the number of doublets per spacepoint bin
+struct doublets_per_spM {
+
+    /// Index of the middle spacepoint.
+    sp_location spM;
+
+    /// The number of compatible middle-bottom/top doublets for the middle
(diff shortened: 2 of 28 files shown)


guilhermeAlmeida1:
As a follow up to what I mentioned in today's parallelization meeting. The current implementation is taking an insanely long ammount of time on the host size running the `traccc::device::make_doublet_spM_buffers(...)` function. I currently have no great idea on how to get around this issue

device/common/src/seeding/make_doublet_spM_buffers.cpp
```diff
@@ -0,0 +1,70 @@
...
+    vecmem::copy& copy, vecmem::memory_resource& mr,
+    vecmem::memory_resource* mr_host) {
+
+    // Get the number of doublets per geometric bin.
+    vecmem::vector<device::doublet_counter_header> doublet_counts(
+        (mr_host != nullptr) ? mr_host : &mr);
+    copy(doublet_counter.headers, doublet_counts);
+    vecmem::jagged_vector<device::doublet_counter> doublet_counts_items(
+        (mr_host != nullptr) ? mr_host : &mr);
+    copy(doublet_counter.items, doublet_counts_items);
```
krasznaa, line 31: This is the expensive bit. Copying the jagged vector back to the host is expensive here.

You can speed things up by specifying explicitly that you're doing a `vecmem::copy::type::device_to_host` copy. But we'll need to discuss how the logic here should happen. As doing this copy is just a bad idea in the first place... :frowning:

guilhermeAlmeida1: Yeah, I'm aware this is the major issue with the current implementation, but I can't quite figure out a way to make it work without this

guilhermeAlmeida1: Just for the record as I will mark this as resolved, I took Attila's suggestion of changing the the whole data structure of not only the previous kernel (count doublets) as well as the downstream kernels to spM based jagged vectors. I will now do some performance testing and cleaning up of code to check wether it is worth merging this in.

guilhermeAlmeida1:
I decided to get back to this and have been testing it with far from ideal results.
The kernels themselves become clearly quicker from having spM based jagged vectors, but this comes at the cost of having to go through more computations on the CPU between kernels, resulting in large gaps between kernels.
Closing this PR as I don't really see way of going forward with this.

PR #254, opened 2022-10-17
Update detray version (v0.11.0 -> v0.12.0) and replace track parameter classes

beomki-yeo:
This PR updates the detray version and removes traccc track parameter classes to utilize detray track parameter classes


device/common/include/traccc/seeding/device/impl/estimate_track_params.ipp
```diff
@@ -35,11 +35,12 @@ inline void estimate_track_params(
     vector3 bfield = {0, 0, 2};
 
     const seed& this_seed = seeds_device.at(globalIndex);
-    bound_vector& param = params_device[globalIndex].vector();
+    // bound_vector param = params_device[globalIndex].vector();
```
krasznaa, line 38: I guess this line could be removed.

krasznaa (commented):
It looks good, we just need to update it to the latest `main` version...

device/common/include/traccc/seeding/device/impl/estimate_track_params.ipp
beomki-yeo: Yes

krasznaa (approved):
Let's do it! :smile:

PR #256, opened 2022-10-20
Changed seeding classes to include previously known positions of doublets & triplets

guilhermeAlmeida1:
The current implementation of find doublets, count triplets, and find triplets contain a for loop in each which finds the positions of  doublets to be used in those kernels. In this PR, these are eliminated by including this information in the items created in previous kernels, to then be reused.

However, despite the apparent reduced workload in these kernels, their performance seems to remain essencially the same.

### DIFF
--- a/core/include/traccc/seeding/detail/doublet.hpp
+++ b/core/include/traccc/seeding/detail/doublet.hpp
@@ -22,6 +22,10 @@ struct doublet {
     sp_location sp1;
     // bottom (or top) spacepoint location in internal spacepoint container
     sp_location sp2;
+
+    // Position of the mid top doublets for this which share this spM
+    unsigned int m_mt_start_idx = 0;
+    unsigned int m_mt_end_idx = 0;
 };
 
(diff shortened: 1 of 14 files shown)


guilhermeAlmeida1:
This is somewhat of an attempt at using some of the concepts of #252 (which, so far, is showing poor performance) in a way which doesn't change as much of our current code

guilhermeAlmeida1:
Now added same logic for triplets. This significantly boosted the speed of update triplet weights (some 10x faster). Unfortunately, this same logic can't currently be used for select_seeds as there we go back to using all spM sharing triplets instead of sharing both spM and spB. That's where the approach in #252 performs better than this one.

beomki-yeo:
This PR and #252 are compatible, or #252 will be closed once this PR is accepted?

guilhermeAlmeida1:
@beomki-yeo a part of this PR will become redundant and not necessary if we merge in #252, the one for the doublets, but the changes to triplets will remain useful

beomki-yeo (approved):
OK let's merge this in!

PR #258, opened 2022-10-26
Improved count/find doublets performance

guilhermeAlmeida1:
* Changed doublet finding compatibility checking function to always run the same ammount of if statements. This reduces divergence in between threads improving performance.

* Removed repeated usage of .size() of resizable vector which was very expensive.

This resulted in an improvement of ~8% in count doublets and ~20% in find doublets.

### DIFF
--- a/core/include/traccc/seeding/doublet_finding_helper.hpp
+++ b/core/include/traccc/seeding/doublet_finding_helper.hpp
@@ -45,46 +45,29 @@ bool doublet_finding_helper::isCompatible(
     const internal_spacepoint<spacepoint>& sp1,
     const internal_spacepoint<spacepoint>& sp2, const seedfinder_config& config,
     bool bottom) {
-    bool result = true;
+
     if (bottom) {
         scalar deltaR = sp1.radius() - sp2.radius();
-        // if r-distance is too big, try next SP in bin
-        if (deltaR > config.deltaRMax) {
-            result = false;
+        scalar cotTheta = sp1.z() - sp2.z();
+        scalar zOrigin = sp1.z() * deltaR - sp1.radius() * cotTheta;
+        if (deltaR > config.deltaRMax || deltaR < config.deltaRMin ||
+            std::fabs(cotTheta) > config.cotThetaMax * deltaR ||
+            zOrigin < config.collisionRegionMin * deltaR ||
+            zOrigin > config.collisionRegionMax * deltaR) {
+            return false;
         }
-        // if r-distance is too small, continue because bins are NOT r-sorted
-        if (deltaR < config.deltaRMin) {
-            result = false;
-        }
-        scalar cotTheta = (sp1.z() - sp2.z()) / deltaR;
-        if (std::fabs(cotTheta) > config.cotThetaMax) {
-            result = false;
-        }
-        scalar zOrigin = sp1.z() - sp1.radius() * cotTheta;
-        if (zOrigin < config.collisionRegionMin ||
-            zOrigin > config.collisionRegionMax) {
-            result = false;
-        }
-    } else if (!bottom) {
+    } else {
         scalar deltaR = sp2.radius() - sp1.radius();
-        if (deltaR > config.deltaRMax) {
-            result = false;
-        }
-        // if r-distance is too small, continue because bins are NOT r-sorted
-        if (deltaR < config.deltaRMin) {
-            result = false;
-        }
-        scalar cotTheta = (sp2.z() - sp1.z()) / deltaR;
-        if (std::fabs(cotTheta) > config.cotThetaMax) {
-            result = false;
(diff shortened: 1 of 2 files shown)


core/include/traccc/seeding/doublet_finding_helper.hpp
```diff
@@ -45,45 +45,46 @@ bool doublet_finding_helper::isCompatible(
...
-            return false;
+            result = false;
         }
         scalar zOrigin = sp1.z() - sp1.radius() * cotTheta;
         if (zOrigin < config.collisionRegionMin ||
             zOrigin > config.collisionRegionMax) {
-            return false;
+            result = false;
         }
     } else if (!bottom) {
```
krasznaa, line 68: This is a very weird statement here.

I know that this is not something that you were editing here, but should we not just remove that second explicit check on the value of `bottom`? :thinking:

krasznaa (commented):
The change in `find_doublets.ipp` is a no-brainer.

I wonder though if we should do things a bit differently in `doublet_finding_helper.hpp`. :thinking: This implementation with a fixed number of if statements is definitely more efficient on the GPU. But I fear that it hurts us on the host. :frowning:

What happens if you construct one (or maybe two including `if(bottom)...`) huge `if` statement? Can it even be done? :thinking:

But in principle, if that's not appropriate, then we should just go with this...

core/include/traccc/seeding/doublet_finding_helper.hpp
```diff
@@ -45,45 +45,46 @@ bool doublet_finding_helper::isCompatible(
...
         if (deltaR > config.deltaRMax) {
-            return false;
+            result = false;
         }
         // if r-distance is too small, continue because bins are NOT r-sorted
         if (deltaR < config.deltaRMin) {
-            return false;
+            result = false;
         }
         scalar cotTheta = (sp1.z() - sp2.z()) / deltaR;
```
guilhermeAlmeida1, line 59: I was just thinking that we might have an issue here. While before if deltaR were 0 we would never get here, now there's theoretically the possibility of doing a division by 0. But adding an additional 
`if ( deltaR == 0) { return 0; }`

is probably not that good as well.

guilhermeAlmeida1: I just realized there's an easy fix, just multiply the other term instead of dividing this one.

core/include/traccc/seeding/doublet_finding_helper.hpp
guilhermeAlmeida1: This is definately a worthwhile which improved performance significantly. Performance improvement went from 8/20% to 44/31% for count/find doublets.

guilhermeAlmeida1:
> What happens if you construct one huge `if` statement?

It's definately possible and didn't seem to make any changes to performance on the GPU. I've not really been looking into how it affects the CPU.

beomki-yeo:
It's interesting that changes in `doublet_finding_helper.hpp` make performance better. I doubted that removing divergence there would not reduce the workloads. Maybe variable caching gets more efficient?

guilhermeAlmeida1:
Accumulating all of the if statements into a single large one seems to have preserved the performance boost on the GPU, while not harming our CPU performance (while the original version of just running all ifs was having a bad impact on CPU perf.)

I also changed some conditions to compare using multiplications instead of divisions to ensure division by zero cannot happen.

krasznaa (approved):
Interesting. :thinking: It's very good that you don't see any host performance drop with these changes. So let's push things in like this.

But I'm actually a bit surprised that there's no visible performance change. :confused: Since in this setup the host code still needs to do a few more operations than it did before. That is probably offset by your very good change of turning divisions into multiplications. :thinking:

One final comment: It is not for this PR, but if we really want to get the maximum performance out of this... The check on the value of the `bottom` variable should not be happening at runtime. In all the cases we know at compile-time what the value of `bottom` is for a given call. (We use "hardcoded" `true` and `false` statements in the call to this helper function.) So let's express it as a compile-time decision! :wink:

You can find an example here for how we did this in the old code: https://github.com/acts-project/acts/blob/main/Plugins/Sycl/src/Seeding/DupletSearch.hpp#L78-L84 (We used some templating. But if you can force the compile-time evaluation of this `if` statement in some other way, I'm open to suggestions.)

guilhermeAlmeida1:
> The check on the value of the `bottom` variable should not be happening at runtime. 

I like using templating to avoid this first if statement, but I wonder if it isn't a bit more simple to just separate this into 2 functions is CompatibleBottom & isCompatibleTop ?

krasznaa:
That's of course possible, but the two functions only differ a bit. They could share more code with templating. But I'm not too strong minded about the details in the end.

PR #259, opened 2022-10-26
Container Comparison Update, main branch (2022.10.26.)

krasznaa:
This one needs a bit of an explanation...

The whole development was triggered by 2 things:
  - @beomki-yeo's PR from this morning (#254) had an unfortunate side-effect. It resulted in the track parameter matching rate going to `0` in our example applications. (Because we removed the specialised comparison operator for `traccc::bound_track_parameters`.)
  - A report this morning from some CMS students on Intel GPUs performing less precisely in their tests than NVIDIA ones.

I especially wanted to follow up on the latter, as I incorrectly told them in that meeting that we don't see the thing that they reported. Just to realise how wrong I was, right after the meeting. :frowning: Since the iGPU that we have at CERN performs quite a bit worse than the NVIDIA GPU. :thinking:

The code in this PR allows me to quantify these differences like the following:
  - On our NVIDIA GPU this PR's code produces the following output:

```
 > ./bin/traccc_seq_example_sycl --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --input_directory=tml_full/ttbar_mu200/ --run_cpu=1 --events=5
Running ./bin/traccc_seq_example_sycl tml_detector/trackml-detector.csv tml_full/ttbar_mu200/ 5
Running Seeding on device: NVIDIA GeForce RTX 2060
event 0
 number of seeds (cpu): 15424
 number of seeds (sycl): 15424
 measurements matching rate: 0.999557
 spacepoint matching rate: 0.96945
Number of spacepoints: 92504 (host), 92504 (device)
  Matching rate(s):
    - 76.3945% at 0.001% uncertainty
    - 99.3114% at 0.1% uncertainty
    - 99.3135% at 1% uncertainty
    - 99.3135% at 5% uncertainty
 seed matching rate: 0.95883
 track parameters matching rate: 0
...
```

  - Our iGPU in the meanwhile does this:

```
 > SYCL_DEVICE_FILTER=level_zero ./bin/traccc_seq_example_sycl --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --input_directory=tml_full/ttbar_mu200/ --run_cpu=1 --events=5
Running ./bin/traccc_seq_example_sycl tml_detector/trackml-detector.csv tml_full/ttbar_mu200/ 5
Running Seeding on device: Intel(R) UHD Graphics 630 [0x3e98]
event 0
 number of seeds (cpu): 15424
 number of seeds (sycl): 15424
 measurements matching rate: 0.886783
 spacepoint matching rate: 0.796398
Number of spacepoints: 92504 (host), 92504 (device)
  Matching rate(s):
    - 62.2557% at 0.001% uncertainty
    - 84.8796% at 0.1% uncertainty
    - 89.3442% at 1% uncertainty
    - 93.5981% at 5% uncertainty
 seed matching rate: 0.732171
 track parameters matching rate: 0
...
```

The slightly surprising thing is that I didn't observe the same difference between using the Level-0 and OpenCL driver that the CMS student reported. :confused:

```
 > SYCL_DEVICE_FILTER=opencl ./bin/traccc_seq_example_sycl --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --input_directory=tml_full/ttbar_mu200/ --run_cpu=1 --events=5
Running ./bin/traccc_seq_example_sycl tml_detector/trackml-detector.csv tml_full/ttbar_mu200/ 5
Running Seeding on device: Intel(R) UHD Graphics 630 [0x3e98]
event 0
 number of seeds (cpu): 15424
 number of seeds (sycl): 15424
 measurements matching rate: 0.886589
 spacepoint matching rate: 0.796117
Number of spacepoints: 92504 (host), 92504 (device)
  Matching rate(s):
    - 62.247% at 0.001% uncertainty
    - 84.8677% at 0.1% uncertainty
    - 89.3486% at 1% uncertainty
    - 93.6057% at 5% uncertainty
 seed matching rate: 0.73042
 track parameters matching rate: 0
...
```

Our code performs equally badly with both on the iGPU. :stuck_out_tongue:

Before going further with this PR, I wanted to get some feedback. Are you guys on board with quantifying the performance of our device algorithms like this?

P.S. The `is_same_angle` function is not actually used just yet. But I suspect that at one point it could become useful. Also, eventually I would of course put a **lot** of unit tests around all this. But I wanted to get some feedback before putting effort into that.


beomki-yeo:
I definitely like this!

krasznaa:
I forgot to mention: I moved the existing `traccc::performance` sources into a top-level directory in the repository, and added my new code into that library. (Instead of adding it to `traccc::core` as I did at first.) Since I plan to add some "centralised code" for time measurements soon, that "reformed" library seemed like a good place to put that code eventually as well.

performance/src/performance/details/is_same_angle.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+}  // namespace
+
+namespace traccc::details {
+
+bool is_same_angle(scalar lhs, scalar rhs, scalar unc) {
+
+    return (std::abs(wrap_to_pi(lhs - rhs)) <=
+            (unc *
+             ((std::abs(wrap_to_pi(lhs)) + std::abs(wrap_to_pi(rhs))) / 2.f)));
```
guilhermeAlmeida1, line 40: I think this should be
`wrap_to_pi( std::abs(lhs) + std::abs(rhs) ) / 2`

performance/include/traccc/performance/details/is_same_object.hpp
```diff
@@ -0,0 +1,51 @@
...
+// Project include(s).
+#include "traccc/definitions/common.hpp"
+#include "traccc/definitions/primitives.hpp"
+
+// System include(s).
+#include <functional>
+
+namespace traccc::details {
+
+/// Generic unary functor for checking equality with a certain object
```
guilhermeAlmeida1, line 19: *unitary

performance/include/traccc/performance/details/is_same_object.hpp
```diff
@@ -0,0 +1,51 @@
...
+/// equality operator, and to specialise it for cases where the uncertainty
+/// of the equality can be set by the user. (To account for floating point
+/// uncertainties in the algorithms.)
+///
+template <typename T>
+class is_same_object {
+
+    public:
+    /// Constructor with a reference object, and an allowed uncertainty
+    is_same_object(const T& ref, scalar /*unc*/ = float_epsilon);
```
guilhermeAlmeida1, line 31: this commented part slipped in

performance/src/performance/details/is_same_object.cpp
```diff
@@ -0,0 +1,100 @@
...
+    const bound_track_parameters& ref, scalar unc)
+    : m_ref(ref), m_unc(unc) {}
+
+bool is_same_object<bound_track_parameters>::operator()(
+    const bound_track_parameters& obj) const {
+
+    return (is_same_scalar(obj.local()[0], m_ref.get().local()[0], m_unc) &&
+            is_same_scalar(obj.local()[1], m_ref.get().local()[1], m_unc) &&
+            is_same_angle(obj.phi(), m_ref.get().phi(), m_unc) &&
+            is_same_scalar(obj.theta(), m_ref.get().theta(), m_unc) &&
```
guilhermeAlmeida1, line 93: Shouldn't theta also be treated as an angle?

beomki-yeo: I think he treated it as scalar because phi is in [-pi, pi] but theta is in [0,pi].
I rather wonder if there would be case where phi is defined out of the [-pi,pi] range

performance/src/performance/details/is_same_object.cpp
```diff
@@ -0,0 +1,100 @@
...
+/// @{
+
+is_same_object<bound_track_parameters>::is_same_object(
+    const bound_track_parameters& ref, scalar unc)
+    : m_ref(ref), m_unc(unc) {}
+
+bool is_same_object<bound_track_parameters>::operator()(
+    const bound_track_parameters& obj) const {
+
+    return (is_same_scalar(obj.local()[0], m_ref.get().local()[0], m_unc) &&
```
beomki-yeo, line 90: Can you also compare the output of `surface_link()` for bound track parameters

performance/src/performance/details/is_same_object.cpp
krasznaa: Indeed. With $\theta$ we don't "flip around". One extreme is in the direction of one of the particle beams, the other extreme is in the direction of the other particle beam. There's no need to worry about the "two extremes being close to each other".

As for why I added a few more `wrap_to_pi` calls in the code than necessarily required: This is mainly out of precaution. While the ATLAS standard is and has been for a long time to define $\phi$ in the [- $\pi$, $\pi$] range, some experiments rather define in in the [0, 2 $\pi$] range. The code that I add in this PR (stolen from https://gitlab.cern.ch/atlas/athena/-/blob/master/Control/CxxUtils/CxxUtils/phihelper.h) tries to give correct results for any combination of values.

And now that I looked at the ATLAS code again, I see that I probably should've gone with

https://gitlab.cern.ch/atlas/athena/-/blob/master/Control/CxxUtils/CxxUtils/phihelper.h#L48-65

to calculate the averages. (I have to admit I didn't think the formula through too much, but if ATLAS does it like that, we might as well do the same in this project.)

performance/include/traccc/performance/details/is_same_object.hpp
krasznaa: I did this very much on purpose. I believe the ATLAS coding standard even encourages this. (To give a name to unused variables so that they could be referred to in the documentation.)

core/include/traccc/utils/impl/container_comparator.ipp
```diff
@@ -0,0 +1,101 @@
...
+    // Calculate the agreements at various uncertainties.
+    std::vector<scalar> agreements;
+    agreements.reserve(m_uncertainties.size());
+    for (scalar uncertainty : m_uncertainties) {
+        // The number of matched items between the containers.
+        std::size_t matched = 0;
+        // If the two containers differ in size, only compare the first N
+        // "inner vectors", that both of them have.
+        const std::size_t cont_size =
+            std::min(lhs_cont.get_items().size(), rhs_cont.get_items().size());
```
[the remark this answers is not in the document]
krasznaa: Maybe. 🤔 I was thinking about it, but thought that a low "agreement value" would be enough of a warning in that case.

performance/include/traccc/performance/details/is_same_object.hpp
krasznaa: 😛 While this naming practically died out of modern C\+\+ by now, this is how such functors used to be called.

https://en.cppreference.com/w/cpp/utility/functional/unary_function

performance/src/performance/details/is_same_object.cpp
krasznaa: Will do. :+1:

performance/src/performance/details/is_same_angle.cpp
guilhermeAlmeida1: edit: in fact, I don't really understand why the abs is here at all, maybe they should not even be here?

core/include/traccc/utils/impl/container_comparator.ipp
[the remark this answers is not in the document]
guilhermeAlmeida1: I would still be slightly afraid of somehow having creating a jagged vector with bigger size than what is actually filled in, and then having a bunch of garbage vectors at the end. Then there would be a correct agreement

performance/src/performance/details/is_same_angle.cpp
krasznaa: I'll replace this with the "Athena recipe", but to explain myself: I was afraid of a value just below $\pi$ and another just above $-\pi$ averaging to around zero. Which is not what we want. (We want them to average to around $\pi$. With whatever sign.)

But yeah, I'll just go with the "Athena formula".

performance/src/performance/details/is_same_object.cpp
krasznaa: This was eye-opening. As it turns out, it's exactly this comparison that made the `==` operator defined in Detray report none of the host and device objects matching each other.

You see, the surface link variable is not set anywhere in the traccc code right now. So far, so good. But while in the host algorithm this means that the default constructor would explicitly set the link to an invalid value, in device memory we don't call on the constructor to initialise any of the variables. So this variable was undefined in the output of the device algorithm. In practice being 0 with CUDA.

I set both the host and the device code to set the surface link to `0`. But now I'm tinking that that may have been a silly solution. Let me see if I can make it better on the device side...

krasznaa: Right... there we go. This check made the device code quite a bit more robust. :wink:

core/include/traccc/utils/impl/container_comparator.ipp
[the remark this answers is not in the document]
krasznaa: I changed the formulas to be a bit more conservative. Now the agreement should definitely be low when one of the containers has (some) garbage in it.

performance/src/performance/details/is_same_scalar.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <cmath>
+
+namespace traccc::details {
+
+bool is_same_scalar(scalar lhs, scalar rhs, scalar unc) {
+
+    // The difference of the two values is meant to be smaller than
+    // their average times the uncertainty.
+    return (std::abs(lhs - rhs) <=
+            (unc * ((std::abs(lhs) + std::abs(rhs)) / 2.f)));
```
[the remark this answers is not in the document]
krasznaa: I do not believe this is relevant here. We're not trying to propagate uncertainties on values. (Which is what your formula is for.) For these comparisons just the differences in values (the absolute values of the differences to be exact) should be appropriate.

performance/src/performance/details/is_same_object.cpp
beomki-yeo: Oh I didn't notice that surface link was not assigned in track parameter estimation.
Assigning it to `geometry_id` (header of spacepoint) will do the job correctly.

beomki-yeo (approved):

PR #260, opened 2022-10-27
Moved device syncronisation statements to as late as possible

guilhermeAlmeida1:
This one's a simple one, moving device synchronisations to a later stage allows for simultaneous computations on the host while the device is busy.

The impact on performance is nothing major, but visible.


krasznaa (approved):
Sounds good. Hand to heart that you checked carefully that the data used after the kernel launches does not depend on these kernels, right? (I didn't look carefully myself...)

guilhermeAlmeida1:
I just double checked, and there's no dependency @krasznaa .
However, something to take into account is that with this change we will be launching the creation of prefix sums without waiting for the filling of doublets/triplets because currently those are created with fixed size, and we're only using the get_size() function, not the actual contents. However, if at a later stage we revert back to using push_backs we would also have to revert this.

krasznaa:
I see. Yeah, we need to keep that in mind...

PR #261, opened 2022-10-27
Memory Resource Cleanup (Part 1), main branch (2022.10.26.)

krasznaa:
This is the beginning of a set of updates, which I should've looked at long ago. Having the end goal of only using `vecmem::host_memory_resource` and `vecmem::<type>::device_memory_resource` for memory allocations in all our (main) executables.

As a first step, made `traccc::sycl::clusterization_algorithm` expect device memory as input. Very sneakily the clusterization code so far was running off of host memory. Which technically worked as long as the host memory was allocated with `vecmem::sycl::host_memory_resource`. It "just" provided a significant overhead for the runtimes of those kernels.

For now let's keep it as a draft, as I'll also want to clean up what we do in the track parameter estimation algorithm(s).

@guilhermeAlmeida1, I believe this is the issue that you were running into recently. :wink:

### DIFF
--- a/device/cuda/include/traccc/cuda/clusterization/clusterization_algorithm.hpp
+++ b/device/cuda/include/traccc/cuda/clusterization/clusterization_algorithm.hpp
@@ -9,24 +9,21 @@
 
 // Project include(s).
 #include "traccc/edm/cell.hpp"
-#include "traccc/edm/cluster.hpp"
-#include "traccc/edm/measurement.hpp"
 #include "traccc/edm/spacepoint.hpp"
 #include "traccc/utils/algorithm.hpp"
 #include "traccc/utils/memory_resource.hpp"
 
 // VecMem include(s).
-#include <vecmem/memory/memory_resource.hpp>
 #include <vecmem/utils/copy.hpp>
 
-//// Traccc library include(s).
-//#include "traccc/utils/memory_resource.hpp"
+// System include(s).
+#include <memory>
 
 namespace traccc::cuda {
 
 class clusterization_algorithm
     : public algorithm<spacepoint_container_types::buffer(
-          const cell_container_types::host&)> {
+          const cell_container_types::const_view&)> {
 
     public:
     /// Constructor for clusterization algorithm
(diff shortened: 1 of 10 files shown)


beomki-yeo:
> vecmem::host_memory_resource and vecmem::<type>::device_memory_resource for memory allocations in all our (main) executables.

Since when has this been possible?

krasznaa:
😕 This has worked since always.

There's a very widespread misconception that H->D memory copies need to start from "device/language specific" host memory allocations. But that's really not true.

What `vecmem::cuda::host_memory_resource`, `vecmem::hip::host_memory_resource` and `vecmem::sycl::host_memory_resource` do are the following:
  - They allow asynchronous copies to device memory without any intervention necessary from the CPU. But since all our operations in this project happen synchronously, that's irrelevant (for now).
  - They allow their memories to be accessed by kernels running on the devices. But of course accessing host memory from a device kernel is very-very slow.

We have a couple of mistakes in our applications that have been masked by using these device/language specific host memory resources. But these are hurting the performance of the examples.

beomki-yeo:
Thanks for the explanation. I learned something new :p

krasznaa:
Okay, I de-scoped this PR a bit. For now I only bothered with the "input side" of the algorithm sequences, and left the track parameter estimation algorithms alone.

Once #259 goes in it will be a lot easier to update `traccc::cuda::track_params_estimation` and `traccc::sycl::track_params_estimation` not to copy its results back to the host all the time.

So for now let's just get this in like this. :wink:

krasznaa:
It's worth noting that this update seemingly slows down clusterization. :facepalm:

In the new code the H->D copy of the cells is counted as part of the device clusterization time. Which is not unfair actually.

But this takes a very measurable amount of time.

image

Previously there was no such actual copy, it was the CSV file reading that was slower because of using the CUDA/SYCL specific host memory resources. The overall execution time of the applications did become shorter. But the time measured by the applications for "clusterization" went up with these changes. Just something to keep in mind...

krasznaa:
Could we push this in? Somebody would need to bless this...

stephenswat (approved):

PR #262, opened 2022-10-28
Add measurement reader and writer

beomki-yeo:
This PR adds read and write functions for measurement data file.
The measurement file should not be used for the full track reconstruction pipeline but it will be useful in truth tracking

One thing to note is that I only assigned `geometry_id` in `cell_module` object because `transform3` object should not be defined in the EDM objects. Instead,  `transform3` (along with corresponding the coordinate type) should be found in detray detector object using `geometry_id`.


krasznaa (approved):
Looks good. After a rebase it should be good to go.

At one point we should move the implementations of these functions into `.cpp` files. My OCD is screaming with all these non-templated functions sitting in the headers. 😛

krasznaa (approved):
:+1:

PR #263, opened 2022-10-31
Changed doublet finding helper functions to preprocessing top/bottom conditionals

guilhermeAlmeida1:


core/include/traccc/seeding/doublet_finding_helper.hpp
```diff
@@ -101,7 +111,12 @@ lin_circle doublet_finding_helper::transform_coordinates(
...
     //
-    int bottomFactor = 1 * (int(!bottom)) - 1 * (int(bottom));
+    int bottomFactor;
+    if constexpr (otherSpType == details::spacepoint_type::bottom) {
+        bottomFactor = -1;
+    } else {
+        bottomFactor = 1;
+    }
     // cot_theta = (deltaZ/deltaR)
     scalar cot_theta = deltaZ * iDeltaR * bottomFactor;
```
krasznaa, line 121: I'm not a big fan of this setup... It made sense with the flag arriving at runtime. But if we do it at compilation time, I think this would make more sense:

```c++
scalar cot_theta = deltaZ * iDeltaR;
if constexpr (otherSpType == details::spacepoint_type::bottom) {
    cot_theta = -cot_theta;
}
```

core/include/traccc/seeding/doublet_finding.hpp
```diff
@@ -11,35 +11,40 @@
...
+          const sp_grid&, const sp_location&)> {
 
     /// Constructor for the doublet finding
     ///
     /// @param seedfinder_config is the configuration parameters
     /// @param isp_container is the internal spacepoint container
-    doublet_finding(const seedfinder_config& config) : m_config(config) {}
+    doublet_finding(const seedfinder_config& config) : m_config(config) {
+        static_assert(otherSpType == details::spacepoint_type::bottom ||
+                      otherSpType == details::spacepoint_type::top);
```
krasznaa, line 34: At this point I'm just using this PR as a teaching tool, this change is not really all that important: In this case the struct itself should never be instantiated with `traccc::details::spacepoint_type::middle`. So the `static_assert(...)` should also be put at the "struct level". Like we do here for instance:

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/data/jagged_vector_buffer.hpp#L44-L52

krasznaa (changes requested):
Other than my very petty comment, this is all good for me already.

core/include/traccc/seeding/doublet_finding.hpp
krasznaa: What I mean is that it should **only** be put at the "struct level". (Re-reading now I see that I wrote it in a very misunderstand-able way.)

guilhermeAlmeida1: I see, I wasn't aware this was possible.

guilhermeAlmeida1:
Should be good to go now

krasznaa (approved):
The final lesson with this PR is that if you'll want to write code in the future that would be a bit more compatible with earlier C\+\+ standards, you should use this formalism of `static_assert(...)`:

```c++
static_assert(i == 0, "The variable i must be zero");
```

As the version without an error message only became available with C\+\+17.

https://en.cppreference.com/w/cpp/language/static_assert

In this project we require C\+\+17, so your code is absolutely good. It's just something that you should know for the future...

PR #264, opened 2022-11-02
Kalman Fitter on CPU

beomki-yeo:
This is a (very preliminary) PR for Kalman filtering on CPU. The detray pull requests also should be merged before we accept this PR.

Let me explain one by one as the PR is pretty big...

### EDM
There are two EDM classes added. `track_candidate`, which takes geometry link and measurement object, is an input to kalman filtering and potentially an output of combinatorial kalman filtering. `track_state` is the output of kalman filtering, which takes the fitted track parameters, measurement and geometry link.

### Algorithm structure
There is a `fitting_algorithm` class which runs the fitting algorithm for `track_candidate` container input (Doesn't have to be for single event). `kalman_fitting`, which is wrapped by `fitting_algorithm`, runs the fitting algorithm for a single track. It runs `detray::propagator` inside a given geometry which triggers `actor_chain` for every surface intersection.

The actor chain is made of the following actors: (1) parameter transporter -> (2) material interactor -> (3) kalman actor -> (4) parameter resetter. 
(1) is for calculating the jacobian from the last surface to the current surface
(2) is for calculating the effect of energy loss and multiple scattering
(3) is for kalman filtering
(4) is for resetting the track parameters for the next surface intersection

(1), (2), (4) are already defined in detray so people can check them out by themselves.

### Kalman filtering (kalman actor)
The updating or filtering of the predicted track parameters is done by the kalman actor  The algorithm is pretty much the same with one implemented in ACTS main repository. Techincally, eq. 7 - 8b of [R.Frühwirth](https://www.sciencedirect.com/science/article/pii/0168900287908874) is implemented in `gain_matrix_updater`

### Kalman smoothing
Given that kalman filtering is an iterative algorithm on the time series of measurements, the filtered state at the last surface contains all information from the previous surfaces while the first measurement only contains the information of its own surface. Since we are mostly interested in the fitted state of the first surface, we need to refine the fitted parameters using the smoothing algorithm. There are two ways for smoothing: (1) For a smoothed parameter of n-th surface, remove the process noise using the predicted and smoothed parameters of the n+1-th surface. Since the smoothed parameters of the last surface is equal to the filtered parameter, the algorithm should iterate backwardly (This is what implemented in ACTS)  (See Eq. 3.36 - 3.37 of [this book](https://link.springer.com/book/10.1007/978-3-030-65771-0) (2) Do one more kalman filtering backwardly from the last surface to the first surface and take an average of two kalman filtering result. (See Eq. 3.38 - 3.40 of  [this book](https://link.springer.com/book/10.1007/978-3-030-65771-0).)

I tried to implemented (2) method at the begining but ended up with implementing (1), because the backward navigation has some issues right now.

### Test Results:

**Setup**
- Telescope geometry implemented in detray (9 surfaces and 5 cm gap) aligned along z axis. 
- 10000 tracks, p = 1 GeV, theta = pi/6, phi = pi/6
- 50 um measurement resolution for local0 and local1
- The seed was smeared by standard deviation of (x = 1 mm, y = 1mm, z = 1mm, n_x = 0.01, n_y = 0.01, n_z = 0.01, qop = 0.1)
- Precision was **double**

Following figures show the pull distributions of bound track parameters at the first surface. I don't know why d0 and z0 (local0 and local1) show unreasonable results (@asalzburger @paulgessinger Do you guys have any idea ?)

One thing to note is that the fitting with a single precision makes wrong results after the smoothing where chi square gets negative.

### What to do next
1. The kalman filtering in this PR was tested in a very ideal detector setup. In real cases, lots of exceptions will occur during the fitting, which we need to handle very carefully.

2. Haven't tested GPU yet, but I have a strong feeling that we will have a bad performance unless we achieve the thread synchronization over the actor chain triggering.

3. As mentioned above, single precision is not working correctly for smoothing. I think it is because the residual covariance is not always positive definite. The ACTS main repository is handling this with the single value decomposition (SVD) to make the covaraince positive definite. We need to think about implementing SVD in algebra-plugins

4. I made another (very unsatisfactory) event map class for this truth tracking validation. I need to come up with a better way to store the truth information of events.


paulgessinger:
First of all this looks like **fantastic** work! Thanks @beomki-yeo!

To understand why the d0/z0 pulls look so narrow, it might help to look at the residual and the error estimate separately? It might be that e.g. the error is overestimated.

asalzburger:
This is impressive work!

Note that d0 and z0 are the only ones somehow that are estimated directly (by the measurements) and not indirectly (by the trajectory), so, Indeed it looks like the errors are not correctly defined.

It does help to print the residuals/pulls per measurement surface ...

beomki-yeo:
Yeah They are smoothed parameter. There is definitely something wrong because chi square gets bigger with smoothed parameters :/  Let me dig into it

beomki-yeo:
Nahh I was comparing with smeared local position instead of the truth local position. Hopefully that fixes the issue...

beomki-yeo:
PR is ready for review. Let me leave `event_map` as it is for now...

beomki-yeo:
Wow. I just found that KF with cmath plugin doesn't work :/ ..

beomki-yeo:
All bugs are fixed now but I still need to wait for acts-project/detray#360 and acts-project/algebra-plugins#80 to be merged.

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+        // triggered only for sensitive surfaces
+        if (navigation.is_on_module() &&
+            navigation.current()->sf_id == detray::surface_id::e_sensitive) {
+
+            auto& trk_state = actor_state();
+
+            // Abort if the propagator fails to find the next measurement
+            if (navigation.current_object() != trk_state.surface_link()) {
+                printf("Aborting... \n");
```
krasznaa, line 90: Is this debug message meant to stay here?

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+    private:
+    std::unique_ptr<detector_type> m_detector;
+    fitter_info<transform3_type> m_fit_info;
+    vecmem::vector<track_state<transform3_type>> m_track_states;
+
+    typename aborter::state m_aborter_state{};
+    typename transporter::state m_transporter_state{};
+    typename interactor::state m_interactor_state{};
+    typename resetter::state m_resetter_state{};
```
krasznaa, line 196: I don't like this. :frowning: We should not have a state on the fitter object.

Any "state" that is needed for the fitting should be put into its own object, which the client of the fitter would manage in memory. So that multiple clients could call on the same fitter object, each with their own "state objects".

I.e. you should introduce something like a `traccc::kalman_fitter::state` struct, which would have all these variables. And then all the clients would need to pass instances of that state type to the `fit(...)`, `filter(...)` and `smooth(...)` functions.

tests/io/test_event_map.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "traccc/io/event_map2.hpp"
+
+// GTest include(s).
+#include <gtest/gtest.h>
+
+// Test generate_particle_map function
+TEST(event_map2, event_map2) {
+
+    const std::string path =
+        "detray_simulation/telescope/kf_validation/1.000000_GeV_0.000000_phi/";
```
krasznaa, line 18: I'm so very confused... Where should this directory be coming from? :confused:The CI seems to have found it. :confused:

krasznaa (changes requested):
I have some reservations against the current interface of the fitter code. We should really make it MT friendly from the start. Otherwise I fear that it will be even more painful to make it MT friendly later on.

core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
```diff
@@ -0,0 +1,107 @@
...
+        const matrix_type<2, 2>& V = cur_state.measurement_covariance();
+        const matrix_type<2, 1> residual = meas_local - H * smt_vec;
+        const matrix_type<2, 2> R =
+            V - H * smt_cov * matrix_operator().transpose(H);
+        const matrix_type<1, 1> chi2 = matrix_operator().transpose(residual) *
+                                       matrix_operator().inverse(R) * residual;
+
+        cur_state.smoothed_chi2() = matrix_operator().element(chi2, 0, 0);
+
+        return true;
```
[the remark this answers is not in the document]
beomki-yeo: Currently, `void` output type doesn't work with type unrolling in detray [`tuple_container` ](https://github.com/acts-project/detray/blob/main/core/include/detray/core/detail/tuple_container.hpp)

```
/home/beomki/projects/traccc/traccc_build/_deps/detray-src/core/include/detray/core/detail/tuple_container.hpp:177:48: error: compound literal of non-object type 
```

I am not sure the boolean return type would be useful in the future though :p

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
beomki-yeo: Good point. It is a leftover

core/include/traccc/edm/track_candidate.hpp
```diff
@@ -0,0 +1,26 @@
...
+namespace traccc {
+
+/// Track candidate contains the measurement object and its surface link
+struct track_candidate {
+    geometry_id surface_link;
+    measurement meas;
+};
+
+/// Declare a track candidates container type
+using track_candidates_container_types =
```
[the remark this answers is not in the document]
beomki-yeo: Fixed

tests/io/test_event_map.cpp
beomki-yeo: I sneakily added detray simulation data in traccc-data repository

core/include/traccc/fitting/fitting_algorithm.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+            trk_states.push_back(std::move(fitter.get_fitter_info()),
+                                 std::move(fitter.get_track_states()));
+        }
+
+        return trk_states;
+    }
+
+    private:
+    std::unique_ptr<detector_type> m_detector;
```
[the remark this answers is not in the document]
beomki-yeo: Hmm you are right. We should not make a copy here.  
I think using a plain pointer would be better, namely:
`const detector_type* m_detector;`

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
beomki-yeo: I made the changes based on suggestion.
Please note that the kalman fitter state and kalman actor state will be updated again when I implement the GPU kalman fitting.

core/include/traccc/edm/track_candidate.hpp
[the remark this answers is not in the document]
krasznaa: :+1:

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+            // Run kalman smoother
+            mask_store.template call<gain_matrix_smoother<transform3_type>>(
+                surface.mask(), *it, *(it - 1));
+        }
+    }
+
+    private:
+    // Detector object
+    const detector_type* m_detector;
```
krasznaa, line 197: Let's make things fairly technical here. :stuck_out_tongue:
  - This way of holding onto a pointer to the detector is not a bad one. I myself used to prefer this way of coding as well.
  - But what I learned since then is that in modern C\+\+ one should instead choose between the following 2 options instead:
    * If the type (in this case `traccc::kalman_fitter`) is meant to be copy-able, it's slightly more elegant to use the [std::reference_wrapper](https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper) type to hold on to the reference/pointer. That way it is very clear in the code that this is **always** meant to point at something valid. Whereas with a pointer you may need to add some lines of documentation to clarify this.
    * If the type is not meant to be copy-able, then one should just use a reference member variable. I.e.

```c++
const detector_type& m_detector;
```

Since I think for this particular class (and the others that use `detector` as well) copying is a valid operation, [std::reference_wrapper](https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper) could be an elegant thing to make use of.

But as I started, this is definitely not something mandatory. The pointer works functionally exactly the same. :wink:

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
```diff
@@ -0,0 +1,202 @@
...
+    kalman_fitter(const detector_type& det) : m_detector(&det) {}
+
+    /// Kalman fitter state
+    struct state {
+
+        /// State constructor
+        ///
+        /// @param track_states the vector of track states
+        state(vecmem::vector<track_state<transform3_type>>&& track_states)
+            : m_fit_actor_state(std::move(track_states)) {}
```
krasznaa, line 76: As always, I'm a bit suspicious about this move operation. Is it really not meant to be possible to construct a "state object" without destroying a vector of track states in the process? :confused: The "move constructor" could be one good constructor for this type. But I could definitely imagine that a copy-constructor could also be necessary.

Though I definitely do not know how these objects are actually meant to interact with each other. :thinking:

tests/cpu/test_kalman_fitter.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include <vecmem/memory/host_memory_resource.hpp>
+
+// GTest include(s).
+#include <gtest/gtest.h>
+
+// ROOT include(s).
+#include <TF1.h>
+
+// System include(s).
+#include <filesystem>
```
krasznaa, line 34: I don't think this include is needed here.

tests/cpu/test_kalman_fitter.cpp
```diff
@@ -0,0 +1,193 @@
...
+    /********************
+     * Pull value test
+     ********************/
+
+    std::array<std::string, 5> pull_names{"pull_d0", "pull_z0", "pull_phi",
+                                          "pull_theta", "pull_qop"};
+
+    auto f = fit_performance_writer.get_file();
+
+    auto gaus = new TF1("gaus", "gaus", -5, 5);
```
krasznaa, line 165: You shouldn't create this object on the heap. It just goes and leaks away like this.

You could just declare it on the stack with:

```c++
TF1 gaus{"gaus", "gaus", -5, 5};
```

Or you could even make it a member variable of `KalmanFittingTests`...

examples/simulation/simulate_telescope.cpp
```diff
@@ -0,0 +1,128 @@
...
+// detray include(s).
+#include "detray/detectors/create_telescope_detector.hpp"
+#include "detray/simulation/simulator.hpp"
+#include "detray/simulation/track_generators.hpp"
+
+// VecMem include(s).
+#include <vecmem/memory/host_memory_resource.hpp>
+
+// System include(s).
+#include <filesystem>
```
krasznaa, line 24: :confused: Wouldn't you need to link against `-lstdc++fs` to make this work with GCC 9? Is our CI not building this code? Since I thought we used GCC 9 in most of our CI builds. :confused:

I'm generally not the biggest fan of [std::filesystem](https://en.cppreference.com/w/cpp/filesystem) exactly because it used to be very finicky to use in practice. I believe with the latest versions of GCC and Clang it's a lot easier to use. But I still wonder if we should use it in this project over [boost::filesystem](https://www.boost.org/doc/libs/master/libs/filesystem/doc/index.htm). Since the latter is a lot more reliable vis-a-vis compiler versions.

core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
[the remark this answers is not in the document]
krasznaa: If there is a technical reason for using a non-void type here, then sure, let's use it. Though we may want to re-think what we do in the future, if Detray forces us to use nonsensical return types to make its own logic work... :thinking:

krasznaa (changes requested):
Updates are still necessary to make this compatible with the main branch.

But generally I'm a lot happier about the API of the new classes now.

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
beomki-yeo: >  But I could definitely imagine that a copy-constructor could also be necessary.

I am not sure. It is not very required for the moment but it could be useful in the future

tests/cpu/test_kalman_fitter.cpp
beomki-yeo: Fixed

tests/cpu/test_kalman_fitter.cpp
beomki-yeo: That was dumb :p Fixed.

examples/simulation/simulate_telescope.cpp
beomki-yeo: > Is our CI not building this code? 

It is not a part of CI. Maybe that's why the CI passed. I didn't know that `std::filesystem` has that kind of issue. I changed to use boost filesystem.

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
beomki-yeo: > That way it is very clear in the code that this is always meant to point at something valid.

Is it like `std::reference_wrapper` won't be compiled if the wrapped type is not copy-able?
Anyway I changed to `std::reference_wrapper` as I don't have any objection on this.

beomki-yeo: Ah OK, it doesn't allow the null value

beomki-yeo:
@krasznaa Could you check the PR?

guilhermeAlmeida1:
@beomki-yeo Attila's on his Christmas break (as most already are), so he might only get around to this in January

krasznaa:
Not quite next year, but by now I'm at 1100m next to a ski slope. With only my phone for electronics.

Will have a look once I'm down from the mountain at the end of the week. 😄

krasznaa (approved):
Only very technical comments on my part. As far as I'm concerned, it could go in like this. But if you think that some of the technical changes I proposed would be reasonable, please go ahead with them. :wink:

performance/include/traccc/utils/helpers.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include "TROOT.h"
+
+namespace traccc {
+
+namespace plot_helpers {
+/// @brief Nested binning struct for booking plots
+struct binning {
+    binning();
+
+    binning(std::string b_title, int bins, float b_min, float b_max);
```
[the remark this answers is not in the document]
beomki-yeo: Hmm let me do this in the separate PR. I don't have any objection with that but this will require change in `eff_plot_tool.hpp` and `duplicate_plot_tool.hpp` as well.

core/include/traccc/fitting/fitting_algorithm.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+        // Iterate over tracks
+        for (std::size_t i = 0; i < n_tracks; i++) {
+
+            // Seed parameter
+            const auto& seed_param = track_candidates[i].header;
+
+            // Make a vector of track state
+            auto& cands = track_candidates[i].items;
+            vecmem::vector<track_state<transform3_type>> track_states;
```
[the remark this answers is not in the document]
beomki-yeo: Right. I changed the variable names.

core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp
```diff
@@ -0,0 +1,110 @@
...
+        track_state<algebra_t>& trk_state,
+        propagator_state_t& propagation) const {
+
+        auto& stepping = propagation._stepping;
+
+        // Some identity matrices
+        const matrix_type<6, 6> I66 =
+            matrix_operator().template identity<e_bound_size, e_bound_size>();
+        const matrix_type<2, 2> I22 =
+            matrix_operator().template identity<2, 2>();
```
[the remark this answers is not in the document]
beomki-yeo: I changed to `static const` - `constexpr` seems require a change in algebra-plugins.

performance/include/traccc/resolution/res_plot_tool.hpp
```diff
@@ -0,0 +1,81 @@
...
+namespace traccc {
+
+class res_plot_tool {
+
+    public:
+    /// @brief The nested configuration struct
+    struct config {
+        /// parameter sets to do plots
+        std::vector<std::string> param_names = {"d0",    "z0",  "phi",
+                                                "theta", "qop", "t"};
```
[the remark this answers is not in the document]
beomki-yeo: Yes

performance/include/traccc/resolution/res_plot_tool.hpp
```diff
@@ -0,0 +1,81 @@
...
+             plot_helpers::binning("r_{z0} [mm]", 100, -0.5, 0.5)},
+            {"residual_phi",
+             plot_helpers::binning("r_{#phi} [rad]", 100, -0.01, 0.01)},
+            {"residual_theta",
+             plot_helpers::binning("r_{#theta} [rad]", 100, -0.01, 0.01)},
+            {"residual_qop",
+             plot_helpers::binning("r_{q/p} [c/GeV]", 100, -0.1, 0.1)},
+            {"residual_t",
+             plot_helpers::binning("r_{t} [s]", 100, -1000, 1000)}};
+    };
```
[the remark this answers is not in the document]
beomki-yeo: Yes

beomki-yeo:
@krasznaa Could you approve again if the changes are OK?

krasznaa (approved):
I'm on board with the latest changes. Have to admit, didn't re-read all of the changes from scratch. :smile:

PR #265, opened 2022-11-02
Draft: SYCL2020 v5 Compatibility, main branch (2022.11.02.)

krasznaa:
Made the SYCL code compatible with [SYCL2020 v5](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html).

Switched to using [sycl::local_accessor](https://sycl.readthedocs.io/en/latest/iface/local-accessor.html) for local memory access, and stopped using [sycl::gpu_selector](https://sycl.readthedocs.io/en/latest/iface/device-selector.html) in the example application. (Switching the code to the new-style [sycl::gpu_selector_v](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-selector) was just not worth it.)

Also updated to the latest version of [VecMem](https://github.com/acts-project/vecmem), which was made compatible with SYCL2020 v5 as well.

This is all to make the project build (and work) with [intel/llvm 2022-09](https://github.com/intel/llvm/releases/tag/2022-09). :wink:

### DIFF
--- a/device/sycl/src/seeding/seed_finding.sycl
+++ b/device/sycl/src/seeding/seed_finding.sycl
@@ -255,11 +255,9 @@ seed_collection_types::buffer seed_finding::operator()(
         details::get_queue(m_queue).submit([&](::sycl::handler& h) {
             // Array for temporary storage of triplets for comparing within seed
             // selecting kernel
-            ::sycl::accessor<scalar, 1, ::sycl::access::mode::read_write,
-                             ::sycl::access::target::local>
-                local_mem(m_seedfilter_config.compatSeedLimit *
-                              weightUpdatingLocalSize,
-                          h);
+            ::sycl::local_accessor<scalar, 1> local_mem(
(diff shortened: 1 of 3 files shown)


krasznaa:
:frowning: Let's give up on this for now.

The very latest version of the compiler ([2022-09](https://github.com/intel/llvm/releases/tag/2022-09)) prints a warning about using local memory in an outdated way.

But the "new way" for doing it, through `sycl::local_accessor`, doesn't even work with the one previous version of the compiler ([2022-06](https://github.com/intel/llvm/releases/tag/2022-06)). That's some bull#&%*...

So in local builds with the latest Intel compiler version we should just put up with the warnings for now. As I just refuse to add logic to this project that would select the right type based on the compiler's capabilities. (I can justify adding such logic to VecMem, but not in this project. Here we are really not meant to be focusing on such technical issues.)

PR #266, opened 2022-11-02
Added comparison of bins with acts

guilhermeAlmeida1:
Additional testing added to `compare_with_acts` to ensure binning is the same.

### DIFF
--- a/tests/cpu/compare_with_acts_seeding.cpp
+++ b/tests/cpu/compare_with_acts_seeding.cpp
@@ -247,6 +247,38 @@ TEST_P(CompareWithActsSeedingTests, Run) {
     // create grid with bin sizes according to the configured geometry
     std::unique_ptr<Acts::SpacePointGrid<SpacePoint>> grid =
         Acts::SpacePointGridCreator::createGrid<SpacePoint>(gridConf);
+
+    // Currently traccc is using grid2. check if acts grid dimensionality is the
+    // same
+    EXPECT_EQ(grid->numLocalBins().size(), 2);
+
+    detray::axis::circular axis0 = internal_spacepoints_per_event.axis_p0();
+    detray::axis::regular axis1 = internal_spacepoints_per_event.axis_p1();
+
+    auto acts_axis0 = grid->axes()[0];
+    auto acts_axis1 = grid->axes()[1];
+
+    EXPECT_EQ(axis0.bins(), acts_axis0->getNBins());
+    EXPECT_EQ(axis1.bins(), acts_axis1->getNBins());
+
+    auto axis0_borders = axis0.all_borders();
+    auto axis1_borders = axis1.all_borders();
+    auto acts_axis0_borders = acts_axis0->getBinEdges();
+    auto acts_axis1_borders = acts_axis1->getBinEdges();
+
+    EXPECT_EQ(axis0_borders.size(), axis0.bins() + 1);
+    EXPECT_EQ(axis1_borders.size(), axis1.bins() + 1);
+
+    EXPECT_EQ(axis0_borders.size(), acts_axis0_borders.size());
+    EXPECT_EQ(axis1_borders.size(), acts_axis1_borders.size());
+
+    for (unsigned int i = 0; i < axis0.bins() + 1; ++i) {
+        EXPECT_NEAR(axis0_borders[i], acts_axis0_borders[i], 0.01);
+    }
+    for (unsigned int i = 0; i < axis1.bins() + 1; ++i) {
+        EXPECT_NEAR(axis1_borders[i], acts_axis1_borders[i], 0.01);
+    }
+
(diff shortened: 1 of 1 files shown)


tests/cpu/compare_with_acts_seeding.cpp
```diff
@@ -247,6 +247,46 @@ TEST_P(CompareWithActsSeedingTests, Run) {
...
+
+    std::size_t border_matches = 0;
+    for (unsigned int i = 0; i < axis0.bins() + 1; ++i) {
+        if (abs(axis0_borders[i] - acts_axis0_borders[i]) /
+                (traccc::scalar)axis0_borders[i] <
+            traccc::float_epsilon) {
+            border_matches++;
+        }
+    }
+    EXPECT_EQ(border_matches, axis0.bins() + 1);
```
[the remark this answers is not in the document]
krasznaa: Not to mention that the way you "normalized" the difference allows for the value to become a negative number. The `abs(...)` would've needed to be called around the entire expression. :stuck_out_tongue:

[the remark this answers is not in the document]
guilhermeAlmeida1: I think this might be because traccc has different point precision from acts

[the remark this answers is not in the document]
krasznaa: Let's not use float_epsilon here then. I'm anyway generally a bit suspicious about that variable. (I fear that we're over-using it.) something like 0.01 would be fine for the delta here.

tests/cpu/compare_with_acts_seeding.cpp
```diff
@@ -247,6 +247,33 @@ TEST_P(CompareWithActsSeedingTests, Run) {
     // create grid with bin sizes according to the configured geometry
     std::unique_ptr<Acts::SpacePointGrid<SpacePoint>> grid =
         Acts::SpacePointGridCreator::createGrid<SpacePoint>(gridConf);
+
+    const auto numLocalBins = grid->numLocalBins();
+    // Currently traccc is using grid2. check if acts grid dimensionality is the
+    // same
+    EXPECT_EQ(numLocalBins.size(), 2);
```
krasznaa, line 254: Since this variable is only used here, let's not make it into an lvalue. Let's just write:

```c++
EXPECT_EQ(grid->numLocalBins().size(), 2);
```

Sidenote: The C\+\+23 literal for `std::size_t` (for this case: `2UZ`) can not get here quickly enough... Since note that this particular test would fail like this with MSVC. But we're not going to build traccc with MSVC, so this is just something to note, not something to worry about...

tests/cpu/compare_with_acts_seeding.cpp
```diff
@@ -247,6 +247,33 @@ TEST_P(CompareWithActsSeedingTests, Run) {
...
+    auto acts_axis1 = grid->axes()[1];
+
+    EXPECT_EQ(axis0.bins(), acts_axis0->getNBins());
+    EXPECT_EQ(axis1.bins(), acts_axis1->getNBins());
+
+    auto axis0_borders = axis0.all_borders();
+    auto axis1_borders = axis1.all_borders();
+    auto acts_axis0_borders = acts_axis0->getBinEdges();
+    auto acts_axis1_borders = acts_axis1->getBinEdges();
+
```
krasznaa, line 269: It might be a bit too verbose, but I think it would make sense to add the following tests as well:

```
EXPECT_EQ(axis0_borders.size(), axis0.bins() + 1);
EXPECT_EQ(axis0_borders.size(), acts_axis0_borders.size());

EXPECT_EQ(axis1_borders.size(), axis1.bins() + 1);
EXPECT_EQ(axis1_borders.size(), acts_axis1_borders.size());
```

So that the assumptions used in the tests in lines 270-275 would be explicitly tested.

krasznaa (changes requested):
Just some more minor comments to annoy you with. :stuck_out_tongue:

tests/cpu/compare_with_acts_seeding.cpp
guilhermeAlmeida1: Is it not safe to assume that the size of the borders vector is = #bins + 1?

If so, these tests are redundant as there's already one which tests that #bins is the same.

` EXPECT_EQ(axis0.bins(), acts_axis0->getNBins());`
` EXPECT_EQ(axis1.bins(), acts_axis1->getNBins());`

krasznaa: That is just how the grids are supposed to work. The goal of a unit test is exactly to test these sort of things. As far as this test is concerned, the objects returned for the bin borders have nothing to do with the values returned bu the `bins()` and `getNBins()` calls.

krasznaa (approved):
Forgot about this one. Let's just push this in.

PR #268, opened 2022-11-08
SYCL2020 v5 Compatibility, main branch (2022.11.08.)

krasznaa:
Let me revive #265. I actually tried to re-open that one, but GitHub didn't think so. :confused:

I removed the changes with the usage of "local memory". This way the code at least compiles with the latest version of the Intel compiler. Even if it emits some warnings while doing so.

As before, stopped using `sycl::gpu_selector` in the example application, and updated to the latest version of VecMem. Which was made compatible with SYCL2020 v5 as well.


stephenswat (approved):
Looks harmless.

stephenswat (approved):

PR #267, opened 2022-11-04
Updated Acts dependency to current main branch patch

guilhermeAlmeida1:
Replacing #228 
Dependency on #266 & Acts [#1650](https://github.com/acts-project/acts/pull/1650)

Wanted to get an opinion before continuing with this on what our plans are for all of the optional filters added to Acts such as:

- [different deltaR boundaries for Bot/Top spacepoints](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Seeding/SeedFinderConfig.hpp#L39-L46)
- [useVariableMiddleSPRange with vector of vectors in config (I assume this might be tricky for our gpu code)](https://github.com/acts-project/acts/blob/main/Core/include/Acts/Seeding/SeedFinderConfig.hpp#L52-55)
- and other such changes which come further down like seedConfirmation, the use of Delegates, etc.

Taking these various optionals out of the equation, this should give us a correct synchronisation with acts.

PS: I'm still trying to figure out why the seq_example's are now going to very small ammount of found seeds.

PPS: Currently I put the Acts dependency on a branch from my own fork of Acts because we need the aforementioned PR merged into Acts before making that dependence onto Acts itself.

PR #269, opened 2022-11-09
Cell Read Error Checks, main branch (2022.11.09.)

krasznaa:
Introduced error detection into the cell reading. So that it would catch issues like the one we've been chasing for a long-long time with the Acts update (#267).

Pinging @guilhermeAlmeida1.


guilhermeAlmeida1 (approved):
This definately would have been a great help for my troubleshooting. A worthwhile addition.

PR #270, opened 2022-11-10
Kokkos spacepoint binning

Yhatoh:
Hi!

In this PR, I did code the spacepoint binning algorithm in Kokkos, so I created a folder in `device` named kokkos, that was the code for spacepoint binning.

Now, I did create a test (that is very similar of the cuda test) for seeding algorithm, but for now only does the part of spacepoint binning. And for the next PR, that test will test the complete seeding algorithm.

Cheers Gabriel


device/cuda/include/traccc/cuda/utils/make_prefix_sum_buff.hpp
```diff
@@ -28,4 +28,4 @@ vecmem::data::vector_buffer<device::prefix_sum_element_t> make_prefix_sum_buff(
     const std::vector<device::prefix_sum_size_t>& sizes, vecmem::copy& copy,
     const traccc::memory_resource& mr);
 
-}  // namespace traccc::cuda
\ No newline at end of file
+}  // namespace traccc::cuda
```
krasznaa, line 31: While all these updates in the CUDA and SYCL files are not bad, you shouldn't mix in those changes into this PR.

device/kokkos/include/traccc/kokkos/seeding/seeding_algorithm.hpp
```diff
@@ -0,0 +1,56 @@
+/** TRACCC library, part of the ACTS project (R&D line)
```
krasznaa, line 1: Let's not put this class into this PR at all. Without the `seed_finding` class existing it can't do anything yet anyway. You should instead instantiate `traccc::kokkos::spacepoint_binning` itself in the example code.

device/kokkos/src/seeding/spacepoint_binning.cpp
```diff
@@ -0,0 +1,93 @@
...
+    sp_grid_buffer grid_buffer(
+        m_axes.first, m_axes.second, std::vector<std::size_t>(grid_bins, 0),
+        std::vector<std::size_t>(grid_capacities_host.begin(),
+                                 grid_capacities_host.end()),
+        m_mr.main, m_mr.host);
+    m_copy->setup(grid_buffer._buffer);
+    sp_grid_view grid_view = grid_buffer;
+
+    // Populate the grid.
+    Kokkos::parallel_for("count_grid_capacities", team_policy(num_blocks, num_threads),
```
krasznaa, line 82: The string name should be `"populate_grid"` here.

examples/run/kokkos/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+
+#
+# (c) 2021-2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+traccc_add_executable( seeding_example_kokkos "seeding_example_kokkos.cpp"
+   LINK_LIBRARIES vecmem::core vecmem::cuda traccc::io traccc::performance
```
krasznaa, line 8: As the CI pointed out, `vecmem::cuda` should not be used here.

examples/run/kokkos/seeding_example_kokkos.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include "traccc/io/writer.hpp"
+#include "traccc/options/common_options.hpp"
+#include "traccc/options/handle_argument_errors.hpp"
+#include "traccc/options/seeding_input_options.hpp"
+#include "traccc/performance/collection_comparator.hpp"
+#include "traccc/seeding/seeding_algorithm.hpp"
+#include "traccc/seeding/track_params_estimation.hpp"
+
+// VecMem include(s).
+#include <vecmem/memory/cuda/device_memory_resource.hpp>
```
krasznaa, line 24: At this point let's just set this test up to work on the host. Let's not touch CUDA specific code yet. (Also because the CI is not yet set up to allow for that.)

krasznaa (changes requested):
You'll need to format all the code with https://github.com/acts-project/traccc/blob/main/.github/check_format.sh. Note that if you use [VS Code](https://code.visualstudio.com/) to edit the code, you can pretty much just press Ctrl-Shift-I and the editor can format the source files in the way that we want them to be formatted.

Other than fixing the CI issues, you should simplify the PR a bit. Kick out the `seeding_algorithm` files, and simplify the `seeding_example_kokkos.cpp` code a little as a start. (So that it would just run spacepoint binning on Kokkos together with full seed-finding on the host. Without really comparing the two just yet.)

krasznaa:
Gabriel, I see that you once again opened a PR from the `main` branch of your fork. :frowning: You should use the same procedure for this repository that we use for [atlas/athena](https://gitlab.cern.ch/atlas/athena).

https://atlassoftwaredocs.web.cern.ch/gittutorial/

I.e. please only commit modifications into "feature branches" in your fork. It's really not good to put "development commits" into the `main` branch directly. Even in a fork.

PR #271, opened 2022-11-11
Created timer class for examples execution time measurements.

guilhermeAlmeida1:
This change cleans up our code a bit, and should also simplify implementation of further performance measurement code moving forward.


performance/include/traccc/performance/timer.hpp
```diff
@@ -0,0 +1,82 @@
...
+    private:
+
+    constexpr static std::size_t precision = 10;
+    constexpr static std::size_t nameLength = 25; 
+
+    /// Name to be printed out from timer
+    std::string name;
+
+    /// Shared pointer to printout string
+    std::shared_ptr<std::string> printout;
```
stephenswat, line 75: Please don't store this as a pointer to a string. Please either store a file handle so you can write directly to `stderr` or `stdout`, or write the data to a map.

stephenswat (changes requested):

performance/include/traccc/performance/timer.hpp
guilhermeAlmeida1: May I ask what is wrong with using the pointer to a string? (not doubting that in any way, it's more of a educational question)

guilhermeAlmeida1: I initially went with this approach of the shared pointer to allow more than one timer to be running at the same time with the same "printout string" as well as being easy to add the functionality for incrementing same "name" calls together

stephenswat: If you consider the two ideas I mentioned before, they each have upsides and downsides; the upside of writing directly to a file handle is immediacy in the sense that the data becomes available to the user immediately. The downside is that the data is unstructured and immalleable: it is (essentially) impossible to retrieve or modify individual points of data. The approach of using a map is the opposite: it is non-immediate (requires dumping it all at the end of the computation, but it is highly structured. The approach you have picked here has the worst of both worlds: it is non-immediate and unstructured, and I personally do not see any upsides to it. The unstructured nature of it requires fragile operations like lookups in a string. Also, this keeps a lot of (shared) state, which is almost always undesirable.

guilhermeAlmeida1: Thanks for the explanation, I'll probably take the map route then. Pun intended

guilhermeAlmeida1:
Things should be a lot cleaner now

krasznaa (changes requested):
I don't really like it. :frowning:

I had something a bit different in mind. Instead of making one complicated type (because I think your `traccc::performance::timer` class is a bit too complicated), I would propose to create at least 2 types. (Though due to performance considerations we may want to introduce a 3rd one as well...)
  - I would imagine having a `traccc::performance::timing_info` class that is pretty much just a pure data storage class. Like:

```c++
struct timing_info {
   std::vector<std::pair<std::string, std::chrono::nanoseconds> > data;
};
std::ostream& operator<<(std::ostream& out, const timing_info& info);
```

  - Then I would introduce a separate `traccc::performance::timer` class, which does the actual timing. Like:

```c++
class timer {
   public:
   timer(timing_info& info, std::string_view name);
   ~timer();
   private:
   std::chrono::high_resolution_clock::time_point m_start;
   std::chrono::nanoseconds& m_measurement;
};
```

My idea would be to use these types like the following in the example applications:

```c++
traccc::performance::timing_info timing_info;
...
{
    traccc::performance::timer t{timing_info, "File reading (cpu)");
    // File reading code...
   ...
}
...
std::cout << timing_info << std::endl;
```

In this setup we would still suffer from string-lookups on every timing measurement. Which is not brilliant. So we could go with a similar approach to what we do with [SG::auxid_t](https://gitlab.cern.ch/atlas/athena/-/blob/master/Control/AthContainersInterfaces/AthContainersInterfaces/AuxTypes.h#L29) in the offline code. But I'll write about that separately...

guilhermeAlmeida1:
@krasznaa this approach you suggested with measuring time with constructors and destructors was what I initially went with. However, because the actual functions we measure return container objects, to put them within separate brackets would make things more complicated with having to create the result object beforehand instead of what we currently have such as:

```c++
elapsedTimes.start("File reading (cpu)");
// Read the cells from the relevant event file into host memory.
const traccc::cell_container_types::host cells_per_event =
traccc::read_cells_from_event(event, common_opts.input_directory,
                                          common_opts.input_data_format,
                                          surface_transforms, digi_cfg,
                                          host_mr);
elapsedTimes.stop("File reading (cpu)");
```

If I understand correctly the only real difference other than that is having a `stop(string)` instead of a `}` at the end of a function, but the actual functionality would be the same with the same calculations and overhead.

krasznaa:
It's more of a design question for me. With separate start and stop functions you open the door to typos. And I just don't like the idea of a single object keeping track of multiple ongoing measurements at the same time. I much more like the idea of individual "timer objects" measuring their individual times, and then "upload" these to some central repository when their measurement is done.

You're right, this would mean creating empty result objects at first, and then assigning values to them inside of brackets. Could you make a quick measurement to see if this affects the performance of the test applications at all? It shouldn't. As long as the VecMem buffer types are implemented reasonably. But we may not have tried this yet... :thinking:

guilhermeAlmeida1:
I had to be quite explicit, using:
```c++
traccc::container_types::buffer buffer{{0, *(mr.host)}, {{}, *(mr.host), mr.host}};
buffer = algorithm(...);
```
 but it worked and did not seem to affect performance. I used all host memory as this buffer was going to be overwritten, I assume that doens't change anything, right? 
Should I go with this and the constructor/destructor type timer? I personnally don't see much gain from this but I get the point

krasznaa:
Yes, go ahead like this for now. I started working on a VecMem update in the meanwhile that will make it easier to do this. (Allowing the default construction of buffers.)

guilhermeAlmeida1:
@krasznaa my earlier question was about the fact that in your suggestion you used 
```c++
class timer {
   public:
   timer(timing_info& info, std::string_view name);
```

where as I used 
```c++
std::shared_ptr<timing_info> info 
```
So this would survive the destruction of the timer instance, not entirely sure if you had something different in mind as Stephen had briefly suggested to stay away from shared pointers as I had already done something similar in a previous iteration of this PR

examples/run/sycl/seq_example_sycl.sycl
```diff
@@ -129,143 +119,118 @@ int seq_run(const traccc::full_tracking_input_config& i_cfg,
...
 
-        traccc::track_params_estimation::output_type params;
-        if (run_cpu) {
-            /*time*/ auto start_tp_estimating_cpu =
-                std::chrono::system_clock::now();
+            // CPU
 
-            params = tp(spacepoints_per_event, seeds);
+            if (run_cpu) {
+                traccc::performance::timer tp_timer_cpu("Track params  (cpu)",
```
krasznaa, line 228: Giving elaborate names for the variables is not bad, but you might as well just call them `t`. Since we don't need to pass them anywhere where their names would become useful.

krasznaa (changes requested):
You should really separate the new code into `timing_info.hpp`, `timing_info.cpp`, `timer.hpp` and `timer.cpp`. You may want to argue about inlining some functions, but on first order let's not inline absolutely everything.

performance/include/traccc/performance/timer.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+/// Struct for storing time measurements collected in timer class
+///
+struct timing_info {
+    std::vector<std::pair<std::string, std::chrono::nanoseconds>> data;
+};
+
+std::ostream& operator<<(std::ostream& out, const timing_info& info){
+    for(auto i : info.data){
+        out << std::setw(30) << std::right << i.first << "  " << std::setw(12) << std::left << (i.second.count() / 1.e9) << std::endl;
```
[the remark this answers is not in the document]
guilhermeAlmeida1: The performance dir wasn't added to the check-format script. I'll add it in this PR

performance/include/traccc/performance/timer.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    private:
+    /// Start time (measured at construct time)
+    std::chrono::high_resolution_clock::time_point m_start;
+
+    /// Name of measurement
+    std::string_view m_name;
+
+    /// Shared ptr to timing info where to store elapsed time
+    timing_info* m_timing_info;
```
krasznaa, line 43: Why are you using a pointer? I was suggesting a non-const reference from the start...

With a pointer you should always check whether the received pointer is `nullptr` or not. If `nullptr` is not a valid option, then the constructor should not even allow providing one to it.

Plus, `traccc::performance::timer` is not meant to be copy-able. If you make this member a non-const reference, you get "copy prevention" for free.

performance/include/traccc/performance/timer.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+    /// End time measurement
+    ~timer();
+
+    private:
+    /// Start time (measured at construct time)
+    std::chrono::high_resolution_clock::time_point m_start;
+
+    /// Name of measurement
+    std::string_view m_name;
```
krasznaa, line 40: So... [std::string_view](https://en.cppreference.com/w/cpp/string/basic_string_view) is an excellent function argument. It allows calls with `const char*` literals to have a zero overhead. Let's say when you instantiate the class with:

```c++
traccc::performance::timer t{"Foo", bar};
```

But as the name implies, this type is merely a view of a string whose ownership is managed somewhere else. It is **very** similar to the VecMem view types in this sense.

So to make `traccc::performance::timer` safe, its member variable should actually be a simple old [std::string](https://en.cppreference.com/w/cpp/string/basic_string) instead. So that in case it's created using a temporary string, it would still work well. Let's say in the following messed up situation:

```c++
traccc::performance::timer t{"Foo"s, bar};
```

See the notes towards the bottom of: https://en.cppreference.com/w/cpp/string/basic_string_view

So:
  - Constructor parameter: `std::string_view`
  - Member variable: `std::string`

performance/src/performance/timer.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <algorithm>
+
+namespace traccc::performance {
+
+/// Start time measurement
+/// @param timer_name name to be printed out identifying what is measured
+/// @param t_info shared_ptr to timing_info where to store timings
+timer::timer(const std::string_view timer_name, timing_info* t_info)
+    : m_name(timer_name), m_timing_info(t_info) {
+    m_start = std::chrono::high_resolution_clock::now();
```
krasznaa, line 21: You should just do this with the same formalism that you use to initialise the other two members. I.e.

```c++
timer::timer(const std::string_view timer_name, timing_info* t_info)
    : m_start(std::chrono::high_resolution_clock::now()),
      m_name(timer_name), m_timing_info(t_info) {}
```

krasznaa (changes requested):
Now in the final stretch... :stuck_out_tongue:

examples/run/cuda/seq_example_cuda.cpp
```diff
@@ -98,142 +89,117 @@ int seq_run(const traccc::full_tracking_input_config& i_cfg,
...
+        // Instantiate host containers/collections
+        traccc::cell_container_types::host cells_per_event;
         traccc::clusterization_algorithm::output_type measurements_per_event;
         traccc::spacepoint_formation::output_type spacepoints_per_event;
+        traccc::seeding_algorithm::output_type seeds;
+        traccc::track_params_estimation::output_type params;
 
-        if (run_cpu) {
+        // Instantiate cuda containers/collections
+        traccc::spacepoint_container_types::buffer spacepoints_cuda_buffer{
```
guilhermeAlmeida1, line 106: @krasznaa in line with your suggestion in the other PR, should I change all these to ouput types? I'm not sure I like it much because the constructors being called aren't exactly "generic" themselves, and it seems slightly more readable like this.

krasznaa: Fair question... :thinking: Let's leave them with this type for now. Once https://github.com/acts-project/vecmem/pull/208 arrives in this project, we will switch to creating the buffers with their default constructors. At that point these examples should switch to using `output_type` types. Since they'll be default-constructing the objects themselves.

But for now this naming/formalism is fine with me.

krasznaa (approved):
Okay, let's stop at this point. The `const std::string_view` parameter is still not perfect (the `const` qualifier is not really meaningful there), but let's not drag this out any longer.

PR #272, opened 2022-11-14
Changed track_params_estimation return type to collection buffer

guilhermeAlmeida1:
The track parameters estimation algorithm was returning a host collection. Changed this to a buffer created on the device. This will be used in further steps when introducing fitting on the device.

### DIFF
--- a/examples/run/cuda/seeding_example_cuda.cpp
+++ b/examples/run/cuda/seeding_example_cuda.cpp
@@ -145,7 +145,8 @@ int seq_run(const traccc::seeding_input_config& i_cfg,
         /*time*/ auto start_tp_estimating_cuda =
             std::chrono::system_clock::now();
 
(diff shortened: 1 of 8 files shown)


examples/run/cuda/seeding_example_cuda.cpp
```diff
@@ -145,7 +145,8 @@ int seq_run(const traccc::seeding_input_config& i_cfg,
         /*time*/ auto start_tp_estimating_cuda =
             std::chrono::system_clock::now();
 
-        auto params_cuda = tp_cuda(spacepoints_cuda_buffer, seeds_cuda_buffer);
+        auto params_cuda_buffer =
```
krasznaa, line 148: Instead of using `auto`, using `traccc::cuda::track_params_estimation::output_type` would sit better with me. Though this is more of a personal taste question.

paulgessinger:
Not sure. Maybe the API call back from policybot was delayed?

PR #273, opened 2022-11-15
File reading small improvement

guilhermeAlmeida1:
This additional check reduces computation by only looking for the cell's module if it's not a repeat of the previous one. Elapsed time for file reading time is reduced by ~35%.

### DIFF
--- a/io/include/traccc/io/csv.hpp
+++ b/io/include/traccc/io/csv.hpp
@@ -251,9 +251,9 @@ inline cell_container_types::host read_cells(
         const cell_container_types::host::header_vector& headers =
             result.get_headers();
 
-        auto it = std::find(headers.begin(), headers.end(), module);
+        auto rit = std::find(headers.rbegin(), headers.rend(), module);
 
-        if (it == headers.end()) {
+        if (rit == headers.rend()) {
 
             if (tfmap != nullptr) {
                 if (!tfmap->contains(module.module)) {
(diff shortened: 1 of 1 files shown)


beomki-yeo:
It looks good to me but I don't know why it fails with Kokkos

krasznaa:
Kokkos is irrelevant there. The failing tests are pure c\+\+ ones. (It's just by chance that this was the first platform on which this test failed.)

Guilherme, I suspect those tests should fail for you locally as well. You'll need to debug tomorrow what's going wrong.

guilhermeAlmeida1:
The problem was actually quite simple, but I ended up simplifying this whole PR by just changing to using a reverse iterator which also solves the issue. Also added this functionality to the other csv reading functions. Hopefully this should work now but let's wait for the CI

io/include/traccc/io/csv.hpp
```diff
@@ -288,7 +288,7 @@ inline cell_container_types::host read_cells(
 
             result.push_back(module, vecmem::vector<cell>({c}));
         } else {
-            auto idx = it - headers.begin();
+            auto idx = std::distance(headers.begin(), rit.base()) - 1;
```
krasznaa, line 291: This is correct. But you should add just one line of explanation of why the `-1` is necessary. It's actually pretty educational to understand this. (Do you? :wink:)

krasznaa (changes requested):
I like it! The switch to reverse iterators in the search is very elegant! :smile:

krasznaa (approved):
:+1:

PR #274, opened 2022-11-15
Changed csv cell reading to counting + filling cell container

guilhermeAlmeida1:
Similarly to what is done in device kernels for avoiding using dynamic memory allocation, this PR changes cell reading to counting the number of members needed for the jagged vector and then filling it. This change reduces this function's time by some ~5%. When working on this, I stumbled upon the much more significant change in #273 which leads to an additional 35% reduction (this is also already implemented in this PR).

### DIFF
--- a/io/include/traccc/io/csv.hpp
+++ b/io/include/traccc/io/csv.hpp
@@ -11,6 +11,7 @@
 #include "traccc/definitions/primitives.hpp"
 #include "traccc/edm/cell.hpp"
 #include "traccc/edm/cluster.hpp"
+#include "traccc/edm/container.hpp"
 #include "traccc/edm/measurement.hpp"
 #include "traccc/edm/particle.hpp"
 #include "traccc/edm/spacepoint.hpp"
@@ -222,6 +223,14 @@ inline std::map<geometry_id, transform3> read_surfaces(
     return transform_map;
 }
 
+struct cell_counter {
+    geometry_id module = 0;
+    unsigned int nCells = 0;
+};  // struct cell_counter
+
+/// Declare all cell counter collection types
+using cell_counter_collection_types = collection_types<cell_counter>;
+
 /// Read the collection of cells per module and fill into a collection
 ///
 /// @param creader The cellreader type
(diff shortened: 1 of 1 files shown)


io/include/traccc/io/csv.hpp
```diff
@@ -235,83 +236,111 @@ inline cell_container_types::host read_cells(
     const Acts::GeometryHierarchyMap<digitization_config>* digi_cfg = nullptr,
     unsigned int max_cells = std::numeric_limits<unsigned int>::max()) {
 
-    cell_container_types::host result(&resource);
+    // Create cell counter vector
+    cell_counter_collection_types::host cell_counts(&resource);
+    // Create cell vector
+    cell_collection_types::host allCells(&resource);
```
guilhermeAlmeida1, line 242: Annoyingly, [the csv file reading object we're currently](https://eic.github.io/doxygen/da/dae/classdfe_1_1io__dsv__impl_1_1NamedTupleDsvReader.html) using does not allow for getting the number of lines in the file, despite the fact that the actual file reader it holds has this, which could've been used here to immediately reserve the number of cells needed. 
Anyways, it would not make a significant difference to the time taken.

io/include/traccc/io/csv.hpp
```diff
@@ -235,83 +236,103 @@ inline cell_container_types::host read_cells(
...
+        }
+    }
 
-            if (digi_cfg != nullptr) {
-                Acts::GeometryIdentifier id(module.module);
+    // Construct the size vectors.
+    const std::size_t size = cell_counts.size();
+    std::vector<std::size_t> numCells(size);
+    std::transform(cell_counts.begin(), cell_counts.end(), numCells.begin(),
+                   [](const cell_counter& counter) { return counter.nCells; });
```
krasznaa, line 279: This transformation seems a bit unnecessary, isn't it? You only access single elements from this `numCells` vector. At that point you might as well just get the single numbers from the original `cell_counts` object in all those places.

core/include/traccc/edm/cell.hpp
```diff
@@ -74,6 +74,11 @@ struct cell_module {
 
 };  // struct cell_module
 
+struct cell_counter {
+    geometry_id module = 0;
+    unsigned int nCells = 0;
+};  // struct cell_counter
```
krasznaa, line 80: Should this really be here? :thinking: For now I'd rather just put it into the CSV reading code itself. Until we need to use it in more places it's not worth it putting it into such a public place.

krasznaa (changes requested):
I generally like it. With a bit of tweaking I think it will be worth the effort. (For context to others, I was not entirely sure about this change when it turned out to only help with a 5% speedup.)

krasznaa:
Probably should've done it the other way around, but since you already had 2 different PRs, I thought it may be clearer for the repository's history if we merge both in. This one just doing the "read-ahead".

io/include/traccc/io/csv.hpp
krasznaa: I was trying to see if I could find a way of counting the lines in a text file in C\+\+ without having to read the content of the whole file in the process, but didn't find any. :confused: (Thinking a bit more, probably it's not possible to do that...)

So let's just accept this for now. One thing that you could try is to reserve some large-ish number of elements in these vectors. (As we discussed yesterday, using [std::vector::reserve](https://en.cppreference.com/w/cpp/container/vector/reserve).) Since remember, [std::vector](https://en.cppreference.com/w/cpp/container/vector) doubles its allocation every time it runs out of memory. So if we start with let's say 50k elements, the number of re-allocations should stay small even for very large input files.

krasznaa (approved):
It's generally acceptable as-is, but trying with some pre-allocation for the containers could be worth looking at.

guilhermeAlmeida1:
`So if we start with let's say 50k elements, the number of re-allocations should stay small even for very large input files.` this really didn't make a significant difference, but I don't see any harm in adding it anyways.

PR #275, opened 2022-11-15
I/O Redesign, main branch (2022.11.15.)

krasznaa:
I have long been annoyed with the I/O code being implemented all in header files. This is a proposal for how I think we should re-write that code.

As a start I only updated how binary file writing works. (As I thought that would have the fewest clients.) I would propose re-designing the rest of the code in a similar way as well. Hiding most of the I/O code into the `io/src/` directory, and only exposing a well defined set of public functions to the other parts of the project.

At first I'm just curious if I'll get any push-back on this. If not, I'll continue with the updates in this PR.


krasznaa:
Curious. So far the clang-format on my development machine worked the same way as the one that we use in the CI. But I guess this code managed to unearth some behavior differences... Sigh...

beomki-yeo:
Looks good to me so far

krasznaa:
This is not great... :confused: The Intel linker freaks out if `ActsPluginJson` is not linked against `traccc::core`. :frowning: Even though that link is **really** not needed.

Not something to sort out in this PR, but this surprised me a lot...

krasznaa:
Okay, this monster is ready for prime time finally. Just a couple of highlights:
  - All the CSV and JSON specific code was moved into "private" code under `io/src`. So `traccc::io` does not need to publicly depend on the libraries used for those types anymore.
  - I generally split the CSV code into a bunch of separate files. This may be a bit overboard, but organisation-wise I liked this setup better than to co-locate multiple things into single files.
  - I found a very sneaky bug in the cell reading code while re-writing it. :frowning: @guilhermeAlmeida1, the code that's currently in the main branch would break if cells from the same module would not be put consecutively in the input file. As the one for-loop that moves the `cell` objects into the result container blindly assumes that cells for a single module are consecutive in the `allCells` container. Which they may not be.
    * So I modified that function a bit during the refactoring...
  - @beomki-yeo this PR will of course clash with #276. But while toying with the idea, I decided not to implement everything from that PR in this one. (I did pull some of that code "in a way" into this PR though.)

It's a ridiculously large PR in the end. :frowning: Sorry about that... (I didn't want to break it into smaller blocks because I was sure that I would never finish it if I made a partial PR at first.)

krasznaa:
Just to add: I made some explicit tests. The performance of CSV reading in our "algorithm sequence examples" remains the same (within uncertainty) after this rewrite. This is what I expected, I didn't make the code any faster, it was still nice to see it. :wink:

examples/io/create_binaries.cpp
```diff
@@ -31,36 +38,36 @@ int create_binaries(const std::string& detector_file,
...
 
         // Read the cells from the relevant event file
-        traccc::cell_container_types::host cells_csv =
-            traccc::read_cells_from_event(event, common_opts.input_directory,
-                                          common_opts.input_data_format,
-                                          surface_transforms, digi_cfg,
-                                          host_mr);
+        traccc::cell_container_types::host cells_csv = traccc::io::read_cells(
+            event, common_opts.input_directory, common_opts.input_data_format,
+            &surface_transforms, &digi_cfg, &host_mr);
```
stephenswat, line 43: Hmm, I'm not a huge fan of using pointers here instead of references, although I understand the use in allowing us to read cells without digitization configs for example. Out of curiosity, do we use this mechanism – passing nullpointers here – anywhere in the code?

examples/run/sycl/seq_example_sycl.sycl
```diff
@@ -69,11 +68,11 @@ int seq_run(const traccc::full_tracking_input_config& i_cfg,
             const traccc::common_options& common_opts, bool run_cpu) {
 
     // Read the surface transforms
-    auto surface_transforms = traccc::read_geometry(i_cfg.detector_file);
+    auto surface_transforms = traccc::io::read_geometry(i_cfg.detector_file);
 
     // Read the digitization configuration file
     auto digi_cfg =
-        traccc::read_digitization_config(i_cfg.digitization_config_file);
+        traccc::io::read_digitization_config(i_cfg.digitization_config_file);
```
stephenswat, line 75: Personally I find the use of `auto` in the IO code a bit excessive. I know it was already like this so I'm not saying we should change it in this PR per sé, but at some point I think we should explicitly use our nice typedefs.

io/include/traccc/io/data_format.hpp
```diff
@@ -9,9 +9,7 @@
 
 namespace traccc {
 
-enum data_format : int {
-    csv = 0,
-    binary = 1,
-};
+/// Format for an input or output file
+enum data_format : int { csv = 0, binary = 1, json = 2 };
```
stephenswat, line 13: Could you put a trailing comma behind the 2 so `clang-format` formats this properly?

io/src/csv/surface.hpp
```diff
@@ -0,0 +1,33 @@
...
+namespace traccc::io::csv {
+
+/// Type to read information into about the detector modules/surfaces
+struct surface {
+
+    uint64_t geometry_id = 0;
+    float cx = 0., cy = 0., cz = 0.;
+    float rot_xu = 0., rot_xv = 0., rot_xw = 0.;
+    float rot_yu = 0., rot_yv = 0., rot_yw = 0.;
+    float rot_zu = 0., rot_zv = 0., rot_zw = 0.;
```
stephenswat, line 25: Do these defaults serve any purpose? Is it possible to read an object of this type without filling all its members with non-default values? :thinking:

io/src/utils.cpp
```diff
@@ -0,0 +1,44 @@
...
+const std::string& data_directory() {
+
+    // Establish the directory name only once for the application.
+    static const std::string data_dir = [] {
+        // Look for an environment variable specifying the directory name.
+        const char* env_dir = std::getenv("TRACCC_TEST_DATA_DIR");
+        if (env_dir == nullptr) {
+            // If an environment variable is not set, rely on the definition
+            // coming from the build system.
+            return std::string(TRACCC_TEST_DATA_DIR) + "/";
```
stephenswat, line 27: Hmm, this strikes me as potentially dangerous; should we add a check whether `TRACCC_TEST_DATA_DIR` is actually defined? Otherwise we might end up accidentally read or writing from `/`, which is a little spooky.

io/src/read.cpp
```diff
@@ -0,0 +1,50 @@
...
+    demonstrator_input result{events, mr};
+
+    // Read in the cell data for all events. In parallel if possible.
+#pragma omp parallel for
+    for (std::size_t event = 0; event < events; ++event) {
+        traccc::cell_container_types::host cells =
+            io::read_cells(event, directory, format, &geom, &digi_cfg, mr);
+
+        // Update the result container. (In a single thread.)
+#pragma omp critical
```
stephenswat, line 42: Isn't `demonstrator_input` just a vector? Then this critical section has no use; every thread will be writing to a separate location in the vector anyway.

stephenswat (commented):
Looks great overall, just a few comments and questions here and there.

examples/io/create_binaries.cpp
krasznaa: Have a look at `ConnectedComponentAnalysisTests`. :wink: That one does:

```c++
        traccc::cell_container_types::host data =
            traccc::io::read_cells(file_hits);
```

examples/run/sycl/seq_example_sycl.sycl
krasznaa: Yeah... I agree. Maybe in a follow-up PR.

examples/io/create_binaries.cpp
stephenswat: Fair.

io/src/csv/surface.hpp
krasznaa: DFE is surprisingly robust with reading differently formatted files. If you read a file that doesn't provide all of these variables, DFE will silently just not touch the variables corresponding to the missing columns. I was actually looking for this type of a bug in my code for a few hours yesterday when by mistake a piece of code was reading an incorrect file. (Which resulted in some way more subtle errors than I would've expected.)

Long story short, setting some defaults helps at least a little with this...

io/src/utils.cpp
krasznaa: :thinking: If `TRACCC_TEST_DATA_DIR` is not defined as a macro, the compilation should fail. (In that case `TRACCC_TEST_DATA_DIR` becomes an unknown token/symbol during the compilation.)

My idea was that this macro would **always** be defined as a `PRIVATE` macro for the `traccc::io` library. This is one of the main reasons that I moved this function's implementation into a `.cpp` file. To be able to rely on a library-specific macro. (Without it bleeding out into all the clients.)

With all this being said, the way that file/directory names are handled in the code is super fragile. But at this point I think it's not worth to make it more robust than this...

io/src/read.cpp
krasznaa: :thinking: Very true. I just took this code as it was. But you're absolutely right.

io/src/csv/surface.hpp
stephenswat: I'm not sure I would class "failing silently" as "surprisingly robust" but yeah that is a pretty good reason. :+1:

io/src/read.cpp
stephenswat: Ah sorry I did not realize that this was just old copy-pasted code. Could be worth changing though, might improve the performance a little. :smile:

io/include/traccc/io/data_format.hpp
krasznaa: :+1:

io/src/read.cpp
krasznaa: :+1:

stephenswat (approved):

beomki-yeo:
> @beomki-yeo this PR will of course clash with https://github.com/acts-project/traccc/pull/276. But while toying with the idea, I decided not to implement everything from that PR in this one. (I did pull some of that code "in a way" into this PR though.)

No worries. I don't think I need #276 to be merged right now :p

PR #279, opened 2022-11-22
TBB Support / Example, main branch (2022.11.22.)

krasznaa:
Taught the project how to use TBB, and added a trivial example. Even though I still maintain that we are not meant to do anything complicated with "frameworking" in this repository, we should get to some multi-threaded tests relatively soon. Just some fairly trivial ones.

OpenMP will be one option for doing that, but since eventually we'll want to make these algorithms work with TBB in one way or another, I thought I may as well make the bed for doing that.

Note that by default the project is set up to build its own TBB library. We may want to change this later on. Since ROOT requires TBB itself, TBB is **always** in our environment already. But on first order this seemed easier to do. :thinking:

Pinging @guilhermeAlmeida1. As he'll probably need to write some of this. :stuck_out_tongue:

### DIFF
--- a/extern/tbb/CMakeLists.txt
+++ b/extern/tbb/CMakeLists.txt
@@ -1,0 +1,31 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# CMake include(s).
+cmake_minimum_required( VERSION 3.14 )
+include( FetchContent )
+
+# Silence FetchContent warnings with CMake >=3.24.
+if( POLICY CMP0135 )
+   cmake_policy( SET CMP0135 NEW )
+endif()
+
+# Tell the user what's happening.
+message( STATUS "Building TBB as part of the TRACCC project" )
+
+# Declare where to get TBB from.
+set( TRACCC_TBB_SOURCE
+   "URL;https://github.com/oneapi-src/oneTBB/archive/refs/tags/v2021.7.0.tar.gz;URL_MD5;68e617448f71df02d8688c84d53778f6"
+   CACHE STRING "Source for TBB, when built as part of this project" )
+mark_as_advanced( TRACCC_TBB_SOURCE )
+FetchContent_Declare( TBB ${TRACCC_TBB_SOURCE} )
+
+# Options used in the build of TBB.
+set( TBB_TEST FALSE CACHE BOOL "Turn off the TBB tests" )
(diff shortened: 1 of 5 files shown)


examples/run/CMakeLists.txt
```diff
@@ -27,3 +27,5 @@ if (OpenMP_CXX_FOUND)
     add_subdirectory(openmp)
     message(STATUS "OpenMP found! Include /examples/openmp subdirectory")
 endif()
+
+add_subdirectory(tbb)
```
stephenswat, line 31: Are we not hiding these examples behind a configurable `TRACCC_BUILD_TBB` flag?

extern/tbb/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+if( POLICY CMP0135 )
+   cmake_policy( SET CMP0135 NEW )
+endif()
+
+# Tell the user what's happening.
+message( STATUS "Building TBB as part of the TRACCC project" )
+
+# Declare where to get TBB from.
+set( TRACCC_TBB_SOURCE
+   "URL;https://github.com/oneapi-src/oneTBB/archive/refs/tags/v2021.7.0.tar.gz;URL_MD5;68e617448f71df02d8688c84d53778f6"
```
stephenswat, line 21: Why 2021.7.0 and not 2021.7.1?

examples/run/CMakeLists.txt
stephenswat: Just read the description, still think this should be configurable.

stephenswat:
For the record I still strongly disagree with the assessment that we should not do scheduling or placement in this repository, we cannot understand the performance of our heterogeneous algorithms without testing (or modelling them) in true heterogeneous environments with placement and scheduling algorithms. So I think adding TBB is a great move towards letting us model that. :slightly_smiling_face:

examples/run/CMakeLists.txt
krasznaa: Moved the example under `cpu/`. Just to show that TBB should always be available going forward. :smile:

extern/tbb/CMakeLists.txt
krasznaa: Where do you see 2021.7.1? :confused:
  - https://github.com/oneapi-src/oneTBB/releases
  - https://github.com/oneapi-src/oneTBB/tags

I thought I picked up the latest "blessed" version... Though even if I didn't, that shouldn't matter for our project all that much.

stephenswat (approved):
Okay, looks good to go as far as I'm concerned.

PR #280, opened 2022-11-23
Added initial throughput test for CPU full tracking

guilhermeAlmeida1:
Just wanted to get some feedback before I move onto the GPU throughput testing

### DIFF
--- a/examples/run/cpu/throughput_seq_example.cpp
+++ b/examples/run/cpu/throughput_seq_example.cpp
@@ -1,0 +1,148 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+// io
+#include "traccc/io/read.hpp"
+
+// algorithms
+#include "traccc/clusterization/clusterization_algorithm.hpp"
+#include "traccc/clusterization/spacepoint_formation.hpp"
+#include "traccc/seeding/seeding_algorithm.hpp"
+#include "traccc/seeding/track_params_estimation.hpp"
+
+// performance
+#include "traccc/performance/timer.hpp"
+
+// options
+#include "traccc/options/handle_argument_errors.hpp"
+#include "traccc/options/throughput_full_tracking_input_options.hpp"
+
+// VecMem include(s).
+#include <vecmem/memory/host_memory_resource.hpp>
+
+// System include(s).
+#include <time.h>
+
+#include <cstdlib>
+#include <exception>
+#include <iostream>
+
+namespace po = boost::program_options;
+
+int throughput_seq_run(
+    const traccc::throughput_full_tracking_input_config& i_cfg) {
+
+    // Memory resource used by the EDM.
+    vecmem::host_memory_resource host_mr;
+
+    traccc::clusterization_algorithm ca(host_mr);
+    traccc::spacepoint_formation sf(host_mr);
+    traccc::seeding_algorithm sa(host_mr);
+    traccc::track_params_estimation tp(host_mr);
+
+    traccc::demonstrator_input cells_event_vec;
+    cells_event_vec.reserve(i_cfg.loaded_events);
+
+    traccc::performance::timing_info elapsedTimes;
+
+    {
+
+        traccc::performance::timer t("File reading", elapsedTimes);
+
+        cells_event_vec = traccc::io::read(
+            i_cfg.loaded_events, i_cfg.input_directory, i_cfg.detector_file,
+            i_cfg.digitization_config_file, i_cfg.input_data_format, &host_mr);
+    }
+    // Seed random number generator
+    std::srand(std::time(0));
+
+    // Dummy count uses output of tp algorithm to ensure the compiler
+    // optimisations don't skip this step
+    std::size_t dummy_count = 0;
+
+    // Loop over events
+    for (int i = 0; i < i_cfg.processed_events; ++i) {
+
+        const int event = std::rand() % i_cfg.loaded_events;
+
+        {
+
+            traccc::performance::timer t("Events processing", elapsedTimes);
+
+            traccc::cell_container_types::host& cells_per_event =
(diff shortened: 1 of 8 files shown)


guilhermeAlmeida1:
I concede that the naming of some stuff isn't great and subject to change

performance/include/traccc/performance/timing_info.hpp
```diff
@@ -15,10 +15,14 @@
...
 
+using timing_info_pair = std::pair<std::string, std::chrono::nanoseconds>;
+
 /// Struct for storing time measurements collected in timer class
 ///
 struct timing_info {
-    std::vector<std::pair<std::string, std::chrono::nanoseconds>> data;
+    std::vector<timing_info_pair> data;
+
+    double get_time(const std::string_view timer_name);
```
krasznaa, line 25: You should return an `std::chrono` object here. That's the whole point of those types. To avoid having to always rely on documentation for knowing what unit a given value is expressed in.

examples/run/cpu/seq_batch_example.cpp
```diff
@@ -0,0 +1,147 @@
...
+              Seeding algorithm
+              -----------------------*/
+
+            auto seeds = sa(spacepoints_per_event);
+
+            /*----------------------------
+              Track params estimation
+              ----------------------------*/
+
+            auto params = tp(spacepoints_per_event, seeds);
```
krasznaa, line 99: I **think** this code is too complicated for the compiler to be quite so aggressive with its optimisations, but generally one must do "something" with the results of a calculation. Otherwise the compiler can say that, hey, nobody actually used `params` for anything. So... let's not even run the code that would create that object.

Just adding something like a "count" variable, is usually what I do in such cases. Like:

```c++
std::size_t dummy_count = 0;
...
    dummy_count += tp(spacepoints_per_event, seeds).size();
...
std::cout << "dummy_count: " << dummy_count << std::endl;
```

It's awkward as hell, I agree. But this way one can be sure that all parts of the code ran for sure.

examples/run/cpu/seq_batch_example.cpp
```diff
@@ -0,0 +1,147 @@
...
+            i_cfg.detector_file, i_cfg.digitization_config_file,
+            common_opts.input_data_format, &host_mr);
+    }
+    // Seed random number generator
+    std::srand(std::time(0));
+
+    // Loop over events
+    for (unsigned int i = 0; i < i_cfg.batch_events; ++i) {
+
+        const unsigned int event = rand() % common_opts.events;
```
krasznaa, line 68: For consistency you should use the `std::rand()` name here. Other than that, this type of simple randomisation is perfect for such applications.

krasznaa (changes requested):
I would indeed not call this "batching". That usually means something a little different. :thinking: 

I would probably change the options a bit more aggressively. So that for the two "event parameters" we would have let's say `--loaded-events` and `--processed-events`. That would make their meaning quite a bit more obvious in my mind.

krasznaa:
I should've added: Overall this is exactly what I had in mind. This is generally how the GPU versions should look like as well. But let's not put those into this PR. Let this PR just be about the "host code". :wink:

examples/run/cpu/throughput_seq_example.cpp
```diff
@@ -0,0 +1,149 @@
...
+    traccc::handle_argument_errors(vm, desc);
+
+    // Read options
+    batch_full_tracking_input_cfg.read(vm);
+
+    std::cout << "Running " << argv[0] << " "
+              << batch_full_tracking_input_cfg.detector_file << " "
+              << batch_full_tracking_input_cfg.input_directory << " "
+              << batch_full_tracking_input_cfg.loaded_events << " "
+              << batch_full_tracking_input_cfg.processed_events << std::endl;
```
krasznaa, line 146: I never liked these printouts. :thinking: Let's come up with a better printout for the new applications at least.

How about you define an

```c++
std::ostream& operator<<(std::ostream&, const traccc::throughput_full_tracking_input_options&);
```

"printout operator". And then we just rely on that to produce some meaningful output at the start of the applications.

examples/run/cpu/throughput_seq_example.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+// The main routine
+//
+int main(int argc, char* argv[]) {
+    // Set up the program options
+    po::options_description desc("Allowed options");
+
+    // Add options
+    desc.add_options()("help,h", "Give some help with the program's options");
+    traccc::throughput_full_tracking_input_config batch_full_tracking_input_cfg(
```
krasznaa, line 130: Since this variable is only used in these few lines, I would definitely make its name shorter. Plus, it still has "batch" in it. :stuck_out_tongue:

krasznaa (approved):
In principle this is okay as-is. I only have rather subjective comments at this point. You should still consider them. :wink:

examples/run/cpu/throughput_seq_example.cpp
```diff
@@ -0,0 +1,149 @@
...
+              << i_cfg.loaded_events << " loaded events." << std::endl;
+
+    std::cout << "File reading throughput (ms/event) = "
+              << elapsedTimes.get_time("File reading").count() * 1.e-6 /
+                     (double)i_cfg.loaded_events
+              << std::endl;
+    std::cout << "Processing throughput   (ms/event) = "
+              << elapsedTimes.get_time("Events processing").count() * 1.e-6 /
+                     (double)i_cfg.processed_events
+              << std::endl;
```
krasznaa, line 117: :thinking: Let's also add a printout in terms of "events per second". After all, that's the actual throughput!

beomki-yeo:
Do you guys have a further plan for multi thread CPU throughput test? (Maybe using TBB?)

guilhermeAlmeida1:
yes @beomki-yeo , that's exactly the plan. I hope to have that ready in the coming days

krasznaa:
> Do you guys have a further plan for multi thread CPU throughput test? (Maybe using TBB?)

image

:smile: See #283,

PR #281, opened 2022-11-23
Added throughput tests for sycl/cuda backends

guilhermeAlmeida1:
Depends on #280

### DIFF
--- a/examples/run/cuda/throughput_seq_example_cuda.cpp
+++ b/examples/run/cuda/throughput_seq_example_cuda.cpp
@@ -9,10 +9,10 @@
 #include "traccc/io/read.hpp"
 
 // algorithms
-#include "traccc/clusterization/clusterization_algorithm.hpp"
-#include "traccc/clusterization/spacepoint_formation.hpp"
-#include "traccc/seeding/seeding_algorithm.hpp"
-#include "traccc/seeding/track_params_estimation.hpp"
+#include "traccc/cuda/clusterization/clusterization_algorithm.hpp"
+#include "traccc/cuda/seeding/seeding_algorithm.hpp"
+#include "traccc/cuda/seeding/track_params_estimation.hpp"
+#include "traccc/device/container_h2d_copy_alg.hpp"
 
 // performance
 #include "traccc/performance/throughput.hpp"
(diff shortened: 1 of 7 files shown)


examples/run/cuda/throughput_seq_example_cuda.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+            auto seeds_buffer = sa(spacepoints_buffer);
+
+            /*----------------------------
+              Track params estimation
+              ----------------------------*/
+
+            auto params_buffer = tp(spacepoints_buffer, seeds_buffer);
+
+            traccc::bound_track_parameters_collection_types::host params_copy;
```
krasznaa, line 124: It could be worth defining this container outside of the for-loop. Would reduce the number of re-allocations a bit.

examples/run/cuda/throughput_seq_example_cuda.cpp
```diff
@@ -0,0 +1,166 @@
...
+            copy(params_buffer, params_copy);
+            dummy_count += params_copy.size();
+        }
+    }
+    std::cout << "dummy_count = " << dummy_count << std::endl;
+    std::cout << "Cold runs = " << i_cfg.cold_run_events << std::endl;
```
krasznaa, line 130: I guess it's obvious, but you should move this printout into the operator printing `traccc::throughput_options`.

examples/options/include/traccc/options/throughput_full_tracking_input_options.hpp
```diff
@@ -25,6 +25,7 @@ struct throughput_full_tracking_input_config {
     std::string digitization_config_file;
     int loaded_events;
     int processed_events;
+    int cold_run_events;
```
krasznaa, line 28: Let's make it into an `std::size_t` as well. :wink:

krasznaa (changes requested):
Really just one comment on my part, other than the fact that you'll need to modify the code a bit to work on top of #283...

krasznaa (approved):
You know what, let's accept it as-is. We can add the separate measurements of the performance of the warm-up events in a follow-up PR. :smile:

PR #287, opened 2022-11-25
Throughput Test Refactor, main branch (2022.11.25.)

krasznaa:
To cut down on the code duplication with the throughput executables, introduced 2 templated functions that implement the single-threaded and multi-threaded throughput tests on top of the "full chain algorithm" that is given to them.

The full chain algorithm can (at least as far as I can see so far) have the same interface for all runner backends. So the templated functions really only need a type and nothing else. But if it turns out later on that the "full chain algorithms" will need custom constructors, the code could be adjusted pretty easily to that as well.

I also renamed the executables a bit. I'm not super happy about these names either though, so feedback is welcome!

Also, the way that the templated code is set up, is quite fragile. In a good setup that code would be described by an `INTERFACE` library in the CMake configuration. But I just really didn't want to make things even more complicated... In the end none of that code is "public", so I thought this could be good enough.


krasznaa:
Also note that there is **no** multi-threaded SYCL executable in this PR. Because that just does not want to work. :frowning: At least not with multiple threads.

The `mutable` in the algorithms would seem like a good smoking gun for that, but I fear that there is a deeper issue with multi-threaded access to Intel devices at the moment. :frowning:

HadrienG2:
I'm pretty sure you are not supposed to use SYCL queues from multiple threads, which I think I heard you saying you did with CUDA streams earlier...

HadrienG2:
Never mind, I cross checked and they are supposed to be

krasznaa:
:thinking: I added a multi-threaded SYCL throughput test after all. As with an NVIDIA backend (in this case specifically when using version [2022-06](https://github.com/intel/llvm/releases/tag/2022-06) of the Intel compiler, in case that turns out to be important) it seems to work correctly. Its performance is quite terrible, just like the native CUDA `traccc_throughput_mt_cuda` application's. But at least it seems to work reliably.

So we'll need to see with Intel why the "Intel backend" is running into issues in a multi-threaded job.

guilhermeAlmeida1 (approved):
Looks good to me.

PR #288, opened 2022-11-30
Created cell reader into collection of cells + collection of modules. 

guilhermeAlmeida1:
This one might need a bit of discussing @krasznaa @stephenswat .

To be used the cuda specific CCA algorithm, implemented a new cell reader which returns a collection of a new struct:

```c++
struct alt_cell {
    cell cell;
    cell_module_link_type module_link;
};

using alt_cell_collection_types = collection_types<alt_cell>;
```
instead of:
```c++
container_types<cell, cell_module>
```

Also added a test for correctness of this algorithm.


io/src/csv/read_cells_alt.cpp
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    // Transform the cellCounts.nCells vector into a prefix sum for accessing
+    // positions in the result vector.
+    std::partial_sum(cellCounts.begin(), cellCounts.end(), cellCounts.begin(),
+                     [](const cell_counter& lhs, cell_counter& rhs) {
+                         rhs.nCells += lhs.nCells;
+                         return rhs;
+                     });
```
stephenswat, line 145: Maybe I am misunderstanding, but why is this necessary? If we are reading everything into a flat array there is no need to calculate the positions in the result vector, right? If we simply read the cells into memory in the same order as they appear in the file there should be no issue, I think.

stephenswat (commented):
Great work, thanks for getting started on this. However, I am a little confused as to why this needs to be as complicated as it is. Am I missing something?

io/src/csv/read_cells_alt.cpp
guilhermeAlmeida1: The problem is the sorting.  The algorithm relies on the fact that the cells are sorted by both module and channel1. The purpose of this is to not rely on the input file already being sorted.

guilhermeAlmeida1: If we do assume input data is already sorted by module that would make this all very simple indeed. Even more so if we also assume they're sorted by channel1 as well.

stephenswat: I think it's fair to make that assumption. In the end, we can always check for this sorted-ness and sort the data (on the GPU) if necessary.

core/include/traccc/edm/cell.hpp
```diff
@@ -85,4 +85,17 @@ using cell_collection_types = collection_types<cell>;
...
 using cell_container_types = container_types<cell_module, cell>;
 
+/// Alternative cell structure which contains both the cell and its module.
+///
+/// This can be used for storing all information in a single collection, whose
+/// objects need to have both the header and item information from the container
+/// types defined above.
+struct alt_cell {
+    cell c;
+    cell_module module;
```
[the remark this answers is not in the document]
guilhermeAlmeida1: Actually this was quite a useful comment. I did not use a link because in our current implementation the cell_module struct contains:
```c++
channel_id range0[2] = {std::numeric_limits<channel_id>::max(), 0};
channel_id range1[2] = {std::numeric_limits<channel_id>::max(), 0};
```
which is defined as the limits of the cells' positions belonging to this module, so you'd need to hold the module in memory and update these values as you read new cells. 
However, it turns out that we're not using these anywhere, so we might as well just entirely remove this from our EDM, no?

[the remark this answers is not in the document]
stephenswat: Seems rather useless to me. If you want the actual size (i.e. number of pixels) of the module you can get those as static properties of the modules themselves, the range in which activations lie seems&mdash;to me&mdash;rather useless.

io/src/csv/read_cells_alt.cpp
guilhermeAlmeida1: @krasznaa would be good to have your two cents on this. iirc we talked about this when a sometime ago and you didn't seemed very willing to depend on this assumption.

guilhermeAlmeida1: I ended up leaving the (overcomplicated) code to allow input data to not come in sorted by module, since we're currently not too concerned about file reading times and it is safer as is.

guilhermeAlmeida1:
Also added similar functionality for reading spacepoints (needed for the seeding_examples). You might notice that here I did not need to create a new EDM struct with the header information, which is because the information we're storing in the spacepoints container headers is not actually being used anywhere.

io/src/csv/read_cells_alt.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include <utility>
+#include <vector>
+
+namespace {
+
+/// Comparator used for sorting cells. This sorting is one of the assumptions
+/// made in the clusterization algorithm
+const auto comp = [](const traccc::alt_cell& c1, const traccc::alt_cell& c2) {
+    return c1.c.channel1 < c2.c.channel1;
+};
```
stephenswat, line 25: Is there a good reason for this to be declared using anonymous function syntax? Why not just have a straight-up function?

stephenswat (approved):
Excellent work, good to go as far as I am concerned.

io/src/csv/read_cells_alt.cpp
guilhermeAlmeida1: I thought this was the "cleanest" way of doing defining a custom comparator to call in std::sort but maybe something to do differently in the future?

stephenswat: In principle I would consider this to be a less clean solution which might also have more overhead than simply using a function pointer. That being said the compiler should be more than able to defunctionalize this.

I hope.

stephenswat: Either way let's not worry about it now.

PR #289, opened 2022-11-30
Removed redundant information from cell EDM type

guilhermeAlmeida1:
Stumbled upon this while working on #288 . It turns out that we had some information being stored in our cell modules which isn't being used anywhere other than some testing. I don't really see any reason to keep it.


stephenswat (approved):
Very good.

stephenswat (approved):

PR #291, opened 2022-12-05
File Read Fix, main branch (2022.12.05.)

krasznaa:
Made `traccc::io::read` with any kind of (host) memory resource. Since I had to realise while working on something that will soon receive its own PR, currently it's not possible to pass let's say a `vecmem::cuda::host_memory_resource` to this function. :frowning:


stephenswat (approved):
:+1:

PR #294, opened 2022-12-06
integrated nvtx tools with timer class

Xola101:
Hi Stephan.

Please delete the previous pull request. I think this should work


performance/CMakeLists.txt
```diff
@@ -4,6 +4,10 @@
 #
 # Mozilla Public License Version 2.0
 
+# Project include(s).
+include( traccc-compiler-options-cpp )
+include( traccc-compiler-options-cuda )
```
krasznaa, line 9: `traccc-compiler-options-cpp` is a good addition (we actually need to add it in a few other places as well), but `traccc-compiler-options-cuda` is not needed here. That is only needed in subdirectories where `.cu` files are being compiled. That is not going to be the case here.

performance/CMakeLists.txt
```diff
@@ -43,3 +47,9 @@ traccc_add_library( traccc_performance performance TYPE SHARED
    "src/performance/throughput.cpp" )
 target_link_libraries( traccc_performance
    PUBLIC ROOT::Core ROOT::RIO ROOT::Tree ROOT::Hist traccc::core )
+
+find_package(CUDAToolkit)
+if(CUDAToolkit_FOUND)
+        target_link_libraries( traccc_performance PRIVATE CUDA::cudart )
+                target_compile_options( traccc_performance PRIVATE )
```
krasznaa, line 54: Please remove the `target_compile_options` line. It's not doing anything.

performance/src/performance/timer.cpp
```diff
@@ -11,6 +11,9 @@
 // System include(s).
 #include <algorithm>
 
+// Nividia tool extensions library
+#include "nvtx3/nvToolsExt.h"
```
krasznaa, line 15: You will need to "protect" this code from being seen in cases where CUDA is not available. That is the "meat" of this PR in the end.

In the CMake configuration you should have something like:

```cmake
find_package(CUDAToolkit)
if (CUDAToolkit_FOUND)
   target_link_libraries( traccc_performance PRIVATE CUDA::cudart )
   target_compile_definitions( traccc_performance PRIVATE TRACCC_HAVE_NVTX )
endif()
```

Then you can surround all these statements in `timer.cpp` like:

```c++
#ifdef TRACCC_HAVE_NVTX
#  include <nvtx3/nvToolsExt.h>
#endif  // TRACCC_HAVE_NVTX
```

performance/src/performance/timer.cpp
```diff
@@ -36,6 +41,7 @@ timer::~timer() {
     } else {
         pos->second += totalTime;
     }
+    nvtxRangePop();
```
krasznaa, line 44: Move this before all of the code in the destructor. It may be either before the

```c++
const auto end = std::chrono::high_resolution_clock::now();
```

line, or just after it. (That could be debated actually.) But definitely not after everything else.

cmake/traccc-compiler-options-cuda.cmake
```diff
@@ -31,6 +31,9 @@ traccc_add_flag( CMAKE_CUDA_FLAGS "--expt-relaxed-constexpr" )
 # build.
 traccc_add_flag( CMAKE_CUDA_FLAGS_DEBUG "-G" )
 
+#Allow the use of Nvidia extensions tools.
+traccc_add_flag( CMAKE_CUDA_FLAGS_DEBUG "-lnvToolsExt" )
```
krasznaa, line 35: Does one need to link against the `libnvToolsExt.so` library to use `<nvtx3/nvToolsExt.h>`? :confused: I thought the `nvtxRangePush` / `nvtxRangePop` functions didn't need anything to be linked into the code. :confused:

We **could** use `CUDA::nvtx3`.

https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#nvtx3

Just privately linking `traccc_performance` against that I mean. But:
  - That will force us to use CMake 3.25 in our builds;
  - That still only talks about a "header only" library. So I still don't understand why this linking is needed. :confused:

krasznaa (changes requested):
Things will need to be cleaned up a bit. Not much point in letting the CI run at this point yet...

performance/CMakeLists.txt
Xola101: Hi Attila. Just removed it

performance/CMakeLists.txt
Xola101: Just removed this

performance/src/performance/timer.cpp
Xola101: Made this adjustment

cmake/traccc-compiler-options-cuda.cmake
Xola101: Looks like its working perfectly fine without it. I think I forgot to remove the link after inserting:
`find_package(CUDAToolkit)
if (CUDAToolkit_FOUND)
   target_link_libraries( traccc_performance PRIVATE CUDA::cudart )
   target_compile_definitions( traccc_performance PRIVATE TRACCC_HAVE_NVTX )
endif()`
in performance/CMakeLists.txt 😑

performance/src/performance/timer.cpp
Xola101: Adjustment made

krasznaa (approved):
As long as the CI succeeds, we should just go ahead with this.

As we discussed yesterday, we will need to split `traccc::performance` into two in a follow-up PR. Separating the timing (computing performance) code from the physics performance one.

But that can come a little later.

performance/src/performance/timer.cpp
```diff
@@ -19,10 +24,13 @@ namespace traccc::performance {
 timer::timer(const std::string_view timer_name, timing_info& t_info)
     : m_start(std::chrono::high_resolution_clock::now()),
       m_name(timer_name),
-      m_timing_info(t_info) {}
+      m_timing_info(t_info) {
+      nvtxRangePushA(timer_name.data());
```
krasznaa, line 28: The problem that we're discussing in Mattermost is coming from here. You have to surround the `nvtxRangePushA(...)` and `nvtxRangePop()` function calls with

```c++
#ifdef TRACCC_HAVE_NVTX
...
#endif
```

statements.

krasznaa (changes requested):
Also make sure that the clang-format formatting "agrees with your code".

performance/src/performance/timer.cpp
Xola101: I used clang-format to do the formatting. It passed on my folded repository.

PR #292, opened 2022-12-05
Async CUDA Clusterization, main branch (2022.12.05.)

krasznaa:
As noted in #290 the main bottleneck for the throughput tests at this point is that synchronous operations from multiple CPU threads can not work in parallel. So we pretty much serialize all GPU operations regardless of how many CPU threads we use.

This PR is meant to be a first step in getting rid of this bottleneck. It introduces some "basic machinery" code for handling CUDA streams in the project, and then it updates `traccc::cuda::clusterization_algorithm` to perform all of its memory operations and kernel launches asynchronously.

Now... This still makes the amount of "asynchronicity" in the code pretty small. But as I was very pleased to see in some profiles, there **is** some parallelism in GPU operations now.

image

Not a lot, but at least something. :smile: The (4-thread) MT throughput on my development system goes from:

```
[bash][atspot01]:build-x86_64_ubuntu2004_gcc9 > ./bin/traccc_throughput_mt_cuda --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --input_directory=tml_full/ttbar_mu200/ --cold_run_events=100 --processed_events=1000 --threads=4

Multi-threaded CUDA GPU throughput tests

>>> Throughput options <<<
Input data format   : csv
Input directory     : tml_full/ttbar_mu200/
Detector geometry   : tml_detector/trackml-detector.csv
Digitization config : tml_detector/default-geometric-config-generic.json
Loaded event(s)     : 10
Cold run event(s)   : 100
Processed event(s)  : 1000
>>> Multi-threading options <<<
CPU threads: 4

Using CUDA device: NVIDIA GeForce RTX 2060 [id: 0, bus: 1, device: 0]
Reconstructed track parameters: 15976769
Time totals:
                  File reading  1068 ms
            Warm-up processing  753 ms
              Event processing  6392 ms
Throughput:
            Warm-up processing  7.53544 ms/event, 132.706 events/s
              Event processing  6.39269 ms/event, 156.429 events/s
[bash][atspot01]:build-x86_64_ubuntu2004_gcc9 >
```

To something like:

```
[bash][atspot01]:build-x86_64_ubuntu2004_gcc9 > ./bin/traccc_throughput_mt_cuda --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --input_directory=tml_full/ttbar_mu200/ --cold_run_events=100 --processed_events=1000 --threads=4

[...]

PR #297, opened 2022-12-14
Asynchronous Container Copies, main branch (2022.12.14.)

krasznaa:
As I noted in #292, the cell H->D copies were done synchronously in the code so far. This PR makes those copies asynchronous in the CUDA throughput tests.

This required a few updates in [vecmem](https://github.com/acts-project/vecmem), so the PR updates the build to [vecmem-2.22.0](https://github.com/acts-project/vecmem/releases/tag/v0.22.0).

The MT throughput with CUDA increases just a little with this update. But the **very** promising thing is that in my latest profiles I observed quite a few things running in parallel. :smile:

image

Note how a couple of kernels manage to run in parallel as well! So I have some renewed hopes for what we may get once we make all of the algorithms run asynchronously. :wink:

### DIFF
--- a/device/common/include/traccc/device/container_d2h_copy_alg.hpp
+++ b/device/common/include/traccc/device/container_d2h_copy_alg.hpp
@@ -40,7 +40,7 @@ class container_d2h_copy_alg
         const typename CONTAINER_TYPES::const_view&)>::output_type;
 
     /// Constructor with the needed resources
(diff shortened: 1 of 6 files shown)


krasznaa:
In the new year somebody should give this a look! :wink:

stephenswat (approved):

PR #299, opened 2022-12-21
Changed device clusterization to flat EDM & cell-level parallelism

guilhermeAlmeida1:
~~This depends on #288~~ now merged
This very extensive PR changes the device clusterization algorithm to a handling single vectors' only, without any jaggeds.

The main changes in this PR are:
* Introduced Partitioning algorithm (on the CPU), which separates the cells' vector into similarly sized groups for the device clusterization algorithm.
* Introduced `alt_measurement`. Similarly to `alt_cell`, this holds a measurement and a link to a module to store all information in a single collection rather than a container with item + header.
* Removed all old common device kernels in clusterization and introduced new ones working with the new EDM types: `reduce_problem_cell`and `aggregate_cluster`.
* Changed clusterization algorithm to use new EDM types and no jaggeds. This comes with the small reversal that a part of our code is no longer single source, as the usage of barriers and group voting requires backend specific functions.

Further changes are mostly minor, and some are almost just duplication of code. Feel free to be somewhat less thourough in reviewing these.
* Changed seeding to take as input a collection of spacepoints rather than a container. (Seeding algorithm, seed finding, and the various seeding kernels.)
* Changed the sycl/cuda executables (seq_examples, seeding_examples, and full_chain_algorithms) to use new clusterization.
* Created alternative throughput tests which use new clusterization. This duplicates the already existing throughput_st / mt code as the CPU and GPU algorithms can no longer run on the same ones.

There's also still some stuff missing which I opted not to add in this PR so as not to make it even longer, namely creating "comparator" code for comparing a collection with a container. Adding a "read_alt" function which reads more than one event in parallel using TBB with the new alt_cell_collection. Create physics performance testing code using the new EDM types. Check whether CPU/Core algorithms benefit from using the jagged clusterization rather than a non-jaggged one. If it does not, this would be quite useful, as we could just ditch this duplicate code and delete the "old" edm by replacing it with the "alt".


guilhermeAlmeida1:
I ran our throughput tests at ttbar_mu300 with 10K events and it seems that (running on a NVIDIA RTX A4000) we're getting benefits on the single-threaded throughput of about 60% on both SYCL and CUDA. However, on the multi-threaded application I only saw an improvement of about 20% for CUDA and less than 10% for SYCL.

stephenswat:
Brilliant, very good! When we get this in, can we get rid of the old FastSV implementation? I assume this supersedes it.

guilhermeAlmeida1:
For sure. The only real difference between the version I implemented here and that on the CCA directory is that I did chose not to have a distinction between number of threads and number of cells per partition, so that all threads only ever deal with 1 cell, unlike what happened in the original algorithm. From my experiments, there was no gain performancewise in having the distinction between these 2.

guilhermeAlmeida1:
The CI is currently running into some issues with sycl as I was using the 2022 version on my end. Any plans to update the dependencies on the CI machine to also run the newer version or should I change my code to be compatible with the older version?

guilhermeAlmeida1:
As an additional note, these changes should only be made after ensuring the CPU-only algorithms do not incur in any performance loss from changing to this flat EDM at event read-in (the change from cells to alt-cells introduced in #288 , any other EDM types internal to clusterization don't need to change if there are losses there), as we can't really have 2 different ways of reading input data depending on whether said event is going to run on a CPU or a GPU.
I fully expect not to see any performance drop, though I'll still verify this in the coming days.

guilhermeAlmeida1:
I think this should be reasonably ready for a review at this point. For whoever is assigned this heroic task, I suggest taking a look at the initial comment I made when opening this PR which explains / organises things nicely

examples/run/common/throughput_mt_alt.hpp
```diff
@@ -0,0 +1,37 @@
...
+/// @param argc The count of command line arguments (from @c main(...))
+/// @param argv The command line arguments (from @c main(...))
+/// @param use_host_caching Flag specifying whether host-side memory caching
+///                         should be used
+/// @return The value to be returned from @c main(...)
+///
+template <typename FULL_CHAIN_ALG,
+          typename HOST_MR = vecmem::host_memory_resource>
+int throughput_mt_alt(std::string_view description, int argc, char* argv[],
+                      bool use_host_caching);
```
krasznaa, line 32: :confused: You had to update all of the `full_chain_algorithm`-s to use the new/alternate EDM. So why didn't you just update the existing `traccc::throughput_st` and `traccc::throughput_mt` functions? The "old ones" become unused by this PR, no?

examples/run/cuda/full_chain_algorithm.cpp
```diff
@@ -73,28 +73,37 @@ full_chain_algorithm::~full_chain_algorithm() {
...
+    m_copy(vecmem::get_data(cells), cells_buffer);
+    cell_module_collection_types::buffer modules_buffer(modules.size(),
+                                                        *m_cached_device_mr);
+    m_copy(vecmem::get_data(modules), modules_buffer);
+    ccl_partition_collection_types::buffer partitions_buffer(
+        partitions.size(), *m_cached_device_mr);
+    m_copy(vecmem::get_data(partitions), partitions_buffer);
+
+    // Synchronize assynchronous copies.
+    m_stream.synchronize();
```
krasznaa, line 95: It's a pity that the clusterization lost its asynchronicity by this PR and now you need to synchronize before calling the clusterization. :frowning:

examples/run/cuda/seeding_example_cuda.cpp
```diff
@@ -159,12 +170,12 @@ int seq_run(const traccc::seeding_input_config& i_cfg,
...
 
             // Compare the seeds made on the host and on the device
-            traccc::collection_comparator<traccc::seed> compare_seeds{
+            /* traccc::collection_comparator<traccc::seed> compare_seeds{
                 "seeds", traccc::details::comparator_factory<traccc::seed>{
                              traccc::get_data(spacepoints_per_event),
                              traccc::get_data(spacepoints_per_event)}};
             compare_seeds(vecmem::get_data(seeds),
-                          vecmem::get_data(seeds_cuda));
+                          vecmem::get_data(seeds_cuda)); */
```
krasznaa, line 178: Doesn't the comparison work between the host- and device-created `alt_seed`-s?

krasznaa (changes requested):
Wow... is this a monster PR...

core/include/traccc/edm/ccl_partition.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace partitioning {
+// Define number of cells to put together in each partition. Equal to the number
+// of threads per block in the CCL kernel.
+static constexpr std::size_t MAX_CELLS_PER_PARTITION = 1024;
+}  // namespace partitioning
+
+/// CCL partition used for dividing cell collection into similarly sized
+/// fragments
+struct ccl_partition {
```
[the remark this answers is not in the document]
guilhermeAlmeida1: Sounds good to me, I might even generalise it a bit more and make it `traccc::partition` . As I suspect it might come in handy in the future for other types other cells.

device/common/include/traccc/clusterization/device/form_spacepoints.hpp
```diff
@@ -25,21 +27,20 @@ namespace traccc::device {
...
+/// @param[out] spacepoints_view    Collection of spacepoints
 ///
 TRACCC_HOST_DEVICE
 inline void form_spacepoints(
-    std::size_t globalIndex,
-    measurement_container_types::const_view measurements_view,
-    vecmem::data::vector_view<const device::prefix_sum_element_t>
-        measurements_prefix_sum_view,
-    spacepoint_container_types::view spacepoints_view);
+    const unsigned int globalIndex,
```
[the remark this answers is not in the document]
guilhermeAlmeida1: Not a great reason, just thought I'd reduce the memory usage given that in reality we won't ever have it be larger than an `unsigned int`, but I'm happy to keep it at `std::size_t`

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -89,181 +276,64 @@ clusterization_algorithm::clusterization_algorithm(
...
-    m_copy(cluster_sizes_buffer, cluster_sizes,
-           vecmem::copy::type::copy_type::device_to_host);
-    m_stream.synchronize();
+    // Copy number of measurements to host
+    unsigned int num_measurements_host;
+    CUDA_ERROR_CHECK(cudaMemcpy(&num_measurements_host,
+                                num_measurements_device.get(),
+                                sizeof(unsigned int), cudaMemcpyDeviceToHost));
+    // I doubt this is necessary here
+    CUDA_ERROR_CHECK(cudaDeviceSynchronize());
```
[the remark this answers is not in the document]
guilhermeAlmeida1: Just thought there wouldn't really be a big difference between having a synchronous copy or having an assynchronous one immediately followed by a synchronisation point, given that it's just a single 32bit copy. And being synchronous just makes the code slightly more compact.

examples/run/common/throughput_mt_alt.hpp
guilhermeAlmeida1: The old ones are still used in the CPU throughput applications. They will also be replaced as soon as the cpu algorithms using the new edm are fully implemented.

examples/run/cuda/seeding_example_cuda.cpp
guilhermeAlmeida1: The host still functions on `seed` rather than `alt_seed`. I could add code which compares between the two but it seems unnecessary as it would be unused very shortly.

device/cuda/src/clusterization/clusterization_algorithm.cu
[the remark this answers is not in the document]
krasznaa: With a synchronous copy you need to:
  - Wait for the kernel that produced the value on the device, before the copy;
  - Schedule the copy, and wait for its results;

If you set it up asynchronously, you can run the CPU code that schedules the copy, while the GPU is still in the process of producing the output. (You don't need to have a synchronization point before the copy.) You do need a synchronization point **after** the copy, since you want to make use of the value copied back to the host. But you will need to do fewer things on the CPU while the GPU is idle. (Since the setup of the copy is done in that case while the GPU is ideally already busy.)

examples/run/cuda/seeding_example_cuda.cpp
krasznaa: Okay, never mind then. You'll re-activate the comparison when the host code is also migrated to the new EDM.

examples/run/common/throughput_mt_alt.hpp
krasznaa: :+1:

core/include/traccc/edm/alt_measurement.hpp
```diff
@@ -0,0 +1,33 @@
...
+namespace traccc {
+
+/// Alternative measurement structure which contains both the measurement and
+/// a link to its module (held in a separate collection).
+///
+/// This can be used for storing all information in a single collection, whose
+/// objects need to have both the header and item information from the
+/// measurement container types.
+struct alt_measurement {
+    measurement meas;
```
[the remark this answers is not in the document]
guilhermeAlmeida1: Sounds good, I'll have to make the same changes to `alt_cell`, but since the code for that is already merged in, I'll do so in a separate PR.

core/include/traccc/edm/ccl_partition.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+// traccc include(s).
+#include "traccc/edm/container.hpp"
+
+namespace traccc {
+
+namespace partitioning {
+// Define number of cells to put together in each partition. Equal to the number
+// of threads per block in the CCL kernel.
+static constexpr std::size_t MAX_CELLS_PER_PARTITION = 1024;
```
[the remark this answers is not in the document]
guilhermeAlmeida1: This is indeed not a good place. I'm wondering if I should just add a new header (e.g. `cell_partitioning_options.h`) with just this information to core or also move the whole partitioning algorithm to the device directory. Despite this being a cpu algorithm, it really is only used in the device pipelines.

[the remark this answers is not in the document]
krasznaa: Functions that need this information to interpret/process their inputs correctly, should receive the parameter explicitly.

At which point the algorithm classes/structs should have their own settings. I.e. the host code may want to use something larger than 1024, while Intel GPUs for instance can definitely only use a maximum of 256. So it should be an algorithm specific value. And possibly not even a fixed, but configurable value for the algorithm(s).

device/common/include/traccc/clusterization/device/reduce_problem_cell.hpp
```diff
@@ -0,0 +1,39 @@
...
+/// @param[in] tid      Current thread id
+/// @param[in] part     Current partition
+/// @param[out] ajc     Number of adjacent cells
+/// @param[out] ajv     Indices of adjacent cells
+///
+TRACCC_HOST_DEVICE
+inline void reduce_problem_cell(
+    const alt_cell_collection_types::const_device& cells,
+    const unsigned short tid, const ccl_partition& part, unsigned char& adjc,
+    unsigned short adjv[8]);
```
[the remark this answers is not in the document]
guilhermeAlmeida1: Sounds good, will change to this after https://github.com/acts-project/vecmem/issues/215 is addressed

core/include/traccc/edm/ccl_partition.hpp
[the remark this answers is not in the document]
guilhermeAlmeida1: > I.e. the host code may want to use something larger than 1024, while Intel GPUs for instance can definitely only use a maximum of 256

In the host code I don't see any point in using partitioning, hence I was suggesting moving this to device
The part of having such a low limit for Intel GPUs might be more crucial though. At high pile-up, with the current way the partitioning algorithm works, I think we'd run into the breaking case where it's not possible to partition with this small size. I was planning on making a change to the partitioning algorithm which would prevent this later but I might instead do so in this PR.

guilhermeAlmeida1:
I have addressed the comments above. Also added the functionality in #308 but, instead of entirely removing the partition class, I made it into an alias of unsigned int. I am, however, fine with just removing it.

guilhermeAlmeida1:
I also intend on adding #309 on top of this but in a separate PR.

core/include/traccc/edm/alt_measurement.hpp
```diff
@@ -0,0 +1,40 @@
...
+/// measurement container types.
+struct alt_measurement {
+    /// Local 2D coordinates for a measurement on a detector module
+    point2 local{0., 0.};
+    /// Variance on the 2D coordinates of the measurement
+    variance2 variance{0., 0.};
+    // TODO: We should consider removing the cluster link, as it is only used in
+    // a CI test.
+    /// Cluster link
+    std::size_t cluster_link = std::numeric_limits<std::size_t>::infinity();
```
stephenswat, line 31: Let's remove this and the test that uses it.

core/include/traccc/edm/alt_seed.hpp
```diff
@@ -0,0 +1,50 @@
...
+struct alt_seed {
+
+    using link_type = spacepoint_collection_types::host::size_type;
+
+    link_type spB_link;
+    link_type spM_link;
+    link_type spT_link;
+
+    scalar weight;
+    scalar z_vertex;
```
stephenswat, line 26: We can discuss this later, perhaps on a different PR, but this information should not really live in the seed EDM.

device/common/src/clusterization/partitioning_algorithm.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace {
+// Check if two cells have at least 1 full empty column between them or belong
+// to different modules in order to be partitioned. This assumes sorting by
+// channel1 already done
+bool isFarEnough(const traccc::alt_cell& c1, const traccc::alt_cell& c2) {
+    if (c1.module_link != c2.module_link || c1.c.channel1 + 1 < c2.c.channel1) {
+        return true;
+    }
+    return false;
```
stephenswat, line 22:
```suggestion
    return c1.module_link != c2.module_link || c1.c.channel1 + 1 < c2.c.channel1;
```

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -11,13 +11,9 @@
 #include "traccc/cuda/utils/definitions.hpp"
 
 // Project include(s)
-#include "traccc/clusterization/device/connect_components.hpp"
-#include "traccc/clusterization/device/count_cluster_cells.hpp"
-#include "traccc/clusterization/device/create_measurements.hpp"
-#include "traccc/clusterization/device/find_clusters.hpp"
+#include "traccc/clusterization/device/aggregate_cluster.hpp"
```
stephenswat, line 14: I am going to assume that this is just a copy of the CUDA code in `component_connection.cu`, so I will skip over it.

device/sycl/src/clusterization/clusterization_algorithm.sycl
```diff
@@ -9,45 +9,325 @@
 #include "../utils/get_queue.hpp"
```
stephenswat, line 9: Do we have this code hooked up to our CCA tests? Would be useful to see whether this code does what it's supposed to do. I trust that it's a faithful translation but I am worried about subtle differences in memory and execution models between CUDA and SYCL.

examples/run/kokkos/seeding_example_kokkos.cpp
```diff
@@ -123,10 +124,26 @@ int seq_run(const traccc::seeding_input_config& i_cfg,
...
+
             {  // Spacepoin binning for kokkos
-                m_spacepoint_binning(traccc::get_data(spacepoints_per_event));
+                traccc::performance::timer t("Spacepoint binning (kokkos)",
+                                             elapsedTimes);
+                m_spacepoint_binning(
+                    vecmem::get_data(alt_spacepoints_per_event));
             }
 
+            std::cout << "here " << __LINE__ << std::endl;
```
stephenswat, line 145: I am guessing this is a debugging leftover. :slightly_smiling_face:

examples/run/kokkos/seeding_example_kokkos.cpp
```diff
@@ -123,10 +124,26 @@ int seq_run(const traccc::seeding_input_config& i_cfg,
...
+            {
+                traccc::performance::timer t("Alt hit reading  (cpu)",
+                                             elapsedTimes);
+                // Read the hits from the relevant event file
+                alt_spacepoints_per_event = traccc::io::read_spacepoints_alt(
+                    event, common_opts.input_directory, surface_transforms,
+                    common_opts.input_data_format, &host_mr);
+            }  // stop measuring hit reading timer
+
+            std::cout << "here " << __LINE__ << std::endl;
```
stephenswat, line 136: See below.

examples/run/sycl/full_chain_algorithm.hpp
```diff
@@ -82,8 +89,12 @@ class full_chain_algorithm
...
     /// @{
 
-    /// Host->Device cell copy algorithm
-    device::container_h2d_copy_alg<cell_container_types> m_host2device;
+    /// The number of cells to put together in each partition.
+    /// Equal to the number of threads in the clusterization kernels.
+    /// Adapt to different GPUs' capabilities.
+    unsigned short m_max_cells_per_partition;
+    /// Partitioning algorithm
+    device::partitioning_algorithm m_partitioning;
```
stephenswat, line 97: Do we need to have this as a dedicated partitioning algorithm rather than an algorithm that is embedded in the clusterization algorithm? It's not like the partitioning code has any use outside of clusterization.

core/include/traccc/edm/alt_measurement.hpp
guilhermeAlmeida1: Okay I'm removing this from the alt_measurement class and it when we move from measurement to alt_measurement I'll also remove the test.

examples/run/kokkos/seeding_example_kokkos.cpp
guilhermeAlmeida1: whoops

examples/run/sycl/full_chain_algorithm.hpp
guilhermeAlmeida1: Yeah this was a poor design choice from me. I'll change it up to just being a part of the clusterization alg.

stephenswat: Okay, that sounds good.

device/sycl/src/clusterization/clusterization_algorithm.sycl
guilhermeAlmeida1: This would definately be good but if I understand correctly we currently don't have a way of executing SYCL code in our CI tests. I can however double check this by doing some further testing locally.

examples/run/cuda/full_chain_algorithm.cpp
guilhermeAlmeida1: This was my mistake. This synchronisation point was unnecessary and has been removed

beomki-yeo:
I have a question in `alt_measurement` implementation. How would I retrieve the set of measurements in a certain module, for example in combinatorial kalman filtering?

guilhermeAlmeida1:
@beomki-yeo even with the current EDM we are not using the detector modules for anything outside of clustering, and this information is not passed down to seeding. If we need the detector modules for CKF that's something we'd need to add at that point.

Regarding the performance gain, it's down to both, as the CCA algorithm heavily relies on having a flat EDM, and I don't see why we would want to make it less efficient by sticking to the old one.

beomki-yeo:
> the CCA algorithm heavily relies on having a flat EDM
@stephenswat can clarify this but I would think that a flat EDM and CCA are orthogonal. Could you educate me about how a flat EDM contributes to the performance gain? Also have you tested the original algorithm with a flat EDM and compare with the current performance? 

And.. is it necessary to remove the original algorithm? :thinking:

guilhermeAlmeida1:
@beomki-yeo on top of jagged vectors just being inherently slightly slower to handle with copying, our "prefix sums", etc. the module-based jaggedness results in very differently sized subvectors. By having all cells in a single vector it eases the creation of these partitions which all have around the same size which makes the algorithm a lot faster by fully using all threads.
I've nothing against keeping the old algorithm in, but I don't really see any major benefit to it.

guilhermeAlmeida1:
`Also have you tested the original algorithm with a flat EDM and compare with the current performance?`
I have not, this could be interesting to look at though

stephenswat (approved):
Okay this has been sitting here for more than long enough, if the CI is happy with the changes let's put them in.

PR #300, opened 2023-01-04
Alpaka first steps

StewMH:
Adds the basic infrastructure for Alpaka. No actual tracking code in there so far, just a very basic test. May need some help adding it to CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,8 @@ option( TRACCC_BUILD_FUTHARK "Build the Futhark sources included in traccc"
    FALSE )
 option( TRACCC_BUILD_KOKKOS "Build the Kokkos sources included in traccc"
    FALSE )
+option( TRACCC_BUILD_ALPAKA "Build the Alpaka sources included in traccc"
+   FALSE )
 option( TRACCC_BUILD_TESTING "Build the (unit) tests of traccc" TRUE )
 option( TRACCC_BUILD_EXAMPLES "Build the examples of traccc" TRUE )
 
@@ -135,6 +137,20 @@ if( TRACCC_SETUP_KOKKOS )
       find_package( Kokkos REQUIRED )
    else()
       add_subdirectory( extern/kokkos )
+   endif()
+endif()
+
+# Set up Alpaka.
+option( TRACCC_SETUP_ALPAKA
+   "Set up the Alpaka library" TRUE )
+option( TRACCC_USE_SYSTEM_ALPAKA
+   "Pick up an existing installation of Alpaka from the build environment"
+   ${TRACCC_USE_SYSTEM_LIBS} )
+if( TRACCC_SETUP_ALPAKA )
+   if( TRACCC_USE_SYSTEM_ALPAKA )
+      find_package( alpaka REQUIRED )
+   else()
+      add_subdirectory( extern/alpaka )
    endif()
 endif()
 
--- a/tests/alpaka/CMakeLists.txt
+++ b/tests/alpaka/CMakeLists.txt
@@ -1,0 +1,12 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+traccc_add_test( alpaka
+   alpaka_main.cpp
+   LINK_LIBRARIES
+   GTest::gtest
+   alpaka
+)
--- a/tests/alpaka/alpaka_main.cpp
+++ b/tests/alpaka/alpaka_main.cpp
@@ -32,7 +32,6 @@ struct VectorOpKernel {
         // but for simplicity we just repeat the type here
         if(globalThreadIdx < n)
         {
-            std::cout << "Hello from " << globalThreadIdx << std::endl;
             result[globalThreadIdx] = process(acc, globalThreadIdx);
         }
         
@@ -55,14 +54,13 @@ GTEST_TEST(AlpakaBasic, VectorOp) {
     using Queue = Queue<Acc, Blocking>;
     auto queue = Queue{devAcc};
 
-    uint32_t n = 16;
+    uint32_t n = 10000;
 
     uint32_t blocksPerGrid = n;
     uint32_t threadsPerBlock = 1;
     uint32_t elementsPerThread = 4;
     using WorkDiv = WorkDivMembers<Dim, Idx>;
     auto workDiv = WorkDiv{blocksPerGrid, threadsPerBlock, elementsPerThread};
-    std::cout << workDiv << std::endl;
 
     // Create a device for host for memory allocation, using the first CPU available
     auto devHost = getDevByIdx<DevCpu>(0u);
--- a/tests/alpaka/alpaka_main.cpp
+++ b/tests/alpaka/alpaka_main.cpp
@@ -16,37 +16,37 @@
 #include <cstdint>
 #include <iostream>
 
-template<typename Acc>
-ALPAKA_FN_ACC float process(Acc const & acc, uint32_t idx)
-{
+template <typename Acc>
+ALPAKA_FN_ACC float process(Acc const& acc, uint32_t idx) {
     return alpaka::math::sin(acc, idx);
 }
 
 struct VectorOpKernel {
-    template<typename Acc>
-    ALPAKA_FN_ACC void operator()(Acc const & acc, float* result, uint32_t n) const {
+    template <typename Acc>
+    ALPAKA_FN_ACC void operator()(Acc const& acc, float* result,
+                                  uint32_t n) const {
         using namespace alpaka;
-        auto const globalThreadIdx = getIdx<alpaka::Grid, alpaka::Threads>(acc)[0u];
+        auto const globalThreadIdx =
+            getIdx<alpaka::Grid, alpaka::Threads>(acc)[0u];
 
         // It is possible to get index type from the accelerator,
         // but for simplicity we just repeat the type here
-        if(globalThreadIdx < n)
-        {
+        if (globalThreadIdx < n) {
             result[globalThreadIdx] = process(acc, globalThreadIdx);
         }
-        
     }
 };
 
-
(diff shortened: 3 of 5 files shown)


tests/alpaka/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+#
+# (c) 2022 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+traccc_add_test( alpaka
+   alpaka_main.cpp
+   LINK_LIBRARIES
+   GTest::gtest
+   alpaka
```
krasznaa, line 11: The `alpaka` target (I believe) does work in all circumstances, but just to show a bit more clearly that this is something "external", I would prefer to use the `alpaka::alpaka` target instead.

https://github.com/alpaka-group/alpaka/blob/develop/cmake/alpakaCommon.cmake#L154

tests/alpaka/alpaka_main.cpp
```diff
@@ -0,0 +1,97 @@
...
+        auto const globalThreadIdx =
+            getIdx<alpaka::Grid, alpaka::Threads>(acc)[0u];
+
+        // It is possible to get index type from the accelerator,
+        // but for simplicity we just repeat the type here
+        if (globalThreadIdx < n) {
+            result[globalThreadIdx] = process(acc, globalThreadIdx);
+        }
+    }
+};
```
krasznaa, line 38: :thinking: For this simple test I would rather just go with a simple function.

```c++
template <typename Acc>
void VectorOpKernel(Acc const& acc, float* result, uint32_t n) {...}
```

tests/alpaka/alpaka_main.cpp
```diff
@@ -0,0 +1,97 @@
...
+/// Copy vector to device, set element[i] = sin(i) on device, then compare with
+/// same calculation on host
+GTEST_TEST(AlpakaBasic, VectorOp) {
+    using namespace alpaka;
+    using Dim = DimInt<1>;
+    using Idx = uint32_t;
+
+    using Acc = ExampleDefaultAcc<Dim, Idx>;
+    std::cout << "Using alpaka accelerator: " << alpaka::getAccName<Acc>()
+              << std::endl;
```
krasznaa, line 49: Please remove this printout. It makes the output of the test "less clean than I'd like". :stuck_out_tongue:

```
[bash][celeborn]:build > ./bin/traccc_test_alpaka 
[----------] 1 test from AlpakaBasic
[ RUN      ] AlpakaBasic.VectorOp
Using alpaka accelerator: AccCpuSerial<1,j>
[       OK ] AlpakaBasic.VectorOp (0 ms)
[----------] 1 test from AlpakaBasic (0 ms total)

[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 1 test.
[bash][celeborn]:build >
```

tests/alpaka/alpaka_main.cpp
```diff
@@ -0,0 +1,97 @@
...
+    alpaka::memcpy(queue, bufHost, bufAcc);
+    // Calculate on the host and compare result
+    for (uint32_t i = 0u; i < n; i++) {
+        bool testPassed = std::abs(bufHost[i] - std::sin(i)) < 1e-3;
+        if (!testPassed) {
+            std::cout << "bufHost[" << i << "]: " << bufHost[i] << std::endl;
+            std::cout << "bufAcc[" << i << "]: " << bufAcc[i] << std::endl;
+            std::cout << "std::sin(" << i << "): " << std::sin(i) << std::endl;
+        }
+        ASSERT_TRUE(testPassed);
```
krasznaa, line 89: I'm not a fan of this. :frowning: I would rather vote for:

```c++
EXPECT_FLOAT_EQ(bufHost[i], std::sin(i));
```

https://github.com/google/googletest/blob/main/googletest/include/gtest/gtest.h#L1960

We can leave it to GTest to give us useful printouts about failures.

tests/alpaka/alpaka_main.cpp
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+int main(int argc, char** argv) {
+    testing::InitGoogleTest(&argc, argv);
+    int r = RUN_ALL_TESTS();
+    return r;
+}
```
krasznaa, line 97: You shouldn't have a `main()` function in here. Instead, just add `GTest::gtest_main` to the libraries linked against the test. I.e.

```diff
diff --git a/tests/alpaka/CMakeLists.txt b/tests/alpaka/CMakeLists.txt
index f581cca..e44dfa4 100644
--- a/tests/alpaka/CMakeLists.txt
+++ b/tests/alpaka/CMakeLists.txt
@@ -8,5 +8,6 @@ traccc_add_test( alpaka
    alpaka_main.cpp
    LINK_LIBRARIES
    GTest::gtest
-   alpaka
+   GTest::gtest_main
+   alpaka::alpaka
 )
diff --git a/tests/alpaka/alpaka_main.cpp b/tests/alpaka/alpaka_main.cpp
index 2997745..78fbd4f 100644
--- a/tests/alpaka/alpaka_main.cpp
+++ b/tests/alpaka/alpaka_main.cpp
@@ -89,9 +89,3 @@ GTEST_TEST(AlpakaBasic, VectorOp) {
         ASSERT_TRUE(testPassed);
     }
 }
-
-int main(int argc, char** argv) {
-    testing::InitGoogleTest(&argc, argv);
-    int r = RUN_ALL_TESTS();
-    return r;
-}
```

krasznaa (changes requested):
The "build" of Alpaka looks good. (I didn't even know up until now that Alpaka is a header-only package...) My comments are mainly about the test executable.

On top of all of those, I would also like to rename `alpaka_main.cpp` to `alpaka_basic.cpp`. In line with the `AlpakaBasic` test group that you define in that file.

In a future PR we should also teach the CI how to test the Alpaka code. But this PR should not worry about that yet.

tests/alpaka/alpaka_main.cpp
krasznaa: Scratch this one. Alpaka is not as flexible with "function objects" as SYCL... :frowning:

```
[100%] Building CXX object tests/alpaka/CMakeFiles/traccc_test_alpaka.dir/alpaka_main.cpp.o
/home/krasznaa/ATLAS/projects/traccc/traccc/tests/alpaka/alpaka_main.cpp:68:5: error: no matching function for call to 'exec'
    alpaka::exec<Acc>(queue, workDiv, VectorOpKernel,
    ^~~~~~~~~~~~~~~~~
/home/krasznaa/ATLAS/projects/traccc/build/_deps/alpaka-src/include/alpaka/kernel/Traits.hpp:269:25: note: candidate template ignored: couldn't infer template argument 'TKernelFnObj'
    ALPAKA_FN_HOST auto exec(TQueue& queue, TWorkDiv const& workDiv, TKernelFnObj const& kernelFnObj, TArgs&&... args)
                        ^
1 error generated.
make[2]: *** [tests/alpaka/CMakeFiles/traccc_test_alpaka.dir/build.make:76: tests/alpaka/CMakeFiles/traccc_test_alpaka.dir/alpaka_main.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:4987: tests/alpaka/CMakeFiles/traccc_test_alpaka.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

StewMH:
Thanks Attila, I incorporated all your suggestions. Not sure if FLOAT_EQ will match if we do the calculation on a GPU (at the moment I can only get CPU backends to work) but we can figure that out when we get to it.

krasznaa:
> Not sure if FLOAT_EQ will match if we do the calculation on a GPU (at the moment I can only get CPU backends to work) but we can figure that out when we get to it.

If `EXPECT_FLOAT_EQ` turns out to be too strict for these tests, we can go to `EXPECT_NEAR` with some slightly larger uncertainty.

What worries me much more at the moment is the Boost version requirement of Alpaka. :frowning: Apparently we'll need to update our CI system before we could go ahead with this. :frowning:

StewMH:
@krasznaa anything I can do here? Can we update the Boost version?

stephenswat:
I will open a pull request on the ACTS machines repository to get an environment with a more modern Boost version.

StewMH:
Thanks @stephenswat - looks like just moving to ubuntu2204 would be enough?

stephenswat:
Indeed, that was exactly what I was about to do! :smile:

stephenswat:
Actually, the problem will be slightly more complex; until we move all out CI jobs to Ubuntu 22.04 (something which I am not opposed to doing per se) we will still compile most of the code with Ubuntu 20.04. If we compile the Alpaka sources by default (even when `TRACCC_BUILD_ALPAKA` is disabled) we will run into this error. We'll need to make sure that `TRACCC_SETUP_ALPAKA` is disabled if `TRACCC_BUILD_ALPAKA` is not enabled.

CMakeLists.txt
```diff
@@ -138,6 +140,20 @@ if( TRACCC_SETUP_KOKKOS )
    endif()
 endif()
 
+# Set up Alpaka.
+option( TRACCC_SETUP_ALPAKA
+   "Set up the Alpaka library" TRUE )
```
stephenswat, line 145:
```suggestion
   "Set up the Alpaka library" ${TRACCC_BUILD_ALPAKA} )
```

Something like this would do the trick, although CMake cache variables might become incoherent if we solve it like this... I would still suggest going with this solution though.

stephenswat:
Okay so with that change made, the code should compile. But sneakily the Alpaka code will not be enabled by default. We will need to add a new CI job to actually build the Alpaka code. I suggest we start without compiling with a CUDA backend by adding a line to the build matrix in `.github/workflows/builds.yml` that looks like this:

```yaml
          - name: ALPAKA
            container: ghcr.io/acts-project/ubuntu2204:v33
```

StewMH:
Done, thanks!

stephenswat:
Did you force push away the change to the CMake build system that I had suggested?

stephenswat (approved):
Looks good to me!

PR #304, opened 2023-01-19
Changed CPU clustering to use flat vectors of alt_cells as input and alt_measurements internally.

guilhermeAlmeida1:
Depends on #299 yet to be merged  (hence the amount of displayed changed loc. In truth it's only a few hundred).

This is the first half of the changes to the CPU clustering to use flat vectors. The second part will contain changing the output to flat spacepoints as output.

Performance-wise this actually seems to have improved things.


stephenswat:
Can you rebase this?

beomki-yeo:
I am sorry but Could you please leave the CPU clustering intact? I am doing something on top of current measurement container type

guilhermeAlmeida1:
> I am sorry but Could you please leave the CPU clustering intact? I am doing something on top of current measurement container type

Sure, I'll keep this as a draft for now, but is there some time frame for this? The plan was to move all of the CPU algorithm to use the same new EDM as the GPU very soon (containers of cells, measurements, and spacepoints would all change to collections of alt_cells, alt_measurements, and spacepoints respectively).

beomki-yeo:
Let's do discuss at the next meeting if I can join...

beomki-yeo:
I couldn't attend the last meeting unfortunately but let's merge this in. We need to think of how to collect the cells in a module at some point. (A buffer from partition algorithm could be used I hope)

guilhermeAlmeida1:
> I couldn't attend the last meeting unfortunately but let's merge this in. We need to think of how to collect the cells in a module at some point. (A buffer from partition algorithm could be used I hope)

I could see this being done with not-too-many changes to our current code. We would need to change the way partitioning happens to ensure no module would be split into 2 
(basically remove the channel1 check from here: https://github.com/acts-project/traccc/blob/main/device/cuda/src/clusterization/clusterization_algorithm.cu#L182-L185 )
Then add an aditional vector of size equal to the number of modules which would store the positions of the measurements belonging to that same module. 

This could then be calculated in a similar way to what we're doing in count/find doublets/triplets where by atomically counting the number of items in each bin (in this case, module) we can also define the positions in the output vector in which they'll be filled.

for reference, see:
https://github.com/acts-project/traccc/blob/main/device/common/include/traccc/seeding/device/impl/count_doublets.ipp#L114-L120
https://github.com/acts-project/traccc/blob/main/device/common/include/traccc/seeding/device/impl/find_doublets.ipp#L66-69

guilhermeAlmeida1:
I'm closing this PR as I opted to take a different route and change the whole thing at once. A PR which replaces this one will follow in the coming days.

PR #308, opened 2023-01-25
Remove CCL partition data structure

stephenswat:
As it stands, the partition data structure keeps track of the starting index and the size of the partition, but this is in fact unnecessary; if we keep the partitions in order, we can simply look at the next partition to find out the size of the current partition. This allows us to save space and simplify the code.


guilhermeAlmeida1 (approved):
This looks clean, I'll make the same changes to #299

PR #309, opened 2023-01-25
Implement cell partitioning in CUDA

stephenswat:
Previously, the cell partitioning was done on a CPU, necessitating an unwanted transfer of partition data from the host to the device. This commit implements partitioning purely in CUDA code, allowing it to run entirely on a GPU without the CPU being involved.

### DIFF
--- a/device/cuda/src/cca/component_connection.cu
+++ b/device/cuda/src/cca/component_connection.cu
@@ -14,6 +14,7 @@
 #include "vecmem/containers/vector.hpp"
 #include "vecmem/memory/allocator.hpp"
 #include "vecmem/memory/binary_page_memory_resource.hpp"
+#include "vecmem/memory/cuda/device_memory_resource.hpp"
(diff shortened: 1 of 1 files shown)


guilhermeAlmeida1:
I'm wondering if the sorting of the inital "temporary" partitions is even necessary. Nothing stands out to me as problematic about having just a bunch of unordered cells belonging to different modules in the same partition at the end.

stephenswat:
@guilhermeAlmeida1 there is no need to remove your comment, your concerns and comments are valuable; also to people reading back PRs later. :slightly_smiling_face:

stephenswat:
Following a conversation I had with Guilherme about this the merging of partitions, which was previously done sequentially by the lead thread in each block, is now done more efficiently and elegantly in a parallel fashion.

guilhermeAlmeida1 (approved):
Looks great to me! A very clean implementation

stephenswat:
> Looks great to me! A very clean implementation

PR #310, opened 2023-01-31
Fix unused parameter in CCA code

stephenswat:
Somehow, #309 introduced a CI error that caught an unused variable. This is odd, because the error was not caught by CI before. This commit fixes the error by removing (or rather, removing the name) for this parameter.

### DIFF
--- a/device/cuda/src/cca/component_connection.cu
+++ b/device/cuda/src/cca/component_connection.cu
@@ -963,7 +963,7 @@ __global__ void partition_sorting_kernel(unsigned* out,
 }
 
 std::tuple<vecmem::unique_alloc_ptr<unsigned[]>, std::size_t> partition_gpu(
-    const cell_container_types::host& data, vecmem::memory_resource& mem,
+    const cell_container_types::host&, vecmem::memory_resource& mem,
     const details::cell_container cells) {
     /*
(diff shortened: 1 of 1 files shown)


device/cuda/src/cca/component_connection.cu
```diff
@@ -963,7 +963,7 @@ __global__ void partition_sorting_kernel(unsigned* out,
 }
 
 std::tuple<vecmem::unique_alloc_ptr<unsigned[]>, std::size_t> partition_gpu(
-    const cell_container_types::host& data, vecmem::memory_resource& mem,
+    const cell_container_types::host&, vecmem::memory_resource& mem,
```
guilhermeAlmeida1, line 966:
```suggestion
    vecmem::memory_resource& mem,
```
I don't see why not just remove this entirely

stephenswat: This way it shares the same API as the `partition_cpu` function. Not sure if we should keep them both.

guilhermeAlmeida1: I see. I like keeping both in, since this is sort of a separate thing to our main example chains. Might as well leave it

guilhermeAlmeida1 (approved):

PR #312, opened 2023-02-02
Implementation of Nvidia NVTX functions into timer class to use for performance studies

Xola101:


krasznaa (approved):
I think this looks good now. As long as the CI is happy, let's finally merge it in. :+1:

PR #313, opened 2023-02-06
Fixed partition merging in Cuda CCA

guilhermeAlmeida1:
This fixes a bug in the partitioning kernel implemented in #309 in which the output partitions can have slightly larger size than the maximum allowed.
I also changed the host code to have the `slots` variable be greater than `max_cells_per_partition`. If not, the algorithm will output partitions with sizes equal or smaller than the first, rather then the latter.

### DIFF
--- a/device/cuda/src/cca/component_connection.cu
+++ b/device/cuda/src/cca/component_connection.cu
@@ -817,24 +817,25 @@ __global__ void partition_kernel(const cell_container cells, unsigned* out,
      * actual CCL kernel.
      *
      * This code works by overriding the existing array of partition indices.
-     * The `old_idx` variable denotes the end of the old array, the `base_idx`
+     * The `old_size` variable denotes the end of the old array, the `base_idx`
      * variable denotes the current starting index in the old array, and the
      * `tmp_idx` variable denotes the index we write partitions to in the new
      * array. The old and new array are actually the same memory, but the
      * writing index for the new points will always be behind the reading
      * indices in the old part, so this is safe!
      */
-    const unsigned old_idx = tmp_idx;
+    const unsigned old_size = tmp_idx;
 
     __syncthreads();
+
+    unsigned base_idx = 0;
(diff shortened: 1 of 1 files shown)


device/cuda/src/cca/component_connection.cu
```diff
@@ -986,8 +983,11 @@ std::tuple<vecmem::unique_alloc_ptr<unsigned[]>, std::size_t> partition_gpu(
      * it is sometimes desirable to process more than one cell per thread. This
      * slots variable determines the number of cells that is examined per
      * block.
+     * Slots should always be larger or equal to max_cells_per_partition.
      */
-    const unsigned slots = 512;
+    const float mult = 1.9;
```
beomki-yeo, line 988: May not be so important but why 1.9 here?

guilhermeAlmeida1: It's an interesting behaviour of this algorithm.
If you chose an integer number, because most partitions end up being just a bit smaller than the maximum allowed value, this will result in every thread block having to create a very small partition with it's last few elements.
Choosing something like 1.9, 2.9, etc. results in the final partition still being close to the maximum.

guilhermeAlmeida1: I'm currently playing around a bit more with this and I hope to come back to this by seing how the `mult` value affects performance between this algorithm and the subsequent clustering.

beomki-yeo: OK. but we have to be careful if the optimal `mult` is specific to the generic (trackML) detector

beomki-yeo (approved):

device/cuda/src/cca/component_connection.cu
guilhermeAlmeida1: In general I expect this behaviour to be true in any scenario, altough it can be finetuned depending on the detector and event characteristics.

beomki-yeo (approved):

PR #315, opened 2023-02-06
Fixed multithreaded application to run different events

guilhermeAlmeida1:
Turns out our multi-threaded application was running the same event repeatedly instead of using the intended random order.


krasznaa (approved):
**Very** curious. :confused:

Yes, the lambda was set up to grab a temporary variable. Which would have gone out of scope by the time that the lambda actually got called.

But boy would a warning from the compiler have been useful...

guilhermeAlmeida1:
The wacky internet connection right now is repeatedly making CI jobs fail due to not being able to download external packages. I can keep pushing empty commits until this succeeds but I think we can just merge this in without it at this point.

krasznaa (approved):
:+1:

PR #316, opened 2023-02-15
Fuse partitioning and clusterization CUDA kernels

stephenswat:
Previously, we had a dedicated partitioning kernel which created partitions for the clusterization to use. Thanks to a brilliant idea by @guilhermeAlmeida1, this can actually be skipped. We can more efficiently partition the cells right inside of the component connection kernel!


guilhermeAlmeida1 (approved):
Looks great

PR #317, opened 2023-02-15
Improved performance of clusterization

guilhermeAlmeida1:
This PR approximately doubles the performance of the ccl kernel, by making two core changes:

- Making each thread handle more than one cell.
-- This is done by introducing an additional variable, `cells_per_thread`, which is currently declared as `static constexpr`. Not sure whether we should also have this as an input parameter of the application (like we did for `max_cells_per_partition`, or hidden to the user. This makes so the block size of the ccl_kernel is equal to `max_cells_per_partition / cells_per_thread`.
- Optimizing the `aggregate_cluster function` to break out of it's loop earlier whenever possible.

### DIFF
--- a/device/cuda/src/clusterization/clusterization_algorithm.cu
+++ b/device/cuda/src/clusterization/clusterization_algorithm.cu
@@ -27,6 +27,8 @@ namespace {
 /// These indices in clusterization will only range from 0 to
 /// MAX_CELLS_PER_PARTITION, so we only need a short
 using index_t = unsigned short;
+
+static constexpr int CELLS_PER_THREAD = 8;
 }  // namespace
 
 namespace kernels {
@@ -49,8 +51,10 @@ namespace kernels {
 /// @param[in] adjv     Vector of adjacent cells
 /// @param[in] tid      The thread index
 ///
-__device__ void fast_sv_1(index_t* f, index_t* gf, unsigned char adjc,
-                          index_t adjv[8], index_t tid) {
+__device__ void fast_sv_1(index_t* f, index_t* gf,
+                          unsigned char adjc[CELLS_PER_THREAD],
+                          index_t adjv[CELLS_PER_THREAD][8], index_t tid,
+                          const index_t blckDim) {
     /*
      * The algorithm finishes if an iteration leaves the arrays unchanged.
      * This varible will be set if a change is made, and dictates if another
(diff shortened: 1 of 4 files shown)


guilhermeAlmeida1:
I would like to note that I do think CELLS_PER_THREAD must stay as a constexpr int, as it is used to create arrays in private memory (local memory in CUDA) which must be of static size.

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -72,13 +76,17 @@ __device__ void fast_sv_1(index_t* f, index_t* gf, unsigned char adjc,
          * cluster ID if it is lower than ours, essentially merging the two
          * together.
          */
-        __builtin_assume(adjc <= 8);
-        for (unsigned char k = 0; k < adjc; ++k) {
-            index_t q = gf[adjv[k]];
+        for (index_t tst = 0; tst < CELLS_PER_THREAD; ++tst) {
+            const index_t cid = tst * blckDim + tid;
```
stephenswat, line 80: Why do you pass `blckDim` as a variable, why not use the builtin `blockDim.x`?

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -310,11 +345,11 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
         cudaMemset(num_measurements_device.get(), 0, sizeof(unsigned int)));
 
     // Launch ccl kernel. Each thread will handle a single cell.
-    kernels::
-        ccl_kernel<<<num_partitions, m_max_cells_per_partition,
-                     2 * m_max_cells_per_partition * sizeof(index_t), stream>>>(
-            cells, modules, partitions, m_max_cells_per_partition,
-            measurements_buffer, *num_measurements_device);
+    kernels::ccl_kernel<<<
+        num_partitions, m_max_cells_per_partition / CELLS_PER_THREAD,
```
stephenswat, line 349: Please make sure that this works if `m_max_cells_per_partition` is not a multiple of `CELLS_PER_THREAD`.

device/sycl/src/clusterization/clusterization_algorithm.sycl
```diff
@@ -364,8 +394,16 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
 
     const unsigned short max_cells_per_partition = m_max_cells_per_partition;
     ::sycl::nd_range ndrange(
-        ::sycl::range<1>(num_partitions * max_cells_per_partition),
-        ::sycl::range<1>(max_cells_per_partition));
+        ::sycl::range<1>(num_partitions * max_cells_per_partition /
+                         CELLS_PER_THREAD),
```
stephenswat, line 398: Same comment here as for CUDA.

device/sycl/src/clusterization/clusterization_algorithm.sycl
```diff
@@ -364,8 +394,16 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
...
+        ::sycl::range<1>(max_cells_per_partition / CELLS_PER_THREAD));
```
stephenswat, line 399: And once more.

stephenswat (changes requested):

device/cuda/src/clusterization/clusterization_algorithm.cu
guilhermeAlmeida1: Just thought it would look slightly cleaner in SYCL to call `item.get_local_range(0)` only once and store it as a variable instead of calling it everywhere, and equivalently just "copied" this into CUDA

device/cuda/src/clusterization/clusterization_algorithm.cu
guilhermeAlmeida1: that's a good idea

stephenswat (approved):
Looks good to me now. Thanks for the changes.

PR #318, opened 2023-02-16
Fuse partitioning and CCA kernels and remove CPU partitioning algorithm

guilhermeAlmeida1:
This was built on top of #317 . I'll leave it to the reviewer's discretion whether they want to review these two PR's separatly or the full changes here, where the changes are still separated into 2 commits, and merge it in one go.

The new changes in this PR remove the need for creating the partitions in a separate step to the actual ccl_kernel, as was done for the separate cuda specific cca chain in #316 .

Performance-wise, this makes the ccl_kernel a bit slower but for the total performance I have not seen any significant difference (when comparing to the one observed in #317). This was to be expected as the CPU partitioning algorithm itself was very quick, but this makes things a bit cleaner by removing the need for using the CPU in the chain as well as the additional copy that comes with it.


stephenswat:
@guilhermeAlmeida1 could you rebase this?

guilhermeAlmeida1:
Should be good to go now @stephenswat . I wanted to try out a slightly smarter way of handling the loop unrolling but it really doesn't make any difference to the kernel performance, so I'll leave it as it is which is simpler.

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -344,14 +373,23 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
     CUDA_ERROR_CHECK(
         cudaMemset(num_measurements_device.get(), 0, sizeof(unsigned int)));
 
+    const unsigned short max_cells_per_partition =
+        m_target_cells_per_partition *
+        (MAX_CELLS_PER_THREAD / (float)TARGET_CELLS_PER_THREAD);
```
stephenswat, line 378: This is the only thing I don't like. Floating point numbers are evil. Would suggest the following:

```suggestion
    const unsigned short max_cells_per_partition = (m_target_cells_per_partition * MAX_CELLS_PER_THREAD + TARGET_CELLS_PER_THREAD - 1) / TARGET_CELLS_PER_THREAD);
```

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -367,16 +405,13 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
         *num_measurements_host, m_mr.main);
 
     // For the following kernel, we can now use whatever the desired number of
-    // threads
-    const unsigned int num_blocks =
-        (*num_measurements_host + m_max_cells_per_partition - 1) /
-        m_max_cells_per_partition;
+    // threads per block
+    const unsigned int num_blocks = (*num_measurements_host + 1024 - 1) / 1024;
```
stephenswat, line 409: Can we make this magic number configurable?

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -367,16 +405,13 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
...
 
     // Turn 2D measurements into 3D spacepoints
-    kernels::
-        form_spacepoints<<<num_blocks, m_max_cells_per_partition, 0, stream>>>(
-            measurements_buffer, modules, *num_measurements_host,
-            spacepoints_buffer);
+    kernels::form_spacepoints<<<num_blocks, 1024, 0, stream>>>(
```
stephenswat, line 412: Same here, configurability please.

device/sycl/src/clusterization/clusterization_algorithm.sycl
```diff
@@ -392,25 +441,40 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
         .memset(num_measurements_device.get(), 0, sizeof(unsigned int))
         .wait_and_throw();
 
-    const unsigned short max_cells_per_partition = m_max_cells_per_partition;
+    const unsigned short max_cells_per_partition =
+        m_target_cells_per_partition *
+        (MAX_CELLS_PER_THREAD / (float)TARGET_CELLS_PER_THREAD);
```
stephenswat, line 446: Same comment as for the CUDA code.

device/sycl/src/clusterization/clusterization_algorithm.sycl
```diff
@@ -392,25 +441,40 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
...
+        m_target_cells_per_partition;
+
     ::sycl::nd_range ndrange(
         ::sycl::range<1>(num_partitions * threads_per_partition),
         ::sycl::range<1>(threads_per_partition));
 
     // Check if device is capable of allocating sufficient local memory
     assert(sizeof(index_t) * 2 * max_cells_per_partition +
-               sizeof(unsigned int) <
+               3 * sizeof(unsigned int) <
```
stephenswat, line 462: Why do we need this factor three here? I don't think we multiply by three for the CUDA code, do we?

stephenswat (changes requested):

device/sycl/src/clusterization/clusterization_algorithm.sycl
guilhermeAlmeida1: in the CUDA code we only dynamically allocate the 2 `index_t `vectors. Here in SYCL we also need to have the 3 extra shared `unsigned int` variables (as all shared variables) allocated like this. https://github.com/guilhermeAlmeida1/traccc/blob/99ea08e1a68600226531fceb648d611c7a915429/device/sycl/src/clusterization/clusterization_algorithm.sycl#L472-L488

Now that I look at this I could've made the whole thing more compact by just allocating a 3-sized vector of unsigned's instead of 3 separate ones, so let me do that in this PR as well.

stephenswat: Ok, I trust your judgement on this.

stephenswat (approved):
Let's put this in.

PR #319, opened 2023-02-22
Remove headers from doublet and triplet containers.

guilhermeAlmeida1:
We are storing redundant information on the headers of the doublet and triplet containers, and above all, spending valuable time in the `find doublets/triplets` kernels with atomic operations for filling these values, which are just equal to the size. As you might notice, we were already using `.size()` on almost all of our code, I only had to change a single call to a header.

This PR replaces the headers from the doublet/triplet container types with an empty type, `std::monostate`.

### DIFF
--- a/core/include/traccc/seeding/detail/doublet.hpp
+++ b/core/include/traccc/seeding/detail/doublet.hpp
@@ -1,20 +1,19 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2021-2022 CERN for the benefit of the ACTS project
+ * (c) 2021-2023 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
 
 #pragma once
 
+// Project include(s).
 #include "traccc/seeding/detail/singlet.hpp"
 
-namespace traccc {
+// System include(s).
+#include <variant>
 
-/// Header: the number of doublets per spacepoint bin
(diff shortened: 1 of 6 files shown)


krasznaa (approved):
Well, now I learned that [std::monostate](https://en.cppreference.com/w/cpp/utility/variant/monostate) exists. :stuck_out_tongue: It looks to be good enough for this place. (Ideally we would not create a 1D vector for the doublets and triplets at all. But since so much of our code is built around `traccc::container`, I agree that this is a quite good setup. At least for now.)

For a bit of historic context: I believe the header was used in the beginning, when the sizes of resizable jagged vectors were not contiguous in memory yet, to allow copying just the sizes back to the host in a convenient way. But I **think** even that was just for debugging with these types.

Since [vecmem::copy::get_sizes(...)](https://acts-project.github.io/vecmem/classvecmem_1_1copy.html#aa75c50a043a95bbbfd2e9cb0bf5be551) is quite efficient these days, this is definitely redundant information by now.

Good catch!

guilhermeAlmeida1:
I'll take the opportunity to sneak in another one-liner change to find_doublets in this PR.

krasznaa (approved):
Yepp, let's not go to the global memory multiple times for that `internal_spacepoint`. :+1:

stephenswat:
As a member of the `std::monostate` fanclub I approve this message.

PR #325, opened 2023-03-01
Add line information `RelWithDebInfo` CUDA build

stephenswat:
As @guilhermeAlmeida1 and myself have previously noticed, the CUDA builds in traccc do not embed line information for most build types. Line information is very useful for profiling. The only build type that enables this information by default is the `Debug` build, on which profiling does not make much sense.

This commit adds the `-lineinfo` flag to the `RelWithDebInfo` build type, which means that the NVIDIA profiler can be used with the `--import-source` flag. This means that NSight Compute will be able to make more accurate determinations about which PTX lines belong to which CUDA lines.


krasznaa (approved):
:+1:

PR #326, opened 2023-03-02
Change seeding to use flat vectors of doublets and triplets.

guilhermeAlmeida1:
This PR changes the seeding algorithm to use flat vectors (rather than jagged ones) for it's internal storage of doublets and triplets. Note, however, that the spacepoint grid maintains its 2D structure.

The actual kernels themselves change in not-too-complicated ways. The main new difference kernel-wise is the new `reduce_triplet_counts` kernel. This results from the need to compare triplets which share the same middle spacepoint or mid-bottom doublet in the seed filtering stage, so in this PR I created 2 different triplet-counter types. One which counts the number of triplets **per mid-bottom doublet**, and another **per middle spacepoint**, both of which also claiming positions in the ouput array where their triplets will be filled. The reduction kernel atomically sums the number of triplets over all middle spacepoints, and uses this atomic operation to correctly calculate the output positions.

PS: I removed the all the separate `make_buffer` functions for doublets and triplets from separate files as they now become trivially small.


guilhermeAlmeida1:
Note also that these changes to the EDM are all internal, the input & output of seed finding are both unchanged. I initially thought of just naming the new variables `doublet` and `triplet` and putting them in the `device` namespace but thought that could be very error-prone. Hence the `device_doublet` and `device_triplet` naming.

guilhermeAlmeida1:
As requested by @krasznaa, here are some performance plots of the throughput of the full aplication using the currently implemented seeding against the one suggested in this PR.
* Current seeding in green
* Suggested flat seeding in purple
* Baseline CPU throughput in blue
Tested using CPU: AMD EPYC 7413 and GPU: NVIDIA RTX A5000. Presented a few select pile-up values but same results obtained in others not shown.
image
image
image
image

device/common/include/traccc/edm/device/triplet_counter.hpp
```diff
@@ -9,37 +9,40 @@
...
 namespace traccc::device {
 
-/// Header type for the "triplet counter container"
-///
-/// The header stores summary information about the number of triplets found in
-/// a given geometric bin.
-///
-struct triplet_counter_header {
+/// Number of triplets for one specific middle spacepoint.
+struct triplet_counter_spM {
```
krasznaa, line 17: I'm now confused. :frowning: Did you change the logic with the triplets this morning? I'm not a fan of making the data layout even more complicated here...

device/common/include/traccc/seeding/device/count_doublets.hpp
```diff
@@ -28,14 +28,18 @@ namespace traccc::device {
...
-/// @param[out] doublet_view Container storing the number of doublets
+/// @param[out] doublet_view Collection storing the number of doublets for each
+/// spacepoint
+/// @param[out] nMidBot      Total number of middle-bottom doublets
+/// @param[out] nMidTop      Total number of middle-top doublets
 ///
 TRACCC_HOST_DEVICE
 inline void count_doublets(
-    std::size_t globalIndex, const seedfinder_config& config,
+    const std::size_t globalIndex, const seedfinder_config& config,
```
krasznaa, line 38: This is purely a personal preference, but when passing something by value, I prefer to:
  - Not put `const` into the declaration into the function;
  - Put `const` into the implementation of the function.

Of course if these two are not separated, then you would put `const` into the one place where the function appears. But in this case I'd only put `const` into the `.ipp` file.

My justification is that making a "by value" argument `const` is an implementation detail that the users of the function are not affected by. Exactly for this reason the compilers allow you to leave off the `const` declaration from the function's API.

guilhermeAlmeida1:
This should once again be ready to be looked at. I put in a couple of changes verbally suggested in our discussions.

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -155,158 +165,167 @@ seed_finding::output_type seed_finding::operator()(
...
 
+    // Counters for the total number of doublets
+    vecmem::unique_alloc_ptr<unsigned int> num_midBot_device =
+        vecmem::make_unique_alloc<unsigned int>(m_mr.main);
+    vecmem::unique_alloc_ptr<unsigned int> num_midTop_device =
+        vecmem::make_unique_alloc<unsigned int>(m_mr.main);
+    CUDA_ERROR_CHECK(
+        cudaMemset(num_midBot_device.get(), 0, sizeof(unsigned int)));
+    CUDA_ERROR_CHECK(
+        cudaMemset(num_midTop_device.get(), 0, sizeof(unsigned int)));
```
krasznaa, line 187: It's a small thing, but we could use a dedicated type just for these two counters as well. That way we could halve the number of allocations/de-allocations necessary to handle these.

Or in fact, you could declare a `struct` for **all** of the `unsigned int` variables needed by the algorithm, and allocate/de-allocate them all at once. That `struct` should probably go under `traccc/edm/device` of course.

device/common/include/traccc/edm/device/triplet_counter.hpp
guilhermeAlmeida1: No, this is still the same logic we discussed. There are 2 triplet counter types, one corresponding to counts per mid-bottom doublets (as before) and another corresponding to each spacepoint middle (the new one). So you'll have 2 collections of triplet counter types linked to each other.

device/common/include/traccc/edm/device/device_doublet.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+// Project include(s).
+#include "traccc/edm/container.hpp"
+#include "traccc/edm/device/doublet_counter.hpp"
+#include "traccc/seeding/detail/singlet.hpp"
+
+namespace traccc {
+
+/// Doublet of middle-bottom or middle-top spacepoints
+struct device_doublet {
```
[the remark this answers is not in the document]
guilhermeAlmeida1: I'll put them into the device namespace, but I still think it's better to name the 2 types differently for clarity.

PR #328, opened 2023-03-03
Optional ROOT, main branch (2023.03.03.)

krasznaa:
Taught the project what to do if [ROOT](https://github.com/root-project/root) is not available.

Since practically every example and test depended on ROOT at this point, first of all I made some larger updates in `traccc::performance` to make it only depend on ROOT privately. Removing all ROOT types from the public interface of that library.

Then taught `traccc::performance` how to not use ROOT "privately" when it's not available. This required a bunch of `#ifdef`-s, but not in the public interface of the library.

Finally, had to make some changes in the Kalman fitter unit tests, as those were/are also very much relying on ROOT. I ended up making a similar change in `traccc::KalmanFittingTests` as in `traccc::performance`. Which allowed me to restrict the `#ifdef`-s to only `kalman_fitting_test.cpp`.

In this new setup you would see an explicit message when ROOT is found.

```
-- Found ROOT: /software/root/6.26.06/x86_64-ubuntu2004-gcc9-opt/include (found version "6.26.06")
```

And of course also when it's not found...

```
-- Could NOT find ROOT (missing: ROOT_INCLUDE_DIRS ROOT_LIBRARY_DIR ROOT_BINDIR)
```

(Previously CMake was silent about finding ROOT. Which is a long-standing gripe of mine about the behaviour of `ROOTConfig.cmake`...)


beomki-yeo:
Hmm. Before I take a closer look, let me ask a question; The unit tests that require ROOT dependency will get ignored if there is no ROOT library installed in the system?

krasznaa:
Let me start with the last question: I want to make building this project as simple as possible. And the ROOT dependency right now is the single biggest hurdle with building it.

I have a number of tickets open with Intel about issues that I can only observe with this project. (As not even the other R&D projects like VecMem would trigger those issues by themselves.) But non-HEP people pretty much give up with traccc as soon as they are hit with the ROOT dependency. So the goal of this is to allow our friends in the industry to be able to build and test this code.

As for the unit test behaviour: I was pondering about this a bit. My first idea was indeed to explicitly disable the tests that rely on ROOT with `GTEST_SKIP()`. But I saw that these tests still performed some basic checks that did not require ROOT. So in the end I rather kept the tests running, and just disabled the checks of the shapes of the result distributions. All the while printing a warning message.

Which message **is** invisible when using CTest. :frowning: But at least when you run the test by hand, you would see a message like:

```
Running main() from /home/krasznaa/ATLAS/projects/traccc/build-vanilla/_deps/googletest-src/googletest/src/gtest_main.cc
Note: Google Test filter = KalmanFitValidation/KalmanFittingTests.*
[==========] Running 3 tests from 1 test suite.
[----------] 3 tests from KalmanFitValidation/KalmanFittingTests
[ RUN      ] KalmanFitValidation/KalmanFittingTests.Run/0
ROOT file "performance_track_fitting_1.000000_GeV_0.000000_phi.root" is NOT created
Pull value tests not performed without ROOT
[       OK ] KalmanFitValidation/KalmanFittingTests.Run/0 (1150 ms)
[ RUN      ] KalmanFitValidation/KalmanFittingTests.Run/1
ROOT file "performance_track_fitting_10.000000_GeV_0.000000_phi.root" is NOT created
Pull value tests not performed without ROOT
[       OK ] KalmanFitValidation/KalmanFittingTests.Run/1 (1144 ms)
[ RUN      ] KalmanFitValidation/KalmanFittingTests.Run/2
ROOT file "performance_track_fitting_100.000000_GeV_0.000000_phi.root" is NOT created
Pull value tests not performed without ROOT
[       OK ] KalmanFitValidation/KalmanFittingTests.Run/2 (1146 ms)
[----------] 3 tests from KalmanFitValidation/KalmanFittingTests (3441 ms total)

[==========] 3 tests from 1 test suite ran. (3441 ms total)
[  PASSED  ] 3 tests.
```

krasznaa:
Note that I was seriously considering extending the CI's build matrix with a boolean for whether ROOT should be used or not. But in the end decided against it.

The CI takes pretty long as is. And to test the functionality of `-DTRACCC_USE_ROOT=FALSE` properly, the job would first need to kneecap the ROOT installation of the Docker image. Just to make sure that nothing ends up silently using ROOT anyway. And that all seemed a bit too much effort for something that we won't be using in regular development anyway. :thinking:

beomki-yeo (approved):
Looks good to me

PR #329, opened 2023-03-03
Add additional output information to clustering algorithm

guilhermeAlmeida1:
This PR adds additional output information to the clustering algorithm, which allows linking back from the created spacepoints down to the cells they were created from. This is currently not used for anything, but will be necessary for physics performance testing with the new EDM.

### DIFF
--- a/device/common/include/traccc/clusterization/device/aggregate_cluster.hpp
+++ b/device/common/include/traccc/clusterization/device/aggregate_cluster.hpp
@@ -34,7 +34,8 @@ inline void aggregate_cluster(
     const cell_module_collection_types::const_device& modules,
     const vecmem::data::vector_view<unsigned short> f_view,
     const unsigned int start, const unsigned int end, const unsigned short cid,
-    alt_measurement& out);
+    alt_measurement& out, vecmem::data::vector_view<unsigned int> cell_links,
+    const unsigned int link);
 
 }  // namespace traccc::device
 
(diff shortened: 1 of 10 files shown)


guilhermeAlmeida1:
> Nice speedy fix. What's the performance impact of this?

The actual kernel itself doesn't really become slower, which that was to be expected, but the extra allocation on the host seems to make the clustering some 25% slower from looking at profiles, but then again if I run the full throughput application it doesn't become any slower at all. So, I guess we can say there's no impact, sort of?

stephenswat:
Yeah, the caching of memory allocations will handle this nicely in multi-event execution.

I will be around to collect on the bet we made about the performance of this on Monday! :wink:

krasznaa (approved):
We could dress up the data types into some more "reader friendly" classes in the future, but for now this looks good enough. The information we need is actually returned, so :+1:.

krasznaa (approved):
For cleanliness, could we all use the "Update with Rebase" button on the web interface? I'm not a fan of including all these "merge commits" in our history... :frowning: (**cough** *Stephen* **cough**... :stuck_out_tongue:)

stephenswat:
I'm fine with that but then we should probably also rebase PRs instead of merging. :innocent: 

image

stephenswat:
Also, looks like applying PRs by rebasing results in unverified comments if the committer is in vigilant mode... https://github.com/acts-project/traccc/commit/cd54f73583c8e31f43862a7afe6ff7d43060f7e5 :frowning_face:

PR #336, opened 2023-03-09
Add CI functionality to detect FP64 instructions in PTX 

stephenswat:
We generally want to avoid doing double-precision work on GPUs, and as such we want to avoid emitting FP64 instructions in the PTX code that we generate from CUDA. This commit adds a new CI job that automatically checks whether there are any such instructions in the code.

### DIFF
--- a/.github/find_f64_ptx.py
+++ b/.github/find_f64_ptx.py
@@ -1,0 +1,168 @@
+#!/bin/python3
+
+
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2023 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+
+import argparse
+import re
+import collections
+import pathlib
+
+
+class InstructionCounter:
+    """
+    Class for counting the use of certain instructions in translation units.
+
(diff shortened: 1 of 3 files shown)


krasznaa:
I know it's still a draft, but it would be better to attach this step to the existing CUDA-Debug build. It's a bit too wasteful to do the CUDA-Debug build twice...

krasznaa (approved):
Okay, as far as I'm concerned, this is simply 90% magic. :stuck_out_tongue:

Some documentation in the Python code could've been good, but in the end I can live with this as well. Let's put such tests in place. :+1:

stephenswat:
Wait, I'll add a few Python comments. :slightly_smiling_face:

stephenswat:
Okay, I have added comments.

stephenswat:
Just to be clear, this does only NVCC-generated PTX at this point. This means it can pick up FP64 instructions in shared code and dedicated CUDA code, but it will not pick up these things in SYCL code. If we want to expand this logic to SYCL we will need to see if the SYCL compiler can be convinced to emit PTX with line information.

paulgessinger:
I like this! 🍻

krasznaa (approved):
:+1:

PR #338, opened 2023-03-09
Remove return in CCL_kernel

guilhermeAlmeida1:
As @krasznaa noticed when testing our code on one of our available Intel cards, the `ccl_kernel` can hang. 
This comes from the undefined behaviour I had previously identified with using group synchronisation functions while having some returned threads. I had initially thought the problem was with group voting but it seems that even just a barrier can lead to a hang.

### DIFF
--- a/device/sycl/src/clusterization/clusterization_algorithm.sycl
+++ b/device/sycl/src/clusterization/clusterization_algorithm.sycl
@@ -254,8 +254,8 @@ class ccl_kernel {
         unsigned char adjc[MAX_CELLS_PER_THREAD];
 
         // It seems that sycl runs into undefined behaviour when calling
-        // any_of_group when some threads have already run into a return. So can
-        // only do this after running the FastSV algorithm.
+        // group synchronisation functions when some threads have already run
+        // into a return. As such, we cannot use returns in this kernel.
 
 #pragma unroll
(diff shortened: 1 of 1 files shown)


stephenswat:
It would seem prudent to add some additional checks to ensure that threads that are out of range, e.g. where `tid >= size`, do not execute code where `tid` is used and therefore risk access violation errors.

guilhermeAlmeida1:
> It would seem prudent to add some additional checks to ensure that threads that are out of range, e.g. where `tid >= size`, do not execute code where `tid` is used and therefore risk access violation errors.

This is already being done, all the loops which access information look like this:
https://github.com/acts-project/traccc/blob/07fe7e9a0f02ba3a9ead46d4924ef971636f8f50/device/sycl/src/clusterization/clusterization_algorithm.sycl#L302

krasznaa (approved):
Let's go ahead with this.

In the end I'm actually not even 1000% sure if SYCL ever actively executes a user function for a non-existing index. Their documentation has been a bit hazy on this... :thinking:

krasznaa:
Hahh!!! Sometimes the world is kind! :smile:

After fixing the FP64 issues, I was still running into some inexplicable exceptions from SYCL when running the throughput tests on my iGPU. This PR gets rid of those as well. :partying_face:

PR #339, opened 2023-03-09
Replace FP64 by FP32 in gain matrix smoother

stephenswat:
Courtesy of #336, this commit removes a use of double-precision floating point numbers and replaces it by a configurable scalar type as we prefer to do in traccc.

### DIFF
--- a/core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
+++ b/core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
@@ -64,7 +64,7 @@ struct gain_matrix_smoother {
         const matrix_type<6, 6>& cur_filtered_cov = cur_filtered.covariance();
 
         // Regularization matrix for numerical stability
-        static constexpr double epsilon = 1e-13;
+        static constexpr scalar epsilon = 1e-13;
         const matrix_type<6, 6> regularization =
             matrix_operator().template identity<e_bound_size, e_bound_size>() *
             epsilon;


core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
```diff
@@ -64,7 +64,7 @@ struct gain_matrix_smoother {
         const matrix_type<6, 6>& cur_filtered_cov = cur_filtered.covariance();
 
         // Regularization matrix for numerical stability
-        static constexpr double epsilon = 1e-13;
+        static constexpr scalar epsilon = 1e-13;
```
krasznaa, line 67: While I'm not the biggest fan of that variable either, wouldn't it be better to use a single variable for this purpose in the project? I.e. `traccc::float_epsilon`?

https://github.com/acts-project/traccc/blob/main/core/include/traccc/definitions/common.hpp#L19

krasznaa:
I assume this should be merged in before #336, right? :thinking:

stephenswat:
It can go either way: #336 does not fail the CI job if any double precision
operations are found, so the order between these two doesn't really matter.
:)

Regarding the epsilon issue, I don't feel strongly either way. Allowing
algorithms to define their own epsilon values ensures they can set
different precision requirements depending on their needs. Also, if a
particular piece of code needs to compare very large numbers then it might
need to set its own higher epsilon value. But harmonizing it across the
project has value in that it makes everything cleaner.

core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
beomki-yeo: I would keep it as it is because their values (1e-5 vs 1e-13) are too different.

krasznaa: :+1:

krasznaa (approved):
:+1:

beomki-yeo (approved):

PR #341, opened 2023-03-10
Add option to log multi-threaded results to output file

guilhermeAlmeida1:
This makes running performance tests through a wide range of input options simpler, by easily being incorporated with any given script which runs the desired options (various mu values, number of threads, etc.) and printing it all to a csv-compatible file.

### DIFF
--- a/examples/run/common/throughput_mt.ipp
+++ b/examples/run/common/throughput_mt.ipp
@@ -32,6 +32,7 @@
 #include <atomic>
 #include <cstdlib>
 #include <ctime>
+#include <fstream>
 #include <iostream>
 #include <memory>
 #include <vector>
@@ -189,6 +190,19 @@ int throughput_mt(std::string_view description, int argc, char* argv[],
               << performance::throughput{throughput_cfg.processed_events, times,
                                          "Event processing"}
               << std::endl;
+
+    // Print results to log file
+    if (throughput_cfg.log_file != "\0") {
+        std::ofstream logFile;
+        logFile.open(throughput_cfg.log_file, std::fstream::app);
+        logFile << "\"" << throughput_cfg.input_directory << "\""
+                << "," << mt_cfg.threads << "," << throughput_cfg.loaded_events
+                << "," << throughput_cfg.cold_run_events << ","
+                << throughput_cfg.processed_events << ","
+                << times.get_time("Warm-up processing").count() << ","
+                << times.get_time("Event processing").count() << std::endl;
+        logFile.close();
+    }
 
     // Return gracefully.
     return 0;
--- a/examples/options/include/traccc/options/throughput_options.hpp
+++ b/examples/options/include/traccc/options/throughput_options.hpp
@@ -46,6 +46,9 @@ struct throughput_options {
     /// them in the performance measurements
     std::size_t cold_run_events = 10;
 
+    /// Output log file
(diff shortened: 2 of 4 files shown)


examples/run/common/throughput_mt.ipp
```diff
@@ -190,6 +191,20 @@ int throughput_mt(std::string_view description, int argc, char* argv[],
...
+    if (mt_cfg.log_result) {
+        std::ofstream logFile;
+        logFile.open(mt_cfg.log_file, std::fstream::app);
+        logFile << "\"" << throughput_cfg.input_directory << "\""
+                << "," << mt_cfg.threads << ","
+                << throughput_cfg.cold_run_events << ","
+                << throughput_cfg.cold_run_events << ","
+                << throughput_cfg.processed_events << ","
+                << times.get_time("Warm-up processing").count() << ","
+                << times.get_time("Event processing").count() << std::endl;
```
guilhermeAlmeida1, line 204: I also considered directly printing out the events/s measurement but thought it might best to leave the output as "clean" as possible by leaving to the user to make the FP divisions. Also note that the output files don't have any headers. I could make this somewhat more complex by opening the file in read & write mode and first checking if the headers are already present, and if not write them out.

guilhermeAlmeida1:
Addressed the comments above by @krasznaa. With the way the options are setup right now we're not actually using the `common_options` in the throughput executables, but rather the `throughput_options`. I'm fine removing the shared variables between the two and also using `common_options` in the throughput executables as well but for the sake of brevity decided not to do so in this PR.

krasznaa (approved):
That's fair with `traccc::throughput_options`. I thought we were combining these in some way, but you're right, we're not. So let's have it in `traccc::throughput_options` after all.

Is `"\0"` better for some reason than an empty string? :confused: I would've just gone with an empty one.

Not having a header is sort of fine. In a future update we should add the logic though that if the file doesn't exist yet, then a header is added before adding the performance numbers. And if the file exists already, then we only append the numbers.

Also, because this has some time pressure, let's skip updating `throughput_st.ipp` as well. But next week you should also update that code to make use of this new command line parameter.

(If this turns out to actually work for your tests, then I'll make a `v0.3.0` tag over the weekend as well. But now let's run the performance tests first before making the tag.)

guilhermeAlmeida1:
> Is `"\0"` better for some reason than an empty string? 😕 I would've just gone with an empty one.
Just a personal preference, I would say, I thought it could be safer to have it well defined like that.

> (If this turns out to actually work for your tests, then I'll make a `v0.3.0` tag over the weekend as well. But now let's run the performance tests first before making the tag.)
For our performance tests to be coherent but be able to use this already I can just make a "release" on my private fork of traccc with just `v0.2.0` plus this specific commit (without including the fp64 to fp32 changes merged in the meantime).

krasznaa:
It's only the FP64 detection that's there over `v0.2.0` at the moment.

https://github.com/acts-project/traccc/compare/v0.2.0...main

So don't sweat it. Current `main` is still performing the same as `v0.2.0`. It's just for "consistency" that I'll want to make a tag for the exact code version that we actually used in the end.

guilhermeAlmeida1:
Ah right, I thought https://github.com/acts-project/traccc/pull/335 had come after the release, that makes our life even easier

PR #342, opened 2023-03-10
ci: Put test data into tar files, unpack them in the CI

paulgessinger:


krasznaa (changes requested):
Please add some lines into `README.md` on what people should do when they clone the repository.

krasznaa:
Please add the `REDME.md` update.

#@^&# gitlab.cern.ch has completely broken down by now. I can clone the HEAD of its master branch, but as soon as I try to switch to an older hash, things just break. (A direct clone of the repository succeeds to switch to an old hash after a while. But the timeout seems to be too long for git-submodule to accept. :thinking:)

To get the CI back into a working state, it seems that we must accept this PR as quickly as possible...

paulgessinger:
How about this @krasznaa?

krasznaa (approved):
Yepp, let's get this in...

PR #343, opened 2023-03-11
Log `target_cells_per_partition` in throughput measurements

guilhermeAlmeida1:
In #341 I forgot to printout log this piece of information which could be useful in the future.


krasznaa:
In the future yes. But for now let's hold with it for a bit.

I.e. I don't plan to show results with different values of this parameter next week because:
  - We have yet to look at and discuss the behaviour of the code in an Acts meeting first;
  - I won't have enough time for such details anyway.

guilhermeAlmeida1:
I wasn't planning on using this already myself, but it might be crucial in the future. Maybe we merge this after making the v0.3 release

stephenswat:
@guilhermeAlmeida1 are we good to put this in now?

guilhermeAlmeida1:
@stephenswat I'm not particularly interested to have a look at measuring throughput with respect to this parameter at the moment but it could definately be interesting in the future. I don't see why not get this in now.

PR #344, opened 2023-03-11
Throughput Measurement Scripting, main branch (2023.03.11.)

krasznaa:
Added a script helping with running throughput measurements.

In draft for the moment, as I didn't test it under absolutely all circumstances yet. But as a reference, you would use it like:

```
../traccc/scripts/traccc_throughput_mt_profiling.sh -x ./bin/traccc_throughput_mt -c intel-i7-6700.csv 2>&1 | tee intel-i7-6700.log
```

The Python script is even more of a WIP, that one you would use like:

```
./traccc/scripts/traccc_throughput_mt_plotter.py \
   ~/cernbox/ATLAS/traccc/throughput-2023.03.11/nvidia-rtx3080-sycl.csv:"NVIDIA^{#oright} GeForce#trademark RTX 3080 (SYCL)" \
   ~/cernbox/ATLAS/traccc/throughput-2023.03.11/nvidia-rtx3080-cuda.csv:"NVIDIA^{#oright} GeForce#trademark RTX 3080 (CUDA)" \
   ~/cernbox/ATLAS/traccc/throughput-2023.03.11/amd-threadripper-3970x.csv:"AMD^{#oright} Threadripper#trademark 3970x" \
   ~/cernbox/ATLAS/traccc/throughput-2023.03.11/nvidia-rtx3050ti-laptop-sycl.csv:"NVIDIA^{#oright} GeForce#trademark RTX 3050Ti Mobile (SYCL)" \
   ~/cernbox/ATLAS/traccc/throughput-2023.03.11/nvidia-rtx3050ti-laptop-cuda.csv:"NVIDIA^{#oright} GeForce#trademark RTX 3050Ti Mobile (CUDA)"
```

### DIFF
--- a/extras/traccc_throughput_mt_plotter.py
+++ b/extras/traccc_throughput_mt_plotter.py
@@ -1,0 +1,442 @@
+#!/usr/bin/env python3
+#
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2023 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+#
+
+# Python import(s).
+import argparse
+import csv
+
+# ROOT import(s).
+import ROOT
+
+def runOnInputs(inputs, func):
+    '''Function running another function on every row in all input CSVs
+
+    Argument(s):
+       inputs  -- List of CSV file name, platform name pairs
+       func    -- The function-like-object to run on every CSV row
+    '''
+
+    # Loop over the files.
+    for input in inputs:
+        # Open this particular file.
+        with open(input[0], 'r') as csvFile:
+            csvReader = csv.DictReader(csvFile)
+            # Process each row of the file.
+            for row in csvReader:
+                # Execute the received function on this row.
+                func(row)
+                pass
+            pass
+        pass
+    return
+
+def getSampleNames(inputs):
+    '''Get the union of "sample names" from the inputs
+
+    Argument(s):
+       inputs  -- List of CSV file name, platform name pairs
+
+    Return:
+       A list of unique sample names
+    '''
+
+    # The result list.
+    result = []
+
+    def collectDirectory(row):
+        '''Collect the directory names from the CSV rows
+
+        Argument(s):
+           row -- A CSV row coming from a CSV DictReader
+        '''
+        # Use the "result" variable from the parent scope.
+        nonlocal result
+        # If this directory was not seen yet, add it to the result.
+        dirname = row['directory']
+        if not dirname in result:
+            result.append(row['directory'])
+            pass
+        return
+
+    # Collect the directory names.
+    runOnInputs(inputs, collectDirectory)
+
+    # Return the list.
+    return result
+
+def getMaxThreads(inputs):
+    '''Get the maximum number of CPU threads for which (a) measurement(s) exist
+
+    Argument(s):
+       inputs  -- List of CSV file name, platform name pairs
+
+    Return:
+       The maximum number of threads found in the input files
+    '''
+
+    # The result value.
+    result = 1
+
+    def updateThread(row):
+        '''Find the maximum number of threads in the CSV rows
+
+        Argument(s):
+           row -- A CSV row coming from a CSV DictReader
+        '''
+        # Use the "result" variable from the parent scope.
+        nonlocal result
+        # If we see a value larger than what we saw before, let's use it.
+        threads = int(row['threads'])
+        if threads > result:
+            result = threads
+            pass
+        return
+
+    # Find the maximum number of threads.
+    runOnInputs(inputs, updateThread)
+
+    # Return the value.
+    return result
+
+def getMinMaxThroughput(inputs, sample = ''):
+    '''Get the minimum/maximum throughput values
+
+    Argument(s):
+       inputs  -- List of CSV file name, platform name pairs
+       sample  -- A specific sample to look at, or all samples if left empty
+
+    Return:
+       A pair of minimum-maximum throughput values
+    '''
+
+    # The result pair.
+    result = [1E10, 0.]
+
+    def findMinMax(row):
+        '''Find the minimum and maximum throughput value in the input files
+
+        Argument(s):
+           row -- A CSV row coming from a CSV DictReader
+        '''
+        # Use the variable(s) from the parent scope.
+        nonlocal sample
+        nonlocal result
+        # Skip the current row if it's not coming from the appropriate
+        # directory/sample.
+        if sample != '' and row['directory'] != sample:
+            return
+        # Update the min/max values.
+        throughput = getThroughput(row)
+        if throughput < result[0]:
+            result[0] = throughput
+            pass
+        if throughput > result[1]:
+            result[1] = throughput
+            pass
+        return
+
+    # Find the minimum-maximum throughput values.
+    runOnInputs(inputs, findMinMax)
+
+    # Return the minimum-maximum pair.
+    return result
+
+def getThroughput(csvRow):
+    '''Get the throughput in Hz for a given CSV row
+
+    Argument(s):
+       csvRow  -- A CSV row coming from a CSV DictReader
+
+    Returns:
+       The throughput value in Hz
+    '''
+
+    return (float(csvRow['processed_events']) /
+            (float(csvRow['processing_time']) * 1E-9))
+
+def configureProfile(profile, platform, graphicsOptions):
+    '''Configure a TProfile object with the appropriate properties
+
+    Argument(s):
+       profile          -- The TProfile object to configure
+       platform         -- The platform that the profile represents
+       graphicsOptions  -- The graphics options to use for the profile
+
+    Return:
+       The configured TProfile object
+    '''
+
+    # Set all (relevant) properties on the profile object.
+    profile.SetMarkerStyle(graphicsOptions[platform]['MarkerStyle'])
+    profile.SetMarkerColor(graphicsOptions[platform]['MarkerColor'])
+    profile.SetMarkerSize (graphicsOptions[platform]['MarkerSize'])
+    profile.SetLineColor(graphicsOptions[platform]['LineColor'])
+
+    # Return the configured TProfile.
+    return profile
+
+def drawPerMuSamplePlots(inputs, samples, canvas, output, maxThreads,
+                         graphicsOptions):
+    '''Function drawing the per-mu-sample plots
+
+    Argument(s):
+       inputs          -- List of CSV file name, platform name pairs
+       samples         -- Names of the (mu) samples to make plots for
+       canvas          -- The TCanvas to draw on
+       output          -- Output PDF file name
+       maxThreads      -- Maximum number of threads to make the plots for
+       graphicsOptions -- The graphics options to use for the profile
+    '''
+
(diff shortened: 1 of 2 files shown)


guilhermeAlmeida1:
Overall I like it, but I feel it could be improved. The script I'm running is essentially the same thing but I usually do some manual work beforehand to make things a bit better.

I would suggest, however, that for each mu value, this script should first do a "test" run with relatively low number of processed events, read the resulting throughput, then extrapolate the number of events that would result in a sufficiently large processing time for our statistical analysis (I go for 12s).
For the CPU this I just do this with a single thread and then use the calculated number of events for 12s as the base value that gets multiplied by the number of threads. On the GPU, however, I usually just multiply it by a factor of 2, as our results so far have never scaled more than that with increased threads.

krasznaa:
Nope. The CERN git server is on the fritz. ☹️ I was also unable to fresh clone traccc today.

Paul's PR may help. But it is **not** required. (Git always downloads a specific hash with a submodule. Not just "master".)

guilhermeAlmeida1:
> Nope. The CERN git server is on the fritz. ☹️ I was also unable to fresh clone traccc today.
> 
> Paul's PR may help. But it is **not** required. (Git always downloads a specific hash with a submodule. Not just "master".)

Ah right, silly me went around looking for where the traccc-data was getting fetched but it's right on the folder itself.

scripts/traccc_throughput_mt_plotter.py
```diff
@@ -0,0 +1,405 @@
...
+        # Open this particular file.
+        with open(input[0], 'r') as csvFile:
+            csvReader = csv.DictReader(csvFile)
+            # Process each row of the file.
+            for row in csvReader:
+                # Execute the received function on this row.
+                func(row, input)
+                pass
+            pass
+        pass
```
stephenswat, line 36: Why are all these pass statements here? :worried:

scripts/traccc_throughput_mt_plotter.py
```diff
@@ -0,0 +1,405 @@
...
+        if not dirname in result:
+            result.append(row['directory'])
+            pass
+        return
+    
+    # Collect the directory names.
+    runOnInputs(inputs, collectDirectory)
+
+    # Return the list.
+    return result
```
stephenswat, line 72: This way of doing things is pretty wasteful. Doesn't really matter for this particular use-case, I suppose, but the amount of unnecessary file reading being done here is quite large.

scripts/traccc_throughput_mt_plotter.py
```diff
@@ -0,0 +1,405 @@
...
+           row    -- A CSV row coming from a CSV DictReader
+           input  -- The CSV file name, platform name pair for the current row
+        '''
+        # Use the "result" variable from the parent's scope.
+        nonlocal result
+        # If this directory was not seen yet, add it to the result.
+        dirname = row['directory']
+        if not dirname in result:
+            result.append(row['directory'])
+            pass
```
stephenswat, line 65: Remove.

scripts/traccc_throughput_mt_plotter.py
```diff
@@ -0,0 +1,405 @@
...
+    # Collect the platform names, and the minimum-maximum values.
+    platforms = []
+    for thr in max_throughputs:
+        if not thr[1] in platforms:
+            platforms.append(thr[1])
+            pass
+        pass
+
+    # Get the absolute minimum-maximum throughput values.
+    min_max_throughput = getMinMaxThroghput(inputs)
```
stephenswat, line 282: Not important, but "throughput" is misspelled here.

scripts/traccc_throughput_mt_plotter.py
krasznaa: Because most editors (at least Emacs and VSCode) handle these as "closing brackets". And know to indent the next line of the code "one less".

scripts/traccc_throughput_mt_profiler.sh
```diff
@@ -0,0 +1,150 @@
...
+   echo "  -h                   Print this help"
+   echo ""
+   echo "Detailed options:"
+   echo "  -g <detectorGeomFile> Detector geometry CSV file"
+   echo "  -d <digiConfigFile>   Detector digitization JSON file"
+   echo ""
+}
+
+# Parse the command line arguments.
+TRACCC_EXECUTABLE="traccc_throughput_mt"
```
stephenswat, line 37: Let's allow the user to set these as environment variables and only use these defaults if nothing is set.

```suggestion
TRACCC_EXECUTABLE=${TRACCC_EXECUTABLE:-"traccc_throughput_mt}"
```

scripts/traccc_throughput_mt_plotter.py
stephenswat: Okay, sure, but let's not pollute the code with unnecessary nonsense to placate a few specific editors.

stephenswat (commented):
As far as I am concerned this doesn't really need to live in the traccc repository; these kinds of measurement scripts are usually so specialized that sharing them in the main repository makes little sense.

At the same time, having them doesn't _hurt_ us, so I have no qualms with putting them in `/extras/` (let's not create yet another top level directory for these).

scripts/traccc_throughput_mt_profiler.sh
```diff
@@ -0,0 +1,150 @@
...
+         echo "Running test ${COUNTER} / ${COUNT}"
+         ((COUNTER++))
+
+         # Run the throughput test.
+         ${TRACCC_EXECUTABLE}                                                 \
+            --detector_file="${TRACCC_DET_FILE}"                              \
+            --digitization_config_file="${TRACCC_DIGI_FILE}"                  \
+            --input_directory="tml_full/${EVTDIR}/"                           \
+            --threads=${NTHREAD}                                              \
+            --cold_run_events=$((20*${NTHREAD}))                              \
```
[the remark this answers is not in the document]
krasznaa: Why? :confused: The purpose of the warm-up events is to:
  - Make sure that the kernels are loaded into the device;
  - Warm up the memory caches.

Why would either of these depend on the pileup value? I chose 20 because this gives a good chance of encountering all 10 of our events at least once on each thread. One could increase this a bit further, but this seemed to be "good enough".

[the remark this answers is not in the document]
stephenswat: I would identify two reasons; first, smaller events will fill up less of the caches, so if you want to get comparable caching effects you will want to preprocess more of them. More importantly, the warm-up events serve to allow the device to scale up its power states and stabilize its execution frequency. A small number of small events will do a worse job at this than a small number of large events, simply due to the difference in execution time.

As far as I am concerned, the way @guilhermeAlmeida1 does this is the way to go. I think he takes 10% of the total number of events and runs them as warm-up events. That way, as long as the total event count is scaled properly to the pile-up, the warm-up will too.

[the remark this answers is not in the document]
guilhermeAlmeida1: That is indeed how I do things. Your logic sounds correct but I admit my logic mostly came just from empirical data where I generally get more "stable" results this way.

scripts/traccc_throughput_mt_profiler.sh
krasznaa: Sure. That could help in some situations.

scripts/traccc_throughput_mt_plotter.py
krasznaa: The script takes <1 second to process 5 CSV files. So I'm not going to put effort into making this more efficient.

krasznaa:
Apart from 1 still pending naming typo fix, I'm now happy with this.

And no, the `pass` statements are not going anywhere.

scripts/traccc_throughput_mt_profiler.sh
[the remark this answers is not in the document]
krasznaa: Yepp, and you guys should implement all of this in a next step. :+1: For the <% effects that this includes (For small events we run many of them anyway. The warm-up was resulting in pretty negligible effects in my tests for those events.) I was not willing to spend more time on this.

extras/traccc_throughput_mt_plotter.py
```diff
@@ -0,0 +1,441 @@
...
+
+        # Create the profile.
+        profile = configureProfile(ROOT.TProfile(
+            'max_throughput_profile_%i' % counter,
+            'Maximum Throughput;;Events/Second',
+            len(samples), 0.5, len(samples) + 0.5), platform, graphicsOptions)
+        for i in range(len(samples)):
+            profile.GetXaxis().SetBinLabel(i + 1, samples[i])
+            pass
+        profile.GetYaxis().SetRangeUser(40., min_max_throughput[1] * 5.)
```
guilhermeAlmeida1, line 335: If I understand this correctly, wouldn't defining the minimum like this be a better approach?
```suggestion
        profile.GetYaxis().SetRangeUser(min_max_throughput[0]*0.8, min_max_throughput[1] * 5.)
```

guilhermeAlmeida1 (commented):
I must admit some of the details escaped me, but looks like it gets the job done. Just a small comment on my end

extras/traccc_throughput_mt_plotter.py
krasznaa: Yes, I absolutely just hardcoded a value as I wanted to get something reasonable-looking.

I think your proposal of "some reasonable" multiplier is a good one. I'm afraid of making the code "too smart", but this is not that just yet. :stuck_out_tongue:

extras/traccc_throughput_mt_profiler.sh
```diff
@@ -0,0 +1,150 @@
...
+TRACCC_EVT_COUNT["ttbar_mu40"]=$((500*${TRACCC_EVT_MULTI}))
+TRACCC_EVT_COUNT["ttbar_mu60"]=$((250*${TRACCC_EVT_MULTI}))
+TRACCC_EVT_COUNT["ttbar_mu80"]=$((200*${TRACCC_EVT_MULTI}))
+TRACCC_EVT_COUNT["ttbar_mu100"]=$((150*${TRACCC_EVT_MULTI}))
+TRACCC_EVT_COUNT["ttbar_mu140"]=$((90*${TRACCC_EVT_MULTI}))
+TRACCC_EVT_COUNT["ttbar_mu200"]=$((50*${TRACCC_EVT_MULTI}))
+TRACCC_EVT_COUNT["ttbar_mu300"]=$((20*${TRACCC_EVT_MULTI}))
+
+# Put a header on the CSV file.
+echo "directory,threads,loaded_events,cold_run_events,processed_events,warm_up_time,processing_time" \
```
guilhermeAlmeida1, line 120: This will need an update
```suggestion
echo "directory,threads,loaded_events,cold_run_events,processed_events,target_cells_per_partition,warm_up_time,processing_time" \
```

krasznaa: Good catch! :+1:

extras/traccc_throughput_mt_plotter.py
krasznaa: :+1:

stephenswat (approved):
I think this is good to go now, so unless @krasznaa objects let's go ahead.

PR #345, opened 2023-03-15
Make cuda chain use asynchronous copies and kernel launches throughout

guilhermeAlmeida1:
At the moment we can't *really* compare our multi-threaded CUDA vs SYCL performance as we weren't using the asynchronous capabilities of CUDA in most of the chain, where kernel launches and memory copies are happening synchronously.
This PR fixes that by passing a `cuda::queue` object to the `seeding_algorithm`. 

However, as @krasznaa pointed out, we still have a lot of synchronisation points between kernels (and copies) which could be eliminated. I will handle this in an upcoming PR both for CUDA and SYCL.

**Early** results are quite interesting, it seems that at low pile-up the native cuda algorithm significantly outperforms sycl, but at high pile-up the performance becomes essentially the same

### DIFF
--- a/device/cuda/include/traccc/cuda/seeding/seed_finding.hpp
+++ b/device/cuda/include/traccc/cuda/seeding/seed_finding.hpp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2021-2022 CERN for the benefit of the ACTS project
+ * (c) 2021-2023 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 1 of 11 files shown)


krasznaa (approved):
As long as it actually works in runtime tests, this is a good direction. :wink:

Note that SYCL doesn't do **any** asynchronous copies yet. I believe that's the main part of the performance disadvantage at low pileup in your latest tests.

guilhermeAlmeida1:
> Note that SYCL doesn't do **any** asynchronous copies yet. I believe that's the main part of the performance disadvantage at low pileup in your latest tests.

Yes, I should've pointed this out in my initial message, I also intend to take care of this in the separate PR.

PR #347, opened 2023-03-16
Implement CUDA block- and warp-wide odd-even sort

stephenswat:
Sorting small arrays is a relatively common problem in GPGPU programming. Many useful algorithms exist, and some are provided by libraries like CUB. An algorithm close to my heart is odd-even sort because it is exceedingly simply, relatively efficient for small arrays and, importantly, it uses O(1) space. This commit adds new implementations of this sorting algorithm for warp- and block-wide thread groups in CUDA.

### DIFF
--- a/device/cuda/src/utils/sort.cuh
+++ b/device/cuda/src/utils/sort.cuh
@@ -1,0 +1,120 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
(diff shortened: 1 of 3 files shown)


krasznaa (changes requested):
Pretty much the same comments as in #346. With the addition that this code I'll really try to implement in a way so that it could be shared between CUDA and SYCL.

But that's not for this PR. Here we should just make sure that CUDA-specific code is not put into a public `.hpp` file.

stephenswat:
Updated.

stephenswat:
Replaced by #632.

PR #348, opened 2023-03-16
Implement near-transparent layout-swappable arrays

stephenswat:
This code came to me in a dream.

We have recently been doing a lot of research into the viability of struct-of-array (SoA) and array-of-struct (AoS) layouts. This distinction is important in factorized computing on CPUs as well as on GPUs as it can impact locality and the ability for the processor to coalesce memory accesses. The problem is that swapping between these layouts requires a lot of programming effort, and I am _lazy_. Llama gives us some of this, but the programming API is in some ways inconvenient.

This is why I came up with this near transparent model for array wrappers which can be swapped between multiple layouts. Let us first consider a classical, C++-native array-of-structs layout. First, we define the struct:

```cpp
struct vector {
    float x;
    float y;
    float z;
};
```

We can then define an array of this type:

```cpp
vector array[100];
```

And then we can sum up the values inside of it:

```cpp
float acc = 0.f;
for (uint32_t i = 0; i < 100; ++i) {
    acc += array[i].x;
}
```

This works, but switching it to SoA is a pain in the rear. Let's consider what this looks like with the layout schemes implemented in this PR. We will first reclare our struct, in a slightly different fashion:

```cpp
template<template <typename> typename F>
struct vector {
    using tuple_t = std::tuple<float, float, float>;
    F<float> x;
    F<float> y;
    F<float> z;
}
```

It's not identical, but the structure is very similar. This is why I call it a near-transparent scheme. Then we create an array of this type (using some memory resource):

```cpp
using array_t = traccc::array_wrapper<soa, vector>;
array_t::owner owner(mr, 100);
array_t::hande array(owner);
```

And then we can access the data:

```cpp
float acc = 0.f;
for (uint32_t i = 0; i < 100; ++i) {
    acc += array[i].x;
}
```

All we now need to do to change this to an AoS layout is change the type declaration, such that:

```cpp
using array_t = traccc::array_wrapper<aos, vector>;
```

Notice the change from `soa` to `aos`. That's all that's necessary. Code using these structures should be 100% compatible, and able to read and write to the array without issue.

This PR also comes with a CUDA test that sums up the first element of one billion 3-vectors in an AoS and SoA layout. This serves to confirm functionality as well as the performance effects of this choice. On the A5000 in `pcadp04` the operation takes 18.0 milliseconds when using the SoA layout, and 27.2 milliseconds when using AoS: The SoA approach is 51% faster!

### DIFF
--- a/core/include/traccc/utils/array_wrapper.hpp
+++ b/core/include/traccc/utils/array_wrapper.hpp
@@ -1,0 +1,238 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <traccc/definitions/qualifiers.hpp>
+#include <traccc/utils/functor.hpp>
+#include <type_traits>
+#include <utility>
+#include <vecmem/memory/memory_resource.hpp>
+#include <vecmem/memory/unique_ptr.hpp>
+
+namespace traccc {
+/**
+ * @brief Tuple of plain-old-data types.
+ *
+ * This is just a super-simple re-implementation of `std::tuple` that ensures
+ * that the type is trivially constructible.
+ */
+template <typename... Ts>
+struct pod {};
+
+template <>
+struct pod<> {};
+
+template <typename T, typename... Ts>
+struct pod<T, Ts...> {
+    static_assert(std::is_pod_v<T>, "Types in POD-tuple must each be POD.");
+
+    T v;
+    pod<Ts...> r;
+};
+
+/**
+ * @brief Get a given value from a POD tuple.
+ *
+ * @tparam I The index in the POD type to retrieve.
+ * @tparam Ts The types of the POD tuple.
+ *
+ * @param p A valid of the POD tuple type described by `Ts...`.
+ *
+ * @return A value of type `Ts[I]` as contained in `p`.
+ */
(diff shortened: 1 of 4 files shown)


krasznaa:
:thinking: We should talk about this at the next parallelization meeting. I'd definitely like to see some performance numbers from ESI before we go forward with anything like this.

Note that since in ATLAS we use SoA in any case (https://cds.cern.ch/record/2007523/files/ATL-SOFT-SLIDE-2015-161.pdf), So there I will not want to use any absolutely generic thing. Rather I'll want to just choose type-by-type how PoD types would be grouped. (Possibly using small structs of PoDs in our arrays in some cases.)

Also: https://github.com/AIDASoft/podio/ Let's not get into trademark disputes. :stuck_out_tongue:

stephenswat:
This should now pass the CI; turns out the `__reduce_add_sync` intrinsic I used in the tests is only available on Compute Capability 8.0 and higher. Replaced this by a butterfly reduction.

krasznaa (approved):
Okay, I'm curious where we end up with this.

As you wrote, if we need something like this, eventually we should look at LLAMA. But I'm not against experimenting with stuff like this for now...

guilhermeAlmeida1:
@stephenswat can you rebase this since it's already approved and we'll get it in?

stephenswat:
Thanks @guilhermeAlmeida1, I rebased the code; could you re-approve?

guilhermeAlmeida1 (approved):
I didn't actually go through any of this, just re-approving after a rebase.

PR #349, opened 2023-03-16
Add Thrust Scan and Sort tests

beomki-yeo:
Slightly inspired by #347: just to demonstrate that we can also use thrust algorithms on `vecmem::device_vector`
The thrust sort function is possibly slower than the bubble sort but it can be used for arbitrary vector size.


stephenswat (approved):
Very nice, more tests are always welcome. :smile:

PR #350, opened 2023-03-16
Remove deprecated information from CPU seeding EDM

guilhermeAlmeida1:
In #326 I forgot to remove some no longer needed information from the doublet & triplet core EDM given that it's now separated from that of the device.


stephenswat (approved):
Happy to approve this as long as the CI passes.

PR #353, opened 2023-03-20
Data Directory Re-Design, main branch (2023.03.20.)

krasznaa:
After the latest round of issues with https://gitlab.cern.ch/acts/traccc-data, this is an alternative proposal for how the project could get the data files that it needs.

The current version of the files were packaged in the "V1 files" under https://acts.web.cern.ch/traccc/data/. Which could be created and downloaded with the 2 scripts included in this PR.

The scripts themselves are pretty bare bones. But they should do the trick in such a basic setup as well. Note for instance that the retry setup for the downloads with `wget` was absolutely necessary. Apparently all CI jobs starting the download at once is pretty heavy traffic for the webserver. :confused: (If we get yelled at for this as well, I'll get **very** upset.) Also note that the error messages are not all super user friendly. For instance in case of a download error, you would get something like:

```
--2023-03-20 18:25:27--  https://acts.web.cern.ch/traccc/data//traccc-data-v1.tar.gz
Resolving acts.web.cern.ch (acts.web.cern.ch)... 2001:1458:d00:62::100:2e8, 2001:1458:d00:16::41d, 2001:1458:d00:19::12e, ...
Connecting to acts.web.cern.ch (acts.web.cern.ch)|2001:1458:d00:62::100:2e8|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 856290640 (817M) [application/x-gzip]
Saving to: ‘traccc-data-v1.tar.gz’

traccc-data-v1.tar.gz                    100%[===============================================================================>] 816.62M  58.9MB/s    in 14s     

2023-03-20 18:25:41 (56.7 MB/s) - ‘traccc-data-v1.tar.gz’ saved [856290640/856290640]

--2023-03-20 18:25:41--  https://acts.web.cern.ch/traccc/data//traccc-data-v1.md5
Resolving acts.web.cern.ch (acts.web.cern.ch)... 2001:1458:d00:62::100:2e8, 2001:1458:d00:16::41d, 2001:1458:d00:19::12e, ...
Connecting to acts.web.cern.ch (acts.web.cern.ch)|2001:1458:d00:62::100:2e8|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 60
Saving to: ‘traccc-data-v1.md5’

traccc-data-v1.md5                       100%[===============================================================================>]      60  --.-KB/s    in 0s      

2023-03-20 18:25:46 (19.6 MB/s) - ‘traccc-data-v1.md5’ saved [60/60]

Files "traccc-data-v1.md5" to "traccc-data-v1.md5-test" are different.
```

But again, making this better did not seem to be important enough... :thinking:

Note that in the CI setup I couldn't remove the installation of git-lfs yet. Since the CI for this PR still needs to get the HEAD of the main branch first as it seems, before switching to my branch. But once/if this all goes into the main branch, we'll be able to remove the installation of git-lfs from the CI recipe.

### DIFF
--- a/data/.gitignore
+++ b/data/.gitignore
@@ -1,0 +1,15 @@
+#
+# (c) 2023 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+#
+cca_test/
+detray_simulation/
+single_module/
+tml_detector/
+tml_full/
+tml_pixel_barrel/
+tml_pixels/
+two_modules/
+*.tar.gz
+*.md5
--- a/data/traccc_data_get_files.sh
+++ b/data/traccc_data_get_files.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+#
+# (c) 2023 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+#
+# Script downloading the traccc data file(s) through HTTPS, and unpacking them.
+#
+
+# Stop on errors.
+set -e
+set -o pipefail
+
+# Function printing the usage information for the script.
+usage() {
+   echo "Script downloading/unpacking data TGZ/MD5 files"
+   echo ""
+   echo "Usage: traccc_data_get_files.sh [options]"
+   echo ""
+   echo "Options:"
+   echo "  -f <filename>        Name of the data file, without its extension"
+   echo "  -d <webDirectory>    Directory holding the data and MD5 files"
+   echo "  -o <dataDirectory>   Main data directory"
+   echo "  -c <cmakeExecutable> CMake executable to use in the script"
+   echo "  -w <curlExecutable>  CUrl executable to use in the script"
+   echo ""
+}
+
+# Default script arguments.
+TRACCC_DATA_NAME=${TRACCC_DATA_NAME:-"traccc-data-v1"}
+TRACCC_WEB_DIRECTORY=${TRACCC_WEB_DIRECTORY:-"https://acts.web.cern.ch/traccc/data"}
+TRACCC_DATA_DIRECTORY=${TRACCC_DATA_DIRECTORY:-$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)}
+TRACCC_CMAKE_EXECUTABLE=${TRACCC_CMAKE_EXECUTABLE:-cmake}
+TRACCC_CURL_EXECUTABLE=${TRACCC_CURL_EXECUTABLE:-curl}
+
+# Parse the command line argument(s).
+while getopts ":f:d:o:c:wh" opt; do
+   case $opt in
+      f)
+         TRACCC_DATA_NAME=$OPTARG
+         ;;
+      d)
+         TRACCC_WEB_DIRECTORY=$OPTARG
+         ;;
+      o)
+         TRACCC_DATA_DIRECTORY=$OPTARG
+         ;;
+      c)
+         TRACCC_CMAKE_EXECUTABLE=$OPTARG
+         ;;
+      w)
+         TRACCC_CURL_EXECUTABLE=$OPTARG
+         ;;
+      h)
+         usage
+         exit 0
+         ;;
+      :)
+         echo "Argument -$OPTARG requires a parameter!"
+         usage
+         exit 1
+         ;;
+      ?)
+         echo "Unknown argument: -$OPTARG"
+         usage
+         exit 1
+         ;;
+   esac
+done
+
+# Go into the target directory.
+cd "${TRACCC_DATA_DIRECTORY}"
+
+# Download the TGZ and MD5 files.
+"${TRACCC_CURL_EXECUTABLE}" --retry 5 --retry-connrefused --retry-delay 10 \
+   --output "${TRACCC_DATA_NAME}.tar.gz"                                   \
+   "${TRACCC_WEB_DIRECTORY}/${TRACCC_DATA_NAME}.tar.gz"
+"${TRACCC_CURL_EXECUTABLE}" --retry 5 --retry-connrefused --retry-delay 10 \
+   --output "${TRACCC_DATA_NAME}.md5"                                      \
+   "${TRACCC_WEB_DIRECTORY}/${TRACCC_DATA_NAME}.md5"
+
+# Verify that the download succeeded.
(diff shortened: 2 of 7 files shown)


data/.gitignore
```diff
@@ -0,0 +1,15 @@
...
+cca_test/
+detray_simulation/
+single_module/
+tml_detector/
+tml_full/
+tml_pixel_barrel/
+tml_pixels/
+two_modules/
+*.tar.gz
+*.md5
```
stephenswat, line 15: Let's just set this to `*` so it automatically adapts to future additions to the data files.

beomki-yeo:
If others are OK, I am OK. Can random users both download and upload the files? or does it require some permission.

data/traccc_data_get_files.sh
```diff
@@ -0,0 +1,95 @@
...
+# Extract the data files.
+"${TRACCC_CMAKE_EXECUTABLE}" -E tar xf "${TRACCC_DATA_NAME}.tar.gz"
+
+# Clean up.
+"${TRACCC_CMAKE_EXECUTABLE}" -E remove "${TRACCC_DATA_NAME}.tar.gz" \
+   "${TRACCC_DATA_NAME}.md5" "${TRACCC_DATA_NAME}.md5-test"
+
+# Leave the user with a message.
+"${TRACCC_CMAKE_EXECUTABLE}" -E echo \
+   "Files from ${TRACCC_DATA_NAME} unpacked under '${TRACCC_DATA_DIRECTORY}'"
```
stephenswat, line 95: I know that you really love CMake, but does using it for this buy us anything? I would posit that `tar`, `md5sum`, `rm` and `echo` are infinitely more ubiquitous in any kind of system than CMake is, so isn't this just adding an unnecessary dependency?

data/traccc_data_get_files.sh
```diff
@@ -0,0 +1,95 @@
...
+done
+
+# Go into the target directory.
+cd "${TRACCC_DATA_DIRECTORY}"
+
+# Download the TGZ and MD5 files.
+"${TRACCC_WGET_EXECUTABLE}" --tries=5 --retry-connrefused --waitretry=10 \
+   "${TRACCC_WEB_DIRECTORY}/${TRACCC_DATA_NAME}.tar.gz"
+"${TRACCC_WGET_EXECUTABLE}" --tries=5 --retry-connrefused --waitretry=10 \
+   "${TRACCC_WEB_DIRECTORY}/${TRACCC_DATA_NAME}.md5"
```
stephenswat, line 78: Could we split these files up so it becomes possible to download only a part of the testing data? That would reduce the load on wherever we host this, and make it easier as well as quicker to partially download the data (say, in the CI).

krasznaa:
> Can random users both download and upload the files? or does it require some permission.

This is one of the downsides of this. That upload is only available to privileged users. So the procedure would be that people make a new TGZ/MD5 file pair, put them somewhere public (let's say CERNBox) and then somebody with the required privileges puts the files in the appropriate EOS directory. Then, once the file is in place, a PR can be opened that switches to the new file.

data/traccc_data_get_files.sh
krasznaa: :thinking: I'm not a big fan of what that would mean for versioning these files. Yes, the one TGZ file is pretty big. But at least then you only have a single file, with a single version number.

data/traccc_data_get_files.sh
krasznaa: Windows... I'm actually pretty annoyed that I couldn't find a web download command in CMake.

data/.gitignore
krasznaa: But then it would also ignore the scripts in the directory. :confused:

data/traccc_data_get_files.sh
stephenswat: Can't we put the files in a versioned directory, i.e. `https://acts.web.cern.ch/traccc/data/v4/tml_full.tar.gz`?

data/.gitignore
krasznaa: Yeah, no...

data/traccc_data_get_files.sh
stephenswat: Does the Windows Subsystem for Linux or whatever it's called not come with these tools installed?

data/traccc_data_get_files.sh
krasznaa: We could. But again, I don't like the added complication. :frowning:

We could make TGZ files just meant for the CI. :thinking: So that CI jobs only download the files that they really need. While the download script by default downloads the "big file".

data/traccc_data_get_files.sh
stephenswat: Can you run Bash scripts on Windows without WSL these days?

Genuine question I haven't used Windows in 10 years.

krasznaa: Admittedly, not out of the box. You'd have to use something like Cygwin to run an actual BASH script as far as I know.

My idea here is rather that CMake commands like these could be translated into `.bat` files with pretty much a copy-paste if it came to that.

To make the copy-paste even better, Windows 10+ apparently comes with `curl` out of the box. So let me switch to that from `wget`...

data/traccc_data_get_files.sh
krasznaa: I had a quick look, using commands like:

```sh
git grep "tml_pixel_barrel" -- "tests/*"
```

The **only** top level directory that we don't pick up anything from in our tests is `tml_pixels/`. From everything else we pick up at least some stuff.

At this stage it could still be a good idea to make a "CI specific" tarball with just the files that are actually used. This will need some detective work, but it could make the "CI specific" tarball a fraction of the size of the full one. But it's clear that simply splitting the tarballs by the top level directories would win us nothing with the CI.

data/.gitignore
stephenswat: Well, you can lead someone to water I guess. :smile:

stephenswat (approved):

stephenswat (approved):

paulgessinger:
This is what it is.

stephenswat:
It be what it be.

PR #354, opened 2023-03-21
Changes to build system for setup of Thrust

stephenswat:
This commit ensures that Thrust is marked as a system include, cleaning up some warnings generated by the library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,9 @@ set( TRACCC_THRUST_OPTIONS "" CACHE STRING
    "Extra options for configuring how Thrust should be used" )
 mark_as_advanced( TRACCC_THRUST_OPTIONS )
 thrust_create_target( traccc::Thrust ${TRACCC_THRUST_OPTIONS} )
+if( ${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.25.0" )
+    set_target_properties( traccc::Thrust PROPERTIES SYSTEM TRUE )
+endif()
 
 # Set up TBB.
 option( TRACCC_SETUP_TBB


CMakeLists.txt
```diff
@@ -102,7 +102,7 @@ option( TRACCC_SETUP_THRUST
    "Set up the Thrust target(s) explicitly" TRUE )
 option( TRACCC_USE_SYSTEM_THRUST
```
guilhermeAlmeida1, line 103: Shouldn't this be `TRUE` by default instead?

stephenswat: Yes you're absolutely right sorry.

guilhermeAlmeida1 (approved):

krasznaa:
Just to put here what I already told Stephen: We build Thrust ourselves by default, because we use it independent of CUDA.
  - https://github.com/acts-project/traccc/blob/main/core/CMakeLists.txt#L85
  - https://github.com/acts-project/traccc/blob/main/core/include/traccc/edm/details/container_base.hpp#L18

I'm open to changing that, but that will require a bit more work than what's in this PR. As it stands, this PR should **not** go through.

krasznaa (changes requested):
Just to make sure that merge is not hit by mistake.

beomki-yeo:
I thought we fetched Thrust library to use it in host side (@krasznaa Is that correct?)

stephenswat:
Changed this to fetch version 2.1.0 instead, which resolves the issue. I still really don't like pulling in all of Thrust for a few product types, but we can cross that bridge later. Again.

CMakeLists.txt
```diff
@@ -116,6 +116,7 @@ set( TRACCC_THRUST_OPTIONS "" CACHE STRING
    "Extra options for configuring how Thrust should be used" )
 mark_as_advanced( TRACCC_THRUST_OPTIONS )
 thrust_create_target( traccc::Thrust ${TRACCC_THRUST_OPTIONS} )
+set_target_properties( traccc::Thrust PROPERTIES SYSTEM TRUE )
```
krasznaa, line 119: This is generally good, but it only works with CMake 3.25+.

https://cmake.org/cmake/help/latest/prop_tgt/SYSTEM.html

What would be compatible with previous versions as well would be something like:

```cmake
get_target_property( _thurstIncludeDirs traccc::Thrust INTERFACE_INCLUDE_DIRECTORIES )
target_include_directories( traccc::Thrust SYSTEM INTERFACE ${_thurstIncludeDirs} )
```

Didn't actually test it though... :thinking:

stephenswat: Oops, my mistake for thinking that something simple might actually be simple to do in CMake.

image

stephenswat: How about I set the `IMPORTED` property on it?

krasznaa: 😕 Did you remove your comment?

No, `IMPORTED` is not a property that could be set by hand.

https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED.html

And I expect that the `traccc::Thrust` target would have it set to `TRUE` anyway.

PR #355, opened 2023-03-21
Add a very simple pair type constructor

stephenswat:
This negates the need for using Thrust unconditionally, as discussed in PR #354.

### DIFF
--- a/core/include/traccc/edm/details/container_base.hpp
+++ b/core/include/traccc/edm/details/container_base.hpp
@@ -10,12 +10,10 @@
 // Project include(s).
 #include "traccc/definitions/qualifiers.hpp"
 #include "traccc/edm/details/container_element.hpp"
+#include "traccc/types/pair.hpp"
 
 // VecMem include(s).
 #include <vecmem/memory/memory_resource.hpp>
-
-// Thrust include(s).
-#include <thrust/pair.h>
 
 // System include(s).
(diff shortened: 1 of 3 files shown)


stephenswat:
The advantage is that `std::pair` does not mark its functions as `constexpr`, so CUDA complains about them (but still works perfectly). This doesn't lose us anything over `thrust::pair` that I can think of.

guilhermeAlmeida1 (changes requested):

beomki-yeo:
This is the duplicate of #229. I believe you and @krasznaa will discuss enough about this but I think that PR was closed for a good reason.

guilhermeAlmeida1:
Just to keep me in the loop, can I get a brief description on why we opted to keep `thrust::pair` in then?

krasznaa:
> Just to keep me in the loop, can I get a brief explanation on why we opted to keep `thrust::pair` in then?

It's a slightly longer story. But mainly because I didn't fancy implementing our own pair and tuple types, when we can just as well take these from Thrust.

We can discuss in more detail on Friday.

beomki-yeo:
Other point of view; We might need to abandon `thrust::tuple` in detray  because its maximum size is 10 which can be easily reached by a complicated geometry. So.. actually I am not very against implementing our own type but I think this should be done in a low level library instead of traccc

PR #356, opened 2023-03-21
Add concurrency group for `build` CI jobs

stephenswat:
We currently build the entirety of the software whenever someone pushes to a branch that is linked to a PR. This is useful, but there are cases where development outpaces the rate at which the CI runs. In such cases, we run multiple CI jobs for the same PR at the same time, which occupies resources that could be spent elsewhere (thus making the CI run more slowly) and, not unimportantly, wastes electricity.

This commit adds the `build` CI jobs to a concurrency group that ensures that any running builds will be automatically cancelled if a new commit is pushed to an existing pull request.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -12,6 +12,10 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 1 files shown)


krasznaa (approved):
Sure. I'll trust that this is the correct formalism for that. :wink:

PR #357, opened 2023-03-21
WIP: Change sycl chain to use async copying and an ordered queue

guilhermeAlmeida1:
This PR moves towards further asynchronising kernel launches and memory copies in our algorithmic chains.
* `vecmem::sycl::copy` is replaced with `vecmem::sycl::async_copy`.
* `sycl::queue` is constructed with [`::sycl::property::queue::in_order`](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:queue-properties). This allows removing all the synchronisation points in between kernels, only requiring them after device-to-host copies.

Keeping this as a draft as I don't really understand the results right now... Somehow multi-threaded performance becomes worse, and that shouldn't really happen.

PS: a big chunk of the changed LOCs in this PR right now come from the formatting, not so actual changes to the code.

### DIFF
--- a/device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
+++ b/device/sycl/include/traccc/sycl/clusterization/clusterization_algorithm.hpp
@@ -37,12 +37,14 @@ class clusterization_algorithm
     /// Constructor for clusterization algorithm
     ///
     /// @param mr is a struct of memory resources (shared or host & device)
+    /// @param copy The copy object to use for copying data between device
+    ///             and host memory blocks
     /// @param queue is a wrapper for the for the sycl queue for kernel
     /// invocation
     /// @param target_cells_per_partition the average number of cells in each
     /// partition
     clusterization_algorithm(const traccc::memory_resource& mr,
-                             queue_wrapper queue,
+                             vecmem::copy& copy, queue_wrapper queue,
                              const unsigned short target_cells_per_partition);
 
     /// @param cells        a collection of cells
@@ -61,7 +63,7 @@ class clusterization_algorithm
 
     traccc::memory_resource m_mr;
     mutable queue_wrapper m_queue;
-    std::unique_ptr<vecmem::copy> m_copy;
+    vecmem::copy& m_copy;
 };
 
 }  // namespace traccc::sycl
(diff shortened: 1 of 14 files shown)


stephenswat:
I know fuck all about SYCL, but doesn't orderedness of the queue impart additional constraints, thus explaining why the code might be slower? For example, the following task flow (with kernels issued in alphabetical order) might run well if orderedness is not required:

```mermaid
flowchart LR

A --> C
A ---> B
C --> D
D --> E
D --> F
B --> G
E --> G
F --> G
```

Whereas it would take much longer with an orderedness requirement:

```mermaid
flowchart LR

A --> B
A --> C
C --> D
D --> E
D ---> F
B ------> G
E ---> G
F --> G
```

guilhermeAlmeida1:
That's true but our kernels already have an ordereness requirement, because they all depend on the immediately previous one. What we were doing before is having a `wait()` statement after every kernel, which here gets replaced by the ordered queue. Each CPU thread will have a separate queue, and I assume that between different queues the "ordereness" requirement does not do anything.

krasznaa:
🤔 We really need to introduce explicit output checks for the full chain algorithms at this point. As I'm concerned that we might not be producing meaningful output already.

But beside that, NSight should be able to tell us what's happening. I hope at least.

guilhermeAlmeida1:
Closing this in favour of #371 

After quite some investigating, something seems to be going inherently wrong with using [`sycl in_order queues`](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:queue-properties) which I can't quite figure out.

It looks like using an in-order queue, when compiling towards a CUDA device, results in it being translated to a CUDA stream. This should be relatively self-explanatory as CUDA streams are themselves ordered. 
Using an out-of-order sycl::queue, however, results in a new CUDA stream getting created everytime there is a kernel launch.

In our scenario, where all the kernels have a direct dependence on the previous kernel, using an out-of-order means having to wait on each kernel before the next one launches, similarly to what happens in a in-order queue. One would expect it to be quicker, or at least equally fast, to use an in-order queue then, but from all my testing on an A5000 this is not the case.

PR #352, opened 2023-03-18
Add CUDA Track Finding Algorithm based on CKF

beomki-yeo:
Detailed description will be added later...

PR #358, opened 2023-03-21
Update to detray v28

beomki-yeo:
This is a part of changes in #352.
Please mind that this PR includes few changes not related to the detray update (e.g. Kalman fitter data name changes, bug fix in options, etc.).

Also the new data file that I sent to @krasznaa  needs to be uploaded.

### DIFF
--- a/examples/options/include/traccc/options/options.hpp
+++ b/examples/options/include/traccc/options/options.hpp
@@ -21,7 +21,7 @@
 namespace traccc {
 
 namespace {
-constexpr char s_separator = ':';
+const std::string s_separator = ":";
 }
 
 /// A fixed number of real values as one user option.
@@ -41,8 +41,7 @@ void parse_variable(std::istream& is, std::vector<value_t>& values,
     std::string buf;
     is >> buf;
     std::vector<std::string> stringValues;
-    boost::split(stringValues, buf,
-                 boost::is_any_of(std::string(&s_separator)));
+    boost::split(stringValues, buf, boost::is_any_of(s_separator));
     for (const std::string& stringValue : stringValues) {
         values.push_back(convert(stringValue));
     }
@@ -92,7 +91,7 @@ inline std::ostream& operator<<(std::ostream& os,
 
     for (size_t i = 0; i < kSize; ++i) {
         if (0u < i) {
-            os << s_separator;
+            os << s_separator.c_str();
         }
         os << data[i];
     }
--- a/examples/options/include/traccc/options/options.hpp
+++ b/examples/options/include/traccc/options/options.hpp
@@ -21,7 +21,7 @@
 namespace traccc {
 
 namespace {
-const std::string s_separator = ":";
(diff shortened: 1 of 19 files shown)


examples/options/include/traccc/options/options.hpp
```diff
@@ -21,7 +21,7 @@
 namespace traccc {
 
 namespace {
-constexpr char s_separator = ':';
+const std::string s_separator = ":";
```
krasznaa, line 24: Since this thing is in the header file, let's make it into a `static const` variable. I wonder if `static constexpr` would even work... :thinking:

The private namespace does a very similar sort of thing, but I'm just partial to using `static` on such constants.

examples/options/include/traccc/options/options.hpp
```diff
@@ -92,7 +91,7 @@ inline std::ostream& operator<<(std::ostream& os,
 
     for (size_t i = 0; i < kSize; ++i) {
         if (0u < i) {
-            os << s_separator;
+            os << s_separator.c_str();
```
krasznaa, line 94: I think this is a fine change in order to avoid creating a temporary `std::string` every time the function is called. Even if I don't actually care about the performance of the command line processing code...

But not passing the string directly to the output stream as-is, is really just over-doing things. Let's just give the string as-is to the stream. (I.e. don't modify this line.)

krasznaa (commented):
If you feel strongly about it, I can forget about my non-important comments. But if you have the necessary cycles, I would prefer to change those lines. :wink:

examples/options/include/traccc/options/options.hpp
stephenswat: It would need to be a `std::string_view`, then it can be `static constexpr`.

examples/options/include/traccc/options/options.hpp
beomki-yeo: OK

examples/options/include/traccc/options/options.hpp
beomki-yeo: I changed to `static constexpr std::string_view` and also added a simple unit test for option parsing

krasznaa (approved):
:+1:

stephenswat:
Can we possible release v28 and jump to that instead?

krasznaa:
> Can we possibly release v28 and jump to that instead?

Depends on how long that takes. Since I'd really prefer to put the data changes in finally.

beomki-yeo:
That's OK. Do you want to include `detray::tuple` in v28?

stephenswat:
No, I am only looking for the fixes to the CMake configuration; the rest is not immediately relevant. :slightly_smiling_face:

beomki-yeo:
Then let me release v28 and makes the change here accordningly

beomki-yeo:
git is down right now

beomki-yeo:
git is back :p

stephenswat (approved):

PR #360, opened 2023-03-22
Thrust System Includes, main branch (2023.03.22.)

krasznaa:
This is meant as a replacement for #354.

The `traccc::Thrust` target is sort of a virtual target that the Thrust CMake code sets up on top of the low-level / real targets (`Thrust::Thrust`, `CUB::CUB`). So one needs to get the `INTERFACE_INCLUDE_DIRECTORIES` properties from those "real" targets, and then add them as system include directories on `traccc::Thrust`. (It's a long story why one should not actually try to modify `Thrust::Thrust` and `CUB::CUB` themselves...)


stephenswat (approved):

krasznaa:
For the love of all that's holy... If you insist on updating my branch, please use rebase. Let's not introduce unnecessary merge commits for no reason whatsoever...

stephenswat:
I'll start squashing your commits.

stephenswat (approved):

PR #361, opened 2023-03-23
Change CPU clustering to use flat (alt) edm

guilhermeAlmeida1:
This PR almost fully eliminates the use of all the "outdated" jagged EDM in the clusterisation.
It is quite a lengthy PR as it really touches almost everything in our code, but there are still a few bits and pieces in our repository using the old edm, which is why I didn't just delete all the old headers just yet.

I'll admit that I made the physics performance "mapper" functions a bit uglier, but this whole piece of code was already a mess and the solution I implemented was the only one I could see without fully rewriting this from scratch.

### DIFF
--- a/core/include/traccc/edm/internal_spacepoint.hpp
+++ b/core/include/traccc/edm/internal_spacepoint.hpp
@@ -23,16 +23,8 @@ namespace traccc {
 template <typename spacepoint_t>
 struct internal_spacepoint {
 
-    // FIXME: geometry_id is hard-coded here
-    using spacepoint_container_type = host_container<geometry_id, spacepoint_t>;
-    using link_type = typename spacepoint_container_type::link_type;
-
+    using link_type = typename collection_types<spacepoint_t>::host::size_type;
     link_type m_link;
-
-    /// TODO: This is a *VERY* temporary variable which will be removed asap.
-    /// only here while transitioning from the jagged to non-jagged
-    /// clusterization and prevent even more code duplication
-    unsigned int m_link_alt;
 
     scalar m_x;
     scalar m_y;
@@ -42,23 +34,10 @@ struct internal_spacepoint {
 
     internal_spacepoint() = default;
 
-    template <typename spacepoint_container_t>
-    TRACCC_HOST_DEVICE internal_spacepoint(
-        const spacepoint_container_t& sp_container, const link_type& sp_link,
-        const vector2& offsetXY)
-        : m_link(sp_link) {
-        const spacepoint_t& sp = sp_container.at(sp_link);
-        m_x = sp.global[0] - offsetXY[0];
-        m_y = sp.global[1] - offsetXY[1];
-        m_z = sp.global[2];
-        m_r = algebra::math::sqrt(m_x * m_x + m_y * m_y);
-        m_phi = algebra::math::atan2(m_y, m_x);
-    }
-
     TRACCC_HOST_DEVICE internal_spacepoint(const spacepoint_t& sp,
-                                           const unsigned int sp_link,
+                                           const link_type sp_link,
                                            const vector2& offsetXY)
-        : m_link_alt(sp_link) {
+        : m_link(sp_link) {
         m_x = sp.global[0] - offsetXY[0];
         m_y = sp.global[1] - offsetXY[1];
         m_z = sp.global[2];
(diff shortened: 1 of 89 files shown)


beomki-yeo:
> I'll admit that I made the physics performance "mapper" functions a bit uglier, but this whole piece of code was already a mess and the solution I implemented was the only one I could see without fully rewriting this from scratch.

You can mess it up as much as you want. I will rewrite them in the near future

guilhermeAlmeida1:
Performance-wise, the cpu-only throughput doens't change a lot. At low pile-up it actually becomes a bit faster with the flat edm but at high pile-up it's practically the same.

core/include/traccc/edm/internal_spacepoint.hpp
```diff
@@ -23,16 +23,7 @@ namespace traccc {
...
-    using spacepoint_container_type = host_container<geometry_id, spacepoint_t>;
-    using link_type = typename spacepoint_container_type::link_type;
-
-    link_type m_link;
-
-    /// TODO: This is a *VERY* temporary variable which will be removed asap.
-    /// only here while transitioning from the jagged to non-jagged
-    /// clusterization and prevent even more code duplication
-    unsigned int m_link_alt;
+    unsigned int m_link;
```
krasznaa, line 26: I would prefer to keep the

```c++
using link_type = unsigned int;
```

declaration, and then using `link_type` in the code, sort of as before. Would reduce the amount of necessary changes a little.

core/src/clusterization/measurement_creation.cpp
```diff
@@ -36,25 +36,24 @@ measurement_creation::output_type measurement_creation::operator()(
...
-            result.push_back(cell_module(module),
-                             measurement_collection_types::host(&(m_mr.get())));
-        }
-
         // Get the cluster.
         cluster_container_types::host::item_vector::const_reference cluster =
             clusters.at(i).items;
 
+        if (cluster.size() == 0) {
+            // this should never happen, but included for safekeeping.
```
krasznaa, line 44: :thinking: If it should really never happen, then let's put an assertion here. I.e.

```c++
assert(cluster.size() > 0);
```

device/cuda/src/cca/component_connection.cu
```diff
@@ -37,7 +37,7 @@ struct cell_container {
     channel_id* channel1 = nullptr;
     scalar* activation = nullptr;
     scalar* time = nullptr;
-    geometry_id* module_id = nullptr;
+    unsigned int* module_id = nullptr;
```
krasznaa, line 40: Did I miss something? Why did you remove the `geometry_id` type? :confused: Typedefs are generally good in this type of code.

examples/options/include/traccc/options/particle_gen_options.hpp
```diff
@@ -63,7 +63,7 @@ struct particle_gen_options {
         mom_range = vm["gen-mom-gev"].as<Reals<scalar_t, 2>>();
         phi_range = vm["gen-phi-degree"].as<Reals<scalar_t, 2>>();
         const auto eta_range = vm["gen-eta"].as<Reals<scalar_t, 2>>();
-        // TODO: remove the conversion here...
+        //  remove the conversion here...
```
krasznaa, line 66: I believe this change went into the PR by mistake...

krasznaa (changes requested):
I largely like it. Though admittedly I was really only looking for bad code smells in general. The PR is way too big to properly check it line-by-line... (At the same time, I didn't see an obvious way either of how to break it into multiple pieces.)

Hopefully we're now getting close to removing/renaming all the `alt_` prefixed types...

device/cuda/src/cca/component_connection.cu
guilhermeAlmeida1: Because I made this code match the "alt_cell" type, which uses links to a module collection to store the geometry information, not the actual geometry ids

krasznaa: :thinking: Then let's rename the variable to something like `module_link`. As `module_id` would suggest a different content.

guilhermeAlmeida1:
This should be ready to be looked at again.

stephenswat (approved):
Let's go ahead and put this in.

PR #362, opened 2023-03-23
SYCL and Build Updates, main branch (2023.03.23.)

krasznaa:
This is an attempt to make the builds not print a gazillion warnings with modern oneAPI and CUDA compilers.

First off, I switched the CI to using [ghcr.io/acts-project/ubuntu2004_cuda_oneapi:v37](https://github.com/acts-project/machines/pkgs/container/ubuntu2004_cuda_oneapi) and  [ghcr.io/acts-project/ubuntu2004_rocm_oneapi:v37](https://github.com/acts-project/machines/pkgs/container/ubuntu2004_rocm_oneapi). For now I removed the use of [ubuntu2004_oneapi](https://github.com/acts-project/machines/pkgs/container/ubuntu2004_oneapi), because:
  - The very latest version, oneAPI 2023.0.0, has a bug that prevents it building our full project (https://github.com/intel/llvm/issues/8588);
  - The current version used in the image (oneAPI 2022.0.2) is not compatible with the changes in this PR;
  - I did not want to update to oneAPI 2022.2.0. I'm not even entirely sure if it would support `sycl::local_accessor`, but I mainly just wanted to wait for oneAPI 2023.1 instead... (May re-visit this idea later on...)

I retired all the deprecated types from the SYCL code, to avoid warnings about them.

I modified `traccc_examples_sycl` to be an `OBJECT` instead of a `STATIC` library. As this avoids the following type of warning during linking:

```
[ 96%] Linking SYCL executable ../../../bin/traccc_throughput_st_sycl
error: overriding the module target triple with nvptx64-nvidia-cuda [-Werror,-Woverride-module]
```

(Something to be followed up with the Intel developers.)

Finally, I was getting a bunch of warnings from CUDA headers, while building [tests/cuda/test_thrust.cu](https://github.com/acts-project/traccc/blob/main/tests/cuda/test_thrust.cu). Which I believe is one of the problems that @guilhermeAlmeida1 was running into, and what triggered #354 and #360. The problem here was that the CUDA headers were picked up "implicitly" by `nvcc`. By making it explicit that the test should use `CUDA::cudart`, the build treats the CUDA headers as "system headers".

All in all, the code builds without warnings in this setup once again... :partying_face:

### DIFF
--- a/.github/ci_setup.sh
+++ b/.github/ci_setup.sh
@@ -1,6 +1,6 @@
 # TRACCC library, part of the ACTS project (R&D line)
 #
-# (c) 2022 CERN for the benefit of the ACTS project
+# (c) 2022-2023 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 #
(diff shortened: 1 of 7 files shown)


beomki-yeo (approved):

beomki-yeo:
I wonder if #358 will pass the CI after it is merged

krasznaa:
> I wonder if #358 will pass the CI after it is merged

Sort of. Since the open source Intel compiler does not use fast math by default.

And I just now see that I'll need to fix the test runtime. :frowning: Since the "host device" is no more with recent oneAPI versions, we need to turn off the SYCL tests... :frowning:

beomki-yeo (approved):

PR #296, opened 2022-12-09
GPU Track Fitting

beomki-yeo:
The PR is still in WIP (and depends on #264). I didn't even generalize it for both CUDA and SYCL yet. 
But I want to get some advises in advance because I am stuck right now.
The main problem was that I was forced to implement the template classes in the source file:
For example, `traccc::cuda::fitting_algorithm` is declared in `traccc/device/cuda/include/traccc/cuda/fitting/fitting_algorithm.hpp` and implemented in `traccc/device/cuda/src/fitting/fitting_algorithm.cu`
Because `traccc::cuda::fitting_algorithm` is a template class, it was necessary to explicitly instantiate explicitly the class with specific template parameters :/ (See the code below)

```c++
// Explicit template instantiation
using host_detector_type =
    detray::detector<detray::detector_registry::telescope_detector,
                     covfie::field>;
using device_detector_type =
    detray::detector<detray::detector_registry::telescope_detector,
                     covfie::field_view, detray::device_container_types>;
using b_field_t = typename host_detector_type::bfield_type;
using rk_stepper_type = detray::rk_stepper<b_field_t::view_t, transform3,
                                           detray::constrained_step<>>;
using device_navigator_type = detray::navigator<const device_detector_type>;
using device_fitter_type =
    kalman_fitter<rk_stepper_type, device_navigator_type>;
template class fitting_algorithm<device_fitter_type, host_detector_type>;
```
 
Instantiating the classes for every detector class is definitely not I want to do. I wonder if there is a workaround for this ?

PR #363, opened 2023-03-23
Add ability to import CMake target for covfie

stephenswat:
PR #296 introduces a dependency on some of the covfie targets, but does not actually attempt to load the library (and thus the targets). This commit adds functionality to the build system to do so.

### DIFF
--- a/extern/covfie/CMakeLists.txt
+++ b/extern/covfie/CMakeLists.txt
@@ -1,0 +1,38 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2023 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# CMake include(s).
+cmake_minimum_required( VERSION 3.14 )
+include( FetchContent )
+
+# Silence FetchContent warnings with CMake >=3.24.
+if( POLICY CMP0135 )
+   cmake_policy( SET CMP0135 NEW )
+endif()
+
+# Tell the user what's happening.
+message( STATUS "Fetching covfie as part of the traccc project" )
+
+# Declare where to get covfie from.
+set( TRACCC_COVFIE_SOURCE
+   "URL;https://github.com/acts-project/covfie/archive/refs/tags/v0.5.0.tar.gz;URL_MD5;e6683f6eb34b6156f8ac90facc7e538e"
+   CACHE STRING "Source for covfie, when built as part of this project" )
+mark_as_advanced( TRACCC_COVFIE_SOURCE )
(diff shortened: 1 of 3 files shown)


krasznaa (approved):
:+1:

stephenswat:
We're already explicitly linking targets, e.g. https://github.com/acts-project/traccc/blob/main/device/sycl/CMakeLists.txt#L39

stephenswat:
Done.

extern/covfie/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+mark_as_advanced( TRACCC_COVFIE_SOURCE )
+FetchContent_Declare( covfie ${TRACCC_COVFIE_SOURCE} )
+
+# Options used for covfie.
+set( COVFIE_BUILD_EXAMPLES OFF CACHE BOOL "Build covfie examples")
+set( COVFIE_BUILD_TESTS OFF CACHE BOOL "Build covfie tests")
+set( COVFIE_BUILD_BENCHMARKS OFF CACHE BOOL "Build covfie benchmarks")
+
+set( COVFIE_PLATFORM_CPU ON CACHE BOOL "Enable covfie CPU platform")
+set( COVFIE_PLATFORM_CUDA ON CACHE BOOL "Enable covfie CUDA platform")
```
krasznaa, line 32: This should be tied to `${TRACCC_BUILD_CUDA}`.

extern/covfie/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+set( COVFIE_REQUIRE_CXX20 OFF CACHE BOOL "Enable covfie C++20 requirement")
```
krasznaa, line 34: :thinking: You could tie this to `${CMAKE_CXX_STANDARD}` being larger-equal 20. It's good to have a separate variable for this in the Covfie build configuration, but here in this project we could tie the two together.

krasznaa (changes requested):
Just some further complaints. Also to trigger a rebase...

extern/covfie/CMakeLists.txt
stephenswat: I don't personally think this is worth doing because this only specifies a hard requirement, not whether C++20 features are used or not; that is derived automatically through feature flags. Also, as far as I can tell there's not really a good way to check whether the C++ standard is greater than a certain value?

extern/covfie/CMakeLists.txt
stephenswat: Sure, can't hurt. :slightly_smiling_face:

niermann999 (approved):
lgtm

PR #364, opened 2023-03-23
Add Spack development environment specification

stephenswat:
People with dual core laptops for which traccc takes ages to compile, unite!

This commit adds a Spack development environment specification that can be used to reproducibly create an environment in which traccc can be built.

Sadly, this has required me to make quite a few changes to the build system of traccc, algebra-plugins, and detray (including a fix for acts-project/algebra-plugins#36 and an identical bug in detray). Currently, these are Spack patches, but should be incorporated into the actual projects, too. This depends on #363, but can already be reviewed. :smile: 

Environment can be built as follows:

```sh
spack env create traccc spack.yaml
spack env activate traccc
spack repo add --scope env:traccc repo/
spack install
```

Currently supports CPU code and CUDA.

To-do:

- [x] Merge #363
- [x] Incorporate patches into detray and algebra-plugins
- [x] Correctly format package files
- [x] Make versions of packages explicit


spack.yaml
```diff
@@ -0,0 +1,15 @@
+spack:
+  specs:
+    - acts@21.1.0 +json
+    - detray@0.22.0
+    - covfie@0.5.0
+    - algebra-plugins +vecmem +vc +eigen
+    - vecmem +cuda
```
krasznaa, line 7: Why no explicit version specification for vecmem? The very latest version may not always work. (In fact I plan to introduce a non-backwards compatible update soon.)

krasznaa (commented):
I'd much prefer to push these recipes into [spack](https://github.com/spack/spack) itself. And then just provide the `spack.yml` file in this repository...

repo/packages/algebra-plugins/fix_cmake_install.patch
```diff
@@ -0,0 +1,21 @@
...
+-
+ # Print a standard information message about the package being found.
+ include( FindPackageHandleStandardArgs )
+ find_package_handle_standard_args( algebra-plugins REQUIRED_VARS
+@@ -42,4 +36,4 @@ find_package_handle_standard_args( algebra-plugins REQUIRED_VARS
+    VERSION_VAR algebra_plugins_VERSION )
+ 
+ # Include the file listing all the imported targets and options.
+-include( "${algebra_plugins_CMAKE_DIR}/algebra-plugins-config-targets.cmake" )
++include( "${CMAKE_CURRENT_LIST_DIR}/algebra-plugins-config-targets.cmake" )
```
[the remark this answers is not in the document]
stephenswat: This is the issue that I have been having in acts-project/algebra-plugins#36 for the past two years, thankfully I figured out how to solve it this time. As to why it works in vecmem, no clue!

stephenswat:
Oh, agreed these should (eventually) go into the main Spack repository. But while the packages remain relatively volatile it might be a good idea to keep them in a place where we can rapidly update them.

repo/packages/algebra-plugins/fix_cmake_install.patch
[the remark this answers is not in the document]
krasznaa: Okay, will look at it tomorrow. That issue fell off of my radar...

repo/packages/algebra-plugins/package.py
```diff
@@ -0,0 +1,37 @@
...
+class AlgebraPlugins(CMakePackage):
+    homepage = "https://github.com/acts-project/algebra-plugins"
+    git = "https://github.com/acts-project/algebra-plugins.git"
+    list_url = "https://github.com/acts-project/algebra-plugins/tags"
+
+    version("0.17.0", commit="f4e9c1cc37ce6080234f265d121f4bc0a74bef87")
+
+    variant("vc", default=False, description="Enables the Vc plugin")
+    variant("eigen", default=False, description="Enables the Eigen plugin")
+    variant("vecmem", default=False, description="Enables the vecmem plugin")
```
HadrienG2, line 13: Since the CMakeLists for this project manually fiddles with the C++ standard, I would advise adding a `cxxstd` variant here, [in the spirit of what the root package does](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/root/package.py#L194). People around here (including you) are already experimenting with C++20, and since support for it in GCC and clang is currently quite brittle, sticking with C++17 is also a very valid option, so given that there are two valid choices for this build parameter, there should be a variant exposing the choice.

This also applies to the detray package, which similarly manually fiddles with CMAKE_CXX_STANDARD.

HadrienG2: Also, you will want to add support for the other algebra-plugins build options. And for readability/maintainability, alphabetical order in the lists of variants and dependencies would be nice. This is a remark that generally applies to your other packages.

repo/packages/covfie/package.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class Covfie(CMakePackage):
+    homepage = "https://github.com/acts-project/covfie"
+    git = "https://github.com/acts-project/covfie.git"
+    list_url = "https://github.com/acts-project/covfie/tags"
+
+    version("0.5.0", commit="59bd2e0a4861db6e5898606e0167da895505e46b")
+
+    variant("cuda", default=False, description="Enables the CUDA platform")
```
HadrienG2, line 11: If you want this to be usable for covfie development, consider adding tests, examples, benchmarks too.

repo/packages/covfie/package.py
```diff
@@ -0,0 +1,29 @@
...
+    def cmake_args(self):
+        def cmake_bool(p):
+            return "On" if p else "Off"
+
+        spec = self.spec
+
+        args = [
+            "-DCMAKE_BUILD_TYPE=Release",
+            "-DCOVFIE_PLATFORM_CPU=On",
+            "-DCOVFIE_PLATFORM_CPU={}".format(cmake_bool(spec.satisfies("+cuda"))),
```
HadrienG2, line 24: Probably this package is too small to justify a dedicated utility function, but consider using the upstream utility [self.define_from_variant()](https://spack.readthedocs.io/en/latest/spack.build_systems.html#spack.build_systems.cmake.CMakeBuilder.define_from_variant)

repo/packages/detray/package.py
```diff
@@ -0,0 +1,30 @@
...
+    depends_on("nlohmann-json")
+    depends_on("dfelibs")
+    depends_on("algebra-plugins +vecmem +vc +eigen")
+
+    patch("fix_cmake_install.patch", when="@0.22.0:", level=1)
+    patch("nlohmann_target.patch", when="@0.22.0:", level=1)
+
+    def cmake_args(self):
+        args = [
+            "-DCMAKE_BUILD_TYPE=Release",
```
HadrienG2, line 24: I would strongly advise against this.

1. Spack's CMake support already defines a `build_type` variant for you, you can use it to tweak this however you want without hardcoding it in your package.
2. By forcing this to `Release`, you break people's CPU profilers by default and force them into a full rebuild when they figure out they need to profile something, which is not nice. `RelWithDebInfo` is a better default because of this.

This applies to all of your other packages.

repo/packages/detray/package.py
```diff
@@ -0,0 +1,30 @@
...
+            "-DDETRAY_USE_SYSTEM_LIBS=On",
+            "-DDETRAY_SETUP_ACTSVG=Off",
```
HadrienG2, line 26: Eventually this should plug into spack's actsvg package, as you traccc package does.

repo/packages/detray/package.py
HadrienG2, line 1: This could use some variants mapping to build options, but I assume you already have this planned.

HadrienG2 (changes requested):
Not bad for a first try!

stephenswat:
I've incorporated the majority of the comments.

stephenswat:
Updated for acts-project/algebra-plugins#96 and acts-project/detray#419.

repo/packages/algebra-plugins/package.py
```diff
@@ -0,0 +1,43 @@
...
+    depends_on("vc@1.4.0:", when="+vc")
+    depends_on("root@6.18.0:", when="+smatrix")
+
+    patch("fix_cmake_install.patch", when="@:0.17.99", level=1)
+
+    def cmake_args(self):
+        spec = self.spec
+
+        args = [
+            self.define_from_variant("CMAKE_CXX_STANDARD", "cxxstd"),
```
krasznaa, line 35: Does the `CMakePackage` base not take care of this setting itself? :thinking: We use standard CMake variables for these purposes exactly so that package managers could work with minimum effort. (I don't know whether this setting missing from https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/vecmem/package.py is indicative of the setting not being needed, or @HadrienG2 having forgotten about it there. :stuck_out_tongue:)

repo/packages/algebra-plugins/package.py
```diff
@@ -0,0 +1,43 @@
+from spack.package import *
+
+
+class AlgebraPlugins(CMakePackage):
```
krasznaa, line 4: Similar to VecMem's package, I think this one should inherit from the `CudaPackage` base class as well. (https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/vecmem/package.py#L9) As I would hope that that would take care of setting things like `CMAKE_CUDA_STANDARD` automatically. :thinking:

`CMAKE_SYCL_STANDARD` is a custom variable that we'll need to set ourselves, no question about it. But I wonder if `CMAKE_HIP_STANDARD` will be automatically handled by Spack at one point...  Since standard CMake by now implemented pretty much the same thing that I did in VecMem at the beginning. (It's a long story why we don't yet use their code though.)

stephenswat: As I understand it `CudaPackage` just adds support for specifying the CC target as a variant, as well as adding an implicit `cuda` variant to enable or disable CUDA compilation.

repo/packages/algebra-plugins/package.py
stephenswat: It does not. This makes sense because:

1. CMake is not just a C++ build tool (sadly), and baking C++ specific features into the `CMakePackage` mixin would violate the separation of concerns; and
2. Exposing an ABI that is dependent on the C++ standard is a pretty significant anti-pattern that should be avoided, and ideally packages would not care about the C++ standards used to build their children.

stephenswat:
Updated to incorporate algebra-plugins 0.18.0.

beomki-yeo:
So after this, it seems we need to update the version of upstream libraries here and there - Can we specify their versions in a single file to make it easier?

stephenswat:
Ideally this would eventually replace the logic of pulling in packages in CMake itself, but until then I see no easy way of simplifying this. It would be possible to specify our version dependencies in a single YAML file (which could then be imported in `spack.yaml`), but I doubt that CMake has very good YAML parsing (although I could be wrong).

beomki-yeo:
> It would be possible to specify our version dependencies in a single YAML file (

That would be good enough for now if it makes sense to you

stephenswat:
Updated to make sure that detray, dfelibs, and algebra-plugins don't build their tests and examples by default.

Also happy to report that this speeds up configuration + build times by a factor three:

```
before:  353.15s user 31.86s system 768% cpu  50.096 total
after:  1269.32s user 95.01s system 780% cpu 2:54.88 total

stephenswat:
Update to use the new versions of `detray` and `algebra-plugins` without patches. :partying_face:

PR #365, opened 2023-03-24
Fix comment in device triplet

guilhermeAlmeida1:


stephenswat (approved):
Willing to bet this will pass the CI.

PR #366, opened 2023-03-24
Profiler

Xola101:
Sorry for the delay, there was a bug that I was busy trying to find for the longest time. So I added some functionality to get fine grained profiling that can hopefully be useful for finding bottle necks. The command I use to run the profiler is:
`nsys profile --trace="cuda,nvtx" --cudabacktrace=all --cuda-um-cpu-page-faults=true --cuda-um-gpu-page-faults=true --stats=true build/bin/traccc_throughput_st_cuda --detector_file=tml_detector/trackml-detector.csv --input_directory=tml_full/single_muon/ --digitization_config_file=tml_detector/detector-geometric-config-generic.json --processed_events=1`

I also attached the output of the command.
[profile_cmndline_output.txt](https://github.com/acts-project/traccc/files/11060670/profile_cmndline_output.txt)

### DIFF
--- a/device/cuda/CMakeLists.txt
+++ b/device/cuda/CMakeLists.txt
@@ -46,7 +46,7 @@ traccc_add_library( traccc_cuda cuda TYPE SHARED
   "src/fitting/fitting_algorithm.cu")
   
 target_compile_options( traccc_cuda
-  PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr -Xcompiler -finstrument-functions -fPIC -lnvToolsExt -lineinfo> )
+  PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr -Xcompiler -finstrument-functions -fPIC -lnvToolsExt> )
 target_link_libraries( traccc_cuda
   PUBLIC traccc::core detray::core detray::utils vecmem::core covfie::core
   PRIVATE CUDA::cudart traccc::device_common vecmem::cuda )
--- a/device/cuda/CMakeLists.txt
+++ b/device/cuda/CMakeLists.txt
@@ -46,7 +46,7 @@ traccc_add_library( traccc_cuda cuda TYPE SHARED
   "src/fitting/fitting_algorithm.cu")
   
 target_compile_options( traccc_cuda
-  PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr> )
+  PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr -Xcompiler -finstrument-functions -fPIC -lnvToolsExt -lineinfo> )
 target_link_libraries( traccc_cuda
   PUBLIC traccc::core detray::core detray::utils vecmem::core covfie::core
   PRIVATE CUDA::cudart traccc::device_common vecmem::cuda )
--- a/examples/run/cuda/CMakeLists.txt
+++ b/examples/run/cuda/CMakeLists.txt
@@ -29,7 +29,7 @@ target_link_libraries( traccc_examples_cuda
    PUBLIC CUDA::cudart vecmem::core vecmem::cuda traccc::core
           traccc::device_common traccc::cuda )
 
-traccc_add_executable( throughput_st_cuda "throughput_st.cpp"
+traccc_add_executable( throughput_st_cuda "throughput_st.cpp" "inst_nvtx.cpp"
    LINK_LIBRARIES vecmem::core vecmem::cuda traccc::io traccc::performance
                   traccc::core traccc::device_common traccc::cuda
                   traccc::options traccc_examples_cuda )
--- a/examples/run/cuda/inst_nvtx.cpp
+++ b/examples/run/cuda/inst_nvtx.cpp
@@ -1,0 +1,87 @@
+/* Copyright (c) 2014, NVIDIA CORPORATION. All rights reserved.
+ * 
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *  * Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  * Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *  * Neither the name of NVIDIA CORPORATION nor the names of its
+ *    contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <stdio.h>
+#include <dlfcn.h>
+#include <cxxabi.h>
+#include "nvtx3/nvToolsExt.h"
+
+const char* const default_name = "Unknown";
+
+const uint32_t colors[] = { 0xff00ff00, 0xff0000ff, 0xffffff00, 0xffff00ff, 0xff00ffff, 0xffff0000, 0xffffffff };
+const int num_colors = sizeof(colors)/sizeof(uint32_t);
+static int color_id = 0;
+
+extern "C" void __cyg_profile_func_enter(void *this_fn, void *call_site) __attribute__((no_instrument_function));
+extern "C" void __cyg_profile_func_exit(void *this_fn, void *call_site) __attribute__((no_instrument_function));
+
+void rangePush(const char* const name ) __attribute__((no_instrument_function));
+void rangePop() __attribute__((no_instrument_function));
+
+extern "C" void __cyg_profile_func_enter(void *this_fn, void *call_site)
+{
+	Dl_info this_fn_info;
+	if ( dladdr( this_fn, &this_fn_info ) )
+	{
+		int status = 0;
+		rangePush(abi::__cxa_demangle(this_fn_info.dli_sname,0, 0, &status));
+	}
+	else
+	{
+		rangePush(default_name);
+	}
+} /* __cyg_profile_func_enter */
+
+extern "C" void __cyg_profile_func_exit(void *this_fn, void *call_site)
+{
+	rangePop();
+} /* __cyg_profile_func_enter */
+
+void rangePush(const char* const name )
+{
+	nvtxEventAttributes_t eventAttrib = {0};
+	eventAttrib.version = NVTX_VERSION;
+	eventAttrib.size = NVTX_EVENT_ATTRIB_STRUCT_SIZE;
+	eventAttrib.colorType = NVTX_COLOR_ARGB;
+	eventAttrib.color = colors[color_id];
+	color_id = (color_id+1)%num_colors;
+	eventAttrib.messageType = NVTX_MESSAGE_TYPE_ASCII;
+	if ( name != 0 )
+	{
+		eventAttrib.message.ascii = name;
+	}
+	else
+	{
+		eventAttrib.message.ascii = default_name;
+	}
+	nvtxRangePushEx(&eventAttrib);
+}
+
+void rangePop()
+{
+	nvtxRangePop();
+}


device/cuda/CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ traccc_add_library( traccc_cuda cuda TYPE SHARED
   "src/fitting/fitting_algorithm.cu")
   
 target_compile_options( traccc_cuda
-  PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr> )
+  PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr -Xcompiler -finstrument-functions -fPIC -lnvToolsExt> )
```
stephenswat, line 49: I would rather not inject this directly into every build. Perhaps we can inject this into `RelWithDebInfo` bugs, although I am also not very fond of introducing performance effects there that do not exist in `Release` builds. Let's add a `Profile` build type or something similar.

stephenswat, line 49: If possible, we should not add the `-lnvToolsExt` flag directly. Instead, it should be added using `target_link_libraries`.

stephenswat (changes requested):
The code looks fairly solid, but we should be careful not to conditionally enable it.

device/cuda/CMakeLists.txt
```diff
@@ -44,10 +44,18 @@ traccc_add_library( traccc_cuda cuda TYPE SHARED
...
-  PRIVATE CUDA::cudart traccc::device_common vecmem::cuda )
+        PUBLIC traccc::core detray::core detray::utils vecmem::core covfie::core
+        PRIVATE CUDA::cudart traccc::device_common vecmem::cuda -lnvToolsExt)
+else()
+    target_compile_options( traccc_cuda
+            PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr>)
+    target_link_libraries( traccc_cuda
+            PUBLIC traccc::core detray::core detray::utils vecmem::core covfie::core
+            PRIVATE CUDA::cudart traccc::device_common vecmem::cuda)
+endif(TRACCC_SETUP_PROFILER)
```
stephenswat, line 59: This is not the best way to solve this. In an unusual turn of events CMake actually has halfway decent syntax for this, which they refer to as "generator expressions". You should write something like `$<$<TRACCC_SETUP_PROFILER>:-Xcompiler -finstrument-functions -fPIC>`. Or, as we might prefer to generate position-independent code in any case: `-Xcompiler -fPIC  $<$<TRACCC_SETUP_PROFILER>:-finstrument-functions>`. To be clear, the use of this is that we would not need to duplicate this much code between the true and false branches of the if-statement, as we could do without the if entirely.

More documentation on generator expressions is available at https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html.

device/cuda/CMakeLists.txt
```diff
@@ -44,10 +44,18 @@ traccc_add_library( traccc_cuda cuda TYPE SHARED
...
-  
+if(TRACCC_SETUP_PROFILER)
 target_compile_options( traccc_cuda
-  PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr> )
+  PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr -Xcompiler -finstrument-functions -fPIC > )
 target_link_libraries( traccc_cuda
-  PUBLIC traccc::core detray::core detray::utils vecmem::core covfie::core
-  PRIVATE CUDA::cudart traccc::device_common vecmem::cuda )
+        PUBLIC traccc::core detray::core detray::utils vecmem::core covfie::core
+        PRIVATE CUDA::cudart traccc::device_common vecmem::cuda -lnvToolsExt)
```
stephenswat, line 52: The CMake documentation is unclear about whether your current implementation would work but it would be more conformant to the specification (although syntactically less clear) to write ` nvToolsExt` instead of `-lnvToolsExt`.

stephenswat (changes requested):

device/cuda/CMakeLists.txt
Xola101: I have implemented your request

device/cuda/CMakeLists.txt
Xola101: Implemented the generator expression using the following line:

`$<$<BOOL:${TRACCC_SETUP_PROFILER}>:-Xcompiler -finstrument-functions -fPIC>>`

device/cuda/CMakeLists.txt
Xola101: Done

device/cuda/CMakeLists.txt
```diff
@@ -1,53 +1,53 @@
...
+        "src/seeding/seed_finding.cu"
+        "src/seeding/spacepoint_binning.cu"
+        "src/seeding/seeding_algorithm.cpp"
+        "src/cca/component_connection.cu"
+#Clusterization
+        "include/traccc/cuda/clusterization/clusterization_algorithm.hpp"
+        "src/clusterization/clusterization_algorithm.cu"
+#Fitting
+        "include/traccc/cuda/fitting/fitting_algorithm.hpp"
+        "src/fitting/fitting_algorithm.cu")
```
stephenswat, line 46: There are a lot of unrelated formatting changes here... Did you accidentally put this code through clang-format? Please undo these changes.

device/cuda/CMakeLists.txt
```diff
@@ -1,53 +1,53 @@
...
 
+target_compile_options(traccc_cuda PRIVATE
+        $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr
+        $<$<BOOL:${TRACCC_SETUP_PROFILER}>:-Xcompiler -finstrument-functions -fPIC>>)
+target_link_libraries(traccc_cuda PUBLIC
+        traccc::core detray::core detray::utils vecmem::core covfie::core
+        PRIVATE CUDA::cudart traccc::device_common vecmem::cuda nvToolsExt)
```
stephenswat, line 53: Please ensure that `nvToolsExt` is only linked against if `TRACCC_SETUP_PROFILER` is enabled, using the same logic as you did above (using the generator expression).

examples/run/cuda/CMakeLists.txt
```diff
@@ -29,10 +29,10 @@ target_link_libraries( traccc_examples_cuda
    PUBLIC CUDA::cudart vecmem::core vecmem::cuda traccc::core
           traccc::device_common traccc::cuda )
 
-traccc_add_executable( throughput_st_cuda "throughput_st.cpp"
+traccc_add_executable( throughput_st_cuda "throughput_st.cpp" "inst_nvtx.cpp"
```
stephenswat, line 32: We should aim to put this `inst_nvtx.cpp` file into a dedicated target which we can then link against using `target_link_libraries`. I'm not 100% sure how linkage works with `-finstrument-functions` but we should be able to compile this into a new library and then link. E.g.:

```cmake
traccc_add_library(nvtx_instrumentation "inst_nvtx.cpp")
```

And then...

```
traccc_add_executable( throughput_st_cuda
   LINK_LIBRARIES ... nvtx_instrumentation
)
```

examples/run/cuda/CMakeLists.txt
```diff
@@ -29,10 +29,10 @@ target_link_libraries( traccc_examples_cuda
...
    LINK_LIBRARIES vecmem::core vecmem::cuda traccc::io traccc::performance
                   traccc::core traccc::device_common traccc::cuda
-                  traccc::options traccc_examples_cuda )
+                  traccc::options traccc_examples_cuda ${CMAKE_DL_LIBS})
```
stephenswat, line 35: This worries me a little (a lot). Why is this necessary? :worried:

stephenswat (changes requested):
Sorry to be pedantic, but while this is already a lot better it's still not quite what I'm looking for.

examples/run/cuda/CMakeLists.txt
Xola101: The problem here is that the code compiles with no problem on my local machine. However, I get an error saying:
`undefined reference to symbol 'dlclose@@GLIBC_2.2.5`
when compiling on github. I found the fix for this on this link: https://stackoverflow.com/questions/56841870/how-to-fix-undefined-reference-to-symbol-dlcloseglibc-2-2-5-from-glad-c

device/cuda/CMakeLists.txt
Xola101: Just tried running the code without `nvToolsExt` . Looks like it was not necessary

examples/run/cuda/CMakeLists.txt
stephenswat: Ah okay yeah that's fine then.

device/cuda/CMakeLists.txt
Xola101: ok I added the line:
`INTERFACE $<$<BOOL:${TRACCC_SETUP_PROFILER}>: CUDA::nvToolsExt`

seems to be working

stephenswat: Great, thanks!

examples/run/cuda/CMakeLists.txt
Xola101: Should we put it in examples/options instead?

stephenswat: No, just put it in this file. :)

Xola101: Not sure why but when i do:
`add_library(nvtx_instrumentation SHARED
"inst_nvtx.cpp")`

It is not able to find the nvtx.h file.

However, when I put it inside:
`add_library(traccc_examples_cuda)`

then it is able to find the header files
Do you know why this might be the case?

device/cuda/CMakeLists.txt
krasznaa, line 53: Something is just not right here...

First off, `nvToolsExt` will simply add a `-lnvToolsExt` argument to the linking command. Which, if that library is not in your runtime environment will not work. You would need to use `CUDA::nvToolsExt` instead.

But according to https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#nvtoolsext that shared library is not recommended for usage anymore. Apparently one is supposed to use `CUDA::nvtx3`, and the corresponding headers instead. Which we already do. :confused: (The NVTX3 headers are currently picked up implicitly from the same place as the rest of the CUDA headers. Since `CUDA::nvtx3` is only introduced with the latest CMake versions.)

examples/run/cuda/CMakeLists.txt
Xola101: So this is what I did:

`add_library(nvtx_instrumentation SHARED
        "inst_nvtx.cpp" )

target_link_libraries( nvtx_instrumentation
        PUBLIC CUDA::cudart traccc::core
        traccc::device_common traccc::cuda 


traccc_add_executable( throughput_st_cuda "throughput_st.cpp"
        LINK_LIBRARIES vecmem::core vecmem::cuda traccc::io traccc::performance
                  traccc::core traccc::device_common traccc::cuda
                  traccc::options traccc_examples_cuda
                  nvtx_instrumentation ${CMAKE_DL_LIBS})`

But is seems as though the instrumentation feature is not working if `inst_nvtx.cpp` is not added as an executable. Do you know why this might be?

device/cuda/CMakeLists.txt
```diff
@@ -1,53 +1,53 @@
...
+#Clusterization
+        "include/traccc/cuda/clusterization/clusterization_algorithm.hpp"
+        "src/clusterization/clusterization_algorithm.cu"
+#Fitting
+        "include/traccc/cuda/fitting/fitting_algorithm.hpp"
+        "src/fitting/fitting_algorithm.cu")
 
+target_compile_options(traccc_cuda PRIVATE
+        $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr
+        $<$<BOOL:${TRACCC_SETUP_PROFILER}>:-Xcompiler -finstrument-functions -fPIC>>)
```
krasznaa, line 50: This is quite fancy, but not quite right.
  - You don't need to use a generator expression for evaluating `${TRACCC_SETUP_PROFILER}`. A simple

```cmake
if( TRACCC_SETUP_PROFILER )
   target_compile_options(...)
endif()
```

setup works just fine in this case.
  - The `-Xcompiler` flag forwards `-finstument-functions` to the host compiler, no? Then we should do something like:

```cmake
if( TRACCC_SETUP_PROFILER )
   target_compile_options( traccc_cuda PRIVATE
      $<COMPILE_LANGUAGE:CPP>-finstrument-functions>
      $<COMPILE_LANGUAGE:CUDA>:-Xcompiler -finstrument-functions>)
endif()
```
  - The `-fPIC` flag should not be specified like this. Please tell us why you needed to specify that explicitly. Since you're building a shared library here, position independent code generation should be the default from CMake...

examples/run/cuda/inst_nvtx.cpp
```diff
@@ -0,0 +1,88 @@
+/* Copyright (c) 2014, NVIDIA CORPORATION. All rights reserved.
```
krasznaa, line 1: Please explain this file. I can't imagine that any recommendation from NVIDIA would include user code having to carry a copy of such a source file with itself.

I get the sense that you "just" wanted to provide these `rangePush(...)` and `rangePop()` functions to the user code. But you didn't add calls to them anywhere. So I can't imagine that this source file, in its current form, would be doing anything. :confused:

krasznaa (changes requested):
I have some big question marks above my head. :frowning:

examples/run/cuda/inst_nvtx.cpp
stephenswat: The `__cyg_profile_func_enter` and `__cyg_profile_func_exit` functions are called automatically when function instrumentation is enabled. :wink:

device/cuda/CMakeLists.txt
Xola101: Hi Attilla, this issue has already been resolved using what you have recommended. Just need to push it.

device/cuda/CMakeLists.txt
Xola101: On the issue of my use of generator expressions, Stephen had expressed concern over huge chunks of code being repeated in the if statement. Thats why we resolved to use the generator expression instead.

Xola101: We use the -fPIC flag so that we can get the name of the called function in the profiler. But i can try compile the code without -fPIC and see if i can get the same output

examples/run/cuda/CMakeLists.txt
krasznaa: :thinking: I believe compilation-wise you'll need a setup like:

https://github.com/acts-project/vecmem/blob/main/sycl/CMakeLists.txt#L48-L77

Note that that code demonstrates multiple things:
  - There we also had to make sure that a piece of code would get linked **into** the client code, not just "against" it. A static library is a good way to go with that.
  - Static libraries by default are **not** built with position independent code. But since we needed PIC there as well, CMake was told to do so using [POSITION_INDEPENDENT_CODE](https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html). (This formalism makes sure that the correct flag would be used, even for compilers that don't use `-fPIC` for this.)

device/cuda/CMakeLists.txt
krasznaa: Huge chunks of code being repeated? You mean `$<$<COMPILE_LANGUAGE:CUDA>:...>` appearing twice? :confused: Multiple calls to `target_compile_options(...)` is definitely not a bad thing. Not in this context at least. This generator expression here confused me as well at first. Let's not try to be this clever. Let's just write simple, easy to understand code for this.

examples/run/cuda/inst_nvtx.cpp
krasznaa: Okay, code-wise it makes some sense at least then. But there are just multiple things in it that I don't like. The non-static global variable declarations being the very first...

examples/run/cuda/inst_nvtx.cpp
```diff
@@ -0,0 +1,88 @@
...
+extern "C" void __cyg_profile_func_enter(void *this_fn, void *call_site) {
+    (void)this_fn;
+    (void)call_site;
+    Dl_info this_fn_info;
+    if (dladdr(this_fn, &this_fn_info)) {
+        int status = 0;
+        rangePush(abi::__cxa_demangle(this_fn_info.dli_sname, 0, 0, &status));
+    } else {
+        rangePush(default_name);
+    }
```
krasznaa, line 58: This is some prime platform specific crap right here. :frowning: If we do decide to add something like this to the project (which I'm still not convinced of), we'll need to protect it against being run on a platform where these things are not available.

Like:

```cmake
include( CheckIncludeFileCXX )
check_include_file_cxx( "cxxabi.h" TRACCC_HAVE_CXXABI_H )
if( HAVE_CXXABI_H )
   target_compile_definitions( traccc::cuda PRIVATE TRACCC_HAVE_CXXABI_H )
endif()
```

```c++
#ifdef TRACCC_HAVE_CXXABI_H
extern "C" {
#   include <cxxabi.h>
}
#endif // TRACCC_HAVE_CXX_ABI_H
...
#ifdef TRACCC_HAVE_CXX_ABI_H
int status = 0;
char* fname = abi::__cxa_demangle(this_fn_info.dli_sname, 0, 0, &status);
if (status != 0) {
   // Handle error somehow.
}
#else
const char* fname = this_fn_info.dli_sname;
#endif
...
#ifdef TRACCC_HAVE_CXX_ABI_H
free(fname);
#endif
```

I.e. yes, this expression also leaks some memory over here... :frowning:

krasznaa (changes requested):
Let's possibly discuss a bit in the meeting right now...

examples/run/cuda/CMakeLists.txt
Xola101: Hi @krasznaa, so what I understand from the vecmem code you posted is that a check was firstly done on the source code to see if CUDA polyfill can make the test work, if it does then you add a library this file: `"src/utils/sycl/cuda_assert_polyfill.sycl"`. You then set the target properties so that the code uses position-independent code. I implemented all of this using `SHARED` libraries first, but this did not work. I then went to have a look at different types of libraries one can create and saw the option of generating `OBJECT` libraries. This seems to work. 

Here is the code I have so far:

`traccc_add_library(nvtx_instrumentation TYPE OBJECT
        "inst_nvtx.cpp" )`

`target_link_libraries( nvtx_instrumentation
        PUBLIC CUDA::cudart traccc::core
        traccc::device_common traccc::cuda )`

`traccc_add_executable( throughput_st_cuda "throughput_st.cpp"
        LINK_LIBRARIES vecmem::core vecmem::cuda traccc::io traccc::performance
                  traccc::core traccc::device_common traccc::cuda
                  traccc::options traccc_examples_cuda
                  nvtx_instrumentation ${CMAKE_DL_LIBS})`

Xola101: I hope this is fine..?

Xola101: Here is the link to the nvidia blog post you asked for: https://developer.nvidia.com/blog/cuda-pro-tip-generate-custom-application-profile-timelines-nvtx/

device/cuda/CMakeLists.txt
Xola101: Hi @krasznaa. Here is my implementation when using `if()`:

`if(TRACCC_PROFILER)
target_compile_options(traccc_cuda
        PRIVATE $<$<COMPILE_LANGUAGE:CPP>:--expt-relaxed-constexpr
        PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler -finstrument-functions>>)
elseif()
    target_compile_options(traccc_cuda
            PRIVATE $<$<COMPILE_LANGUAGE:CPP>:--expt-relaxed-constexpr>>)
    endif()`

What I meant by repetition of big chunks of code is that I had to use and if-else statement for the case where we need to have `target_compile_options()` for the case where we do not use the profiling code.

stephenswat:
Sorry to say but this PR has been going downhill a bit... The recent changes are really over-complicating things. :disappointed:

Still, it's a hard thing to get right. Would you mind if I make a commit directly to your branch?

stephenswat:
Hi @Xola101, I've pushed a commit to your branch fixing many of the problems with your PR. Please take a look and let me know if you approve. If not, go ahead and discard the changes if you wish.

stephenswat:
Pretty cool, although we may want to think of a way to limit which functions are profiled; currently, the size of the profile files is quite... significant.

image

Xola101:
Yes I thought this might be a problem. I did read somewhere that it can be done. Just don't know which functions to filter at the moment

PR #368, opened 2023-03-24
Change alt_cell to not depend on cell. Remove deprecated cell. 

guilhermeAlmeida1:
Depends on #361.
This PR makes `alt_cell` not depend on `cell`. It also entirely removes `cell.hpp` from our repository.

PS: This will break the Futhark code which still uses `cell`, but since this is not being checked in the CI nor do I think is currently being used by anyone, I decided to let that happen. @stephenswat


stephenswat:
So uh doesn't this mean we can rename `alt_cell` to `cell`? The circle of life?

guilhermeAlmeida1:
> So uh doesn't this mean we can rename `alt_cell` to `cell`? The circle of life?

Yes, my plan was to have a PR which does all the renaming separately, to make it clear that this is the one which actually makes changes to the code, even though they're minor ones, the other would literally be just renaming variables/types

stephenswat:
Works for me, I support whatever you think is more practical. :+1:

krasznaa (approved):
Pretty straight forward stuff in the end.

And yes, things will be a whole lot better when we finally rename `traccc::alt_cell` to `traccc::cell`. :wink:

PR #369, opened 2023-03-26
Add quality reporting for seeds of arbitrary length

stephenswat:
As far as I understand, the seed efficiency tools we have now don't really work with seeds that are not of size 3, and it also appears that they are not currently working for the CUDA code. This PR introduces code that functions similarly, but allows matching and filtering to be configured. It also outputs its data to files that can be analyzed further.

Example output:

```
$ ./build/bin/traccc_seeding_example_cuda --detector_file tml_detector/trackml-detector.csv --events 10 --input_directory tml_full/ttbar_mu300/ --check_performance=1
Running ./build/bin/traccc_seeding_example_cuda tml_detector/trackml-detector.csv tml_full/ttbar_mu300/ 10
==> Seed finding efficiency ...
- Particle filter      : Charged with |η| ≤ 2.70 and pT ≥ 1.000 GeV
- Particle matcher     : Stepped with ≥ 60.0% similarity
- Total seeds          :  294740
- True seeds           :  208530
- False seeds          :   86210
- Total tracks         :   13076
- Matched tracks       :   11632
- Unmatched tracks     :    1444
- Precision            :  70.75%
- Fake rate            :  29.25%
- Recall/Efficiency    :  88.96%
```

### DIFF
--- a/performance/src/efficiency/nseed_performance_writer.cpp
+++ b/performance/src/efficiency/nseed_performance_writer.cpp
@@ -1,0 +1,147 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022-2023 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include <memory>
+#include <optional>
+#include <traccc/efficiency/nseed_performance_writer.hpp>
+#include <traccc/efficiency/track_filter.hpp>
+#include <traccc/efficiency/track_matcher.hpp>
+
+namespace traccc {
+nseed_performance_writer::nseed_performance_writer(
+    const std::string& prefix, std::unique_ptr<track_filter>&& filter,
+    std::unique_ptr<track_matcher>&& matcher)
+    : _prefix(prefix),
+      _filter(std::forward<std::unique_ptr<track_filter>>(filter)),
+      _matcher(std::forward<std::unique_ptr<track_matcher>>(matcher)) {}
+
+void nseed_performance_writer::initialize() {
+    output_seed_file.open(_prefix + "seeds.csv");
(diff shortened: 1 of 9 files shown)


guilhermeAlmeida1:
> it also appears that they are not currently working for the CUDA code. 

This was a result of the difference between CPU and GPU EDM. It is fixed in #361 

Otherwise, I'm on board with trying to simplify our physics performance code.

guilhermeAlmeida1 (approved):
I finally got down to reviewing this. Looks great to me, just needs a rebase (which I don't have the perms to do)

stephenswat:
Thanks @guilhermeAlmeida1, I have rebased the PR. Could you re-approve?

guilhermeAlmeida1 (approved):

performance/src/efficiency/nseed_performance_writer.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+namespace traccc {
+nseed_performance_writer::nseed_performance_writer(
+    const std::string& prefix, std::unique_ptr<track_filter>&& filter,
+    std::unique_ptr<track_matcher>&& matcher)
+    : _prefix(prefix),
+      _filter(std::forward<std::unique_ptr<track_filter>>(filter)),
+      _matcher(std::forward<std::unique_ptr<track_matcher>>(matcher)) {}
+
+void nseed_performance_writer::initialize() {
```
guilhermeAlmeida1, line 22: Actually I entirely forgot about this comment I had. Would it not be safer to call these in the objects constructor and destructor?

guilhermeAlmeida1: meaning initialize() and finalize()

PR #370, opened 2023-03-26
Add a very simple seedin efficiency plotter

stephenswat:
This PR adds a Python script to plot seeding efficiencies as produced by PR #369. Features configurable ranges and outputs to various data formats.

Example output:

eff


stephenswat:
Thinking of it this would benefit from having fake rates and duplication rates included, but that should be easy.

stephenswat:
Updated to fix an important data input error, and to produce additional plots about duplication rates and track counts:

image

beomki-yeo:
I will try to review the PR as soon as I can - btw you have two different efficiency plots in the PR description and comment. I guess one is from the strict matching and the other from 50% matching rate?

stephenswat:
Thanks for taking a look! The second plot is a little different because an earlier version of the script had a bug in it.

extras/plot_seeding_performance/README.md
```diff
@@ -0,0 +1,30 @@
+# Seeding performance plotter
+
+This Python script plots the performance of seeding code if the
+`nseed_performance_writer` class is used to analyse it. The script takes a file
+of seeds and a file of tracks (with default arguments) and optional filter and
+output flags. For example, simply calling the script should work fine in most
+situations (if the CSV files are in the current directory):
+
+```sh
+$ python extras/plot_seeding_performance/plot.py
```
krasznaa, line 10: I like that you use `python3` explicitly in the shebang in the script. But then let's be consistent, and use `python3` in the README as well. :wink:

krasznaa (changes requested):
I know this is a very subjective thing, but could you rather move the script to something like `extras/traccc_plot_seeding_performance.py`? Just adding your description to the existing `extras/README.md` file. Which may force me into writing a few lines about the other two scripts as well... :stuck_out_tongue:

PR #371, opened 2023-03-27
Change SYCL chain to use async memory copies

guilhermeAlmeida1:
Change `sycl::FULL_CHAIN_ALGORITHM` to use a `vecmem::sycl::async_copy` object, similarly to what happens in the `CUDA` chain.
This doesn't seem to affect performance.


stephenswat (approved):
This will be beneficial to performance once we remove the unnecessary waits. :+1:

guilhermeAlmeida1:
And the same could be said about the whole CUDA chain, where we've been using `async_copy` for a while now, apparently without running into issues despite lacking any `vecmem::copy::wait()`.

PR #372, opened 2023-03-28
Remove deprecated seed type. Rename alt_seed to seed.

guilhermeAlmeida1:
This doesn't actually make any algorithmic changes to the code. Only deleting and renaming stuff.


stephenswat (approved):
Happy to merge if the CI passes.

guilhermeAlmeida1:
I decided to also add to this PR the renaming of `alt_cell`. Each of the commits was the result of a simple script which ran through the directory renaming every instance of "alt...".

stephenswat:
Very well! Just be aware that it is usually considered good practice to keep commit and PR titles to 50 characters or less, as this improves compatibility with many git tools.

Not super important, of course, but something to keep in the back of your mind and a good exercise in brevity.

guilhermeAlmeida1:
I screwed up my git workflow a bit in this PR, it's probably best to squash commit when this goes in.

PR #373, opened 2023-03-28
Fix: Sycl algorithms wait for assynchronous operations.

guilhermeAlmeida1:
As @krasznaa pointed out in #371 , my change to an assynchronous copy object was susceptible to error by missing calls to `wait()` on the various operations it does.


krasznaa (approved):
Looks good. Too bad the CI is not collaborating. :frowning:

PR #377, opened 2023-04-12
Change CCL kernel to common code

guilhermeAlmeida1:
Depends on https://github.com/acts-project/vecmem/pull/228

When I first implemented the "new" clustering code I made into fully separate CUDA/SYCL implementations due to some difficulties with harmonising them. This PR brings them back to sharing most of the code.

It also introduces 2 new types, `traccc::sycl::barrier` & `traccc::cuda::barrier`. These allow for easily calling block synchronisation functions within common code using templating.

### DIFF
--- a/device/cuda/include/traccc/cuda/utils/barrier.hpp
+++ b/device/cuda/include/traccc/cuda/utils/barrier.hpp
@@ -1,0 +1,23 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2023 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/definitions/qualifiers.hpp"
+
+namespace traccc::cuda {
+
+struct barrier {
+    TRACCC_DEVICE
+    void blockBarrier() { __syncthreads(); }
+
+    TRACCC_DEVICE
+    bool blockOr(bool predicate) { return __syncthreads_or(predicate); }
+};
+
+}  // namespace traccc::cuda
--- a/device/sycl/include/traccc/sycl/utils/barrier.hpp
+++ b/device/sycl/include/traccc/sycl/utils/barrier.hpp
@@ -1,0 +1,31 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2023 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/definitions/qualifiers.hpp"
+
+namespace traccc::sycl {
+
+struct barrier {
+    barrier(::sycl::nd_item<1> item) : m_item(item){};
+
+    TRACCC_DEVICE
+    void blockBarrier() { m_item.barrier(); }
+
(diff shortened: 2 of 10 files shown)


device/cuda/include/traccc/cuda/utils/barrier.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include "traccc/definitions/qualifiers.hpp"
+
+namespace traccc::cuda {
+
+struct barrier {
+    TRACCC_DEVICE
+    void blockBarrier() { __syncthreads(); }
+
+    TRACCC_DEVICE
+    bool blockOr(bool predicate) { return __syncthreads_or(predicate); }
```
stephenswat, line 20: Can we get these functions to support thread masks?

stephenswat (commented):
Putting aside my objections to this "common code" stuff, this looks pretty good overall. My biggest concern is that this takes the barrier operations from simple, static operations to stateful, object-oriented code. I am curious whether or not it is possible to unify both the CUDA and SYCL barrier semantics under a single unified set of static function calls? All this would really require is getting the group identifier at the call-site, which I assume should be fast (it is in CUDA).

device/sycl/include/traccc/sycl/utils/barrier.hpp
```diff
@@ -0,0 +1,31 @@
...
+// Project include(s).
+#include "traccc/definitions/qualifiers.hpp"
+
+namespace traccc::sycl {
+
+struct barrier {
+    barrier(::sycl::nd_item<1> item) : m_item(item){};
+
+    TRACCC_DEVICE
+    void blockBarrier() { m_item.barrier(); }
```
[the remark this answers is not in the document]
guilhermeAlmeida1: I am aware of the ambiguity here. I'm not particularly attached to either of the terminologies, and happy to stick with whichever most people find clearer.

device/cuda/include/traccc/cuda/utils/barrier.hpp
guilhermeAlmeida1: I'm not sure that would really be the way to go here. I understand it's usefulness, but the whole point of this class was to have the same interface as it's SYCL twin, and SYCL does not provide such a functionality AFAIK

stephenswat: Never mind I need another coffee, block synchronization does not take masks, but this problem will come back later when we add workgroup/warp synchronization.

stephenswat:
I notice you've changed the SYCL barrier to accept a group instead of an item now, and while I don't think that is a bad idea, I think it will make it impossible to add workgroup-synchronization primitives to this class, right? Because you can go from item to group and workgroup, but you cannot go from group to workgroup.

krasznaa:
[vecmem-0.25.0](https://github.com/acts-project/vecmem/releases/tag/v0.25.0) is now available. :wink: You could just update to it in this PR, there's no need to put it into a separate one.

guilhermeAlmeida1:
As far as I can tell from a simple test, this does not seem to impact performance negatively

stephenswat (approved):
Let's put this in then.

guilhermeAlmeida1:
@krasznaa if there is nothing else you'd like to get in for whatever measurements we show at CHEP, can we make a new tag of traccc then?

PR #385, opened 2023-04-19
Add default destructors to virtual classes

guilhermeAlmeida1:
#369 introduced a compiler warning for newer versions of gcc our CI didn't pickup which this PR fixes (sample warning follows).

```
/opt/rh/gcc-toolset-12/root/usr/lib/gcc/x86_64-redhat-linux/12/../../../../include/c++/12/bits/unique_ptr.h:95:2: warning: delete called on 'traccc::track_filter' that is abstract but has non-virtual destructor [-Wdelete-abstract-non-virtual-dtor]
        delete __ptr;
```


stephenswat (approved):
Very good, thanks.

PR #386, opened 2023-04-19
Check results in throughput measurements

guilhermeAlmeida1:
This is very much a draft as I'd like to hear if others' opinions on this.
Doubts on the veracity of our throughput measurements can (have) come up from the fact that we're only printing out the total number of seeds getting produced, but not looking at their actual content.
This initial approach consists of storing some ammount of results into memory and, at the end of the throughput measurement, checking them against some "truth value", which here are getting produced by our usual CPU reconstruction chain (which we know works as intended from our CI test of comparing against ACTS).

The number of stored events was setup to be proportional to the number of processed events used in the measurement, so that comparisons between runs with different ammount of processed events would stay fair.

### DIFF
--- a/examples/run/common/throughput_mt_alt.ipp
+++ b/examples/run/common/throughput_mt_alt.ipp
@@ -21,6 +21,13 @@
 #include "traccc/performance/timer.hpp"
 #include "traccc/performance/timing_info.hpp"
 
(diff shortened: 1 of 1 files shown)


guilhermeAlmeida1:
This will be replaced by a different approach

PR #388, opened 2023-04-21
feat: Test cuda full chain against cpu

guilhermeAlmeida1:
Now that we finally have GPU CI, I thought we could be more thorough in actually testing that the chain is working as intended by comparing the output of the CPU vs GPU algorithms

Note: I setup the tests to only run a single event for each input data type because these already take about 1min30s as they are. These can all be set to 10 events instead if we want to be even more thorough, but that would mean keeping the CI Bridge machine busy for 15min.

### DIFF
--- a/performance/include/traccc/performance/collection_comparator.hpp
+++ b/performance/include/traccc/performance/collection_comparator.hpp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2022 CERN for the benefit of the ACTS project
+ * (c) 2022-2023 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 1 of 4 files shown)


guilhermeAlmeida1:
pinging @paulgessinger

paulgessinger:
Awesome! How about we keep it at this scope and see how busy the CI becomes.

paulgessinger (approved):

stephenswat (approved):
Re-approving.

krasznaa (approved):
Some of the comparator code could be done a little differently as well, but I think this is perfectly fine for this. As long as the CI is still happy, let's go forward with this.

guilhermeAlmeida1:
> Would it be possible to generate the CPU outputs in a separate somehow, to try to free up the GPU machine?

So what are we doing about reducing the ammount of time required on the CI machine introduced in this PR. I don't really like the idea of having the CPU results just available _a priori_ so maybe we should reduce the number of test cases?

guilhermeAlmeida1:
This needs a bit of fixing after the rebase as well, I'll take care of that

krasznaa:
I'm fine with removing a number of test cases. But I completely agree that I wouldn't want to complicate things with pre-calculated CPU results.

PR #389, opened 2023-04-21
Remove git-lfs from actions. Update to checkout@v3

guilhermeAlmeida1:
Minor update.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -55,12 +55,7 @@ jobs:
       run:
         shell: bash
     steps:
-      - name: Dependencies
-        run: apt-get install -y git-lfs wget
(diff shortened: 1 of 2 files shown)


stephenswat:
Well spotted, the only thing I am not sure about is whether CMake will still work (with the wget / curl wrapper) if wget is not installed. But we will see if the CI passes.

guilhermeAlmeida1:
> Well spotted, the only thing I am not sure about is whether CMake will still work (with the wget / curl wrapper) if wget is not installed. But we will see if the CI passes.

It actually slipped me that I was removing that as well, but it seems like it's not necessary

stephenswat (approved):
Seems like it works, so let's go

PR #391, opened 2023-04-28
Cleanup: Rename alt types.

guilhermeAlmeida1:
Similar to https://github.com/acts-project/traccc/pull/372

This doesn't change any algorithmic code, only renaming / deleting stuff.
The only "alt" that remains anywhere in our code after this should be the `alt_measurement` type which is on hold until we get CKF in.


stephenswat (approved):
image

PR #392, opened 2023-05-03
Remove synchronization from cuda chain

guilhermeAlmeida1:
Removes a bunch of synchronization points in the CUDA code which are not necessary.

Note: In SYCL all of these waits are still there, given that the queue object is out-of-order.

### DIFF
--- a/device/cuda/src/clusterization/clusterization_algorithm.cu
+++ b/device/cuda/src/clusterization/clusterization_algorithm.cu
@@ -152,7 +152,6 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
         spacepoints_buffer);
 
     CUDA_ERROR_CHECK(cudaGetLastError());
-    m_stream.synchronize();
 
     return {std::move(spacepoints_buffer), std::move(cell_links)};
 }
--- a/device/cuda/src/seeding/seed_finding.cu
+++ b/device/cuda/src/seeding/seed_finding.cu
@@ -193,7 +193,6 @@ seed_finding::output_type seed_finding::operator()(
         doublet_counter_buffer, (*globalCounter_device).m_nMidBot,
         (*globalCounter_device).m_nMidTop);
     CUDA_ERROR_CHECK(cudaGetLastError());
-    m_stream.synchronize();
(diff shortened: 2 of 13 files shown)


device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -152,7 +152,6 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
         spacepoints_buffer);
 
     CUDA_ERROR_CHECK(cudaGetLastError());
-    m_stream.synchronize();
```
krasznaa, line 155: I think this is a good change. But you should add some explicit mention about this in the documentation of this algorithm.

Namely that the users of the algorithm **must not** dismiss the return value of the algorithm, or delete it before the `form_spacepoints` kernel would've finished running.

The way that vecmem buffers work, it **is** possible to pass around objects between components that are in the process of being filled. We'll just need to do this carefully.

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -347,7 +333,6 @@ seed_finding::output_type seed_finding::operator()(
                                       triplet_counter_midBot_buffer,
                                       triplet_buffer, seed_buffer);
     CUDA_ERROR_CHECK(cudaGetLastError());
-    m_stream.synchronize();
```
krasznaa, line 350: The same here. The user of the algorithm will need to take care of not deleting the received buffer until the kernel filling it would've finished.

krasznaa (commented):
I'm a bit torn about whether to add synchronization points at the end of algorithms or not. :thinking: I'm curious about the opinions of others as well...

guilhermeAlmeida1:
> I'm a bit torn about whether to add synchronization points at the end of algorithms or not. 🤔 I'm curious about the opinions of others as well...

You raise a good point, it's definately a small concern but if we want to maximize the performance of our full chain we'll probably benefit from not having that extra synchronisation. And you actually reminded me that in the seq_example I'll need to add explicit wait statements outside the algorithhms which are missing.

PR #393, opened 2023-05-04
Fix: use correct memory resources for host vectors

guilhermeAlmeida1:
Throughout our code frequently use assignment operations to define host vectors, which easily leads to creating these objects on default memory.
This PR addresses this issue. The largest of these changes is in the IO interface, where the result objects are now passed by reference to the file reading functions, e.g.
```c++
cells_collection cells = io::read_cells(filename, mem_resource)
```
Now becomes:
```c++
cells_collection cells(mem_resource)
io::read_cells(cells, filename)
```

### DIFF
--- a/device/cuda/src/seeding/seed_finding.cu
+++ b/device/cuda/src/seeding/seed_finding.cu
@@ -195,8 +195,11 @@ seed_finding::output_type seed_finding::operator()(
     CUDA_ERROR_CHECK(cudaGetLastError());
 
     // Get the summary values.
-    device::seeding_global_counter globalCounter_host;
-    CUDA_ERROR_CHECK(cudaMemcpyAsync(&globalCounter_host,
+    vecmem::unique_alloc_ptr<device::seeding_global_counter>
+        globalCounter_host =
+            vecmem::make_unique_alloc<device::seeding_global_counter>(
+                (m_mr.host != nullptr) ? *(m_mr.host) : m_mr.main);
+    CUDA_ERROR_CHECK(cudaMemcpyAsync(globalCounter_host.get(),
                                      globalCounter_device.get(),
                                      sizeof(device::seeding_global_counter),
                                      cudaMemcpyDeviceToHost, stream));
@@ -204,10 +207,10 @@ seed_finding::output_type seed_finding::operator()(
 
     // Set up the doublet counter buffers.
     device::device_doublet_collection_types::buffer doublet_buffer_mb = {
-        globalCounter_host.m_nMidBot, m_mr.main};
+        globalCounter_host->m_nMidBot, m_mr.main};
     m_copy.setup(doublet_buffer_mb);
     device::device_doublet_collection_types::buffer doublet_buffer_mt = {
-        globalCounter_host.m_nMidTop, m_mr.main};
+        globalCounter_host->m_nMidTop, m_mr.main};
     m_copy.setup(doublet_buffer_mt);
 
     // Calculate the number of threads and thread blocks to run the doublet
@@ -232,7 +235,7 @@ seed_finding::output_type seed_finding::operator()(
     m_copy.setup(triplet_counter_spM_buffer);
     m_copy.memset(triplet_counter_spM_buffer, 0);
     device::triplet_counter_collection_types::buffer
-        triplet_counter_midBot_buffer = {globalCounter_host.m_nMidBot,
+        triplet_counter_midBot_buffer = {globalCounter_host->m_nMidBot,
                                          m_mr.main,
                                          vecmem::data::buffer_type::resizable};
     m_copy.setup(triplet_counter_midBot_buffer);
@@ -241,7 +244,7 @@ seed_finding::output_type seed_finding::operator()(
     // counting kernel for.
     const unsigned int nTripletCountThreads = WARP_SIZE * 2;
     const unsigned int nTripletCountBlocks =
-        (globalCounter_host.m_nMidBot + nTripletCountThreads - 1) /
+        (globalCounter_host->m_nMidBot + nTripletCountThreads - 1) /
         nTripletCountThreads;
 
     // Count the number of triplets that we need to produce.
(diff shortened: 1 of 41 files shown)


guilhermeAlmeida1:
Some positive performance plots which come from this change:

stephenswat:
Out of curiosity, do you have any Nsight Systems data on how much time we spend copying data to and from the device before and after this change?

guilhermeAlmeida1:
> Out of curiosity, do you have any Nsight Systems data on how much time we spend copying data to and from the device before and after this change?

Nsight Systems GPU Summary (Kernels/MemOps), only showing MemOps, using:
Non-pinned host memory: 
 | 	Time	 | 	Total Time	 | 	Instances	 | 	Avg	 | 	Operation	 | 
 | 	---	 | 	---	 | 	---	 | 	---	 | 	---	 | 
 | 	5.80%	 | 	161.568 ms	 | 	9900	 | 	16.320 μs	 | 	[CUDA memcpy DtoH]	 | 
 | 	18.50%	 | 	512.505 ms	 | 	4400	 | 	116.478 μs	 | 	[CUDA memcpy HtoD]	 | 
 | 	0.40%	 | 	12.449 ms	 | 	8800	 | 	1.414 μs	 | 	[CUDA memset]	 | 

Pinned host memory:
 | 	Time	 | 	Total Time	 | 	Instances	 | 	Avg	 | 	Operation	 | 
 | 	---	 | 	---	 | 	---	 | 	---	 | 	---	 | 
 | 	5.90%	 | 	154.254 ms	 | 	9900	 | 	15.581 μs	 | 	[CUDA memcpy DtoH]	 | 
 | 	14.50%	 | 	381.746 ms	 | 	4400	 | 	86.760 μs	 | 	[CUDA memcpy HtoD]	 | 
 | 	0.60%	 | 	16.629 ms	 | 	8800	 | 	1.889 μs	 | 	[CUDA memset]	 |

device/cuda/src/seeding/seed_finding.cu
```diff
@@ -195,19 +195,22 @@ seed_finding::output_type seed_finding::operator()(
     CUDA_ERROR_CHECK(cudaGetLastError());
 
     // Get the summary values.
-    device::seeding_global_counter globalCounter_host;
-    CUDA_ERROR_CHECK(cudaMemcpyAsync(&globalCounter_host,
+    vecmem::unique_alloc_ptr<device::seeding_global_counter>
+        globalCounter_host =
+            vecmem::make_unique_alloc<device::seeding_global_counter>(
+                *(m_mr.host));
```
krasznaa, line 201: :thinking: We currently use the algorithms with separate host and device memory resources. But it would be nice to keep the ability of using a single `vecmem::cuda::managed_memory_resource`. Mainly for testing purposes.

While a shared memory resource in general would require quite a few code changes (since explicit copies are not needed when using that sort of a thing), at least for now we should check whether `m_mr.host` is a `nullptr` or not. We can think about taking a different code path with different memory resource setups in the future.

device/sycl/src/clusterization/clusterization_algorithm.sycl
```diff
@@ -137,32 +137,33 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
         .wait_and_throw();
 
     // Copy number of measurements to host
-    unsigned int num_measurements_host;
+    vecmem::unique_alloc_ptr<unsigned int> num_measurements_host =
+        vecmem::make_unique_alloc<unsigned int>(*(m_mr.host));
```
krasznaa, line 141: Same comment here. And in a few more places, so I won't point out all of them...

examples/run/common/throughput_mt.ipp
```diff
@@ -80,17 +80,17 @@ int throughput_mt(std::string_view description, int argc, char* argv[],
...
         performance::timer t{"File reading", times};
-        for (unsigned int event = 0; event < throughput_cfg.loaded_events;
-             ++event) {
-            input = io::read(
-                throughput_cfg.loaded_events, throughput_cfg.input_directory,
-                throughput_cfg.detector_file,
-                throughput_cfg.digitization_config_file,
-                throughput_cfg.input_data_format, &uncached_host_mr);
+        for (std::size_t i = 0; i < throughput_cfg.loaded_events; ++i) {
+            input.push_back(&uncached_host_mr);
```
krasznaa, line 88: Let's be a bit more explicit with this code. I know what is happening here, but to a lay reader it would be pretty confusing. (Since an auto-conversion is taking place here.)

io/include/traccc/io/reader_edm.hpp
```diff
@@ -19,6 +19,9 @@ namespace traccc::io {
 struct cell_reader_output {
     cell_collection_types::host cells;
     cell_module_collection_types::host modules;
+
+    cell_reader_output() {}
+    cell_reader_output(vecmem::memory_resource* mr) : cells(mr), modules(mr) {}
```
krasznaa, line 24: Doesn't a single constructor suffice? :thinking:

```c++
cell_reader_output(vecmem::memory_resource* mr = nullptr) : cells(mr), modules(mr) {}
```

krasznaa (commented):
I am not against this type of API change,  but I don't think it is absolutely necessary. In my testing copy elision took care of making the

```c++
auto some_container = traccc::io::read_some_container(...);
```

type of code produce a container with the correct memory resource. It would be worth playing with this a bit more before we commit to the API change.

With that being said, this API is definitely less error prone. (As copy elision can only go so far. It would not work when assigning something to a pre-existing object.) 

And if we go with the new API, we should (in a follow-up PR) just rename all the functions to `traccc::io::read(...)` in my mind. :smile: Similar to what we do already with `traccc::io::write(...)`. (This was just not possible to do with the previous API...)

guilhermeAlmeida1:
> In my testing copy elision took care of making the `auto some_container = traccc::io::read_some_container(...);` type of code produce a container with the correct memory resource. 

My initial comment was, in fact, not correct, as this does seem to work (at least sometimes). But like you said it's prone to error. I find the new API safer and not particularly complicated, in fact, so I think we should change it.

guilhermeAlmeida1:
All other comments have been adressed

stephenswat (approved):
What a shame to see a return-based API replaced by an output-argument-based API... :cry: 

But if it is the only way to go, then so be it.

PR #394, opened 2023-05-08
Record fitting statistics (Ndf, Chi2) in KF and Add its plot tool

beomki-yeo:
This PR records the Ndf and Chi2 of fitted track into the header of `track_state_container`

Ndf is the summation of the measurement dimension subtracted by 5 and (global) Chi2 is the summation of (local) chi2 of each measurement. The KF unit test is also enhanced by checking the (Ndf == 2 * number of planes - 5)

Here is how the plots look like...

There is an issue that the pvalue distribution is not uniform but I think I know how to fix it. I will handle the issue in the next PRs


beomki-yeo:
A pvalue distribution is not uniform :woman_facepalming:

krasznaa (approved):
From a technical perspective I don't have objections with this one. If you are happy with it @beomki-yeo, then feel free to merge it in. :wink:

PR #395, opened 2023-05-09
Add tests for container copy

beomki-yeo:


krasznaa (commented):
Took me a while to remember where the previous test went. Finally I remembered that it was removed in #368, when the cell EDM got changed.

Using a test EDM to test the algorithms is a very good idea. I'm generally in favour of this. But I'd prefer to re-introduce this test with a "CPU test" at the same time...

beomki-yeo:
Wow - Didn't know that there was a test for this.
What do you exactly mean with the "CPU test" btw. I mean how do we test d2h or h2d with the CPU?

krasznaa:
> Wow - Didn't know that there was a test for this. What do you exactly mean with the "CPU test" btw. I mean how do we test d2h or h2d with the CPU?

You could copy the container from one `::host` object to a `::buffer` object, and vice versa. The buffer would still be in host memory, but the copy itself could still test if the "3rd object" in this chain is the same as the first one.

But I'm mostly just thinking out loud. Since I myself also only added a CUDA test in #238... :thinking:

beomki-yeo:
OK I made a test for host-to-host copy as suggested

beomki-yeo:
@krasznaa Could you approve this :p

krasznaa (approved):
:+1:

PR #397, opened 2023-05-11
External packages only build when needed

guilhermeAlmeida1:
Minor improvement to the build reducing the fetched content to only what is actually getting used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,7 +140,7 @@ endif()
 
 # Set up Kokkos.
 option( TRACCC_SETUP_KOKKOS
-   "Set up the Kokkos library" TRUE )
+   "Set up the Kokkos library" ${TRACCC_BUILD_KOKKOS} )
 option( TRACCC_USE_SYSTEM_KOKKOS
    "Pick up an existing installation of Kokkos from the build environment"
    ${TRACCC_USE_SYSTEM_LIBS} )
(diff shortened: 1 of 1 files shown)


stephenswat (changes requested):
Unless you have a good reason to do it this way, why not just go to https://github.com/acts-project/traccc/blob/main/CMakeLists.txt#L147 and change `if( TRACCC_SETUP_KOKKOS )` to `if( TRACCC_SETUP_KOKKOS AND TRACCC_BUILD_KOKKOS)`?

guilhermeAlmeida1:
> Unless you have a good reason to do it this way, why not just go to https://github.com/acts-project/traccc/blob/main/CMakeLists.txt#L147 and change `if( TRACCC_SETUP_KOKKOS )` to `if( TRACCC_SETUP_KOKKOS AND TRACCC_BUILD_KOKKOS)`?

Oh, I had somehow missed these two "SETUP" flags.

krasznaa:
Not even `AND` conditions! Just make the default value of the `SETUP` variables depend on the appropriate `BUILD` variables.

guilhermeAlmeida1:
Thank you both for the help, I've changed this to the correct approach now

stephenswat (dismissed):
Parece bom para mim

stephenswat (approved):
image

PR #399, opened 2023-05-11
Clean up unit_vectors.hpp

beomki-yeo:
Most of functions defined in `unit_vectors` are either unused or already defined in algebra-plugins. So I cleaned them up and rename it to `ranges.hpp`.

Also added a unit test for `eta_to_theta_range()` function


beomki-yeo:
`ranges.hpp` is moved to `perofrmance/utils` because its function is not used in core directory

stephenswat (approved):

PR #400, opened 2023-05-12
Track candidate with `detray::barcode`

beomki-yeo:
Still not sure if we have to unify `geometry_id` with `unit64_t` or `detray::geometry::barcode` over the entire project.

But let's do this for track_candidate at the moment for uniformity with `track_state`


stephenswat (approved):
Let's go ahead with this.

PR #403, opened 2023-05-16
Remove seeding internal unit system and refactor interface

beomki-yeo:
I started this PR to replace `Acts::Units` with `detray::units` but extended its scope eventually. The following is the changes:

1. Replacing `Acts::Unit` with `detray::unit`
2. Remove internal unit system
3. Remove the functions that creates config objects (e.g. `default_seed_finder_config()`)

I had to refactor the constructors of seeding algorithms to while removing the config generating functions.

There has been also a bug in calculating `pTperHelixRadius`: https://github.com/acts-project/traccc/blob/ceb8f3c794c4108cdf233ab91e1e06977a315041/core/src/seeding/seeding_algorithm.cpp#L27-L29 

The right equation is `rho [mm] / pT [MeV] = 0.0033356 / B[kT] <=>  pT[MeV] / rho[mm] = 299.79 * B[kT]`, so the factor of 300 is wrong
(Reference: https://en.wikipedia.org/wiki/Rigidity_(electromagnetism) , this is possibly the wikipedia page that I visited the most :stuck_out_tongue:... )

The same bug exists in ACTS Core as well.
https://github.com/acts-project/acts/blob/1f7a6c0697bf067030c727f5f763dee829b18fbd/Core/include/Acts/Seeding/SeedFinderConfig.hpp#L265-266

Therefore, after fixing the bug, the `compare_with_acts_seeding` fails. Need to think about how to bypass the test failures...

The PR also detaches the core library from ACTS Core - I moved `digitization_config` to io library


core/include/traccc/seeding/detail/seeding_config.hpp
```diff
@@ -117,29 +114,28 @@ struct seedfinder_config {
...
+                   std::sqrt(radLengthPerSeed) *
+                   (1 + 0.038 * std::log(radLengthPerSeed));
+
+        float maxScatteringAngle = highland / minPt;
+        maxScatteringAngle2 = maxScatteringAngle * maxScatteringAngle;
+
+        // pT [GeV] / R[mm] = B [T] in natrual unit
+        // Original equation with a bug
+        // pTPerHelixRadius = bFieldInZ * 0.0003 /
+        // traccc::unit<traccc::scalar>::T;
```
beomki-yeo, line 130: This commented lines are equal to what is calculated in main branch

stephenswat (commented):
I'm happy to approve this, but I'm curious whether you have checked the physics performance. Is it unchanged? Also, it might be useful to add a docstring to the configuration objects, and anywhere where we accept floating point numbers describing physical quantities, what the units are supposed to be.

beomki-yeo:
>  Is it unchanged?

Yes It is changed slightly because there is a change in the constant 300->299.7. I don't think it will affect the actual tracking performance meaningfully

> what the units are supposed to be.

We don't have to specify the unit with the docstring. (I don't know how to put this clearly) In the ACTS unit system, the physical quantities are just numbers without a unit. The original internal unit system was what needed such docstring actually

tests/cpu/compare_with_acts_seeding.cpp
```diff
@@ -336,7 +316,7 @@ TEST_P(CompareWithActsSeedingTests, Run) {
     float seed_match_ratio = float(n_seed_match) / seeds.size();
 
     // Ensure that ACTS and traccc give the same result
-    EXPECT_EQ(seeds.size(), seedVector.size());
+    EXPECT_NEAR(seeds.size(), seedVector.size(), seeds.size() * 0.001);
```
beomki-yeo, line 319: I mitigated the condition to pass the test

core/src/seeding/seeding_algorithm.cpp
```diff
@@ -12,60 +12,24 @@
...
 seeding_algorithm::output_type seeding_algorithm::operator()(
     const spacepoint_collection_types::host& spacepoints) const {
 
-    return m_seed_finding(spacepoints, m_spacepoint_binning(spacepoints));
+    /// Sub-algorithm performing the spacepoint binning
+    spacepoint_binning binning_alg(m_finder_config, m_grid_config, m_mr);
+    /// Sub-algorithm performing the seed finding
+    seed_finding finding_alg(m_finder_config, m_filter_config);
+
+    return finding_alg(spacepoints, binning_alg(spacepoints));
```
krasznaa, line 32: Why? Why would you want to (re-)create the `traccc::spacepoint_binning` and `traccc::seed_finding` algorithms on every call of the operator? :confused:

Let's keep the setup exactly like we had before. The only thing that I can see being useful is to extend the constructor of `traccc::seeding_algorithm` to receive non-standard configurations. Since right now there is no way to specify a non-standard configuration. :frowning:

But we should keep the `m_spacepoint_binning` and `m_seed_finding` members as they are. Let's not get rid of them!

device/cuda/include/traccc/cuda/seeding/seeding_algorithm.hpp
```diff
@@ -54,11 +54,17 @@ class seeding_algorithm : public algorithm<seed_collection_types::buffer(
...
+    seedfinder_config m_finder_config;
+    spacepoint_grid_config m_grid_config;
+    seedfilter_config m_filter_config;
 
+    /// Memory resource used by the algorithm
+    traccc::memory_resource m_mr;
+    /// The copy object to use
+    vecmem::copy& m_copy;
+    /// The CUDA stream to use
+    stream& m_stream;
```
krasznaa, line 67: The same here. Instead of adding unnecessary members to this class, let's just make sure that user-supplied "config objects" can be given to the algorithm's constructor. And then passed on directly to the underlying algorithms.

device/sycl/include/traccc/sycl/seeding/seeding_algorithm.hpp
```diff
@@ -50,10 +50,17 @@ class seeding_algorithm : public algorithm<seed_collection_types::buffer(
...
+    seedfinder_config m_finder_config;
+    spacepoint_grid_config m_grid_config;
+    seedfilter_config m_filter_config;
+
+    /// Memory resource used by the algorithm
+    traccc::memory_resource m_mr;
+    /// The copy object to use
+    vecmem::copy& m_copy;
+    /// The SYCL Queue
+    queue_wrapper m_queue;
```
krasznaa, line 63: You know the drill by now...

tests/cpu/compare_with_acts_seeding.cpp
krasznaa: Not great, but I can live with it for now.

krasznaa (changes requested):
I like the PR. But the fundamental setup of the `seeding_algorithm` classes should really not change. Just how they pass the configuration objects to the underlying algorithm objects.

device/cuda/include/traccc/cuda/seeding/seeding_algorithm.hpp
stephenswat: Are you talking about putting the memory resources and copy objects and streams into the config objects? Because we should absolutely **not** be doing that. :worried:

krasznaa: No, I'm not.

Currently all `seeding_algorithm` classes have 2 members: A "sub-algorithm" doing the spacepoint binning, and another "sub-algorithm" doing the seed finding.

The main algorithm's **only** purpose is to call those 2 sub-algorithms in sequence.

The configurations, memory resources, etc. are **only** needed by the sub-algorithms. The main algorithm doesn't use these directly ever.

So the setup in the current code is that the constructor or `seeding_algorithm` passes all the appropriate values to the constructors of the sub-algorithms, and then just holds on to the sub-algorithms. Not all the parameters that the sub-algorithms use internally.

Very long story short, in none of the `seeding_algorithm` classes should the list of members change in this MR. We should just extend the `seeding_algorithm(...)` constructors to be able to receive user-defined config objects. Along with the user-defined streams, memory resources, etc.

stephenswat: Ok good.

core/src/seeding/seeding_algorithm.cpp
beomki-yeo: I reverted the change but seeding will take config parameters as inputs instead. Please tell me if you are OK with that.
This will require some changes in `full_chain_algorithm` constructor and etc.

core/include/traccc/seeding/seeding_algorithm.hpp
```diff
@@ -43,6 +46,8 @@ class seeding_algorithm : public algorithm<seed_collection_types::host(
     spacepoint_binning m_spacepoint_binning;
     /// Sub-algorithm performing the seed finding
     seed_finding m_seed_finding;
+    /// Memory resource used by the algorithm
+    vecmem::memory_resource& m_mr;
```
krasznaa, line 50: This variable is still not needed here.

core/include/traccc/seeding/seeding_algorithm.hpp
```diff
@@ -28,7 +28,10 @@ class seeding_algorithm : public algorithm<seed_collection_types::host(
     ///
     /// @param mr The memory resource to use
     ///
-    seeding_algorithm(vecmem::memory_resource& mr);
+    seeding_algorithm(const seedfinder_config& finder_config,
+                      const spacepoint_grid_config& grid_config,
+                      const seedfilter_config& filter_config,
+                      vecmem::memory_resource& mr);
```
krasznaa, line 34: If you want to avoid modifying the client code in this MR, you should add this as an additional constructor on top of the existing one, which only takes a memory resource. (Default function arguments don't really play here, as I would also very much prefer the config objects coming before the memory resource when we want to define everything.)

krasznaa (changes requested):
Yes, this is what I had in mind. You should just still allow the algorithms to be created without a configuration object in case the user doesn't want to override their default settings. Just to avoid writing a whole lot of filler code when setting up these algorithms in all our our examples.

beomki-yeo:
> You should just still allow the algorithms to be created without a configuration object in case the user doesn't want to override their default settings. Just to avoid writing a whole lot of filler code when setting up these algorithms in all our our examples.

I don't get it. If someone wants to change the default parameter they can fix the default values in `seeding_config.hpp` :thinking:

core/include/traccc/seeding/detail/seeding_config.hpp
```diff
@@ -117,35 +117,43 @@ struct seedfinder_config {
...
+        rMax = finder_config.rMax;
+        zMax = finder_config.zMax;
+        zMin = finder_config.zMin;
+        deltaRMax = finder_config.deltaRMax;
+        cotThetaMax = finder_config.cotThetaMax;
+        impactMax = finder_config.impactMax;
+        phiMax = finder_config.phiMax;
+        phiMin = finder_config.phiMin;
+        phiBinDeflectionCoverage = finder_config.phiBinDeflectionCoverage;
+    }
```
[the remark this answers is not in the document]
beomki-yeo: Makes sense. Fixed

krasznaa:
> I don't get it. If someone wants to change the default parameter they can fix the default values in `seeding_config.hpp` 🤔

I don't care about it a whole lot for these R&D classes. But for a "production class", which is expected to 90% be used with its default configuration, and only be specifically configured in 10% of the cases, one should definitely rather go with:

```c++
algorithm(vecmem::memory_resource& mr) : algorithm(config_type{}, mr) {}
algorithm(const config_type& config, vecmem::memory_resource& mr) {...}
```

Just to save on every user having to write useless padding code in 90% of the cases.

So fine, since you already updated all the client code, none of which modify the default parameters, let's go with your changes...

examples/run/cpu/full_chain_algorithm.hpp
```diff
@@ -62,6 +65,11 @@ class full_chain_algorithm
     /// Track parameter estimation algorithm
     track_params_estimation m_track_parameter_estimation;
 
+    /// Configs
+    seedfinder_config m_finder_config;
+    spacepoint_grid_config m_grid_config;
+    seedfilter_config m_filter_config;
```
krasznaa, line 71: :frowning: Why...?

Much like how the different `seeding_algorithm`-s didn't need to hold on to any configuration objects themselves, the `full_chain_algorithm`-s don't need to do it either. They just need to pass these on to the "sub-algorithms" in their constructors, and that's it. No new members should be added to these classes.

examples/run/cuda/full_chain_algorithm.hpp
```diff
@@ -94,6 +97,11 @@ class full_chain_algorithm
     /// Track parameter estimation algorithm
     track_params_estimation m_track_parameter_estimation;
 
+    /// Configs
+    seedfinder_config m_finder_config;
+    spacepoint_grid_config m_grid_config;
+    seedfilter_config m_filter_config;
```
krasznaa, line 103: Same here.

examples/run/sycl/full_chain_algorithm.hpp
```diff
@@ -96,6 +99,11 @@ class full_chain_algorithm
     /// Track parameter estimation algorithm
     track_params_estimation m_track_parameter_estimation;
 
+    /// Configs
+    seedfinder_config m_finder_config;
+    spacepoint_grid_config m_grid_config;
+    seedfilter_config m_filter_config;
```
krasznaa, line 105: And here...

krasznaa (changes requested):
I don't like to be this guy by now, but as long as I held out for cleaning up these sort of issues before, they should be fixed in this latest update as well... :frowning:

examples/run/cpu/full_chain_algorithm.hpp
beomki-yeo: Otherwise, I cannot pass the config parameters in the copy constructor. Do you have an idea?

```
      m_seeding(
          parent.m_finder_config, parent.m_grid_config, parent.m_filter_config,
          memory_resource{*m_cached_device_mr, &m_host_mr}, m_copy, m_stream),
```

krasznaa: Ahh, crud... 😦 

The correct way of doing this would of course be to implement proper copy-constructors for all of those sub-algorithms. And have `full_chain_algorithm` just rely on those copy-constructors.

But fine, that's a fair amount of extra code that this PR would have to collect. Okay, let's just go with this for now then, we can clean up the code a little more in a future PR.

krasznaa (approved):

PR #405, opened 2023-05-18
Add functions to calculate pvalue

beomki-yeo:
Depends on #403.

This PR adds functions required for pvalue calculation.
I am thinking of using the pvalue for the Kalman filternig improvement without introducing ROOT library into traccc core


stephenswat:
Is there a reason to use a custom Γ function rather than `std::tgamma`? Unless physicists have a different Γ function that I am not familiar with.

beomki-yeo:
It is the same gamma and I didn't know that `std::tgamma` exists. 
Thing is that I am not sure if the lower and upper integral can be obtained from std functions.

And yes, I am thinking of using it in the device code.

PR #410, opened 2023-05-26
Implement seed finding Thunderdome in CUDA

stephenswat:
This draft pull request implements a Thunderdome-like fight to the death for seeds: _"Two seeds enter, one seed leaves"_ . The idea is to allow seeds to "eat" other seeds, growing bigger themselves and eliminating other seeds from the seed list. This is designed to significantly reduce the combinatorics that the CKF has to deal with, thereby improving performance.

image

### DIFF
--- a/core/include/traccc/edm/nseed.hpp
+++ b/core/include/traccc/edm/nseed.hpp
@@ -10,6 +10,7 @@
 #include <traccc/edm/container.hpp>
 #include <traccc/edm/seed.hpp>
 #include <traccc/edm/spacepoint.hpp>
+#include <traccc/definitions/qualifiers.hpp>
 
 namespace traccc {
 /**
@@ -32,18 +33,20 @@ struct nseed {
      */
     using link_type = spacepoint_collection_types::host::size_type;
 
+    nseed() = default;
+
     /**
      * @brief Construct a new n-seed object from a 3-seed object.
      *
      * @param s A 3-seed.
      */
-    nseed(const seed& s)
+    TRACCC_HOST_DEVICE nseed(const seed& s)
         : _size(3), _sps({s.spB_link, s.spM_link, s.spT_link}) {}
 
     /**
      * @brief Get the size of the seed.
      */
-    std::size_t size() const { return _size; }
+    TRACCC_HOST_DEVICE std::size_t size() const { return _size; }
 
     /**
      * @brief Get the first space point identifier in the seed.
@@ -55,7 +58,7 @@ struct nseed {
      */
     const link_type* cend() const { return &_sps[_size]; }
 
-    private:
+    public:
     std::size_t _size;
     std::array<link_type, N> _sps;
 };
--- a/device/cuda/include/traccc/cuda/seed_merging/seed_merging.hpp
+++ b/device/cuda/include/traccc/cuda/seed_merging/seed_merging.hpp
@@ -1,0 +1,34 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2023 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/cuda/utils/stream.hpp"
+#include "traccc/edm/seed.hpp"
+#include "traccc/edm/nseed.hpp"
+#include "traccc/utils/algorithm.hpp"
+#include "traccc/utils/memory_resource.hpp"
+#include <vecmem/utils/copy.hpp>
+#include <vecmem/memory/unique_ptr.hpp>
+
+
+namespace traccc::cuda {
+
+class seed_merging : public algorithm<std::pair<vecmem::unique_alloc_ptr<nseed<20>[]>, uint32_t>(const seed_collection_types::buffer&)> {
+    public:
+    seed_merging(const traccc::memory_resource& mr, stream& str);
+
+    output_type operator()(
+        const seed_collection_types::buffer&) const override;
+
+    private:
+    traccc::memory_resource m_mr;
+    stream& m_stream;
+};
+
+}  // namespace traccc::cuda
--- a/device/cuda/src/seed_merging/seed_merging.cu
+++ b/device/cuda/src/seed_merging/seed_merging.cu
@@ -1,0 +1,157 @@
+#include "traccc/cuda/seed_merging/seed_merging.hpp"
+#include "traccc/edm/seed.hpp"
+#include "traccc/edm/nseed.hpp"
+#include "traccc/cuda/utils/definitions.hpp"
+
+#include <iostream>
+
+namespace traccc::cuda {
+namespace kernels {
+template<std::size_t N>
+__global__ void convert_to_nseeds(seed_collection_types::view vf, nseed<N> * out, unsigned long long * out_n) {
+    seed_collection_types::device vfd(vf);
+
+    for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < vf.size(); i += gridDim.x * blockDim.x) {
+        out[i] = vfd[i];
+        atomicAdd(out_n, 1ULL);
+    }
+}
+
+template<std::size_t N, std::size_t M>
+__global__ void merge_nseeds(const nseed<N> * in, const unsigned long long * in_c, nseed<M> * out, unsigned long long * out_c) {
+    __shared__ nseed<M> out_seeds[32];
+    __shared__ uint32_t num_seeds;
+    __shared__ uint32_t num_consm;
+    __shared__ uint32_t out_index;
+
+    if (threadIdx.x == 0) {
+        num_seeds = 0;
+        num_consm = 0;
+    }
+
+    __syncthreads();
+
+    for (int i = threadIdx.x; i < *in_c; i += blockDim.x) {
+        if (i == blockIdx.x) {
+            continue;
+        }
+
+        bool compat, consumed;
+
+        if (in[blockIdx.x].size() == in[i].size()) {
+            compat = true;
+            consumed = true;
+
+            for (int j = 0; j < in[i].size() - 1; ++j) {
+                if (in[blockIdx.x]._sps[j+1] != in[i]._sps[j]) {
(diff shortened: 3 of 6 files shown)


core/include/traccc/edm/nseed.hpp
```diff
@@ -55,7 +58,7 @@ struct nseed {
      */
     const link_type* cend() const { return &_sps[_size]; }
 
-    private:
+    public:
     std::size_t _size;
     std::array<link_type, N> _sps;
```
krasznaa, line 63: If client code is meant to access these variables now, they should really not be prefixed with `_`.

device/cuda/include/traccc/cuda/seed_merging/seed_merging.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include "traccc/edm/nseed.hpp"
+#include "traccc/utils/algorithm.hpp"
+#include "traccc/utils/memory_resource.hpp"
+#include <vecmem/utils/copy.hpp>
+#include <vecmem/memory/unique_ptr.hpp>
+
+
+namespace traccc::cuda {
+
+class seed_merging : public algorithm<std::pair<vecmem::unique_alloc_ptr<nseed<20>[]>, uint32_t>(const seed_collection_types::buffer&)> {
```
krasznaa, line 22: Why `nseed<20>`? I thought we were only going to merge 2 seeds together in this algorithm. At least for now. So why now `nseed<4>`? 😕 

At the same time the return type should really instead be `vecmem::data::vector_buffer<nseed<4> >`. Since that type is literally what you wrote here. (An array with a known size.)

device/cuda/src/seed_merging/seed_merging.cu
```diff
@@ -0,0 +1,157 @@
...
+namespace kernels {
+template<std::size_t N>
+__global__ void convert_to_nseeds(seed_collection_types::view vf, nseed<N> * out, unsigned long long * out_n) {
+    seed_collection_types::device vfd(vf);
+
+    for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < vf.size(); i += gridDim.x * blockDim.x) {
+        out[i] = vfd[i];
+        atomicAdd(out_n, 1ULL);
+    }
+}
```
krasznaa, line 18: I don't understand this kernel... 😕

What's up with the `atomicAdd(...)`? We know exactly how many (3-spacepoint) seeds we start with. As long as the kernel code doesn't have a bug in it, that's exactly how large `out_n` is supposed to become.

At the same time, you'll need to put some amount of explanation here about that for-loop. I assume you're going for optimal cache line usage with it. Though on first look I don't understand why this access would yield better caching than just processing a single seed in every thread. 😕

device/cuda/include/traccc/cuda/seed_merging/seed_merging.hpp
```diff
@@ -0,0 +1,34 @@
...
+    public:
+    seed_merging(const traccc::memory_resource& mr, stream& str);
+
+    output_type operator()(
+        const seed_collection_types::buffer&) const override;
```
krasznaa, line 27: The input parameter needs to be `seed_collection_type::const_view`. The algorithm doesn't need to know whether it's a buffer or something else.

device/cuda/src/seed_merging/seed_merging.cu
```diff
@@ -0,0 +1,157 @@
...
+}
+
+seed_merging::seed_merging(const traccc::memory_resource& mr, stream& str) : m_mr(mr), m_stream(str) {
+}
+
+seed_merging::output_type seed_merging::operator()(const seed_collection_types::buffer&i) const {
+    vecmem::unique_alloc_ptr<nseed<20>[]>
+        arr1 = vecmem::make_unique_alloc<nseed<20>[]>(m_mr.main, 1000000),
+        arr2 = vecmem::make_unique_alloc<nseed<20>[]>(m_mr.main, 1000000);
```
krasznaa, line 122: What's with the 1M fixed numbers? 😕 Just ask the input **view** how many elements it has. I.e.

```c++
const unsigned int nseeds = m_copy.get_size(input_view);
```

Assuming that you switch to using a view, and that you ask the user to provide a `vecmem::copy` object to the algorithm.

device/cuda/src/seed_merging/seed_merging.cu
```diff
@@ -0,0 +1,157 @@
...
+
+    vecmem::unique_alloc_ptr<unsigned long long>
+        siz1 = vecmem::make_unique_alloc<unsigned long long>(m_mr.main),
+        siz2 = vecmem::make_unique_alloc<unsigned long long>(m_mr.main);
+
+    kernels::convert_to_nseeds<20><<<2048, 256>>>(i, arr1.get(), siz1.get());
```
krasznaa, line 128: We should try to write new code with asynchronicity in mind. But this is just to note this, I'm fine with this PR not caring about asynchronicity yet.

device/cuda/src/seed_merging/seed_merging.cu
```diff
@@ -0,0 +1,157 @@
...
+    CUDA_ERROR_CHECK(cudaGetLastError());
+    CUDA_ERROR_CHECK(cudaDeviceSynchronize());
+
+    unsigned long long rc;
+
+    CUDA_ERROR_CHECK(cudaMemcpy(&rc, siz1.get(), sizeof(unsigned long long), cudaMemcpyDeviceToHost));
+
+    std::cout << "Step 0 has " << rc << " seeds." << std::endl;
+
+    for (std::size_t i = 0; i < 5; ++i) {
```
krasznaa, line 139: I imagine 5 is a number that you came up with after some testing. That is fine, but it has to be made a configurable property of the algorithm...

device/cuda/src/seed_merging/seed_merging.cu
```diff
@@ -0,0 +1,157 @@
...
+        CUDA_ERROR_CHECK(cudaMemset(siz2.get(), 0, sizeof(unsigned long long)));
+        kernels::merge_nseeds<20, 20><<<rc, 256>>>(arr1.get(), siz1.get(), arr2.get(), siz2.get());
```
krasznaa, line 141: Curious. I thought the compiler would be able to deduce those template parameters automatically... 🤔

device/cuda/src/seed_merging/seed_merging.cu
```diff
@@ -0,0 +1,157 @@
...
+
+        CUDA_ERROR_CHECK(cudaGetLastError());
+        CUDA_ERROR_CHECK(cudaDeviceSynchronize());
+
+        std::swap(arr1, arr2);
+        std::swap(siz1, siz2);
```
krasznaa, line 147: I like this. 😄 You should've mentioned that you're taking queues from the discussion we had with Beomki about AdePT's memory handling. 😛

krasznaa (changes requested):
It might not look like it from my comments, but I like this proposal. I believe it will prove very useful once we get to making CKF as fast as we can.

I'm just a stickler for some of the technicalities...

core/include/traccc/edm/nseed.hpp
stephenswat: Absolutely correct, I plan on renaming these and making some more changes to the `nseed` class during the development of this PR. Well spotted. :+1:

device/cuda/include/traccc/cuda/seed_merging/seed_merging.hpp
stephenswat: Ah okay, sorry if I did not explain properly. The five-iteration loop you commented on below means that seeds can become at most five spacepoints longer than they are now. For example, seed ABC can be merged with BCD to form ABCD in step 1, then that can merge with BCDE in step 2, etc. So In principle the seed capacity needs to be 3 + _n_ where _n_ is the number of merging steps.

device/cuda/src/seed_merging/seed_merging.cu
stephenswat: You're absolutely correct that this doesn't make sense right now. What I want to do in the future is turn this kernel into a seed binning kernel that stores seeds into _φ_-bins to reduce the amount of work done. When I get that sorted I will need to do atomic space reservations. For now, with only one bin, this is indeed not really necessary.

device/cuda/include/traccc/cuda/seed_merging/seed_merging.hpp
stephenswat: :+1:

device/cuda/src/seed_merging/seed_merging.cu
stephenswat: So the thing with this is that it is _possible_ for the number of seeds to increase, at least in early steps of the algorithms. So I need a little bit of buffer space for that. But you're right I can do two times the number of initial seeds, for example.

device/cuda/src/seed_merging/seed_merging.cu
stephenswat: I'll make sure to add the appropriate synchronization points later. This algorithm is also a prime candidate for Dynamic Parallalism 2, which might reduce the number of synchronization points. :slightly_smiling_face:

device/cuda/src/seed_merging/seed_merging.cu
stephenswat: Absolutely.

device/cuda/src/seed_merging/seed_merging.cu
stephenswat: It probably can, actually! I added them explicitly for no real good reason, honestly.

device/cuda/src/seed_merging/seed_merging.cu
stephenswat: Not to toot my own horn but [I've been on this train for a while](https://github.com/acts-project/traccc/pull/230/files#diff-3786648a39e32623f3757c0a32cf10666485600f80acfb39c106906989788050R812). :angel:

stephenswat:
Thanks a lot for taking a look, glad you like it! It's still very early development but you've raised some valid points.

PR #412, opened 2023-05-30
d2h copy with a size instead of capacity

beomki-yeo:
I am trying to make d2h_copy assign the vector size with a size instead of capacity.
But it gives me a segmentation fault error. Is it even a valid approach?

### DIFF
--- a/device/common/include/traccc/device/impl/container_d2h_copy_alg.ipp
+++ b/device/common/include/traccc/device/impl/container_d2h_copy_alg.ipp
@@ -38,7 +38,8 @@ container_d2h_copy_alg<CONTAINER_TYPES>::operator()(input_type input) const {
     std::vector<std::size_t> capacities(size, 0);
     std::transform(input.items.host_ptr(), input.items.host_ptr() + size,
                    capacities.begin(),
-                   [](const auto& view) { return view.capacity(); });
+                   [](const auto& view) { return view.size(); });
+
     typename CONTAINER_TYPES::buffer hostBuffer{{size, *host_mr},
                                                 {capacities, *host_mr}};
     m_hostCopy.setup(hostBuffer.headers);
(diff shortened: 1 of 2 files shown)


guilhermeAlmeida1:
I'm not sure this is intended funcionality (@krasznaa ) but it seems that the size is not accessible to `host_ptr`
e.g. if you try:
```c++
    auto b = input.items.host_ptr()[0].capacity();
    printf("before %u\n", b);
    auto a = input.items.host_ptr()[0].size();
    printf("%u\n", a);
```

the seg fault comes at line 3.

Regardless of that, making the copy algorithm just change to using the size instead of the capacity might not be the most correct approach, and this somewhat links back to https://github.com/acts-project/vecmem/issues/95

krasznaa:
This is sort of a longer story. But in short, let's not try to be super smart here. I experimented with the performance of these things quite a bit before coming up with this exact code.

Technically the problem is indeed that the sizes of the inner vectors of a resizable jagged vector buffer are not accessible from the host. This is why [vecmem::copy::get_sizes(...)](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/utils/copy.hpp#L136-L139) exists.

But you should use that call to get the exact sizes either. You see, you have two options with your input container in this case.
  1. It is a non-resizable container, or a resizable one that was filled to the brim. In this case `capacity()` will be equivalent to `size()` anyway.
  2. It's a resizable container that was not filled fully.

In the second case your update would seem like it could help with performance. But... if we accept that we copy more data than absolutely necessary, we can perform the copy of the entire payload of the input jagged vector with a single copy operation. Since the memory allocated for the jagged vector **is** contiguous.

If on the other hand we insist on using the concrete sizes of the inner vectors, we end up with having to perform N separate, small copies instead.

Of course the phase space for when one is better than the other is not a completely trivial one. But I decided a while ago that instead of trying to be super smart about it (trying to minimise some complex expression with the number of copies and the sizes of the individual copies), I would always go for minimising the number of copies.

So long story short, let's just not bother with this here...

PR #415, opened 2023-06-09
Fix: Handle empty inputs gracefully

### LINKED ISSUE #414: CUDA algorithms should handle empty inputs gracefully
Related to #413, the CUDA code currently does not check whether the inputs passed to it are empty. In some cases (e.g. in the seeding) this causes the host to request a kernel launch with zero blocks, which causes a launch failure and breaks the application. For each kernel, we should aim to determine the appropriate behaviour in case the input size is zero, so that we can handle these edge cases gracefully.

guilhermeAlmeida1:
Fixes #414 
Depends on https://github.com/acts-project/vecmem/pull/232

Note: this should pass the CI but **should not** be merged in before https://github.com/acts-project/vecmem/pull/232 and updating the external dependency on vecmem

### DIFF
--- a/device/cuda/src/clusterization/clusterization_algorithm.cu
+++ b/device/cuda/src/clusterization/clusterization_algorithm.cu
@@ -91,6 +91,11 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
     const cell_collection_types::view::size_type num_cells =
         m_copy.get_size(cells);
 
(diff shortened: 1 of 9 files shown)


stephenswat (approved):

krasznaa (approved):
Let's quickly merge this in before something invalidates it again... 😛

Issue #413, opened 2023-06-07
Single muon events fail to construct any seeds


stephenswat:
As discussed in the EF Tracking meeting this Monday, Shima (whose GitHub handle I do not have) has been running into an issue where very simple single-muon events are failing to generate seeds. Spacepoint files which reproduce this failure are [event000000280-spacepoints.csv](https://github.com/acts-project/traccc/files/11676453/event000000280-spacepoints.csv) and [event000000114-spacepoints.csv](https://github.com/acts-project/traccc/files/11676454/event000000114-spacepoints.csv). Both of these tracks have very similar parameter, namely θ = 1.42 and 1.34, φ = 0.32 and 0.36, and z<sub>0</sub> = 98.2mm and 97.9mm. We should aim to understand why these (trivial) tracks fail to reconstruct.

Shima has hinted that these seeds fail to be made due to the `dCotThetaMinusError2` cut. This could be the result of some catastrophic cancellation, but the cause is not clear to me at this time.

shimasnd:
A minor additional info of this issue is that the setting for the seed filtering (i.e. seeding_config.hpp values) are loosened to cover the detector geometry, in which these single muon events are simulated. 
![image](https://github.com/acts-project/traccc/assets/84299713/a14c25de-789e-45c9-b9f3-2021ead12a16)
Many thanks for your investigation.

beomki-yeo:
How often does this happen? And how do we can reproduce the input spacepoint file?

stephenswat:
My understanding is this happens deterministically for the given spacepoint files.

shimasnd:
These are actually ATLAS single muon events. The global coordinates of the ATLAS pixel clusters are treated as spacepoints and are used to test the traccc seeding process. They are not in traccc-data. I got these 2 events from 500 single muon events, which are generated in 0.1<eta<0.3, 0.3<phi<0.5, and z0 ranging over +-300mm.

beomki-yeo:
Thanks for the info. Could you also check the truth z0 of the particles which fail the reconsturction? I guess the default seeding config parameter should be changed to cover 300 mm range of z0

shimasnd:
As Stephen wrote in his first comment, the truth track values of these two events are:
"namely θ = 1.42 and 1.34, φ = 0.32 and 0.36, and z0 = 98.2mm and 97.9mm"
They fail at the triplet production. 
[...]

PR #416, opened 2023-06-09
Add tests for seeding edge cases

beomki-yeo:
NOTE: This PR is not meant to be merged but to handle the issue from #413

Pinging @shimasnd

The two cases that fail in #413 can be reconstructed by adjusting `maxPtScattering`. 
I added unit tests for the cases.

Now I am confused why the parameter is named as `maxPtScattering` instead of `minPtScattering` ...  but whatever  :p


beomki-yeo:
@shimasnd Out of curiosity, what were the pT and p of muons that failed the reconstruction?

shimasnd:
Here are the muon p and pT of them:
event000000114:: true p=16.62GeV, pT=16.42 GeV
event000000280:: true p = 1.85 GeV, pT=1.80 GeV
Hope it helps.

beomki-yeo:
Now I think it is worth merging this in :p

PR #417, opened 2023-06-13
Change input parameter for kalman fitting test

beomki-yeo:
Decoupled from #352 for easier review


krasznaa (approved):
Without fully knowing every nook and cranny of this code, this all looks very reasonable. :wink: As long as it simplifies #352 a bit, I'm all for it. :smile:

PR #418, opened 2023-06-14
Add B field as an input for track parameter estimation

beomki-yeo:
Update: Instead of making an alternative one, I just make the original algorithm take the b field as an input argument

---------------------------
~~Adding a track parameter estimation with detector object which holds B field. I am not replacing the original one for faster development.~~

I am trying to reconstruct the momentum of spacepoints provided by @shimasnd and @stephenswat but current algorithm giving me a wrong value when I assumed 2 T magnetic field. I think there is a bug in the current algorithm but could you guys confirm the bfield strength?

### DIFF
--- a/core/include/traccc/seeding/track_params_estimation_helper.hpp
+++ b/core/include/traccc/seeding/track_params_estimation_helper.hpp
@@ -86,9 +86,10 @@ inline TRACCC_HOST_DEVICE bound_vector seed_to_bound_vector(
     // The projection of the top space point on the transverse plane of
     // the new frame
     scalar rn = local2[0] * local2[0] + local2[1] * local2[1];
-    // The (1/tanTheta) of momentum in the new frame,
+    // The (1/tanTheta) of momentum in the new frame
+    static constexpr scalar G = static_cast<scalar>(1.f / 24.f);
     scalar invTanTheta =
-        local2[2] * std::sqrt(1.f / rn) / (1.f + rho * rho * rn);
+        local2[2] * std::sqrt(1.f / rn) / (1.f + G * rho * rho * rn);
 
     // The momentum direction in the new frame (the center of the circle
     // has the coordinate (-1.*A/(2*B), 1./(2*B)))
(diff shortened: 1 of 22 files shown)


beomki-yeo:
Think I will just refactor the current algorithm without detector template

core/include/traccc/seeding/track_params_estimation_helper.hpp
```diff
@@ -83,12 +83,13 @@ inline TRACCC_HOST_DEVICE bound_vector seed_to_bound_vector(
...
     // Curvature (with a sign) estimate
     scalar rho = -2.0f * B / getter::perp(vector2{1., A});
+
     // The projection of the top space point on the transverse plane of
     // the new frame
     scalar rn = local2[0] * local2[0] + local2[1] * local2[1];
     // The (1/tanTheta) of momentum in the new frame,
     scalar invTanTheta =
-        local2[2] * std::sqrt(1.f / rn) / (1.f + rho * rho * rn);
+        local2[2] * std::sqrt(1.f / rn) / (1.f + rho * rho * rn / 24.);
```
beomki-yeo, line 92: Think I spotted a bug - without this term track_parameter_estimation test fails :p

guilhermeAlmeida1:
I think in the end this should be set in the constructor of the algorithm with a `covfie::field_view` (which should come from `detector::get_bfield()` ). Definately not just a vector as we want to be able to handle non-const bfields.

beomki-yeo:
OK I like that idea

tests/cpu/compare_with_acts_seeding.cpp
```diff
@@ -70,10 +70,10 @@ inline bool operator==(const Acts::BoundVector& acts_vec,
             traccc::float_epsilon * 10 &&
         std::abs(acts_vec[Acts::eBoundTheta] -
                  traccc::getter::element(traccc_vec, traccc::e_bound_theta,
-                                         0)) < traccc::float_epsilon * 10 &&
+                                         0)) < traccc::float_epsilon * 1000 &&
```
beomki-yeo, line 73: I had to mitigate the condition for equality operation on momentum direction as the specific bug is not fixed in ACTS yet.
I will revert the change once acts-project/acts#2211 is included in any of next releases

beomki-yeo:
Can it be approved soon?

tests/cpu/compare_with_acts_seeding.cpp
guilhermeAlmeida1: This can be reverted now with https://github.com/acts-project/acts/releases/tag/v27.0.0

PR #411, opened 2023-05-28
Add CKF examples with detray toy detector

beomki-yeo:
CKF examples with more realistic setup. Also includes quite many bug fixes in CPU/CUDA CKF.
The first benchmark result is not bad - for 10000 tracks, cuda is 10 times faster than cpu (single thread, RTX 2070)

```
Running ./bin/traccc_truth_finding_example_cuda detray_simulation/toy_detector/n_particles_10000/ 1
===>>> Event 0 <<<===
Track candidate matching Rate: 1
==> Statistics ... 
- created (cuda) 28107 found tracks
- created (cuda) 28107 fitted tracks
- created  (cpu) 28048 found tracks
- created  (cpu) 28048 fitted tracks
==>Elapsed times...
         Track finding  (cuda)  325 ms
         Track fitting  (cuda)  61 ms
          Track finding  (cpu)  3574 ms
          Track fitting  (cpu)  907 ms
```          

And fitting resolution might be suffering from high duplication rate

PR #421, opened 2023-06-19
Seeding with detray - Version1

beomki-yeo:
### UPDATE

This PR is for running seeding examples with detray geometry. (Geometry will be needed in spacepoint formation of CCL)

There was an issue in track parameter estimation where the outputs from ACTS and traccc are different. 

1. Why is there a discrepancy?
ACTS core track parameter estimation (TPE) gets local track parameter by doing global-to-local transformation while traccc one just reads the measurement file. There should not be difference _if we run the full chain from CCL_. Things get different if we start from seeding where spacepoint object are the truth hit:  TPE outputs of ACTS core is the global-to-local transform of truth hit without smearing and those of traccc will be the smeared measurements -> This fails the `compare_with_acts_seeding`.

2. Differece between this PR and #428
#428 makes both outputs the same by changing the TPE as ACTS one, and this PR doesn't have such change but disable the TPE part of `compare_with_acts_seeding`

3. Which one is better
I think we should rely on the measurement file (this PR) instead of doing global-to-local transformation (#428) even though it is different from ACTS core. Anyway the outputs should be the same for full chain algorithm and the computation is less massive. Above all, if we want to do the same as ACTS does, we will need to add template parameter of detector class into TPE to do the global to local transform properly, which I don't want to do. (#428 is pretty simplified for pixel only detector) Of course if we want to exclude any other measurement type in TPE, we can get by with what is in #428 

But people might have different opinions - Anyway I made two versions to make it easier for people to compare the different designs. So please share your thought. 

--------------------------------------------

Now depends on #424

--------------------------------------------

~~Depends on #411~~


examples/run/cpu/seeding_example.cpp
```diff
@@ -85,12 +109,26 @@ int seq_run(const traccc::seeding_input_config& i_cfg,
...
+
+            else if (i_cfg.run_detray_geometry == true) {
+                traccc::event_map2 evt_map(event, common_opts.input_directory,
+                                           common_opts.input_directory,
+                                           common_opts.input_directory);
+                sd_performance_writer.write(
+                    vecmem::get_data(seeds),
+                    vecmem::get_data(spacepoints_per_event), readOut.modules,
+                    evt_map);
+            }
```
krasznaa, line 131: This verbosity in the code may not be a good thing in the end... :thinking: I would much rather just have:

```c++
if (i_cfg.run_detray_geometry) {
    traccc::event_map2 evt_map(event, common_opts.input_directory,
                                common_opts.input_directory,
                                common_opts.input_directory);
    sd_performance_writer.write(
        vecmem::get_data(seeds),
        vecmem::get_data(spacepoints_per_event), readOut.modules,
        evt_map);
} else {
    traccc::event_map evt_map(event, i_cfg.detector_file,
                              common_opts.input_directory,
                              common_opts.input_directory, host_mr);
    sd_performance_writer.write(
        vecmem::get_data(seeds),
        vecmem::get_data(spacepoints_per_event), evt_map);
}
```

I.e. I **think** the compiler can be smart enough to recognize that one of these options will always happen, but the way that you wrote it, I believe the compiler does need to  check the value of that variable twice...

krasznaa (commented):
The comment goes for all cases where you check the value of the new command line parameter variable.

But I guess I'll need to listen to your talk today to better understand what these PRs are about exactly... :thinking:

examples/run/cuda/seeding_example_cuda.cpp
```diff
@@ -22,6 +23,11 @@
 #include "traccc/seeding/seeding_algorithm.hpp"
 #include "traccc/seeding/track_params_estimation.hpp"
 
+// Detray include(s).
+#include "detray/detectors/create_toy_geometry.hpp"
+#include "detray/io/json/json_reader.hpp"
+#include "detray/io/json/json_writer.hpp"
```
niermann999, line 29:
```suggestion
#include "detray/detectors/detector_metadata.hpp"
#include "detray/io/json/json_reader.hpp"
```

examples/run/sycl/seeding_example_sycl.sycl
```diff
@@ -28,6 +29,11 @@
 #include "traccc/options/handle_argument_errors.hpp"
 #include "traccc/options/seeding_input_options.hpp"
 
+// Detray include(s).
+#include "detray/detectors/create_toy_geometry.hpp"
+#include "detray/io/json/json_reader.hpp"
+#include "detray/io/json/json_writer.hpp"
```
niermann999, line 35:
```suggestion
#include "detray/detectors/detector_metadata.hpp"
#include "detray/io/json/json_reader.hpp"
```

io/src/csv/read_spacepoints.cpp
```diff
@@ -44,20 +60,21 @@ void read_spacepoints(spacepoint_reader_output& out, std::string_view filename,
...
-        // Find the local<->global transformation for the spacepoint's detector
-        // module.
-        const transform3& placement = geom[iohit.geometry_id];
-
         // Construct the global 3D position of the spacepoint.
         const point3 pos{iohit.tx, iohit.ty, iohit.tz};
 
         // Construct the local 3D(2D) position of the measurement.
-        const point3 lpos = placement.point_to_local(pos);
+        // const point3 lpos = placement.point_to_local(pos);
```
niermann999, line 67: Should this be removed?

tests/io/mock_data/event000000000-measurements.csv
```diff
@@ -0,0 +1,4 @@
+measurement_id,geometry_id,local_key,local0,local1,phi,theta,time,var_local0,var_local1,var_phi,var_theta,var_time
```
niermann999, line 1: Should this file be in the change set?

io/include/traccc/io/read_geometry.hpp
```diff
@@ -27,4 +27,19 @@ namespace traccc::io {
 geometry read_geometry(std::string_view filename,
                        data_format format = data_format::csv);
 
+/// Read in the detector geometry description from a detector object
+template <typename detector_t>
+geometry alt_read_geometry(const detector_t& det) {
```
niermann999, line 32: So , this is not in fact using the detray geometry, but just copying the data frm detray into a traccc data structure?

io/include/traccc/io/read_geometry.hpp
```diff
@@ -27,4 +27,19 @@ namespace traccc::io {
...
+
+    const auto transforms = det.transform_store();
+    const auto surfaces = det.surfaces();
+    std::map<traccc::geometry_id, traccc::transform3> maps;
```
niermann999, line 36: The ```traccc::geometry_id``` is just some hash right? Because the ```sf.barcode().value()``` is pretty much meaningless if not used in the barcode type. Using ```std::map<detray::barcode, traccc::transform3>``` would retain much more information, but I don't know how feasible it is at this point. I am guessing, that the ```geometry``` type is still something traccc specific that can't be changed?

core/include/traccc/edm/seed.hpp
```diff
@@ -24,6 +25,25 @@ struct seed {
     scalar weight;
     scalar z_vertex;
 
+    TRACCC_HOST_DEVICE
+    std::array<measurement_link, 3> get_measurements(
+        const spacepoint_collection_types::const_view& spacepoints_view,
+        const cell_module_collection_types::host& modules) const {
+        const spacepoint_collection_types::const_device spacepoints(
+            spacepoints_view);
+        alt_measurement alt_measB = spacepoints.at(spB_link).meas;
```
niermann999, line 34: Why does this have to be done via the ```alt_measurement``` type?

io/include/traccc/io/read_geometry.hpp
beomki-yeo: That's what this function is doing for upstream algorithms (clusterization and seeding).
But track fitting or finding uses `detray::detector`

io/include/traccc/io/read_geometry.hpp
beomki-yeo: Barcode is used in downstream edms (`track_candidate` or `track_state` for KF).
I have not just fixed `traccc::geometry` yet and it is not an urgent issue as far as I concern (geometry information is barely used in clusterization and seeding)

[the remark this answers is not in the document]
beomki-yeo: Don't remember. Maybe we need it for binary search later

tests/io/mock_data/event000000000-measurements.csv
beomki-yeo: Yeah I added this file because of the changes in `read_spacepoints`

io/include/traccc/io/read_geometry.hpp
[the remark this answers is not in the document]
beomki-yeo: Currently I think I am only using it in `read_spacepoints.cpp` for single lookups. So yeah I think it could be unordered map for my use case but it might be used differently in other places

beomki-yeo: BTW, we are recording geometry id of hits as `sf.barcode().value()` in `detray::event_writer` - should this be fixed as well?

core/include/traccc/edm/seed.hpp
beomki-yeo: BTW, I think we will remove current `measurement` and rename `alt_measurement` to `measurement`

io/include/traccc/io/read_geometry.hpp
niermann999: But we are reading it back into a barcode or not? I am also not saying it will break, there is just no strict guarantee about the ordering of the values (Although I believe it still holds by design). It should be possible to sort the values and get all surfaces for one volume in a range, I think, but let's not rely on this.

I am afraid that someone might read a detray barcode into an ACTS geometry ID and get weird garbage out of it.

beomki-yeo: I thought we can create barcode from geometry id. Could you check the following line?

https://github.com/acts-project/traccc/blob/3d8ddbc9182b12887d09d625e88b924ef17432a4/core/include/traccc/utils/seed_generator.hpp#L70

beomki-yeo: There is also a similar line in #430 :

https://github.com/beomki-yeo/traccc/blob/d71faf68cb318d1bce8878728cb2092563d5e37d/core/src/seeding/track_params_estimation.cpp#L41

niermann999: You can transform a detray barcode to a 64bit unsigned integer type and back to a barcode, yes. But it also possible to instantiate a barcode on any other integer value, e.g. one that was encoded by ACTS. Then the detray barcode will decode garbage from that value. As long as ```surface_link``` was written out by detray (e.g. to a csv file), the above lines are OK. But if we use different barcodes/geometryID/source links in the same code and we accidentally mix them up, we might get into trouble

niermann999: Hence, it would be best to save the barcodes as ```detray:::barcode``` and not as an integer

beomki-yeo: But how can I save the full information when simulation data only geometry ID (output of `barcode::value()`)?

niermann999: As I said, as long as the value originally came from a barcode its fine. But this values -> barcode conversion should only happen in the csv IO where we read it data from files that we know are generated by detray. I am more worried about this line https://github.com/acts-project/traccc/blob/3d8ddbc9182b12887d09d625e88b924ef17432a4/core/include/traccc/definitions/primitives.hpp#L26

core/include/traccc/edm/seed.hpp
guilhermeAlmeida1: that's the idea :) if you need any help with that you can @ me, but it should be pretty trivial

io/include/traccc/io/read_geometry.hpp
niermann999: Sounds good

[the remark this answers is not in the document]
niermann999: OK

examples/run/cpu/seeding_example.cpp
```diff
@@ -21,6 +22,11 @@
 #include "traccc/options/handle_argument_errors.hpp"
 #include "traccc/options/seeding_input_options.hpp"
 
+// Detray include(s).
+#include "detray/detectors/create_toy_geometry.hpp"
```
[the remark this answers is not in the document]
beomki-yeo: Fixed

examples/run/cuda/seeding_example_cuda.cpp
beomki-yeo: Fixed

examples/run/sycl/seeding_example_sycl.sycl
beomki-yeo: Fixed

io/src/csv/read_spacepoints.cpp
beomki-yeo: Good catch. I removed it

examples/run/cpu/seeding_example.cpp
```diff
@@ -21,6 +22,11 @@
...
+#include "detray/io/json/json_reader.hpp"
+#include "detray/io/json/json_writer.hpp"
```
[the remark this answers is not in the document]
beomki-yeo: Fixed

niermann999 (approved):

io/include/traccc/io/read_geometry.hpp
niermann999: I think, if possible, we should be able to look the transforms up directly in the detector, instead of copying them and  then searching for them

PR #423, opened 2023-06-21
Minor improvements

guilhermeAlmeida1:
* Remove redundant code in sparseCCL
* Fix examples to correctly print number of obtained work items (cells, spacepoints, etc.).

### DIFF
--- a/core/include/traccc/clusterization/detail/sparse_ccl.hpp
+++ b/core/include/traccc/clusterization/detail/sparse_ccl.hpp
@@ -111,27 +111,22 @@ TRACCC_HOST_DEVICE inline unsigned int sparse_ccl(
     for (unsigned int i = 0; i < n_cells; ++i) {
         L[i] = i;
         unsigned int ai = i;
-        if (i > 0) {
-            for (unsigned int j = start_j; j < i; ++j) {
-                if (is_adjacent(cells[i], cells[j])) {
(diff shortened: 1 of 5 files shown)


core/include/traccc/clusterization/detail/sparse_ccl.hpp
```diff
@@ -111,27 +111,22 @@ TRACCC_HOST_DEVICE inline unsigned int sparse_ccl(
...
-        unsigned int l = 0;
         if (L[i] == i) {
-            l = labels;
-            ++labels;
+            L[i] = labels++;
         } else {
-            l = L[L[i]];
+            L[i] = L[L[i]];
         }
-        L[i] = l;
```
beomki-yeo, line 134: hmm is this correct change?

beomki-yeo: Ah I think it is right.

L[i] will still be `l`which is before ++ operator, right?

guilhermeAlmeida1: yes, that's correct, `i++` returns the value of `i` before incrementing, contrarily to `++i`.
https://en.cppreference.com/w/cpp/language/operator_incdec

beomki-yeo (approved):

PR #424, opened 2023-06-21
Remove multiple cache in performance writer

beomki-yeo:
Seeding and finding performance writer that I added used to allow multiple caches for CPU and GPU performance; This PR removes that because I think it is a bit over-engineering.

Depends on #411


krasznaa (approved):
I'm all in favour of simplifications.

I never really dug into this code to fully understand how it works/worked, so as long as you and the CI are happy (and since the changes technically look okay), I'm happy for this to go in.

PR #428, opened 2023-06-27
Seeding with detray - Version2

beomki-yeo:


beomki-yeo:
Let me close this PR if people are OK with #421

PR #431, opened 2023-07-07
Alpaka: Spacepoint Binning Implementation

CrossR:
Hi All,

To extend on the ground work from @StewMH over in #300, here is a first pass for the spacepoint binning algorithm utilising the alpaka framework.

This isn't 100%, but close enough to get the code looked at. The main things I know aren't quite right are:

- There is I believe a race condition - If I run over 10 `ttbar_mu300` events from `data/tml_full`, everything is fine. If I run over the smaller and therefore quicker to process `ttbar_mu20`, I get occasional memory access errors.

- The second one is that right now, I haven't fully fleshed out the required defines / taken care around some of the copy methods etc, to make sure this can run on both CPU/GPU. It works fine when compiled with CUDA (`-Dalpaka_ACC_GPU_CUDA_ENABLE=ON`), but I've not done any testing on any other backends.

- I've not really attempted to optimise anything here, its just a straight copy based on the CUDA implementation.

All of these issues aren't ideal, but aren't really show stoppers for getting the code merged in hopefully.

I've got the full seeding algorithm 95% complete now, I just need to tidy up some bits around the alpaka model for shared memory, then I can get that PR'd too, though we thought smaller, more modular PRs would be appreciated. Additionally, the optimisation pass and ensuring the code runs across multiple backends is much easier when I have a final output that I can more easily compare to the CPU implementation, which is why I've delayed that until the full seeding algorithm is complete.

Hopefully this looks reasonable! Can also ping me on Mattermost (rycross) and perhaps once I've got the seeding algorithm complete, I can give a few slides on it / developer experience / performance etc.


device/alpaka/src/seeding/spacepoint_binning.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+            // Find the grid bin that the spacepoint belongs to.
+            const internal_spacepoint<spacepoint> isp(sp, globalThreadIdx,
+                                                      config.beamPos);
+            const std::size_t bin_index =
+                phi_axis.bin(isp.phi()) + phi_axis.bins() * z_axis.bin(isp.z());
+
+            // Add the spacepoint to the grid.
+            grid.bin(bin_index).push_back(std::move(isp));
+        }
```
CrossR, line 95: This I need to check again....I was originally getting errors around calling a `__host__` function from  `__host__ __device__` function here I believe, so I inlined it just to test some other parts. I believe I ended up with it fixed in my full seeding example, so I expect I was just doing something else wrong here.

beomki-yeo:
> There is I believe a race condition - If I run over 10 ttbar_mu300 events from data/tml_full, everything is fine. If I run over the smaller and therefore quicker to process ttbar_mu20, I get occasional memory access errors.

I think we need to fix this before merging the PR. you can copy-paste a few hits from `ttbar_mu20` into an empty file accumulatively and run the spacepoint binning until it breaks. or you can comment codes line-by-line until the error message is gone.

StewMH:
> > There is I believe a race condition - If I run over 10 ttbar_mu300 events from data/tml_full, everything is fine. If I run over the smaller and therefore quicker to process ttbar_mu20, I get occasional memory access errors.
> 
> I think we need to fix this before merging the PR. you can copy-paste a few hits from `ttbar_mu20` into an empty file accumulatively and run the spacepoint binning until it breaks. or you can comment codes line-by-line until the error message is gone.

It might be easier to fix with the code in the release, and it won't affect the CI since it doesn't run there.

device/alpaka/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2021-2022 CERN for the benefit of the ACTS project
```
beomki-yeo, line 3: Could you edit the license years of new and changed files

beomki-yeo (commented):
OK let's get this in after minor fix

beomki-yeo (approved):

beomki-yeo:
Once CI is passed, It is good to be merged

CrossR:
Looks like the failure is because I haven't got some guards in place for some CUDA bits in this version of the code.
I've got that fixed in a later test, so I'll bring those changes over here as well, rather than waiting to fix the building on CPUs.

PR #433, opened 2023-07-28
Update detray to v0.38.0

beomki-yeo:
UPDATE: PR now updates the detray version to v0.38.0

PR #434, opened 2023-08-02
Add `return` to Kalman actor abort

beomki-yeo:
Without `return`, `get_surface()` can be called when the track is not on a surface


niermann999 (approved):
I can't follow how this will call ```get_surface()```, but the change seems to make sense in any case

beomki-yeo:
Ah it will be called after #433 merged

PR #444, opened 2023-08-17
Measurement comparison operator only with detray barcode

beomki-yeo:
Let me see if this passes the CI.
I would need this for converting clusterization output (`measurement_collection`) to track finding input (`measurement_container`)


beomki-yeo:
Ah nvm - I guess I can put a custom operator instead of the predefined ones

PR #445, opened 2023-08-19
Make disabling TRACCC_BUILD_EXAMLPE option possible

beomki-yeo:
Currently `TRACCC_BUILD_EXAMPLE=False` makes an error because `traccc_options` in example directory is required for `traccc_test`.

I removed the test for `traccc_options` in traccc test to enable it


krasznaa (approved):
I know it's cheating, but as far as I'm concerned, this is fine. 😉

PR #446, opened 2023-08-19
Track finding with measurement collection input

beomki-yeo:
### UPDATE

Now the measurement collection is sorted in finding_algorithm.

And here is the benchmark result that compares with the `main` branch

This PR:
```
==> Statistics ... 
- read    106619 spacepoints from 26364 modules
- created  (cpu)  113677 seeds
- created (cuda)  113667 seeds
- created  (cpu) 117725 found tracks
- created (cuda) 117678 found tracks
- created  (cpu) 117725 fitted tracks
- created (cuda) 117678 fitted tracks
==>Elapsed times...
            Hit reading  (cpu)  2596 ms
                Seeding (cuda)  16 ms
                Seeding  (cpu)  596 ms
           Track params (cuda)  1 ms
           Track params  (cpu)  22 ms
 Track finding with CKF (cuda)  1640 ms
  Track finding with CKF (cpu)  16201 ms
  Track fitting with KF (cuda)  864 ms
   Track fitting with KF (cpu)  10309 ms
                     Wall time  32269 ms
```

`main`:
```
==> Statistics ... 
- read    106619 spacepoints from 26364 modules
- created  (cpu)  113677 seeds
- created (cuda)  113668 seeds
- created  (cpu) 117725 found tracks
- created (cuda) 117678 found tracks
- created  (cpu) 117725 fitted tracks
- created (cuda) 117678 fitted tracks
==>Elapsed times...
            Hit reading  (cpu)  2878 ms
                Seeding (cuda)  15 ms
                Seeding  (cpu)  602 ms
           Track params (cuda)  2 ms
           Track params  (cpu)  23 ms
 Track finding with CKF (cuda)  1671 ms
  Track finding with CKF (cpu)  17029 ms
  Track fitting with KF (cuda)  948 ms
   Track fitting with KF (cpu)  10730 ms
                     Wall time  33918 ms
```
---------------------------------

This PR refactors CKF with the input of measurement collection.

Please note that the measurements should be sorted based on the detray::barcode before being passed to the CKF algorithm.

I will check the performance difference later but I don't think it is going to make a dramatic change ... :p


core/include/traccc/finding/finding_algorithm.ipp
```diff
@@ -19,29 +19,48 @@ namespace traccc {
...
-
-    std::vector<std::pair<geometry_id, unsigned int>> module_map;
-    const int n_modules = measurements.size();
+    /*****************************************************************
+     * Measurement Operations
+     *****************************************************************/
+
+    // Sort the measurements w.r.t geometry barcode
+    std::sort(measurements.begin(), measurements.end(),
+              measurement_sort_comp());
```
krasznaa, line 33: Okay, so you move the measurement container to avoid making a copy for the sorting... :thinking:

I'm not a fan of this. But I see how one could argue that this could provide an optimisation in the cases where the measurement collection is not needed after the track finding anymore.

Is that the case by the way? Does track fitting not need the measurements anymore? Because if it does, then we're only making the user code around the algorithms more complicated than they need to be. Shifting the copy-making to the user of this algorithm, instead of the algorithm making the copy itself. :thinking:

beomki-yeo: I don't think we need measurement collection after track finding.
In that case, do you think it is better to make a copy of the measurement collection for sorting?

beomki-yeo: Just for update: I changed to use const reference and make a copy for sorted measurements

core/include/traccc/finding/finding_algorithm.ipp
```diff
@@ -20,28 +20,55 @@ template <typename stepper_t, typename navigator_t>
...
-    std::vector<std::pair<geometry_id, unsigned int>> module_map;
-    const int n_modules = measurements.size();
+    /*****************************************************************
+     * Measurement Operations
+     *****************************************************************/
+
+    // Copy the measurements
+    measurement_collection_types::host sorted_measurements;
+    std::copy(measurements.begin(), measurements.end(),
+              std::back_inserter(sorted_measurements));
```
krasznaa, line 35: I think this is perfectly valid, but to be honest I would've just gone with the much simpler:

```c++
measurement_collection_types::host sorted_measurements = 
    measurements;
```

:wink:

krasznaa (approved):
Let's do it like this for now. :thinking: But we should try to remember to make it possible in the future to get rid of the copy in case we can really afford that. I'm just still not sure what would be the most intuitive API for that. :frowning:

core/include/traccc/finding/finding_algorithm.ipp
beomki-yeo: Makes sense

beomki-yeo:
I guess getting const reference is a lot better for users as long as it doesn't take too much time for copy. Anyway it should be much cheaper than sort function

beomki-yeo:
Just update on the benchmarking result.
I just realized that the nubmer of tracks after CUDA CKF is different. Maybe because the output of sort function is not determinisitc? 

This PR:

==> Statistics ... 
- read    106829 spacepoints from 26487 modules
- created  (cpu)  114031 seeds
- created (cuda)  114027 seeds
- created  (cpu) 117480 found tracks
- created (cuda) 117492 found tracks
- created  (cpu) 117480 fitted tracks
- created (cuda) 117492 fitted tracks
==>Elapsed times...
            Hit reading  (cpu)  2650 ms
                Seeding (cuda)  16 ms
                Seeding  (cpu)  605 ms
           Track params (cuda)  2 ms
           Track params  (cpu)  23 ms
 Track finding with CKF (cuda)  1741 ms
  Track finding with CKF (cpu)  18467 ms
  Track fitting with KF (cuda)  1117 ms
   Track fitting with KF (cpu)  11034 ms
                     Wall time  35678 ms

Main branch:

==> Statistics ... 
- read    106829 spacepoints from 26487 modules
- created  (cpu)  114031 seeds
- created (cuda)  114026 seeds
- created  (cpu) 117480 found tracks
- created (cuda) 117493 found tracks
- created  (cpu) 117480 fitted tracks
- created (cuda) 117493 fitted tracks
==>Elapsed times...
            Hit reading  (cpu)  2957 ms
                Seeding (cuda)  15 ms
                Seeding  (cpu)  604 ms
           Track params (cuda)  1 ms
           Track params  (cpu)  23 ms
 Track finding with CKF (cuda)  1679 ms
  Track finding with CKF (cpu)  18586 ms
  Track fitting with KF (cuda)  1213 ms
   Track fitting with KF (cpu)  11209 ms
                     Wall time  36308 ms

PR #447, opened 2023-08-22
Remove module map in track param estimation

beomki-yeo:
PR depends on #446 

As the surface link is already the member of measurements we don't have to pass the module map to track parameter estimation


krasznaa (approved):
Seems simple enough. I trust that the logic is correct as well. :wink: (I have to admit I'm not sure I fully understand the information flow in our EDM at the moment. :frowning: Will play with the code myself as well soon, hopefully things will be more clear then. :wink:)

PR #448, opened 2023-08-23
(Experimental) Decoupling Spacepoint Formation from Clusterization

beomki-yeo:
Depends on #446 and #447.

There are two main points:
1. Spacepoint formatioin is decoupled from the clusterization: The reason for decoupling is that `spacepoint` is only used for seeding, and what we need from clusterization is measurements. 
2. The decoupled spacepoint formation is written in `experimental` directory: The new spacepoint formation is not compatible with the original codes as `detray::detector` is involved. 

In another PR, the clsuterization will also be rewritten in `experimental` directory to make the output type measurements

### DIFF
--- a/device/common/CMakeLists.txt
+++ b/device/common/CMakeLists.txt
@@ -39,6 +39,8 @@ traccc_add_library( traccc_device_common device_common TYPE SHARED
    "include/traccc/seeding/device/populate_grid.hpp"
    "include/traccc/seeding/device/impl/populate_grid.ipp"
    # Seed finding function(s).
+   "include/traccc/seeding/device/experimental/form_spacepoints.hpp"
+   "include/traccc/seeding/device/experimental/impl/form_spacepoints.ipp"
    "include/traccc/seeding/device/count_doublets.hpp"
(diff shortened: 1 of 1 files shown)


krasznaa:
The "Experimental" name is a bit misleading, since we absolutely need to go ahead with this. So let me rebase it, so that #446 and #447 wouldn't show up in it anymore...

krasznaa (approved):
I fear that we'll need to come up with a better way for propagating the detector template type to the device code later on, but for now my only request is to add `form_spacepoints.hpp` and `form_spacepoints.ipp` to `device/common/CMakeLists.txt`. (Technically it's not needed for a successful build, it just helps some IDEs with viewing the project.)

krasznaa:
Sorry, I should've advertised the rebase. 😦 In the future, you could do this more elegantly like: https://stackoverflow.com/questions/1125968/how-do-i-force-git-pull-to-overwrite-local-files

krasznaa (approved):
As long as it compiles, let's go ahead with it like this for now.

beomki-yeo:
Sorry that I don't understand clearly. Do you mean squashing into a single commit?

PR #451, opened 2023-08-31
Alpaka: Seed Finding

CrossR:
This is a follow up to #431, which extends the Alpaka implementation to include seed finding, making the `seeding_example` a touch more complete (though it is missing the track finding / fitting, as was added to the CUDA example recently).

This should now compile under both a GPU and CPU based Alpaka build, and I believe the race condition present in the previous code is fixed now (though I'll know for sure once I get onto the full single and multithreaded examples that I can run for more than 10/100 events). I use the following build commands:

```
# CUDA-based build
cmake -DTRACCC_BUILD_CUDA=ON -DTRACCC_BUILD_ALPAKA=ON -Dalpaka_ACC_GPU_CUDA_ENABLE=ON -B . -S .. -G Ninja

# CPU-based build
cmake -DTRACCC_BUILD_ALPAKA=ON -B . -S .. -G Ninja
```

I've not had an environment to test HIP based builds...so I've not been able to test that (though I would like to!). The Alpaka threaded backend I believe won't work right now, though that is also something I'm working to enable (well, working to understand if its worth enabling, or if CPU vs GPU is enough).

Alpaka CUDA vs real CUDA is fairly competitive, but Alpaka CPU vs CPU is fairly awful (not that its too surprising that CPU with framework overhead vs CPU without framework overhead has a giant performance delta, especially when the Alpaka with CPU backend is going through algorithms optimised for GPUs).

The tests I've been running for this work have just been

```
./bin/traccc_seeding_example_{cuda/alpaka} --detector_file=tml_detector/trackml-detector.csv --input_directory=tml_full/ttbar_mu{20/100/200/300} --events=10 --run_cpu=1
```

I've continued to work on the full chain, so hopefully I can get that finished, then go back and add the bits like track fitting/finding that have been added since I started working on this.


device/alpaka/CMakeLists.txt
```diff
@@ -15,16 +15,27 @@ if(alpaka_ACC_GPU_CUDA_ENABLE)
 
   set_source_files_properties(src/utils/make_prefix_sum_buff.cpp PROPERTIES LANGUAGE CUDA)
   set_source_files_properties(src/seeding/spacepoint_binning.cpp PROPERTIES LANGUAGE CUDA)
+
+  set_source_files_properties(src/seeding/track_params_estimation.cpp PROPERTIES LANGUAGE CUDA)
+  set_source_files_properties(src/seeding/seed_finding.cpp PROPERTIES LANGUAGE CUDA)
+  set_source_files_properties(src/seeding/seeding_algorithm.cpp PROPERTIES LANGUAGE CUDA)
```
krasznaa, line 21: As long as we have a number of files here, it would be a bit more compact to write this as:

```cmake
  set_source_files_properties(
    src/utils/make_prefix_sum_buff.cpp
    src/seeding/spacepoint_binning.cpp
    src/seeding/track_params_estimation.cpp
    src/seeding/seed_finding.cpp
    src/seeding/seeding_algorithm.cpp
    PROPERTIES LANGUAGE CUDA)
```

device/alpaka/include/traccc/alpaka/utils/definitions.hpp
```diff
@@ -11,15 +11,42 @@
...
-#include "vecmem/utils/copy.hpp"
+#include <vecmem/utils/copy.hpp>
 
 namespace traccc::alpaka {
 
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+#define WARP_SIZE 32
+#else
+#define WARP_SIZE 4
+#endif
```
krasznaa, line 25: What is this? :confused: I don't see the `WARP_SIZE` macro used anywhere else in this PR. :confused:

Is `4` something that you would expect that a CPU would have as number of threads? Or where does that value come from?

Even if we decide to make this choice at compile time, which I'm not convinced that we should (it would be better to make this decision at runtime if possible), we should not use a pre-processor macro for this. Rather, I would do something like:

```c++
static constexpr std::size_t WARP_SIZE =
#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
    32
#else
    4
#endif
    ;
```

device/alpaka/include/traccc/alpaka/utils/definitions.hpp
```diff
@@ -11,15 +11,42 @@
...
+template <typename TAcc>
+inline WorkDiv makeWorkDiv(Idx blocksPerGrid,
+                           Idx threadsPerBlockOrElementsPerThread) {
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+    if constexpr (std::is_same_v<TAcc, ::alpaka::AccGpuCudaRt<Dim, Idx>>) {
+        const auto elementsPerThread = Idx{1};
+        return WorkDiv{blocksPerGrid, threadsPerBlockOrElementsPerThread,
+                       elementsPerThread};
+    } else
+#endif
```
krasznaa, line 44: Interesting. Is this piece of code non-functional when CUDA is not being used? I would've hoped that the Alpaka code around CUDA could still be referred to, even if CUDA was "not active". :thinking: Since the whole point of Alpaka in my mind would be to do as little pre-processor magic on our side as possible.

device/alpaka/src/seeding/track_params_estimation.cpp
```diff
@@ -0,0 +1,77 @@
...
+        const vector3 bfield,
+        const std::array<traccc::scalar, traccc::e_bound_size> stddev,
+        bound_track_parameters_collection_types::view params_view) const {
+        auto const globalThreadIdx =
+            ::alpaka::getIdx<::alpaka::Grid, ::alpaka::Threads>(acc)[0u];
+        device::estimate_track_params(globalThreadIdx, spacepoints_view,
+                                      seed_view, modules_view, bfield, stddev,
+                                      params_view);
+    }
+};
```
krasznaa, line 34: It is **not** for this PR, but I wonder... Do we need to define these separate structs? Is Alpaka not able to use lambdas instead? (I genuinely don't know...)

device/common/include/traccc/seeding/device/impl/populate_grid.ipp
```diff
@@ -12,7 +12,7 @@
 
 namespace traccc::device {
 
-TRACCC_DEVICE
```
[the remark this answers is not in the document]
krasznaa: Sure, I don't mind this...

Having Alpaka build this code for the host using `nvcc` will not result in efficient code, but I see how logically it's good to let it happen.

krasznaa (changes requested):
I'm quite happy with this. :wink:

I'll try to run some tests before merging it in, but right now I only have some surface level comments. :smile:

device/alpaka/src/seeding/track_params_estimation.cpp
CrossR: There is some nuance with Kernel Definitions: https://alpaka.readthedocs.io/en/latest/basic/library.html#kernel-definition

For that reason, and because it more closely mimics the layout of the CUDA code, I went with the struct layout for now.
I'm also not sure how the later dynamic shared memory (`extern __shared__`) equivalent would work, which is defined using traits, with an example in the `seed_finding.cpp` code (though

device/alpaka/include/traccc/alpaka/utils/definitions.hpp
krasznaa: Work group sizes are hardcoded in a number of places at the moment. That is generally not a terrible idea, since the ideal work group size is a bit more dependent on the algorithm than on the underlying hardware. (Though it's not a black or white distinction.)

But when it comes to figuring out how many threads would be running in parallel in a work group, so the actual "warp size", there we should avoid hardcoding. For consumer GPUs, and all NVIDIA GPUs, this number is `32` of course. But for instance AMD data center GPUs have a "wavefront width" of 64. (Which, if we don't account for it, can mean a 2x performance loss for our code right away.)

Also keep in mind that CPUs (at least through OpenCL) report their capabilities like:

```
  Platform Name                                   Intel(R) OpenCL
Number of devices                                 1
  Device Name                                     Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
  Device Vendor                                   Intel(R) Corporation
  Device Vendor ID                                0x8086
  Device Version                                  OpenCL 3.0 (Build 0)
  Driver Version                                  2023.15.3.0.20_160000
  Device OpenCL C Version                         OpenCL C 3.0 
  Device Type                                     CPU
  Device Profile                                  FULL_PROFILE
  Device Available                                Yes
  Compiler Available                              Yes
  Linker Available                                Yes
  Max compute units                               16
  Max clock frequency                             3600MHz
  Device Partition                                (core)
    Max number of sub-devices                     16
    Supported partition types                     by counts, equally, by names (Intel)
    Supported affinity domains                    (n/a)
  Max work item dimensions                        3
  Max work item sizes                             8192x8192x8192
  Max work group size                             8192
  Preferred work group size multiple              128
  Max sub-groups per work group                   2048
  Sub-group sizes (Intel)                         4, 8, 16, 32, 64
  Preferred / native vector sizes                 
    char                                                 1 / 32      
    short                                                1 / 16      
    int                                                  1 / 8       
    long                                                 1 / 4       
    half                                                 0 / 0        (n/a)
    float                                                1 / 8       
    double                                               1 / 4        (cl_khr_fp64)
  Half-precision Floating-point support           (n/a)
...
```

I.e. this 8 core / 16 thread CPU will happily tell you to use up to `8192` threads per block if you wish. :stuck_out_tongue: So we'll need to think a bit about how we'd want to run these algorithms on CPUs exactly. (The final answer will probably just be that we wouldn't... :stuck_out_tongue:)

device/alpaka/src/seeding/track_params_estimation.cpp
krasznaa: Okay, let's just go with the structs for now. :wink:

device/alpaka/CMakeLists.txt
CrossR: For now, I've swapped this to utilise the alpaka CMake functions: 

https://github.com/alpaka-group/alpaka/blob/develop/cmake/addExecutable.cmake

https://github.com/alpaka-group/alpaka/blob/develop/cmake/addLibrary.cmake

I'm not necessarily 100% sure with that decision though.

It does mean we remove this annoying block of CMake (made more annoying when you consider that there also needs to be an equivalent HIP block), and it guarantees that all the relevant Alpaka CMake flags are kept and applied....

But it duplicates the `traccc-functions` code, which does not seem ideal, but I'm not necessarily sure what the best other approach is? I guess I could add a smaller Alpaka specific CMake file that basically duplicates the functionality of the alpaka funcs, that we call in our CMake, around the existing `traccc_add_executable` etc.

extern/alpaka/cmake/alpaka-functions.cmake
```diff
@@ -0,0 +1,127 @@
...
+   cmake_parse_arguments( ARG "" "" "LINK_LIBRARIES" ${ARGN} )
+
+   # Set up the executable.
+   set( exe_name "traccc_${name}" )
+   add_executable( ${exe_name} ${ARG_UNPARSED_ARGUMENTS} )
+   if( ARG_LINK_LIBRARIES )
+      target_link_libraries( ${exe_name} PRIVATE ${ARG_LINK_LIBRARIES} )
+   endif()
+
+   target_compile_definitions(${exe_name} PRIVATE ALPAKA_HOST_ONLY)
```
CrossR, line 89: This is the key line in this function.

Without it, the example code will go via whatever compiler makes sense for the current accelerator (i.e. it will go via `nvcc` when you are using CUDA).

That isn't what we want: the examples (and I think all executables, but it is something to consider going forward), should be compiled as regular CXX files, and the device code should all live in the relevant `traccc::XXX` library.

I somehow got away with that in the old code, and this code until recently, but it looks like the recent detray inclusion (and all the templating code in the detray functions) causes `nvcc` to fall over.

This function makes alpaka fine to compile as a normal C++ executable, whilst all the actual library code that we want accelerated is compiled with `nvcc`, as expected, over in `traccc::alpaka`.

(Without it Alpaka will complain that you are using a C++ compiler, and not `nvcc` etc. This is because, well I think because, alpaka is a header only lib, so you end up pulling in all the alpaka headers very quickly, then you hit warnings over in those headers due to the combination of the current compiler + current enabled backend. With this flag, the warnings are not enabled, and the code builds as you'd expect).

device/alpaka/include/traccc/alpaka/utils/definitions.hpp
CrossR: This is certainly something myself and Stewart are talking about, as this situation gets worse when you start adding in `if CUDA if HIP else` etc...

https://github.com/cms-sw/cmssw/blob/70c7fb8ed2ff22abb8a635c438fa7dd2313ebb3b/HeterogeneousCore/AlpakaInterface/interface/workdivision.h#L21-L76

Gives an idea of how they deal with this in CMSSW, but I think the alpaka recommended way does use some small amount of pre-processor defines, when you get into code for a specific accelerator, at least if you want to use code that can only work if you are using the compiler for that accelerator.

Examples on the alpaka side for that:

Define-based : https://github.com/alpaka-group/alpaka/blob/develop/example/kernelSpecialization/src/kernelSpecialization.cpp

Tag Based : https://github.com/alpaka-group/alpaka/blob/develop/example/tagSpecialization/src/tagSpecialization.cpp

CrossR:
I think with the recent changes to make the CI happy, this PR is back up to date, post recent changes to the traccc repo.

Testing this branch in isolation is a bit of pain though, at least for me...Due to the changes to the examples and `data/` files, I can't easily test just this code.

That said, I've got an almost ready branch that includes some further algorithms and the throughput examples, which I have tested, and it based on top of this branch.

I'm not sure if we want to block this until we can test just this branch, or if its worth getting this compiling but not tested separately code in, and then quickly issuing the follow up PR for the later work. (The only real thing I need to do over there is run formatting / remove unused vars).

Once this and the follow up PR are in, I'm going to work with Stewart to get a full chain test done with HIP, as well as working through any bugs / ugly features of the current Alpaka code (like the `vecmem::` code found in the examples needed `#ifdef`s to support different devices).

beomki-yeo (approved):
Probably more experts like @krasznaa needs to look into the PR again but It looks good already

device/alpaka/CMakeLists.txt
```diff
@@ -7,37 +7,31 @@
...
+
 if(alpaka_ACC_GPU_CUDA_ENABLE)
   enable_language(CUDA)
-  find_package( CUDAToolkit REQUIRED )
-
   include( traccc-compiler-options-cuda )
 
-  set_source_files_properties(src/utils/make_prefix_sum_buff.cpp PROPERTIES LANGUAGE CUDA)
-  set_source_files_properties(src/seeding/spacepoint_binning.cpp PROPERTIES LANGUAGE CUDA)
-elseif(alpaka_ACC_GPU_HIP_ENABLE)
```
StewMH, line 18: Will have to put some HIP logic back in but easier to get this merged first

CrossR: Yeah, https://github.com/alpaka-group/alpaka/blob/develop/cmake/addLibrary.cmake deals with some parts, but its possible we will need some extra bits here.

I'm not actually sure how needed the `enable_language(CUDA)` bit is, since that is also in the Alpaka wrapper. Its possible I needed it previously, but since fixing the issue with the examples going via `nvcc`, it may not be needed now.

extern/alpaka/cmake/alpaka-functions.cmake
krasznaa: This then means that `traccc::alpaka` exposes Alpaka in its public interface. :frowning: We need to fix that. Neither `traccc::cuda`, nor `traccc::sycl` expose the clients of those libraries to CUDA and SYCL headers respectively. :thinking: `traccc::alpaka` must also be able to provide an interface that doesn't include any Alpaka headers...

device/alpaka/include/traccc/alpaka/utils/definitions.hpp
```diff
@@ -11,15 +11,40 @@
 #include <alpaka/example/ExampleDefaultAcc.hpp>
```
krasznaa, line 11: I guess this is the smoking gun... :thinking: Alpaka headers should not be included in the public headers of this library. This `definitions.hpp` file could just be moved under `device/alpaka/src/`, right?

krasznaa (changes requested):
The `traccc::alpaka` library's public interface should be cleaned a little, before we merge this in. Also, the branch needs to be updated first anyway.

krasznaa:
Ahh, forgot to add... I would also rename `alpaka-functions.cmake` to `traccc-alpaka-functions.cmake`. To avoid possible name clashes. :thinking:

device/alpaka/include/traccc/alpaka/utils/definitions.hpp
CrossR: Looking over my code, I was closely mimicing the structure of the CUDA code, which has the `#include <cuda.h>` etc in its `include/utils/definitions.hpp`, but they are obviously causing a lot more issues here due to the header-only nature of Alpaka.

I've refactored out basically all of this file into the `src/` folder, leaving only the warpSize definition and the non-alpaka definitions. That seems to have made things much happier, so I've also dropped the `alpaka_test/executable` CMake functions (as we seemingly don't currently need them).

extern/alpaka/CMakeLists.txt
```diff
@@ -44,3 +44,11 @@ if( DEFINED _buildTestingValue )
...
    unset( BUILD_TESTING CACHE )
 endif()
+
+# Include the alpaka-functions.cmake file.
+# This has specific versions of the traccc-functions, with the add_library
+# and add_executable functions swapped to the alpaka versions.
+#
+# These versions ensure Alpaka flags are passed over, and the relevant
+# accelerator backends are enabled, including file properties.
+include( ${CMAKE_CURRENT_LIST_DIR}/cmake/alpaka-functions.cmake )
```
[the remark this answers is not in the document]
CrossR: Sounds reasonable!

I'm having a touch of trouble getting `device/alpaka/CMakeLists.txt` to pickup `traccc-alpaka-functions` though, which may just be that I'm missing something obvious. For now (to test the rest of the changes for your comments), I've updated the CMAKE Path in the main CMake file to include this path, as without it the path only includes the main CMAKE folder and the `_deps` folders in the build output.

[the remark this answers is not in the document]
krasznaa: Very strange. 😕 This code is meant to take care of that:

https://github.com/acts-project/traccc/blob/main/CMakeLists.txt#L39-L41

That should make the `cmake/` directory visible to `include(...)` commands from there on out. Just as the `include(traccc-functions)` call demonstrates. 😕 

My best bet is a typo. 🤔 Since as long as

https://github.com/acts-project/traccc/blob/main/device/alpaka/CMakeLists.txt#L8

works correctly, then

```cmake
include( traccc-alpaka-functions )
```

on the next line should work as well... 🤔

CMakeLists.txt
```diff
@@ -165,6 +165,9 @@ if( TRACCC_SETUP_ALPAKA )
    else()
       add_subdirectory( extern/alpaka )
    endif()
+
+   # Include the Alpaka CMake code.
+   list( APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/extern/alpaka/cmake" )
```
krasznaa, line 170: Wait, wait, wait...

The `traccc-alpaka-functions.cmake` file should be put under `cmake/`. Along with all of our other such CMake files. No reason to separate it like this. 🤔

And as long as you put it under `cmake/`, you also don't need to do anything to make it visible to any sub-directory of the project.

device/alpaka/CMakeLists.txt
```diff
@@ -5,39 +5,35 @@
 # Mozilla Public License Version 2.0
 
 # Project include(s).
+message( STATUS "CMAKE_MODULE_PATH: ${CMAKE_MODULE_PATH}" )
```
krasznaa, line 8: Don't forget to remove the debugging message.

device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * (c) 2023 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s)
+#include "traccc/alpaka/utils/definitions.hpp"
```
krasznaa, line 11: Do you actually need these definitions in the header file? Don't you only need them in the `.cpp` file?

This is what I meant earlier when pondering whether it would be possible to move `definitions.hpp` under `src/`. Since I don't think you need its definitions in the public interface of any of the classes / algorithms.

device/alpaka/src/utils/utils.hpp
```diff
@@ -0,0 +1,37 @@
+/** TRACCC library, part of the ACTS project (R&D line)
```
krasznaa, line 1: Ahh, okay. I just see this new file. Yes, I like it. But is there a reason for keeping `definitions.hpp` alive? Why not move everything (needed) in here from there?

device/cuda/include/traccc/cuda/seeding/track_params_estimation.hpp
```diff
@@ -43,7 +43,7 @@ struct track_params_estimation
     track_params_estimation(const traccc::memory_resource& mr,
                             vecmem::copy& copy, stream& str);
 
-    /// Callable operator for track_params_esitmation
+    /// Callable operator for track_params_estimation
```
krasznaa, line 46: What's going on here? 😕 I can't understand what actual change happened here... 😕

krasznaa (changes requested):
I'm still very much in favour of this. Just after some more annoying requests...

device/cuda/include/traccc/cuda/seeding/track_params_estimation.hpp
StewMH: Is it not just the typo fix (esitmation -> estimation)?

device/alpaka/src/utils/utils.hpp
CrossR: Done this now....

The only reason I'd kept `definitions.hpp` around was as a top level place to define the various includes.
I think really, we will need to think up a sensible / clean way of doing that somewhere though.

(What I mean is conditional includes to include say `vecmem/copy`, but also `vecmem/cuda/copy` when compiled to a CUDA target, or `vecmem/hip/copy` for a HIP target and so on. Right now, the includes and any guards are duplicated in a few different places, which isn't ideal.)

device/cuda/include/traccc/cuda/seeding/track_params_estimation.hpp
CrossR: Apologies, yes this is just a typo fix, I must have fixed this when I was poking around the files (I have the spell-checker in vim bound, so I must have just activated it without realising at some point).

Can undo if needed.

krasznaa: No, this is absolutely fine. It's good that you fixed the typo. I'm just old and blind by now, and couldn't see what the change was. 😛

krasznaa (approved):
Let's finally go ahead with this one...

PR #453, opened 2023-09-06
Remove duplicate detector file option

stephenswat:
This commit removes the duplicately defined detector file option from the full tracking options, so that the examples no longer fail to run.

### DIFF
--- a/examples/options/src/options/full_tracking_input_options.cpp
+++ b/examples/options/src/options/full_tracking_input_options.cpp
@@ -10,15 +10,11 @@
 
 traccc::full_tracking_input_config::full_tracking_input_config(
     po::options_description& desc) {
-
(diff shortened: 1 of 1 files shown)


examples/options/src/options/full_tracking_input_options.cpp
```diff
@@ -10,9 +10,6 @@
 
 traccc::full_tracking_input_config::full_tracking_input_config(
     po::options_description& desc) {
-
-    desc.add_options()("detector_file", po::value<std::string>()->required(),
-                       "specify detector file");
```
beomki-yeo, line 15: I think line 19 also need to be removed?

stephenswat: Good point, done.

beomki-yeo (approved):

PR #455, opened 2023-09-08
Compiler Option Fixes, main branch (2023.09.08.)

krasznaa:
As I mentioned in #454, I found that our builds are not quite as "rigorous" as I thought they would be. "Extended warning flags" were not used until now in many places unfortunately.

After fixing the CMake code to apply all the warning flags correctly, I also had to fix a few of the warnings that showed up as a result. In all cases about variables shadowing each other. (Kokkos had some other warnings as well, but those I silenced by making our build consider the Kokkos headers as "system headers", as it should.)

### DIFF
--- a/cmake/traccc-compiler-options-sycl.cmake
+++ b/cmake/traccc-compiler-options-sycl.cmake
@@ -7,6 +7,12 @@
 # Include the helper function(s).
 include( traccc-functions )
 
+# Only tweak the flags for the Intel compiler.
+if( NOT ( ( "${CMAKE_SYCL_COMPILER_ID}" STREQUAL "IntelLLVM" ) OR
+          ( "${CMAKE_SYCL_COMPILER_ID}" MATCHES "Clang" ) ) )
+   return()
+endif()
+
 # Basic flags for all build modes.
 foreach( mode RELEASE RELWITHDEBINFO MINSIZEREL DEBUG )
    traccc_add_flag( CMAKE_SYCL_FLAGS_${mode} "-Wall" )
--- a/performance/src/performance/timer.cpp
+++ b/performance/src/performance/timer.cpp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2022 CERN for the benefit of the ACTS project
+ * (c) 2022-2023 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -39,8 +39,8 @@ timer::~timer() {
     const std::chrono::nanoseconds totalTime = end - m_start;
     const auto pos =
         std::find_if(m_timing_info.data.begin(), m_timing_info.data.end(),
-                     [&m_name = m_name](const timing_info_pair& element) {
-                         return element.first == m_name;
+                     [&name = m_name](const timing_info_pair& element) {
+                         return element.first == name;
                      });
 
     if (pos == m_timing_info.data.end()) {
--- a/performance/src/performance/timing_info.cpp
+++ b/performance/src/performance/timing_info.cpp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2022 CERN for the benefit of the ACTS project
+ * (c) 2022-2023 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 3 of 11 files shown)


performance/src/performance/timing_info.cpp
```diff
@@ -19,9 +19,10 @@ namespace traccc::performance {
 std::chrono::nanoseconds timing_info::get_time(
     std::string_view timer_name) const {
 
-    auto it = std::find_if(
-        data.begin(), data.end(),
-        [&timer_name](timing_info_pair it) { return it.first == timer_name; });
+    auto it = std::find_if(data.begin(), data.end(),
+                           [&timer_name](timing_info_pair itr) {
+                               return itr.first == timer_name;
+                           });
```
stephenswat, line 25: Not a problem with this PR per se, but why isn't this just `std::find`? 🤨

performance/src/performance/timer.cpp
```diff
@@ -39,8 +39,8 @@ timer::~timer() {
     const std::chrono::nanoseconds totalTime = end - m_start;
     const auto pos =
         std::find_if(m_timing_info.data.begin(), m_timing_info.data.end(),
-                     [&m_name = m_name](const timing_info_pair& element) {
-                         return element.first == m_name;
+                     [&name = m_name](const timing_info_pair& element) {
+                         return element.first == name;
```
stephenswat, line 43:
```suggestion
                     [&m_name](const timing_info_pair& element) {
                         return element.first == m_name;
```

stephenswat: Also should be `std::find` (see below), but no need to fix that in this PR.

cmake/traccc-compiler-options-sycl.cmake
```diff
@@ -7,6 +7,11 @@
 # Include the helper function(s).
 include( traccc-functions )
 
+# Only tweak the flags for the Intel compiler.
+if( NOT "${CMAKE_SYCL_COMPILER_ID}" STREQUAL "IntelLLVM" )
+   return()
+endif()
```
stephenswat, line 13: Going to leave it up to you to decide, but I would mildly prefer inverting this pattern, e.g. putting the options inside a non-negated `if( "${CMAKE_SYCL_COMPILER_ID}" STREQUAL "IntelLLVM" )` block.

stephenswat (approved):

performance/src/performance/timing_info.cpp
krasznaa: I don't understand. :frowning: How should one do this with [std::find](https://en.cppreference.com/w/cpp/algorithm/find)? The elements of `data` are pairs. We want to find the first element whose `first` variable/element is equal to `timer_name`.

I'm pretty sure that `std::find_if` **is** the way to do this in the most compact way. :thinking: Or are you proposing changing the type of `traccc::performance::timing_info_pair` to `std::map<std::string, std::chrono::nanoseconds>`? :thinking: That would probably not be a bad idea indeed. (Since we don't allow multiple uses of the same "name" anyway.)

But that I would not want to do in this PR.

cmake/traccc-compiler-options-sycl.cmake
krasznaa: My thinking was that I'd change to that once/if we start defining specialized flags for other SYCL compilers as well. Until then I would keep this formalism. :thinking:

performance/src/performance/timer.cpp
krasznaa: I had to try this out to remind myself why the code was written in such a weird way to begin with. Unfortunately capture by reference does not like to capture member variables all that much.

```
/data/ssd-1tb/projects/traccc/traccc/performance/src/performance/timer.cpp:42:24: error: 'm_name' in capture list does not name a variable
                     [&m_name](const timing_info_pair& element) {
                       ^
/data/ssd-1tb/projects/traccc/traccc/performance/src/performance/timer.cpp:43:50: error: 'this' cannot be implicitly captured in this context
                         return element.first == m_name;
                                                 ^
/data/ssd-1tb/projects/traccc/traccc/performance/src/performance/timer.cpp:42:30: note: explicitly capture 'this'
                     [&m_name](const timing_info_pair& element) {
                             ^
                             this
```

If I remember correctly, @Xola101 was capturing `this` in the lambda at first. Which I myself didn't like. That's how we ended up with `[&m_name = m_name]`. :thinking:

All in all, I'd rather leave the code as I have it now...

PR #456, opened 2023-09-09
Add measurement dimension and its subspace

beomki-yeo:
The PR adds two measurement properties which are also implemented in ACTS [Measurement](https://github.com/acts-project/acts/blob/main/Core/include/Acts/EventData/Measurement.hpp) class:

1. Dimension of measurement (ex 1 for strip, 2 for pixel). A different part is that measurement dimension is a run time object in this PR
2. Subspace for projection matrix - it's complicated to explain but it is essential for annulus endcap strip

A PR for csv I/O of those variables will follow as well


core/include/traccc/utils/subspace.hpp
```diff
@@ -0,0 +1,119 @@
...
+            m_axes[i] = static_cast<size_type>(indices[i]);
+        }
+    }
+
+    /// Size of the subspace.
+    TRACCC_HOST_DEVICE
+    static constexpr size_type size() { return kSize; }
+    /// Size of the full vector space.
+    TRACCC_HOST_DEVICE
+    static constexpr size_type fullSize() { return kFullSize; }
```
krasznaa, line 63: Interesting... This is very "modern C\+\+", but in the olden days we would've simply done this with:

```c++
static const size_type size = kSize;
static const size_type fullSize = kFullSize;
```

Well, maybe now using `constexpr` instead of `const`. But mainly, defining these as functions seems to be a bit of an overkill.

krasznaa (commented):
Just to stir the pot a bit. :imp:

core/include/traccc/utils/subspace.hpp
beomki-yeo: Fixed

core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
```diff
@@ -44,67 +44,117 @@ struct gain_matrix_smoother {
...
+
+    template <size_type D>
+    struct gain_matrix_smoother_helper<D, typename std::enable_if_t<D == 0u>> {
+
+        TRACCC_HOST_DEVICE inline void operator()(
+            track_state<algebra_t>& /*cur_state*/,
+            const track_state<algebra_t>& /*next_state*/) const {
+            return;
+        }
+    };
```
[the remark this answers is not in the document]
beomki-yeo: The suggested change doesn't seem to work with the specilation of `0u`:

```
In file included from /home/beomki/projects/traccc/traccc/core/include/traccc/fitting/kalman_filter/kalman_actor.hpp:13,
                 from /home/beomki/projects/traccc/traccc/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp:17,
                 from /home/beomki/projects/traccc/traccc/tests/common/tests/kalman_fitting_test.hpp:12,
                 from /home/beomki/projects/traccc/traccc/tests/common/tests/kalman_fitting_test.cpp:16:
/home/beomki/projects/traccc/traccc/core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp:142:15: error: explicit specialization in non-namespace scope ‘struct traccc::gain_matrix_updater<algebra_t>’
  142 |     template <>
      |               ^
/home/beomki/projects/traccc/traccc/core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp:143:12: error: template parameters not deducible in partial specialization:
  143 |     struct gain_matrix_updater_helper<0u> {
```

[There is a way to bypass the error](https://stackoverflow.com/questions/3052579/explicit-specialization-in-non-namespace-scope) but it already makes things too complicated.. :p

core/include/traccc/edm/track_state.hpp
```diff
@@ -71,20 +71,34 @@ struct track_state {
...
+        static_assert(((D == 1u) || (D == 2u)),
+                      "The measurement dimension should be 1 or 2");
+
+        matrix_type<D, 1> ret = matrix_operator().template zero<D, 1>();
+        if constexpr (D == 1u) {
+            matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
+        } else if (D == 2u) {
+            matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
+            matrix_operator().element(ret, 1, 0) = m_measurement.local[1];
+        }
```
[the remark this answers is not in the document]
niermann999: Also, the generation of the zero matrix is probably not needed here?

core/include/traccc/edm/track_state.hpp
```diff
@@ -71,20 +71,34 @@ struct track_state {
...
-        matrix_type<2, 2> ret = matrix_operator().template zero<2, 2>();
-        matrix_operator().element(ret, 0, 0) = m_measurement.variance[0];
-        matrix_operator().element(ret, 1, 1) = m_measurement.variance[1];
+    template <size_type D>
+    TRACCC_HOST_DEVICE inline matrix_type<D, D> measurement_covariance() const {
+        static_assert(((D == 1u) || (D == 2u)),
+                      "The measurement dimension should be 1 or 2");
+
+        matrix_type<D, D> ret = matrix_operator().template zero<D, D>();
+        if constexpr (D == 1u) {
```
niermann999, line 96: Same suggestion as above

core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
```diff
@@ -44,67 +44,117 @@ struct gain_matrix_smoother {
...
-            V - H * smt_cov * matrix_operator().transpose(H);
-        const matrix_type<1, 1> chi2 = matrix_operator().transpose(residual) *
-                                       matrix_operator().inverse(R) * residual;
-
-        cur_state.smoothed_chi2() = matrix_operator().element(chi2, 0, 0);
+        this->operator()(cur_state, next_state);
     }
+
+    template <size_type D = 2u>
+    TRACCC_HOST_DEVICE inline void operator()(
```
niermann999, line 55: This maybe a stupid question, but where is ```D``` passed in the template parameter list when this operator is called in l. 51 and why does this need to be a separate operator from the call operator above?

core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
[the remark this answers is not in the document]
niermann999: I think the specialization for ```D==0``` is needed because ```gain_matrix_smoother_helper``` is recursive. But would a simple switch-case not have done the trick?
```
switch(meas.meas_dim) {
case 1: gain_matrix_smoother_helper<1>()(cur_state, next_state); break;
case 2: gain_matrix_smoother_helper<2>()(cur_state, next_state); break;
case 0: ERROR
};
```
Then ```gain_matrix_smoother_helper``` can even be a simple function

core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp
```diff
@@ -38,71 +38,117 @@ struct gain_matrix_updater {
...
-        const mask_group_t& mask_group, const index_t& index,
+        const mask_group_t& /*mask_group*/, const index_t& /*index*/,
         track_state<algebra_t>& trk_state,
         bound_track_parameters& bound_params) const {
 
-        // Some identity matrices
-        // @Note: Make constexpr work
-        const matrix_type<6, 6> I66 =
-            matrix_operator().template identity<e_bound_size, e_bound_size>();
+        this->operator()(trk_state, bound_params);
```
niermann999, line 45: Same as above:
```suggestion
        switch(trk_state.get_measurement().meas_dim){
        case 1: gain_matrix_updater_helper<1>(trk_state, bound_params); break;
        case 2: gain_matrix_updater_helper<2>(trk_state, bound_params); break;
        case 0: ERROR
        };
```

core/include/traccc/utils/subspace.hpp
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// Check if the given axis index in the full space is part of the subspace.
+    TRACCC_HOST_DEVICE
+    constexpr bool contains(size_type index) const {
+        bool isContained = false;
+        // always iterate over all elements to avoid branching and hope the
+        // compiler can optimized this for us.
+        for (auto a : m_axes) {
+            isContained = (isContained or (a == index));
```
niermann999, line 78:
```suggestion
            isContained |= (a == index);
```

core/include/traccc/utils/subspace.hpp
```diff
@@ -0,0 +1,119 @@
...
+        return m_axes;
+    }
+
+    /// Check if the given axis index in the full space is part of the subspace.
+    TRACCC_HOST_DEVICE
+    constexpr bool contains(size_type index) const {
+        bool isContained = false;
+        // always iterate over all elements to avoid branching and hope the
+        // compiler can optimized this for us.
+        for (auto a : m_axes) {
```
niermann999, line 77:
```suggestion
        for (const auto a : m_axes) {
```

beomki-yeo:
```
I think the specialization for D==0 is needed because gain_matrix_smoother_helper is recursive. But would a simple switch-case not have done the trick?

switch(meas.meas_dim) {
case 1: gain_matrix_smoother_helper<1>(cur_state, next_state); break;
case 2: gain_matrix_smoother_helper<2>(cur_state, next_state); break;
case 0: ERROR
};

Then gain_matrix_smoother_helper can even be a simple function
```

I think it makes sense. I changed into that (I cannot reply to your comment directly)

core/include/traccc/utils/subspace.hpp
beomki-yeo: I just removed the function because it seems we don't need it

core/include/traccc/edm/track_state.hpp
beomki-yeo: I think we need it to be zero matrix

core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
beomki-yeo: Fixed I guess

core/include/traccc/edm/track_state.hpp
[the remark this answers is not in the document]
niermann999: Isn't ```matrix_type<D, 1>``` just a vector and you are setting all of its elements explicitly anyway?

core/include/traccc/edm/track_state.hpp
niermann999: Agreed

core/include/traccc/edm/track_state.hpp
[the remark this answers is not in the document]
beomki-yeo: I need to set them explicitly. otherwise it can have undefined values.
matrix_type<D ,1> is a vector but I need it to be matrix type to use it in matrix operations.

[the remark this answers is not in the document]
niermann999: My point is, aren't you setting them explicitly twice now? First to {0, 0} and then to {loc[0], loc[1]} ? Anyway, it's a minor point, so let's just leave it as it is...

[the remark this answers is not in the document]
beomki-yeo: Understand. Fixed it now

niermann999 (approved):
Just a minor suggestion

PR #458, opened 2023-09-12
Make Csv Edm Public and add `local_key` to Measurement Edm

beomki-yeo:
The PR depends #456. 

Two changes
1) Move the edm files in `src/io/csv` to `include/io/csv` to use in simulation
2) Use `local_key` of `csv/measurement` in measurement reading.

~~Actually, If you look at the change in measurement reading ([io/src/csv/read_measurements.cpp](https://github.com/acts-project/traccc/compare/main...beomki-yeo:refactor-measurement-io?expand=1#diff-2cf443598544e2a2ccc5d7daa10b4e50939fa8b108bac0571c25617692eb566b)), you might think there is no improvement - which I agree.~~

~~I expected that subspace indices and the its dimension size is encrypted in `local_key`. However, it seems only the measurement dimension can be recovered while the indices is always fixed to {0u} for 1D measurement and {0u, 1u} for 2D measurement. (For annulus endcap strip, I think it should be {1u})~~

~~I will update the writing once I get clear on this matter.~~


io/include/traccc/io/csv/measurement.hpp
```diff
@@ -21,7 +21,7 @@ struct measurement {
 
     uint64_t measurement_id = 0;
     uint64_t geometry_id = 0;
-    std::string local_key = "";
+    uint8_t local_key = 0;
```
krasznaa, line 24: Are you sure that a `uint16_t` would not serve us better here? The `uint8_t` type is clearly handled as a `char` by the I/O. Leading to some awkward characters in the CSV files.

I think the larger memory footprint of `uint16_t` is overruled by its better ease of use of. :wink:

krasznaa (approved):
Other than the `uint8_t` / `uint16_t` question I'm happy with this.

io/include/traccc/io/csv/measurement.hpp
beomki-yeo: It's awkward indeed, but that's what is defined in ACTS Core repository: https://github.com/acts-project/acts/blob/ea7aaa5086ebd3618360949df04f885488d3d9ce/Examples/Io/Csv/src/CsvOutputData.hpp#L123

I think we need to keep the same type to read ACTS simulation data properly.

6 bits in the middle are used for each index of bound track parameter (x,y,phi,theta,q/p,t); 1 means the measurement contains the information of the parameter with corresponding index.

If I am not wrong, the first and last digit are dummy values

krasznaa: Okay, that's an acceptable reason for the type then. Objection withdrawn.

krasznaa: > If `uint16_t` is be compatible with `uint8_t` in I/O-wise, then I think it is good to change

No, I think it would hurt us. I just didn't think of the necessary compatibility with Acts...

beomki-yeo:
Let me elaborate on `local_key` for future development

@krasznaa suggested having a larger memory for `local_key`,  and actually ACTS core and traccc needs it definitely.  8 bits of local key is not enough.

What `local_key` can do is literally setting _which parameter I am going to use for measurement_ but it doesn't set _in which order I am going to take measurement_. In other words, current `local_key` does not have permutation information that is required to construct `traccc::subspace` correctly.  

Fortunately, any permutation is not needed for the measurements of our experiment, so this is just a potential bug at the moment

PR #450, opened 2023-08-30
Update detray version to v0.41.0

beomki-yeo:
This PR updates the detray version to v0.41.0 which changes the simulator interface.

PR #459, opened 2023-09-13
Add Simulator from Detray

beomki-yeo:
The PR depends on #458 #450

Reimplementing a simulator which is removed in acts-project/detray#553


krasznaa (changes requested):
Well, the bug is clear enough. :wink: If you really intended to remove the `measurement_hit_if.hpp` file, then you also need to remove the reference to it from its `CMakeLists.txt` file.

beomki-yeo:
Actually I am trying to keep `measurement_hit_id`...

beomki-yeo:
I decided to skip detray v0.43.0 and wait for future updates.

PR #461, opened 2023-09-26
SYCL test in gitlab CI

beomki-yeo:
This PR copy-pastes detray's [gitlab workflow](https://github.com/acts-project/detray/blob/main/.gitlab-ci.yml)
Gitlab CI have a device in its server, meaning that the workflow is not hindered by oneDPL's limitation where the device should be present.
I made the simulation directory of cpu/cuda/sycl unit test different because the same simulation directory was causing the error during `ctest`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -37,3 +37,41 @@ test_cuda:
     - nvidia-smi
     - ctest --output-on-failure -E "^SeedingValidation/CompareWithActsSeedingTests.*"
 
+
+build_sycl:
+  tags: [docker]
+  stage: build
+  image: "ghcr.io/acts-project/ubuntu2004_oneapi:v30"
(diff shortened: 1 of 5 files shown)


beomki-yeo:
@krasznaa
How do you think of disabling github SYCL CI and using gitlab SYCL CI which has a device already? Then we can go with @442

.gitlab-ci.yml
```diff
@@ -37,3 +37,41 @@ test_cuda:
     - nvidia-smi
     - ctest --output-on-failure -E "^SeedingValidation/CompareWithActsSeedingTests.*"
 
+
+build_sycl:
+  tags: [docker]
```
paulgessinger, line 42: this can be removed. It was used to mimic what the shared GitLab runners were configured like, but that recently changed.

PR #460, opened 2023-09-19
Add SYCL CKF

beomki-yeo:
I made the SYCL implementation for CKF, even though PR #442 can not be merged now.

### SYCL Implementation

As seeding and fitting are implemented, SYCL and CUDA implementation share the same kernel functions.
A remarkable chnage is that SYCL implementation is supported by oneDPL instead of thrust library.

### Seeding + Finding + Fitting Result

For a single CPU threads benchmark, SYCL is slower than CUDA

1. SYCL (2000 charged tracks)

```
[beomki@pc-54 traccc_sycl]$ ./bin/traccc_seeding_example_sycl --input_directory=detray_simulation/toy_detector/n_particles_2000/ --run_detray_geometry=true --check_performance=true --detector_file=detray_json/toy_detector_geometry.json --material_file=detray_json/toy_detector_homogeneous_material.json --event=1 --track_candidates_range=3:10 --constraint-step-size-mm=1 --run_cpu=1
Running ./bin/traccc_seeding_example_sycl detray_json/toy_detector_geometry.json detray_simulation/toy_detector/n_particles_2000/ 1
Running on device: NVIDIA GeForce RTX 2070
===>>> Event 0 <<<===
Number of seeds: 11275 (host), 11274 (device)
  Matching rate(s):
    - 5.66741% at 0.01% uncertainty
    - 23.9202% at 0.1% uncertainty
    - 72.9401% at 1% uncertainty
    - 92.2749% at 5% uncertainty
Number of track parameters: 11275 (host), 11274 (device)
  Matching rate(s):
    - 51.184% at 0.01% uncertainty
    - 83.6718% at 0.1% uncertainty
    - 98.6341% at 1% uncertainty
    - 99.9024% at 5% uncertainty
Track candidate matching Rate: 0.999015
==> Statistics ... 
- read    10565 spacepoints from 2553 modules
- created  (cpu)  11275 seeds
- created (sycl) 11274 seeds
- created  (cpu) 11163 found tracks
- created (sycl) 11162 found tracks
- created  (cpu) 11163 fitted tracks
- created (sycl) 11162 fitted tracks
==>Elapsed times...
            Hit reading  (cpu)  234 ms
                Seeding (sycl)  32 ms
                Seeding  (cpu)  56 ms
           Track params (sycl)  0 ms
           Track params  (cpu)  2 ms
 Track finding with CKF (sycl)  308 ms
  Track finding with CKF (cpu)  1895 ms
  Track fitting with KF (sycl)  186 ms
[...]

PR #462, opened 2023-09-30
Asynchronous operation for Track Finding and Fitting

beomki-yeo:
Depends on #460 

This PR is for enabling asynchronous operation on Track finding and fitting, as much as I can.
There are still bunch of memory allocation calls in CKF algorithm so it may not be very efficient at the moment. 
I believe we can reduce the calls by allocating a big memory.   

Meanwhile, not all functions of oneDPL seem to support asynchrnouse operation (for example, `dpl::upper_bound` or `dpl::unique_copy`)


beomki-yeo:
No idea why the oneDPL header is read from oneAPI directory instead of traccc external
```
In file included from /builds/acts/ci-bridge/src/device/sycl/src/finding/finding_algorithm.sycl:11:
In file included from /opt/intel/oneapi/dpl/2021.6.0/linux/include/oneapi/dpl/async:24:
In file included from /opt/intel/oneapi/dpl/2021.6.0/linux/include/oneapi/dpl/internal/async_extension_defs.h:20:
/opt/intel/oneapi/dpl/2021.6.0/linux/include/oneapi/dpl/internal/async_impl/async_utils.h:50:56: error: use of undeclared identifier 'access_mode';
did you mean 'sycl::access_mode'?
        return __my_op(__my_buffer.template get_access<access_mode::read>()[__my_offset], __init);
                                                       ^~~~~~~~~~~
                                                       sycl::access_mode
```

beomki-yeo:
Changing include order solves the problem :/

PR #465, opened 2023-10-11
feat: implement digitization configurator for detray geometries

niermann999:
Adds a digitization configurators that can read a digitization meta configuration and generate a full digitization configuration for a detray geometry. This is mostly useful for geometries that are generated in detray, as this should hopefully become compatible with ACTS generated tracking geometries at some point.


niermann999:
Still very WIP

niermann999:
I have prepared a meta-config file for a toy detector with 3 and 7 endcaps and generated the corresponding digitization config files. The unittest in this PR expects to find the geometry and digi config files for the toy detector in the data directory

krasznaa (approved):
This looks okay. Even if it has a bunch more `auto` declarations in it than I would prefer...

PR #466, opened 2023-10-11
Update to Detray v42

beomki-yeo:
As #459 is closed.

The PR merely migrates simulation file from detray


beomki-yeo:
@krasznaa @niermann999 Could you guys review it. This PR is just migrating the simulation script from detray

io/src/csv/read_measurements.cpp
```diff
@@ -48,10 +48,31 @@ void read_measurements(measurement_reader_output& out,
...
+                        indices[meas.meas_dim++] = ipar;
+                    }; break;
+                    case e_bound_loc1: {
+                        meas.local[1] = iomeas.local1;
+                        meas.variance[1] = iomeas.var_local1;
+                        indices[meas.meas_dim++] = ipar;
+                    }; break;
+                }
+            }
+        }
```
niermann999, line 71: What does this do? It checks if `iomeas.local_key` has the second bit set when `ipar` is 0 (I guess when `iomeas.local_key` is 2 or 6. Are other cases possible, e.g. can it be 3?) and has the third bit set when `ipar` is 1, so either 4 or 6. For either case it then checks if `ipar` is 0 or 1 (are we expecting `e_bound_0` and `e_bound_1` to switch values?) and then I start to get lost because it also starts counting how often the correct bit was encountered in `meas.meas_dim`, i.e. once if `iomeas.local_key` is 2 or 4 and twice if it was 6?
I get that this has to cover many possible combinations: Can this be done in way that is maybe longer but more readable or could you draw a table in the comments on what the expected behaviour is?

simulation/include/traccc/simulation/measurement_smearer.hpp
```diff
@@ -0,0 +1,117 @@
...
+        subspace<transform3, 6u, 2u> subs(indices);
+
+        if (meas_dim == 1u) {
+            const auto proj = subs.projector<1u>();
+            matrix_type<1u, 1u> meas = proj * bound_params.vector();
+
+            if (mask_t::shape::name == "line") {
+                iomeas.local0 =
+                    std::max(std::abs(matrix_operator().element(meas, 0u, 0u)) +
+                                 offset[0],
```
niermann999, line 97: So, offset contains the smearing parameters and this must not become less than 0 because this is the approach distance for line surfaces, where the sign is not recorded?

niermann999 (approved):
Just minor comments, but can go ahead

io/src/csv/read_measurements.cpp
beomki-yeo: First, this is from ACTS Core and not my design. 
My interpretation might have an error but here it is:

Local key is a 8 bit char and first and last bit are dummy value. 2 - 7th bits are for 6 bound track parameters.

Ex1: 0000010 or 2 -> meas dim = 1 and bound_loc0 active -> strip or wire detector 
Ex2: 0000110 or 6 -> meas dim = 2 and bound_loc1 and bound_loc2 active -> pixel
Ex3: 0000100 or 4 -> meas dim = 1 and bound_loc2 active ->  annulus where we measure phi

simulation/include/traccc/simulation/measurement_smearer.hpp
beomki-yeo: yes. In any case, there can not be minus for distance measurement.
This will cause a bias in the final physics but we really don't have to be precise on this because smearing won't be used in real case.

io/src/csv/read_measurements.cpp
niermann999: Ok, I see! Might be good to have that in the comments somewhere?

beomki-yeo: > I get that this has to cover many possible combinations

I thought that too, but I don't see any real use case with more combinations. What current implementation provides will be enough. So I would postpone engineering this further.

beomki-yeo: > Ok, I see! Might be good to have that in the comments somewhere?

Ok

PR #467, opened 2023-10-16
Sort measurements in CCA

beomki-yeo:
I think it is pointless to do measurement-sort in CKF. At least we can reduce a copy operation by moving the sort to CCA.
Of course, note that the best approach would be modifying CCA to yield the sorted measurements directly.

+ Also note that the measurement output of CPU clusterization is always sorted, and it makes sense to unify the output property (sorted-ness) of CPU and GPU clusterization

### DIFF
--- a/core/include/traccc/finding/finding_algorithm.ipp
+++ b/core/include/traccc/finding/finding_algorithm.ipp
@@ -29,30 +29,21 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
      * Measurement Operations
      *****************************************************************/
 
-    // Copy the measurements
-    measurement_collection_types::host sorted_measurements = measurements;
-
-    // Sort the measurements w.r.t geometry barcode
-    std::sort(sorted_measurements.begin(), sorted_measurements.end(),
-              measurement_sort_comp());
-
     // Get copy of barcode uniques
     std::vector<measurement> uniques;
-    uniques.resize(sorted_measurements.size());
+    uniques.resize(measurements.size());
 
-    auto end =
-        std::unique_copy(sorted_measurements.begin(), sorted_measurements.end(),
-                         uniques.begin(), measurement_equal_comp());
+    auto end = std::unique_copy(measurements.begin(), measurements.end(),
+                                uniques.begin(), measurement_equal_comp());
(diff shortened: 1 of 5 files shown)


beomki-yeo:
There is no big difference in CCA time

After
```
==> Statistics ... 
- read    199547 cells from 3859 modules
- created (cpu)  36907 measurements     
- created (cpu)  36907 spacepoints     
- created (cuda) 36907 spacepoints     
- created  (cpu) 6151 seeds
- created (cuda) 6151 seeds
==>Elapsed times...
           File reading  (cpu)  394 ms
         Clusterization (cuda)  0 ms
         Clusterization  (cpu)  7 ms
   Spacepoint formation  (cpu)  1 ms
                Seeding (cuda)  2 ms
                Seeding  (cpu)  173 ms
           Track params (cuda)  0 ms
           Track params  (cpu)  1 ms
                     Wall time  580 ms
```
Before
```
==> Statistics ... 
- read    199547 cells from 3859 modules
- created (cpu)  36907 measurements     
- created (cpu)  36907 spacepoints     
- created (cuda) 36907 spacepoints     
- created  (cpu) 6151 seeds
- created (cuda) 6151 seeds
==>Elapsed times...
           File reading  (cpu)  394 ms
         Clusterization (cuda)  0 ms
         Clusterization  (cpu)  7 ms
   Spacepoint formation  (cpu)  0 ms
                Seeding (cuda)  2 ms
                Seeding  (cpu)  165 ms
           Track params (cuda)  0 ms
           Track params  (cpu)  1 ms
                     Wall time  572 ms
```

krasznaa (approved):
Fine, let's go ahead like this. Though there are a couple of things in `clusterization_algorithm.cu` that I don't like. But we'll need to make vecmem a bit smarter (I think...) to have that code in a way that I'd be more okay with. 🤔

beomki-yeo:
OK. I also want to leave the sort function in track finding like the following:

```c++
track_finding(measurement_collection& measurements, bool doSort=false){
  if (doSort)
     std::sort(measurements.begin(), measurements.end(), measurement_sort_comp());
}
```

This is just for the case where users don't use traccc clusterization algorithm in their chain
But let's think about this later...

PR #468, opened 2023-10-17
Git ignore also data symlinks

fiedlerp:
Test data were not ignored when they were symlinked from another place. Also, the `single_tracks` were not ignored.

### DIFF
--- a/data/.gitignore
+++ b/data/.gitignore
@@ -3,13 +3,14 @@
 #
 # Mozilla Public License Version 2.0
 #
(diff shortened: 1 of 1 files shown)


stephenswat:
I don't have a problem with this at all, although I would like to note that you can also set the `TRACCC_TEST_DATA_DIR` environment variable to direct the executables to the testing data. This would be a lot easier than symbolically linking each of these (many!) directories.

PR #469, opened 2023-10-17
Migrate Simulation Validation Script from Detray

beomki-yeo:
Reviving the script that was removed by acts-project/detray#553


core/include/traccc/utils/subspace.hpp
```diff
@@ -48,10 +48,6 @@ struct subspace {
         for (size_type i = 0u; i < kSize; ++i) {
             assert((indices[i] < kFullSize) and
                    "Axis indices must be within the full space");
-            if (0u < i) {
```
beomki-yeo, line 51: Our measurement dimension is a run time variable so axis indices may not be unique for some cases. (e.g. `indices = {0,0}` for wire measurement)

niermann999 (approved):

PR #470, opened 2023-10-18
Data Files V4, main branch (2023.10.18.)

krasznaa:
On @niermann999's request I created V4 of the data files with the addition of detector description files for the toy detector.

Joana, the new files are under `${TRACCC_TEST_DATA_DIR}/geometries/toy_detector/`. I thought that at one point we'll probably want to clean up the test files a little, giving the directory a bit more "organized layout". But for now I didn't move anything else.

@fiedlerp, since I ended up touching `data/.gitignore` in this PR anyway, I incorporated your changes from #468. We should discuss in that PR how you should probably re-fork the repository afterwards, to get rid of your (now probably unmerged) changes from the `main` branch of your fork. :thinking:

### DIFF
--- a/data/.gitignore
+++ b/data/.gitignore
@@ -3,13 +3,14 @@
 #
 # Mozilla Public License Version 2.0
 #
-cca_test/
-detray_simulation/
-single_module/
-tml_detector/
-tml_full/
-tml_pixel_barrel/
-tml_pixels/
-two_modules/
+cca_test
+detray_simulation
+geometries
+single_module
+tml_detector
+tml_full
+tml_pixel_barrel
+tml_pixels
+two_modules
 *.tar.gz
 *.md5
--- a/data/traccc_data_get_files.sh
+++ b/data/traccc_data_get_files.sh
@@ -27,7 +27,7 @@ usage() {
 }
 
 # Default script arguments.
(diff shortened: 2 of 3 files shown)


beomki-yeo:
@niermann999 Is this fully-spanned digitization configuration? or short version?

I would just take the short version into the git repository _directly_ and encourage people to generate the fully-spanned version by themselves.

niermann999 (approved):
Ok, I updated the path

niermann999:
> @niermann999 Is this fully-spanned digitization configuration? or short version?
> 
> I would just take the short version into the git repository _directly_ and encourage people to generate the fully-spanned version by themselves.

Both, actually. I added the full version mainly because the same data file was added for the tml detector as well. But I can also generate it on the fly in the unittest

niermann999:
We will whenever something changes in the geometry for sure. Then also the geometry, material and grid files have to be updated

beomki-yeo:
Ok. Anyway it is g2g as it is harmless

beomki-yeo (approved):

krasznaa:
> We can go with the current change. I hope we won't have to edit or modify the files

When we need to modify the files, we'll go to V5. I'm not worried about this. Updating the data files once, or even a couple of times a week should not be the thing holding us back. 🤔

beomki-yeo:
Aren't you the only manager of this data folder? I would worry the case when you are away.

PR #472, opened 2023-10-21
Fix: Do not sort measurments in spacepoint reading

beomki-yeo:
Fixing a bug introduced by #467. Seeding example is not woking now

measurement-to-hit id is broken when measurements is shuffled by sorting.

### DIFF
--- a/io/src/csv/read_measurements.hpp
+++ b/io/src/csv/read_measurements.hpp
@@ -20,6 +20,7 @@ namespace traccc::io::csv {
 ///
 /// @param out A measurement & a cell_module (host) collections
 /// @param filename The file to read the measurement data from
+/// @param do_sort Whether to sort the measurements or not
 ///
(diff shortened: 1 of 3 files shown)


beomki-yeo:
@krasznaa Could you take a look into this PR soon?

io/src/csv/read_measurements.hpp
```diff
@@ -22,6 +22,6 @@ namespace traccc::io::csv {
 /// @param filename The file to read the measurement data from
```
krasznaa, line 22: Just to give you some actual comment, you should add a line in the documentation about the new parameter. :wink:

krasznaa (approved):
Sure, this is simple enough.

io/src/csv/read_measurements.hpp
beomki-yeo: Sure I often forget doing it  ;p

PR #474, opened 2023-10-23
Asynchronous CUDA finding and fitting

beomki-yeo:
Adding stream and async_copy to cuda finding and fitting.

Also thrust device function takes cuda_stream as an input for asynchronous operation

### DIFF
--- a/device/cuda/src/finding/finding_algorithm.cu
+++ b/device/cuda/src/finding/finding_algorithm.cu
@@ -6,6 +6,7 @@
  */
 
 // Project include(s).
+#include "../utils/utils.hpp"
 #include "traccc/cuda/finding/finding_algorithm.hpp"
 #include "traccc/cuda/utils/definitions.hpp"
 #include "traccc/definitions/primitives.hpp"
@@ -162,16 +163,9 @@ __global__ void build_tracks(
 
 template <typename stepper_t, typename navigator_t>
 finding_algorithm<stepper_t, navigator_t>::finding_algorithm(
-    const config_type& cfg, const traccc::memory_resource& mr)
-    : m_cfg(cfg), m_mr(mr) {
-
-    // Initialize m_copy ptr based on memory resources that were given
-    if (mr.host) {
-        m_copy = std::make_unique<vecmem::cuda::copy>();
-    } else {
-        m_copy = std::make_unique<vecmem::copy>();
-    }
-};
+    const config_type& cfg, const traccc::memory_resource& mr,
+    vecmem::copy& copy, stream& str)
+    : m_cfg(cfg), m_mr(mr), m_copy(copy), m_stream(str){};
 
 template <typename stepper_t, typename navigator_t>
 track_candidate_container_types::buffer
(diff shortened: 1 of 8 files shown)


beomki-yeo:
Review anyone?

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -498,21 +507,21 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
         nBlocks = (n_tips_total + nThreads - 1) / nThreads;
-        kernels::build_tracks<<<nBlocks, nThreads>>>(
+        kernels::build_tracks<<<nBlocks, nThreads, 0, stream>>>(
             measurements, seeds_buffer, links_buffer, param_to_link_buffer,
             tips_buffer, track_candidates_buffer);
 
         CUDA_ERROR_CHECK(cudaGetLastError());
-        CUDA_ERROR_CHECK(cudaDeviceSynchronize());
     }
+    m_stream.synchronize();
```
krasznaa, line 524: This should stay here for this PR.

But... at one point we'll need to experiment with removing these "final sync points" from the algorithms. Since in principle the returned buffer **can** wait for an asynchronous copy "on the caller's side" as well. (The buffer is moved back to the user. So the asynchronous copy will "find it".)

device/cuda/src/fitting/fitting_algorithm.cu
```diff
@@ -90,12 +87,15 @@ track_state_container_types::buffer fitting_algorithm<fitter_t>::operator()(
         const unsigned int nBlocks = (n_tracks + nThreads - 1) / nThreads;
 
         // Run the track fitting
-        kernels::fit<fitter_t><<<nBlocks, nThreads>>>(
+        kernels::fit<fitter_t><<<nBlocks, nThreads, 0, stream>>>(
             det_view, field_view, m_cfg, navigation_buffer,
             track_candidates_view, track_states_buffer);
         CUDA_ERROR_CHECK(cudaGetLastError());
         CUDA_ERROR_CHECK(cudaDeviceSynchronize());
```
krasznaa, line 94: This line should go. :wink:

krasznaa (changes requested):
Only minor points. :wink:

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -295,19 +296,21 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
         nThreads = WARP_SIZE * 2;
         nBlocks = (n_in_params + nThreads - 1) / nThreads;
-        kernels::apply_interaction<detector_type><<<nBlocks, nThreads>>>(
-            det_view, navigation_buffer, n_in_params, in_params_buffer);
+        kernels::apply_interaction<detector_type>
+            <<<nBlocks, nThreads, 0, stream>>>(det_view, navigation_buffer,
+                                               n_in_params, in_params_buffer);
         CUDA_ERROR_CHECK(cudaGetLastError());
-        CUDA_ERROR_CHECK(cudaDeviceSynchronize());
+        m_stream.synchronize();
```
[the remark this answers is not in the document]
beomki-yeo: You are right - we don't need it

device/cuda/src/finding/finding_algorithm.cu
beomki-yeo: OK

device/cuda/src/fitting/fitting_algorithm.cu
beomki-yeo: Thanks for catching. Fixed

krasznaa (approved):
With those things fixed, I'm happy to wave this one through. :wink:

PR #435, opened 2023-08-02
Add drift chamber simulation and unit test

beomki-yeo:
This PR adds unit test for track fitting with drift chamber (hooray..)

I had to went through a massive refactoring & fix:
- Fix `event_map2` to take one dimensional measurement correctly
- Fix `find_tracks.ipp` to consider the case where there are more candidates than maximum number (CKF on the drift chamber sometimes experience high number of combinatorics)
- Unify the detector type with `default_metadata` 
- Separate kalman fitting test file for different detector 
- Edit `README.md` for tutorial

Following is the truth track finding results with drift chamber:

PR #477, opened 2023-10-28
Add plots for pvalue of measurements

beomki-yeo:
PR is based on #435

image


beomki-yeo:
This PR is merged into #480

PR #478, opened 2023-10-30
Clang & CUDA Fix, main branch (2023.10.30.)

krasznaa:
The `cuda` namespace is not pointing clearly at `traccc::cuda` in the presence of `using namespace traccc`. At least Clang doesn't seem to think so. :thinking:

```
[ 95%] Building CXX object tests/cuda/CMakeFiles/traccc_test_cuda.dir/test_clusterization.cpp.o
/home/krasznaa/ATLAS/projects/traccc/traccc/tests/cuda/test_clusterization.cpp:51:5: error: reference to 'cuda' is ambiguous
    cuda::experimental::clusterization_algorithm ca_cuda(mr, copy, stream,
    ^
/software/cuda/12.0.1/x86_64/include/cuda/std/detail/libcxx/include/type_traits:4861:1: note: candidate found by name lookup is 'cuda'
_LIBCUDACXX_BEGIN_NAMESPACE_STD_NOVERSION
^
/software/cuda/12.0.1/x86_64/include/cuda/std/detail/__config:192:61: note: expanded from macro '_LIBCUDACXX_BEGIN_NAMESPACE_STD_NOVERSION'
#define _LIBCUDACXX_BEGIN_NAMESPACE_STD_NOVERSION namespace cuda { namespace std {
                                                            ^
/home/krasznaa/ATLAS/projects/traccc/traccc/device/cuda/include/traccc/cuda/clusterization/experimental/clusterization_algorithm.hpp:22:19: note: candidate found by name lookup is 'traccc::cuda'
namespace traccc::cuda::experimental {
                  ^
/home/krasznaa/ATLAS/projects/traccc/traccc/tests/cuda/test_clusterization.cpp:55:9: error: use of undeclared identifier 'ca_cuda'
        ca_cuda(vecmem::get_data(cells), vecmem::get_data(modules));
        ^
2 errors generated.
```

At the same time I also renamed one of the `clusterization_algorithm.cu` source files, since it seems that `nvcc` cannot deal with two source files having the same name. (The preprocessor seems to overwrite temporary files with each other in some weird way.) Which seems to be the reason for the intermittent CI failures since #449.

As discussed [on Mattermost](https://mattermost.web.cern.ch/acts/channels/traccc), the "file name clash" happens because of:

https://github.com/acts-project/traccc/blob/main/cmake/traccc-compiler-options-cuda.cmake#L32

So it's a bug in our build setup, and not an inherent bug in CUDA or CMake as it turns out.

### DIFF
--- a/device/cuda/CMakeLists.txt
+++ b/device/cuda/CMakeLists.txt
@@ -43,12 +43,12 @@ traccc_add_library( traccc_cuda cuda TYPE SHARED
   "src/cca/component_connection.cu"
   # Clusterization
   "include/traccc/cuda/clusterization/experimental/clusterization_algorithm.hpp"
-  "src/clusterization/experimental/clusterization_algorithm.cu"
+  "src/clusterization/experimental/clusterization_algorithm_exp.cu"
   "include/traccc/cuda/clusterization/clusterization_algorithm.hpp"
   "src/clusterization/clusterization_algorithm.cu"
   # Finding
   "include/traccc/cuda/finding/finding_algorithm.hpp"
-  "src/finding/finding_algorithm.cu"  
+  "src/finding/finding_algorithm.cu"
   # Fitting
   "include/traccc/cuda/fitting/fitting_algorithm.hpp"
   "src/fitting/fitting_algorithm.cu")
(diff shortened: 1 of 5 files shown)


krasznaa:
Our dedicated CI machine being overloaded certainly doesn't help the optics of this PR. :frowning:

beomki-yeo (approved):
good

PR #481, opened 2023-11-06
Change default value of min track candidates

beomki-yeo:
two is a bad number where ndf gets negative (2*N_candidates - 5 = -1)


krasznaa (approved):
Don't fully understand it, but technically this is a very trivial change, so...

PR #482, opened 2023-11-06
Bug Fixes, main branch (2023.11.06.)

krasznaa:
This is to fix multiple, independent bugs that I ran into just by trying to build the project on my laptop.
  - Stopped `algebra-plugin` from downloading/building Google Benchmark itself. This was more of an annoyance than an actual bug.
  - Fixed the setup of the `traccc::simulation` library. This one confused the heck out of me, as to why this bug was not found before. And then I realized that I only ran into it because I turned off the build/usage of Vc. And with Vc enabled, we do get a dummy shared library for `traccc::simulation`. :dizzy_face:

```
[bash][celeborn]:build > more simulation/CMakeFiles/traccc_simulation.dir/flags.make 
# CMAKE generated file: DO NOT EDIT!
# Generated by "Unix Makefiles" Generator, CMake Version 3.26

[bash][celeborn]:build > more simulation/CMakeFiles/traccc_simulation.dir/link.txt 
/usr/bin/c++ -fPIC -O3 -g -shared -Wl,-soname,libtraccc_simulation.so.0 -o ../lib/libtraccc_simulation.so.0.6.0  -Wl,-rpath,/home/krasznaa/ATLAS/projects/traccc/build/lib:/home/krasznaa/ATLAS/projects/traccc/build/_deps/acts-build/lib: ../lib/libtraccc_io.so.0.6.0 ../lib/libtraccc_core.so.0.6.0 ../lib/libVc.a ../_deps/acts-build/lib/libActsCore.so ../lib/libvecmem_core.so.0.26.0 -Wl,-rpath-link,/home/krasznaa/ATLAS/projects/traccc/build/_deps/acts-build/lib 
[bash][celeborn]:build >
```

While if I turn off Vc, I get the much more reasonable:

```
...
-- Configuring done (6.4s)
CMake Error: CMake can not determine linker language for target: traccc_simulation
CMake Error in simulation/CMakeLists.txt:
  Exporting the target "traccc_simulation" is not allowed since its linker
  language cannot be determined
```

  - Made the code build without ROOT. This is just an oversight. When `TRACCC_USE_ROOT` was introduced, I should've added a CI test for it as well. I now did. :wink:

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -27,6 +27,9 @@ jobs:
           - name: CPU
             container: ghcr.io/acts-project/ubuntu2004:v30
             options:
+          - name: CPU
+            container: ghcr.io/acts-project/ubuntu2004:v30
+            options: -DTRACCC_USE_ROOT=FALSE
           - name: CUDA
             container: ghcr.io/acts-project/ubuntu2004_cuda:v30
             options: -DTRACCC_BUILD_CUDA=TRUE
--- a/extern/algebra-plugins/CMakeLists.txt
+++ b/extern/algebra-plugins/CMakeLists.txt
@@ -18,7 +18,7 @@ message( STATUS "Building Algebra Plugins as part of the TRACCC project" )
 
 # Declare where to get Algebra Plugins from.
 set( TRACCC_ALGEBRA_PLUGINS_SOURCE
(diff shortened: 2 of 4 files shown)


beomki-yeo (approved):
Looks good to me

PR #484, opened 2023-11-14
Fix a bug that default material and grid json input did not work

beomki-yeo:
Discovered by @shimasnd while working on ITk


niermann999 (approved):

beomki-yeo:
Hmm I think there is an issue in CI test which occurs randomly

krasznaa:
> Hmm I think there is an issue in CI test which occurs randomly

Yes, the CI is a lot more flaky than I'd like. 😦 I just didn't have time yet to look at it more closely.

krasznaa (approved):
This is fine by me as well by the way. 👍

PR #488, opened 2023-11-16
Parameter to limit the N track cands per initial seed.

shimasnd:
Suppress the explosion of N track candidates caused by almost identical two tracks. Introduce a new parameter in a finding_config file, setting a Max value on the N track candidates at each step for each initial seed.


core/include/traccc/finding/finding_algorithm.ipp
```diff
@@ -75,9 +75,13 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
 
     // Copy seed to input parameters
     std::vector<bound_track_parameters> in_params;
+    std::vector<unsigned int> n_trks_per_seed;
+    n_trks_per_seed.reserve(seeds.size());
```
beomki-yeo, line 79: I think it is easier to do in this way:
`std::vector<unsigned int> n_trks_per_seed(seeds.size(), 0);`

core/include/traccc/finding/finding_algorithm.ipp
```diff
@@ -100,10 +104,18 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
         const unsigned int previous_step =
             (step == 0) ? std::numeric_limits<unsigned int>::max() : step - 1;
 
+        for (unsigned int i_seed = 0; i_seed < seeds.size(); i_seed++)
+            n_trks_per_seed[i_seed] = 0;
```
beomki-yeo, line 108: One line will be enough:
`std::fill(n_trks_per_seed.begin(), n_trks_per_seed.end(), 0);`

core/include/traccc/finding/finding_algorithm.ipp
```diff
@@ -188,9 +204,12 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
                         static_cast<unsigned int>(links[step].size());
 
                     n_branches++;
+                    n_trks_per_seed[orig_param_id]++;
 
+                    // links[step].push_back(
+                    //     {{previous_step, in_param_id}, item_id});
```
beomki-yeo, line 210: Let's remove this commented lines

core/include/traccc/finding/candidate_link.hpp
```diff
@@ -30,6 +30,9 @@ struct candidate_link {
 
     // Measurement index
     unsigned int meas_idx;
+
+    // Index to the initial seed
+    unsigned int initial_seed;
```
beomki-yeo, line 35: I think `seed_idx` is decripted better

beomki-yeo (commented):
Here are a couple of comments

core/include/traccc/finding/finding_algorithm.ipp
beomki-yeo: `n_trks_per_seed.push_back(0);` will need to be removed together

shimasnd:
Using traccc::performance::timing_info, the measured time for 84k found tracks is not really changed by the use of the new parameter (value is set to 1000). On my laptop (CPU):
w/o the new parameter:                  Track finding  1251609 ms
w/ the new parameter:                  Track finding  1243698 ms
If I should try any better time measurement, please let me know.

beomki-yeo:
Ah also please make the default value of the parameters `std::numeric_limits<unsigned int>::max` _for the moment_

I will change it back to 1000 when I implement the same algorithm to GPU

shimasnd:
Thanks for suggestions. Once the default value is set as `std::numeric_limits<unsigned int>::max()`, can I skip your first suggestion, i.e.  two `finding_algorithm` objects in the test suite? To be honest, I don't understand why n(m+1) should be n*1000, when the max value is set to 1000, and how the max value is related to the number of planes.

beomki-yeo:
> To be honest, I don't understand why n^(m+1) should be n*1000, when the max value is set to 1000

Sorry There should be 2^9 branches per seed because there are 9 planes and two identical seeds - so the max value should be < 512 to test your code.

Or are you asking why it should be the power of number of identical tracks?

beomki-yeo (approved):
Looks good to me!

PR #490, opened 2023-11-16
Add check format local

beomki-yeo:
For people who don't have clang-format-10

### DIFF
--- a/.github/check_format_local.sh
+++ b/.github/check_format_local.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+set -e
+
+if [ $# -ne 1 ]; then
+    echo "\033[31mERROR\033[0m"\
+         "wrong number of arguments"
+    echo "\tusage: check_format.sh <DIR>\n"
+    exit 1
+fi
+
+# Setup some variables
+USER_ID=`id -u`
+GROUP_ID=`id -g`
+
+docker run --rm -ti \
+       -v $PWD:/work_dir:rw \
+       --user "${USER_ID}:${GROUP_ID}" \
+       -w "/work_dir" \
+       ghcr.io/acts-project/format10:v41 \
+       .github/check_format.sh $1


paulgessinger (approved):

beomki-yeo:
@paulgessinger I have question:

There is the line in `check_format.sh` that runs `git diff`:
https://github.com/acts-project/traccc/blob/a217af72d58f6dded61a27031d67a9e8786555f5/.github/check_format.sh#L36

If I run this with `check_format_local.sh`, the line makes an warning:
```
warning: Not a git repository. Use --no-index to compare two paths outside a working tree
usage: git diff --no-index [<options>] <path> <path>
```

I think I have seen this in Acts core as well, do you know how to bypass this? (Maybe add an if condition that checks if pwd is in git repository?)

paulgessinger:
@beomki-yeo I guess it depends if the mounted volume has `.git` in the directory hierarchy. If not it won't work I think.

krasznaa (changes requested):
I don't know... I see that this can help people in some situations, but...

On your personal machine you can just install `clang-format`, and in 99.999999% of the cases it will just work fine with the `check_format.sh` script. I didn't yet run into a situation where `clang-format` on Ubuntu 20.04, Ubuntu 22.04 or RHEL9 would not have produced the same output as `clang-format` in the specific Docker image that we use.

And for the very-very rare cases where this happens, I'm okay with just starting Docker myself. 🤔

If we want to make this script "a thing", it would have to be smarter than this. Right now I can just start it from a "wrong directory", and it will not work.

So... do we really want to write a script like this? Because then it would have to be made a little more robust than this...

beomki-yeo:
> On your personal machine you can just install clang-format, and in 99.999999% of the cases it will just work fine with the check_format.sh script.

Have you tested with other version of clang-format?
`clang-format-10` is not easy to install these days and i believe our CI will break even with `clang-format-11` if I remember correctly.

paulgessinger:
@beomki-yeo @krasznaa 

1. Consider bumping the version once in a while
2. I usually grab statically compiled binaries from here: https://github.com/muttleyxd/clang-tools-static-binaries/releases/tag/master-8f72ab3c

krasznaa:
As I wrote, on Ubuntu I only ever installed the default `clang-format` package, and that worked "pretty much" always the same as the version coming with the Docker image.

Admittedly I apparently spoke too quickly about RHEL9. Apparently on that I only ever used the Docker image. As I'm having a hard time finding a viable package for it on a vanilla system. 😦

But what always works for me, even on Windows: VS Code comes with a built-in code formatter. Which respects the project's `.clang-format` settings. So in 99% of the cases I just use the VS Code shortcut these days. And then do one big run with `check_format.sh` at the end of some bigger updates.

PR #493, opened 2023-11-21
Better thread balancing in `find_tracks` kernel of CKF

beomki-yeo:
This PR improves the thread balancing in `find_tracks` kernel which runs kalman updater over measurments

In the current algorithm, each thread only iterates over measurements associated with a single track, leading to an imbalance between threads. This PR makes each thread iterate over the exact same number of measurements (which might cover more than a single track) by adding a new configuration parameter (`n_measurements_per_thread`) and replacing `count_threads` kernel with `count_measurements` kernel. The concept is described in the below figure.

image

With my local machine, the main branch and this PR takes  575 and 582 ms for track finding, respectively (8000 muons per event, 10 events in total, single precision). Please note that ~90% of run time is occupied by `propagate_to_next_surface` kernel so the changes in run time is not very apparent: 

Update: After removing some unused variables, track finding time is reduced to 577 ms


krasznaa (approved):
I'm happy with this. I have to admit that I only glanced over the implementation, but didn't see anything obviously bad at least. :wink:

PR #494, opened 2023-11-21
Update detray version to v49

beomki-yeo:


niermann999:
Since there are material maps in the geometry now, you will need an accessor that overloads on material slab vectors vs material grids, whenever you access the material store in a visitor, e.g.:
https://github.com/acts-project/detray/blob/fcb1c85ba8bdb8799b2f42a4bba37b1ef35858dd/core/include/detray/propagator/actors/pointwise_material_interactor.hpp#L74-L75

beomki-yeo:
I don't think this is a good design. We should define accessors globally instead of doing this for every kernels

beomki-yeo:
Detray vesion is bumped to 49

tests/common/tests/kalman_fitting_wire_chamber_test.hpp
```diff
@@ -67,6 +67,7 @@ class KalmanFittingWireChamberTests : public KalmanFittingTests {
         vecmem::host_memory_resource host_mr;
 
         detray::wire_chamber_config wire_chamber_cfg;
+        wire_chamber_cfg.mapped_material(detray::vacuum<scalar>());
```
beomki-yeo, line 70: @niermann999  2 GeV test of wire chamber still fails with (berylium + 100 um thickness) setup so I decided to make portal materials empty :/

niermann999: Ok

niermann999 (approved):

PR #496, opened 2023-11-24
Add an option for charge and corresponding tests

beomki-yeo:
Including option and tests for positively charged particles as well


krasznaa (approved):
I did not appreciate how many flippin' tests we had already. But yeah, this looks good, and warranted.

PR #498, opened 2023-11-24
Remove unused geometry writing part

beomki-yeo:
Two changes:
(1) Ckf tests were writing geometry json again (`SetUp()` writes geometry json file first)
(2) Remove `remove_all` to keep simulated file


krasznaa (approved):
Sure. Looks simple enough. 👍

PR #503, opened 2023-11-27
WIP: Add event tree writer

beomki-yeo:


beomki-yeo:
@krasznaa Could you take a look?

krasznaa (changes requested):
:thinking: I believe the code is a bit too over-complicated right now.

But let's start at the beginning. Let's not use a generic name like `event_tree_writer` for this class. It is doing a very specific thing. It writes data about the truth and reconstructed tracks. So let's call it something like: `traccc::root::tracking_performance_file_writer`. (I'm very open to a better name. My point is just to give it a little more concrete of a name.)

Then... I am on board with the general API of the current `traccc::event_tree_writer` class. But its implementation is unnecessarily complicated. Just make `traccc::details::event_tree_writer_data` (or whatever you rename it to) do the heavy lifting itself. There's no need for an extra layer of classes under that data class. It can just be the "data class", which itself would both hold the primitive variables that the output `TTree` is written with, and the code that deals with the `TTree`.

The "cache concept" for this writer is just not necessary. :thinking: 

But sure, with a better name, and some simplification, I'm very open to adding this to the project.

beomki-yeo:
Thanks for the comments! I simplified the code now with different name

PR #507, opened 2023-12-18
Better VSCode Integration, main branch (2023.12.18.)

krasznaa:
Before doing some slightly more serious work in this project once again, I wanted to make my IDE experience a little better. :wink: This PR is very much a follow-up from https://github.com/acts-project/algebra-plugins/pull/109 and of some things learned from https://github.com/acts-project/vecmem/pull/243. 🤔

First off, introduced some CMake presets to the project. To allow users to use VSCode's built-in CMake support pretty much out of the box. (Just building this project with its default options is not appropriate for code development in most cases. And passing custom options to VSCode's build "without presets" is a chore.)

Then, added a `.vscode/setings.json` file to be able to edit `.sycl` files conveniently. I know this is a contentious issue a bit, as VSCode likes to update this file, so we have to keep track of not committing unwanted changes to it by accident, but that's how [vecmem](https://github.com/acts-project/vecmem) has been operating since forever. 🤔 So hopefully this project's developers will be able to adapt as well...

Finally, switched the GitLab CI to use the "cuda" preset, instead of hard-coding configuration parameters itself. (Very interested to see if that would work. As that test only runs on PRs.)

### DIFF
--- a/tests/io/test_binary.cpp
+++ b/tests/io/test_binary.cpp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2022 CERN for the benefit of the ACTS project
+ * (c) 2022-2023 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -29,7 +29,7 @@ TEST(io_binary, cell) {
 
     // Set event configuration
     const std::size_t event = 0;
-    const std::string cells_directory = "tml_full/ttbar_mu200/";
+    const std::string cells_directory = "tml_full/ttbar_mu100/";
 
     // Memory resource used by the EDM.
     vecmem::host_memory_resource host_mr;
@@ -171,7 +171,7 @@ TEST(io_binary, spacepoint) {
 TEST(io_binary, measurement) {
     // Set event configuration
     const std::size_t event = 0;
-    const std::string measurements_directory = "tml_full/ttbar_mu200/";
+    const std::string measurements_directory = "tml_full/ttbar_mu300/";
 
     // Memory resource used by the EDM.
     vecmem::host_memory_resource host_mr;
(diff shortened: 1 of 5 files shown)


krasznaa:
This PR managed to find yet another race condition in our tests. 😦

https://github.com/acts-project/traccc/actions/runs/7252257817/job/19756339748?pr=507

We'll really need to make these tests, and the I/O code in general better. But for now this hacky fix should be good enough...

krasznaa:
@beomki-yeo, @niermann999, please wave this one through. It helps my code editing quite a bit...

tests/io/test_binary.cpp
```diff
@@ -29,7 +29,7 @@ TEST(io_binary, cell) {
 
     // Set event configuration
     const std::size_t event = 0;
-    const std::string cells_directory = "tml_full/ttbar_mu200/";
+    const std::string cells_directory = "tml_full/ttbar_mu100/";
```
niermann999, line 32: Is this meant to be in this PR?

niermann999 (approved):

tests/io/test_binary.cpp
niermann999: OK, nvm. I did not read your comment

krasznaa: As I wrote above, I came across a race condition in the CI.

Those "binary tests" are writing temporary files into the data directory. (Which is pretty bad, as the data directory should really be treated as read-only...) And both the "cell" and "measurement" test write a file with the cell module information. With the same name. So under just the right circumstances the two can overwrite each others' files. (The two tests write exactly the same file. So for the error to appear, they have to ask the OS to flush their writes exactly at the same time in some weird way.)

Now... I could've opened a separate PR for the test fix. But since it was affecting the CI of this PR, I thought it would be fine to just slap it in here. (Since the fix/workaround is just 2 lines in the end.)

PR #509, opened 2023-12-19
Cell Reading with Detray Geometry, main branch (2023.12.19.)

krasznaa:
Made the CPU algorithm sequence use Detray geometries. This is needed for processing data produced for the ODD geometry.

To be able to read those files correctly, taught the CSV cell reading code how to switch from Acts to Detry identifiers in the cell module data on the fly. To make the simulation files usable with the identifiers generated by Detray.

This is all very much a hack at the moment, in order to try to make the full (host) chain work on the ODD geometry.

PR #510, opened 2023-12-21
update to detray v51

beomki-yeo:


krasznaa:
@beomki-yeo, @niermann999, is this the update necessary for the full-chain? 🤔

It seems to be just a `p_mag` -> `p_tot` issue. Will you fix it?

krasznaa (approved):
As long as the CI is happy, this should go in. As it is required for #509. :wink:

krasznaa (changes requested):
Unfortunately the test failures are not something that I could/would fix. 😦

beomki-yeo:
Probably the changes in the `detray::navigator` are conflicting with `traccc::kalman_actor`. @niermann999 any thought?

beomki-yeo:
It might be something else because ckf does not use kalman actor as far as I know

beomki-yeo:
I believe this will fix the errors in the CI, but I cannot guarantee that this will work for ODD.

beomki-yeo:
@krasznaa 

If you think the ODD test is not working in the current setup, I suggest to remove the `detray::stepper_default_policy` from the `rk_stepper_type`. This change should be available for CPU fitting and finding algorithm. 
```
-    using rk_stepper_type = detray::rk_stepper<
-        b_field_t::view_t, typename host_detector_type::transform3,
-        detray::constrained_step<>, detray::stepper_default_policy>;

+   using rk_stepper_type = detray::rk_stepper<
+      b_field_t::view_t, typename host_detector_type::transform3,
+     detray::constrained_step<>>;
```

krasznaa:
Unfortunately as long as the tests keep failing, the PR definitely doesn't go in. 🤔

If anything, your latest changes only made things worse. 😦 Previously the failing tests complained about efficiencies going down a little bit. Now the tests don't find any tracks anymore. 😦

At the same time this gives me hope that the complete lack of track finding in #509 could potentially be fixed with a relatively small change.

krasznaa:
Actually, never mind. Even in #509 some of the tests complain about a complete lack of track finding. 🤔

```
[----------] 1 test from KalmanFitTelescopeValidation2/KalmanFittingTelescopeTests
[ RUN      ] KalmanFitTelescopeValidation2/KalmanFittingTelescopeTests.Run/0
WARNING: mask store has empty collection no. 1
WARNING: mask store has empty collection no. 2
WARNING: mask store has empty collection no. 3
WARNING: mask store has empty collection no. 4
WARNING: mask store has empty collection no. 5
WARNING: mask store has empty collection no. 6
WARNING: mask store has empty collection no. 7
WARNING: material store has empty collection no. 0
WARNING: material store has empty collection no. 1
WARNING: material store has empty collection no. 2
WARNING: material store has empty collection no. 4
WARNING: acceleration data structures store has empty collection no. 1
WARNING: acceleration data structures store has empty collection no. 2
WARNING: acceleration data structures store has empty collection no. 3
WARNING: acceleration data structures store has empty collection no. 4
Detector check: OK
/__w/traccc/traccc/tests/cpu/test_kalman_fitter_telescope.cpp:131: Failure
Expected equality of these values:
  track_candidates.size()
    Which is: 0
  n_truth_tracks
    Which is: 100
[  FAILED  ] KalmanFitTelescopeValidation2/KalmanFittingTelescopeTests.Run/0, where GetParam() = ("telescope_100_GeV_0_phi_muon", { 0, 0, 0 }, { 0, 0, 0 }, { 100, 100 }, { 0, 0 }, { 0, 0 }, -1, 100, 100) (107 ms)
[----------] 1 test from KalmanFitTelescopeValidation2/KalmanFittingTelescopeTests (107 ms total)

[==========] 1 test from 1 test suite ran. (107 ms total)
```

Still, I can reasonably hope that whatever broke those tests, is what makes `traccc_seq_example` not find any tracks either. 🤔

beomki-yeo:
@krasznaa  I adjusted the input parameter to pass the CI test.

tests/cpu/test_kalman_fitter_wire_chamber.cpp
```diff
@@ -149,6 +149,15 @@ TEST_P(KalmanFittingWireChamberTests, Run) {
...
 
+            const auto& seed = track_candidates[i_trk].header;
+            const auto& cands = track_candidates[i_trk].items;
+
+            if (cands.size() < 20) {
+
+                std::cout << "Track momentum (p, pT): " << seed.p() << " "
+                          << seed.pT() << " " << seed.qop() << " " << seed.phi()
+                          << " " << seed.theta() << std::endl;
+            }
```
krasznaa, line 160: Was this meant to stay in, or was this just a debug measure?

tests/cuda/test_ckf_toy_detector.cpp
```diff
@@ -249,8 +249,8 @@ INSTANTIATE_TEST_SUITE_P(
 INSTANTIATE_TEST_SUITE_P(
     CkfToyDetectorValidation1, CkfToyDetectorTests,
     ::testing::Values(std::make_tuple(
-        "toy_n_particles_10000", std::array<scalar, 3u>{0.f, 0.f, 0.f},
+        "toy_n_particles_1000", std::array<scalar, 3u>{0.f, 0.f, 0.f},
         std::array<scalar, 3u>{0.f, 0.f, 0.f},
         std::array<scalar, 2u>{1.f, 100.f}, std::array<scalar, 2u>{-3.f, 3.f},
         std::array<scalar, 2u>{0.f, 2.0f * detray::constant<scalar>::pi}, -1.f,
-        10000, 1)));
+        1000, 1)));
```
krasznaa, line 256: :confused: Why did you reduce the number of particles? The other parameter changes I don't find too worrisome. (Maybe because I don't fully understand them.) But if we cannot reconstruct 10k particles anymore, that's a bit worrisome...

krasznaa (commented):
I'm generally on board, just some simple questions...

tests/cuda/test_ckf_toy_detector.cpp
beomki-yeo: Probably it is Ok with CI.
But my laptop (rtx 2070) does not pass the test due to OOM error, which I don't know why

beomki-yeo: Regarding that this is not supposed to happen, the current CKF performance with GPU will be pretty bad compared to what I have shown before. 

We will definitely need to fix it soon, but let us focus on running ODD at the moment

tests/cpu/test_kalman_fitter_wire_chamber.cpp
beomki-yeo: Good catch. I removed it

tests/cuda/test_ckf_toy_detector.cpp
krasznaa: Hmm... OOM. Just what I like to hear... :frowning:

Fine, let's reduce the number of particles for now then.

tests/cpu/test_kalman_fitter_wire_chamber.cpp
```diff
@@ -193,7 +193,7 @@ INSTANTIATE_TEST_SUITE_P(
     ::testing::Values(std::make_tuple(
         "wire_10_GeV_muon", std::array<scalar, 3u>{0.f, 0.f, 0.f},
         std::array<scalar, 3u>{0.f, 0.f, 0.f},
-        std::array<scalar, 2u>{10.f, 10.f}, std::array<scalar, 2u>{-1.f, 1.f},
+        std::array<scalar, 2u>{10.f, 10.f}, std::array<scalar, 2u>{-0.3f, 0.3f},
```
beomki-yeo, line 196: @krasznaa Just commenting on this change, the eta range is changing from [-1,1] -> [-0.3, 0.3].

Yeah so nothing has been fixed yet and I am avoiding to face the problem in this PR

tests/cuda/test_ckf_toy_detector.cpp
niermann999: I increased the bin capacity in the grids, which will overallocate memory now. I had not expected this to cause an OOM for the toy detector though. I has only 18 grids...

beomki-yeo: It might not be from detray v51 as CI works OK. Instead it could be triggered by other PRs in the past.

beomki-yeo: Therefore, I think the only problem we have is the wire chamber navigation - but anyway it is not even working well in detray helix navigation so it is not very surprising...

PR #511, opened 2024-01-22
Hip updates main

StewMH:
As proposed by @krasznaa in https://github.com/acts-project/vecmem/pull/261, extend the __host__ and __device__ macros for HIP.

Clang wanted TRACCC_HOST_DEVICE to be added to the function definition as well in a couple of places. I could not get compilation to work with gcc. Wasn't clear to me if this was supported at all.

I now get as far as a puzzling linker error:
```
[100%] Linking CXX executable ../../../bin/traccc_seeding_example_alpaka
/cvmfs/sft.cern.ch/lcg/releases/binutils/2.39-393d1/x86_64-centos7/bin/ld: CMakeFiles/traccc_seeding_example_alpaka.dir/seeding_example_alpaka.cpp.o: relocation R_X86_64_32S against symbol `_ZTVN6vecmem4copyE' can not be used when making a PIE object; recompile with -fPIE
/cvmfs/sft.cern.ch/lcg/releases/binutils/2.39-393d1/x86_64-centos7/bin/ld: failed to set dynamic section sizes: bad value
```
but I think this is separate and that these changes are safe.

### DIFF
--- a/core/include/traccc/definitions/qualifiers.hpp
+++ b/core/include/traccc/definitions/qualifiers.hpp
@@ -8,25 +8,25 @@
 
 #pragma once
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) || defined(__HIP__)
 #define TRACCC_DEVICE __device__
 #else
 #define TRACCC_DEVICE
 #endif
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) || defined(__HIP__)
 #define TRACCC_HOST __host__
 #else
 #define TRACCC_HOST
 #endif
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) || defined(__HIP__)
 #define TRACCC_HOST_DEVICE __host__ __device__
 #else
 #define TRACCC_HOST_DEVICE
 #endif
 
-#if defined(__CUDACC__)
+#if defined(__CUDACC__) || defined(__HIP__)
 #define TRACCC_ALIGN(x) __align__(x)
 #else
 #define TRACCC_ALIGN(x) alignas(x)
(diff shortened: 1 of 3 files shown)


krasznaa (commented):
I am **very** surprised about the `TRACCC_HOST_DEVICE` additions. 😕 These functions are clearly not needed by the CUDA code. (Otherwise that build would've failed.) So how are they needed by Alpaka when using the HIP backend? 😕

I knew for a while already that a couple more "common functions" will need to become `TRACCC_HOST_AND_DEVICE` from their current `TRACCC_DEVICE` status when finally we pick those up with Kokkos and Alpaka. But I did not expect any function, which was so far not marked as having to be a "device function" at all, to need such additions.

I guess I'll need to have a closer look at your Alpaka code... 🤔

krasznaa:
> Hi Attila,
> 
> I can reproduce the host device issue standalone: https://godbolt.org/z/qcrcM5E85
> 
> You can see that rocm needs the declaration to match the definition (i.e. `__host__` and `__device__` in both), while nvcc does not.

Ahh... I didn't appreciate that what you're doing here is to fix the inconsistency between the declaration and implementation of these functions. 🤔 Yes, that's definitely a bug in our current code...

krasznaa (changes requested):
As long as you run the changes through `clang-format`, I'll be okay with this. 😉

I.e.

```
docker run -it --rm -v $PWD:$PWD -w $PWD --user "$(id -u):$(id -g)" ghcr.io/acts-project/format10:v11
./traccc/.github/check_format.sh ./traccc/
```

StewMH:
@krasznaa should I merge from main again?

StewMH:
Merged into main to try out my new CI bridge privileges in any case

krasznaa:
My own preference is to use rebasing in such cases, but this works fine as well...

beomki-yeo (approved):
Looks good

beomki-yeo:
I don't know how to remove the WIP block :(

PR #512, opened 2024-02-06
feat: Update to detray v0.53.0

niermann999:
Update to latest detray version, which contains among other things a refactoring of the RKN stepper and the grids, as well as local navigation


niermann999:
Not absolutely sure why the CUDA tests started breaking, but I added a fix that worked locally in acts-project/detray#655

beomki-yeo (approved):

PR #513, opened 2024-02-07
Update algebra-plugins and vecmem versions

StewMH:
Update algebra-plugins and vecmem to latest tags. Largely important for Alpaka HIP developments.

PR #518, opened 2024-02-13
traccc::device::make_barcode_sequence Fix, main branch (2024.02.14.)

krasznaa:
Fixed the logic in the device function, after finding a bug in it during the debugging of #513.

While we put the "unique measurements" at the beginning of the `uniques` vector/view, the size of the vector/view is not updated. 🤔 It is allowed to be larger than the size of the barcodes vector. (Which will be exactly the size of the number of unique measurements.)

Later on we may want to do this in a smarter way, making the measurement buffer update its size properly. But for now this should be good enough. 🤔


beomki-yeo (approved):
Makes sense

PR #520, opened 2024-02-21
Fix chi2 calculation for annulus measurements

shimasnd:
Update of `track_state.hpp` to get annulus measurements properly in track_state, otherwise chi2 is always zero for annulus measurements.

### DIFF
--- a/core/include/traccc/edm/track_state.hpp
+++ b/core/include/traccc/edm/track_state.hpp
@@ -77,10 +77,10 @@ struct track_state {
                       "The measurement dimension should be 1 or 2");
 
         matrix_type<D, 1> ret;
-	if(m_measurement.subs.get_indices()[0]==0)	
-	  matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
-	else
-	  matrix_operator().element(ret, 0, 0) = m_measurement.local[1];
+        if (m_measurement.subs.get_indices()[0] == 0)
+            matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
+        else
+            matrix_operator().element(ret, 0, 0) = m_measurement.local[1];
 
         if constexpr (D == 2u) {
             matrix_operator().element(ret, 1, 0) = m_measurement.local[1];
@@ -95,10 +95,10 @@ struct track_state {
                       "The measurement dimension should be 1 or 2");
 
         matrix_type<D, D> ret;
-	if(m_measurement.subs.get_indices()[0]==0)	
-	  matrix_operator().element(ret, 0, 0) = m_measurement.variance[0];
-	else
-	  matrix_operator().element(ret, 0, 0) = m_measurement.variance[1];
+        if (m_measurement.subs.get_indices()[0] == 0)
+            matrix_operator().element(ret, 0, 0) = m_measurement.variance[0];
+        else
+            matrix_operator().element(ret, 0, 0) = m_measurement.variance[1];
 
         if constexpr (D == 2u) {
             matrix_operator().element(ret, 0, 1) = 0.f;
--- a/core/include/traccc/edm/track_state.hpp
+++ b/core/include/traccc/edm/track_state.hpp
@@ -91,6 +91,10 @@ struct track_state {
             assert(
                 "The measurement index out of e_bound_loc0 and e_bound_loc1 "
                 "should not happen.");
+            matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
+            if constexpr (D == 2u) {
+                matrix_operator().element(ret, 1, 0) = m_measurement.local[1];
+            }
         }
 
         return ret;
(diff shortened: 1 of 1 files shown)


core/include/traccc/edm/track_state.hpp
```diff
@@ -77,7 +77,11 @@ struct track_state {
                       "The measurement dimension should be 1 or 2");
 
         matrix_type<D, 1> ret;
-        matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
+        if (m_measurement.subs.get_indices()[0] == 0)
```
krasznaa, line 80: You should add a line of documentation that explains what this check is actually doing. Since I don't know this code, this line looks like checking some "magic number". :frowning:

krasznaa (changes requested):
@beomki-yeo will have to say whether the change is good or not. Formally, it looks fine. But some comments are desperately needed in the code for the lay reader.

core/include/traccc/edm/track_state.hpp
beomki-yeo: I had to rewind the logic from scratch...
What I think now is that we should not touch `track_state` at all

@shimasnd Have you checked the matrix elements of the projection matrix? https://github.com/acts-project/traccc/blob/fd145d8fa3b5dac40fa88e449d1d235f0dff6b85/core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp#L74

`H` should be [0 1 0 0 0 0] for Annulus measurement so that  `H * predicted_vec` becaomes [e_bound_loc1], which is phi measurement.

beomki-yeo: No I think actually this PR is doing the right thing as the following line should bring the `e_bound_loc1` component when `subs.get_indices()[0] == 1`.

https://github.com/acts-project/traccc/blob/fd145d8fa3b5dac40fa88e449d1d235f0dff6b85/core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp#L77-L78

I guess we should be generous about using the magic number here but we can be more explicit..

```c++
if (m_measurement.subs.get_indices()[0] == 0){
            matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
            if constexpr (D == 2u) {
                 matrix_operator().element(ret, 1, 0) = m_measurement.local[1];
            }
}
else if (m_measurement.subs.get_indices()[0] == 1){
            matrix_operator().element(ret, 0, 0) = m_measurement.local[1];
            if constexpr (D == 2u) {
                 matrix_operator().element(ret, 1, 0) = m_measurement.local[0];
            }
}
```

And the same goes for variances.

shimasnd:
Codes modified according to Beomki's comments. Also modified to use `e_bound_loc0` and `e_bound_loc1` instead of 0 and 1 when checking the value of `subs.get_indices()[0]`. I guess this is more proper, if I correctly understand how subspace indices are set in other places.

krasznaa:
I'll leave it to @beomki-yeo to give the final blessing on this one. I'll approve it once Beomki is happy with it. (I think technically I'll also need to approve it now, after requesting changes on it previously...)

core/include/traccc/edm/track_state.hpp
```diff
@@ -77,10 +77,23 @@ struct track_state {
...
+            matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
+            if constexpr (D == 2u) {
+                matrix_operator().element(ret, 1, 0) = m_measurement.local[1];
+            }
+        } else if (m_measurement.subs.get_indices()[0] == e_bound_loc1) {
+            matrix_operator().element(ret, 0, 0) = m_measurement.local[1];
+            if constexpr (D == 2u) {
+                matrix_operator().element(ret, 1, 0) = m_measurement.local[0];
+            }
+        } else {  // same as e_bound_loc0
```
beomki-yeo, line 90: I think we don't address such case at the moment
Let us put the warning in the if block:
```c++
assert("The measurement index out of e_bound_loc0 and e_bound_loc1 should not happen.");
```

core/include/traccc/edm/track_state.hpp
```diff
@@ -91,11 +104,34 @@ struct track_state {
...
+
+            matrix_operator().element(ret, 0, 0) = m_measurement.variance[1];
+            if constexpr (D == 2u) {
+                matrix_operator().element(ret, 0, 1) = 0.f;
+                matrix_operator().element(ret, 1, 0) = 0.f;
+                matrix_operator().element(ret, 1, 1) =
+                    m_measurement.variance[0];
+            }
+        } else {  // same as e_bound_loc0
+
```
beomki-yeo, line 127: Same comment

```suggestion
assert("The measurement index out of e_bound_loc0 and e_bound_loc1 should not happen.");
```

beomki-yeo (approved):
Looks good to me

beomki-yeo (approved):

PR #516, opened 2024-02-12
Reduce the max track size

beomki-yeo:
This will resolve the OOM issue

PR #522, opened 2024-02-26
Fix the default candidate range

beomki-yeo:
Fix a bug that causes OOM
This should have been fixed in #516 already


krasznaa (approved):
Ouch... 🤦

But sure, this looks safe and correct.

PR #523, opened 2024-02-29
feat: update to detray v0.55.0

niermann999:
Update to detray v0.55.0 and add extra some includes

### DIFF
--- a/core/include/traccc/edm/track_state.hpp
+++ b/core/include/traccc/edm/track_state.hpp
@@ -14,7 +14,7 @@
 #include "traccc/edm/track_candidate.hpp"
 
 // detray include(s).
-#include "detray/propagator/navigator.hpp"
+#include "detray/navigation/navigator.hpp"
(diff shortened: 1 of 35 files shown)


.gitignore
```diff
@@ -33,3 +33,5 @@
 
 # VSCode build output.
 out/
+data/
```
krasznaa, line 36: Nope. The `data/` directory must not be ignored. We continue to have the data download/creation scripts in there...

krasznaa (changes requested):
The code updates I'm okay with. But the `.gitignore` change needs to be removed from this PR. If we need to change that file for some reason, let's do so in a separate PR.

.gitignore
niermann999: Ah, oops, I do this locally everytime, since otherwise code searches in vscode take ages. Will revert

niermann999: Done

krasznaa (approved):
The code changes look okay as far as I can tell.

PR #526, opened 2024-03-12
Thrust Update, main branch (2024.03.12.)

krasznaa:
While trying to build the project on Ubuntu 22.04 with CUDA 12.4 + oneAPI 2024.0.1, I ran into the following issues:

```
[ 95%] Building CUDA object tests/cuda/CMakeFiles/traccc_test_cuda.dir/test_copy.cu.o
/home/krasznaa/software/cuda/12.4.0/x86_64-ubuntu2204/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/__iterator/unreachable_sentinel.h:49:1: error: an attribute list cannot appear here
   49 | [[nodiscard]] friend constexpr bool 
      | ^~~~~~~~~~~~~
/home/krasznaa/software/cuda/12.4.0/x86_64-ubuntu2204/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/__iterator/unreachable_sentinel.h:57:1: error: an attribute list cannot appear here
   57 | [[nodiscard]] friend constexpr bool 
      | ^~~~~~~~~~~~~
/home/krasznaa/software/cuda/12.4.0/x86_64-ubuntu2204/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/__iterator/unreachable_sentinel.h:64:1: error: an attribute list cannot appear here
   64 | [[nodiscard]] friend constexpr bool 
      | ^~~~~~~~~~~~~
/home/krasznaa/software/cuda/12.4.0/x86_64-ubuntu2204/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/__iterator/unreachable_sentinel.h:71:1: error: an attribute list cannot appear here
   71 | [[nodiscard]] friend constexpr bool 
      | ^~~~~~~~~~~~~
4 errors generated.
```

After some amount of debugging, I found that [Thrust 1.5.0](https://github.com/NVIDIA/thrust/releases/tag/1.15.0), which we use at the moment by default, cannot work with these compilers anymore. :frowning: In the end it's not super unexpected, since that version of Thrust was released back in 2021. :thinking:

When updating to [Thrust 2.1.0](https://github.com/NVIDIA/thrust/releases/tag/2.1.0), the current latest version, I had to change the following:
  - The tweaks in [CMakeLists.txt](CMakeLists.txt) for making sure that the Thrust headers would be treated as system headers, are very much tinkered around the exact CMake configuration of Thrust 1.15.0. I tried to update these at first, but since even with Clang I was not seeing any warnings from Thrust 2.1.0, in the end I just removed the tweaks all together. :thinking:
  - [finding_algorithm.cu](device/cuda/src/finding/finding_algorithm.cu) was missing an include, which didn't show up as an issue with the previous Thrust version.
  - Any CUDA source file that uses Thrust (2.1.0) needs to be built with [CUDA_SEPARABLE_COMPILATION](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_SEPARABLE_COMPILATION.html) enabled when using CUDA 11.X. So added this property on `traccc::cuda` and `traccc_test_cuda`. :thinking:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # TRACCC library, part of the ACTS project (R&D line)
 #
-# (c) 2021-2023 CERN for the benefit of the ACTS project
+# (c) 2021-2024 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
@@ -117,13 +117,6 @@ set( TRACCC_THRUST_OPTIONS "" CACHE STRING
    "Extra options for configuring how Thrust should be used" )
 mark_as_advanced( TRACCC_THRUST_OPTIONS )
 thrust_create_target( traccc::Thrust ${TRACCC_THRUST_OPTIONS} )
-# Make sure that Thrust/CUB headers are treated as system headers.
-get_target_property( _thrustIncludeDir Thrust::Thrust INTERFACE_INCLUDE_DIRECTORIES )
-get_target_property( _cubIncludeDir CUB::CUB INTERFACE_INCLUDE_DIRECTORIES )
-target_include_directories( traccc::Thrust
-   SYSTEM INTERFACE ${_thrustIncludeDir} ${_cubIncludeDir} )
-unset( _thrustIncludeDir )
-unset( _cubIncludeDir )
 
 # Set up TBB.
 option( TRACCC_SETUP_TBB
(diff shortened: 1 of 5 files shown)


stephenswat (approved):
Looks fine!

PR #530, opened 2024-04-04
Move to ACTS v27 and adapt test code

StewMH:
Move to ACTS v26, needed for GCC 13 and clang 17 (because of https://github.com/acts-project/acts/pull/2079)

Compiles for me but tests fail due to:
```
[ RUN      ] SeedingValidation/CompareWithActsSeedingTests.Run/0
unknown file: Failure
C++ exception with description "SeedFinderConfig not in ACTS internal units in SeedFinder" thrown in the test body.
```
which I don't know how to solve.

Note also that the atlas_cuts file is now just a copy-paste of https://github.com/acts-project/acts/blob/v25.0.0/Tests/UnitTests/Plugins/Cuda/Seeding/ATLASCuts.hpp


StewMH:
@CarloVarni would you be able to help?

CarloVarni:
> @CarloVarni would you be able to help?

Hi @StewMH that is because for every `Config` and `Options` obejcts you now have to call `toInternalUnits` to assure that the proper Acts units are used. 

See for instance: https://github.com/acts-project/acts/blob/v26.0.0/Examples/Algorithms/TrackFinding/src/SeedingAlgorithm.cpp#L33

krasznaa (approved):
I'm in favour. :thinking: After you rebased it, I'm happy for it to go in.

krasznaa:
Oh, and in case you end up clicking the merge button first, please squash these changes. 9 commits for relatively small changes is really a bit much. :thinking:

beomki-yeo:
So we still don't know what has been dropping the matching rate between ACTS and traccc seeding?

StewMH:
No, I didn't get any further. With v26 + https://github.com/acts-project/acts/pull/2132 I found it was a bit closer but still lower than before. I don't really have the experience in ACTS or traccc to figure out what the drop is from.

beomki-yeo (approved):
OK. Let's go with this for the moment.
Please rebase and squash your commits.

PR #422, opened 2023-06-20
Update acts version and seeding test

beomki-yeo:
It's not working very well. Any help?

```
Note: Google Test filter = *Compare*
[==========] Running 10 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 10 tests from SeedingValidation/CompareWithActsSeedingTests
[ RUN      ] SeedingValidation/CompareWithActsSeedingTests.Run/0
/home/beomki/projects/traccc/traccc/tests/cpu/compare_with_acts_seeding.cpp:347: Failure
The difference between seeds.size() and seedVector.size() is 4044, which exceeds seeds.size() * 0.001, where
seeds.size() evaluates to 17756,
seedVector.size() evaluates to 13712, and
seeds.size() * 0.001 evaluates to 17.756.
/home/beomki/projects/traccc/traccc/tests/cpu/compare_with_acts_seeding.cpp:348: Failure
Value of: seed_match_ratio > 0.999
  Actual: false
Expected: true
/home/beomki/projects/traccc/traccc/tests/cpu/compare_with_acts_seeding.cpp:461: Failure
The difference between acts_params.size() and traccc_params.size() is 4044, which exceeds acts_params.size() * 0.001, where
acts_params.size() evaluates to 13712,
traccc_params.size() evaluates to 17756, and
acts_params.size() * 0.001 evaluates to 13.712.
/home/beomki/projects/traccc/traccc/tests/cpu/compare_with_acts_seeding.cpp:463: Failure
Value of: params_match_ratio > 0.999
  Actual: false
Expected: true
Parameter matching ratio: 0.85216265916824341
```

PR #532, opened 2024-04-09
Draft: Update to ACTS v27

beomki-yeo:
@StewMH  
I think I attempt updating ACTS version in the past.
Maybe the efficiency dropped a little but the seeding comparison worked well
Feel free to steal the PR contents if you think useful


beomki-yeo:
Oh man! Actually it has opened here: https://github.com/acts-project/traccc/pull/422

PR #534, opened 2024-04-09
ODD Full Chain Reconstruction, main branch (2024.04.09.)

krasznaa:
Hahhh!!!! Code cleanup does have its use!

This here is an update of just `examples/run/cpu/seq_example.cpp`, on top of #533. (Once that one is merged in, this PR should become quite readable.)

I don't know exactly what changed since my last attempt. I suspect that https://github.com/acts-project/detray/pull/701 may have helped, but I'm not entirely sure. :thinking: Whatever it was, with the code in this branch, I just now did:

```
[bash][celeborn]:traccc > ./traccc/out/build/default/bin/traccc_seq_example --detector-file=geometries/odd/odd_geometry_detray.json --use-detray-detector --digitization-file=geometries/odd/odd-digi-geometric-config.json --input-directory=odd/muon100GeV-geant4/ 
Running Full Tracking Chain on the Host

>>> Detector Options <<<
  Detector file       : geometries/odd/odd_geometry_detray.json
  Material file       : 
  Surface rid file    : 
  Use detray::detector: yes
  Digitization file   : geometries/odd/odd-digi-geometric-config.json
>>> Input Data Options <<<
  Input data format             : csv
  Input directory               : odd/muon100GeV-geant4/
  Number of input events        : 1
  Number of input events to skip: 0
>>> Clusterization Options <<<
  Target cells per partition: 1024
>>> Track Seeding Options <<<
  None
>>> Track Finding Options <<<
  Track candidates range   : 3:100
  Maximum Chi2             : 30
  Maximum branches per step: 4294967295
>>> Track Propagation Options <<<
  Constraint step size : 3.40282e+38 [mm]
  Overstep tolerance   : -100 [um]
  Mask tolerance       : 15 [um]
  Search window        : 0 x 0
  Runge-Kutta tolerance: 0.0001
>>> Performance Measurement Options <<<
  - Run performance checks: no

WARNING: No material in detector
WARNING: No entries in volume finder
Detector check: OK
WARNING: No material in detector
WARNING: No entries in volume finder
Detector check: OK
==> Statistics ... 
- read    82092 cells from 11131 modules
- created 31587 measurements. 
- created 31587 space points. 
- created 5977 seeds
- found   4388 tracks
- fitted  4388 tracks
[bash][celeborn]:traccc >
```

Where of course the emphasis is on the last 2 lines. :laughing:

@beomki-yeo, let's talk tomorrow a little bit about the code. But generally, as it looks, we can forget about the debugging session that we had planned for. :partying_face:

PR #536, opened 2024-04-13
Fix ambiguity resolution algorithm: add unique measurement_id

SylvainJoube:
This is a fix for @krasznaa PR  [ODD Full Chain Reconstruction, main branch (2024.04.09.) #534](https://github.com/acts-project/traccc/pull/534).

The ambiguity resolution algorithm failed because every `measurement_id` were equal to zero, as they were not initialized when created from clusters during `measurement_creation`.

Since we are running sequentially on the CPU, I think iteratively giving a unique measurement id by incrementing a variable may be okay?

Output example:
```
Running Full Tracking Chain on the Host

>>> Detector Options <<<
  Detector file       : geometries/odd/odd_geometry_detray.json
  Material file       : 
  Surface rid file    : 
  Use detray::detector: yes
>>> Input Data Options <<<
  Input directory               : odd/muon100GeV-geant4/
  Number of input events        : 2
>>> Track Seeding Options <<<
  None
>>> Track Finding Options <<<
  Maximum Chi2             : 30
  Maximum branches per step: 4294967295
  Constraint step size : 3.40282e+38 [mm]
  Overstep tolerance   : -100 [um]
  Mask tolerance       : 15 [um]
  Search window        : 0 x 0
  Runge-Kutta tolerance: 0.0001
>>> Track Ambiguity Resolution Options <<<
  Run ambiguity resolution : yes
  Run performance checks: no

Detector check: OK
Detector check: OK
@greedy_ambiguity_resolution_algorithm: Iteration_count: 6417
@greedy_ambiguity_resolution_algorithm: Checking result validity...
@greedy_ambiguity_resolution_algorithm: OK 1/2: every removed track had at least one commun measurement with another track.
@greedy_ambiguity_resolution_algorithm: OK 2/2: each selected_track shares at most (_config.maximum_shared_hits - 1)(=0) measurement(s)
@greedy_ambiguity_resolution_algorithm: state.selected_tracks.size() = 1728
@greedy_ambiguity_resolution_algorithm: Iteration_count: 5869
@greedy_ambiguity_resolution_algorithm: Checking result validity...
@greedy_ambiguity_resolution_algorithm: OK 1/2: every removed track had at least one commun measurement with another track.
@greedy_ambiguity_resolution_algorithm: OK 2/2: each selected_track shares at most (_config.maximum_shared_hits - 1)(=0) measurement(s)
@greedy_ambiguity_resolution_algorithm: state.selected_tracks.size() = 1751
==> Statistics ... 
- read     163156 cells from 22437 modules
- created  63510 measurements. 
- created  63510 space points. 
- created  11806 seeds
- found    15765 tracks
- fitted   15765 tracks
- resolved 3479 tracks
```

### DIFF
--- a/core/include/traccc/clusterization/detail/measurement_creation_helper.hpp
+++ b/core/include/traccc/clusterization/detail/measurement_creation_helper.hpp
@@ -78,12 +78,11 @@ TRACCC_HOST inline void calc_cluster_properties(
 /// @param[in] cluster is the input cell vector
 /// @param[in] module is the cell module where the cluster belongs to
 /// @param[in] module_link is the module index
-/// @param[in] unique_mid unique measurement id, for ambiguity resolution
 ///
 TRACCC_HOST inline void fill_measurement(
     measurement_collection_types::host& measurements,
     const cell_collection_types::host& cluster, const cell_module& module,
-    const unsigned int module_link, std::size_t unique_mid) {
+    const unsigned int module_link) {
 
     // To calculate the mean and variance with high numerical stability
     // we use a weighted variant of Welford's algorithm. This is a
@@ -117,7 +116,7 @@ TRACCC_HOST inline void fill_measurement(
         // @todo add variance estimation
 
         // For the ambiguity resolution algorithm, give a unique measurement ID
-        m.measurement_id = unique_mid;
+        m.measurement_id = measurements.size();
 
         measurements.push_back(std::move(m));
     }
--- a/core/include/traccc/clusterization/detail/measurement_creation_helper.hpp
+++ b/core/include/traccc/clusterization/detail/measurement_creation_helper.hpp
@@ -78,11 +78,12 @@ TRACCC_HOST inline void calc_cluster_properties(
 /// @param[in] cluster is the input cell vector
 /// @param[in] module is the cell module where the cluster belongs to
 /// @param[in] module_link is the module index
+/// @param[in] unique_mid unique measurement id, for ambiguity resolution
 ///
 TRACCC_HOST inline void fill_measurement(
     measurement_collection_types::host& measurements,
     const cell_collection_types::host& cluster, const cell_module& module,
-    const unsigned int module_link) {
+    const unsigned int module_link, std::size_t unique_mid) {
 
     // To calculate the mean and variance with high numerical stability
     // we use a weighted variant of Welford's algorithm. This is a
@@ -114,6 +115,9 @@ TRACCC_HOST inline void fill_measurement(
             m.variance + point2{pitch[0] * pitch[0] / static_cast<scalar>(12.),
                                 pitch[1] * pitch[1] / static_cast<scalar>(12.)};
         // @todo add variance estimation
+
+        // For the ambiguity resolution algorithm, give a unique measurement ID
+        m.measurement_id = unique_mid;
 
         measurements.push_back(std::move(m));
     }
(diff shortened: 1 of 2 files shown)


core/src/clusterization/measurement_creation.cpp
```diff
@@ -48,7 +50,10 @@ measurement_creation::output_type measurement_creation::operator()(
         const auto &module = modules.at(module_link);
 
         // Fill measurement from cluster
-        detail::fill_measurement(result, cluster, module, module_link);
+        // Give each measurement a unique ID, for the ambiguity resolution
+        // algorithm
+        detail::fill_measurement(result, cluster, module, module_link,
+                                 ++unique_measurement_id);
```
krasznaa, line 56: :thinking: Should the IDs start from `0` (my preference) or `1` (what the code does currently)? To start from zero, you'd need:

```suggestion
                                 unique_measurement_id++);
```

core/include/traccc/clusterization/detail/measurement_creation_helper.hpp
```diff
@@ -115,6 +116,9 @@ TRACCC_HOST inline void fill_measurement(
                                 pitch[1] * pitch[1] / static_cast<scalar>(12.)};
         // @todo add variance estimation
 
+        // For the ambiguity resolution algorithm, give a unique measurement ID
+        m.measurement_id = unique_mid;
```
krasznaa, line 120: Instead of any other change, just make this line into:

```suggestion
        m.measurement_id = measurements.size();
```

? This code is currently not shared with the GPU implementation at all (something that I'll want to improve on a little later on), so there's no real need to get the unique ID from the outside into this function. :thinking:

krasznaa (changes requested):
Thanks a lot for the investigation! I'm very happy that the issue turned out to be a relatively easy one to fix.

I'll spend time next week on making the GPU chain work, which will include some re-shuffling of the clusterization code. Which will probably modify this part as well. But even with that, let's get this fixed as-is. I'll worry about doing this correctly on the device side. :thinking:

core/src/clusterization/measurement_creation.cpp
SylvainJoube: Thanks for your review!

Good luck for making the GPU chain work next week. I will try to make an implementation of the greedy ambiguity solver with SYCL in the coming weeks, but I'm not sure it'll be efficient, even parallelized with small clusters.

SylvainJoube:
Good luck getting the GPU chain to work next week. I'll try to implement the greedy ambiguity solver with SYCL in the coming weeks, but I'm not sure how efficient it will be, even parallelized with small clusters.

I will check the greedy ambiguity resolution code again if anything breaks after your changes!

core/src/clusterization/measurement_creation.cpp
SylvainJoube: Using `.size()` as you mentioned solved this issue. I was a bit worried by the `measurement_id` 0, since it can be confused with an uninitialized value, but it doesn't really matter, I'll implement an additional check on the greedy solver, to warn the user if there are too many zeroes as input IDs.

core/include/traccc/clusterization/detail/measurement_creation_helper.hpp
SylvainJoube: You are absolutely right, that's better this way, and only requires to change a single line! :)

krasznaa (approved):
:+1:

PR #537, opened 2024-04-13
Use 3:3 for default search window

beomki-yeo:
Let's take a good number of search window by default

### DIFF
--- a/examples/options/include/traccc/options/track_propagation.hpp
+++ b/examples/options/include/traccc/options/track_propagation.hpp
@@ -42,7 +42,7 @@ class track_propagation : public interface {
     std::ostream& print_impl(std::ostream& out) const override;
 
     /// Search window (helper variable)
-    value_array<unsigned int, 2> m_search_window = {0u, 0u};
+    value_array<unsigned int, 2> m_search_window = {3u, 3u};
 
 };  // class track_propagation
 


niermann999:
Note though, that 3x3 is not a good default for any detector that is converted from ACTS. It only works well for the toy detector and wire chamber. For all other detectors 0x0 is correct

krasznaa:
:thinking: Since I myself view the toy and wire chamber detectors as the testing ground, while the geometries coming from Acts (ODD, ITk) as the actual thing that we're trying to make to work, I'd vote for keeping the 0x0 window by default then...?

beomki-yeo:
I worry that people (including myself) running the examples _naively_ will keep complaining on the bad tracking efficiency

beomki-yeo:
BTW @niermann999  what does you mean with "it only works well toy and wire"? If we widen the search window, navigation with other detectors such as Itk or ODD will fail?

niermann999:
> BTW @niermann999 what does you mean with "it only works well for toy and wire"? If we widen the search window, navigation with other detectors such as Itk or ODD will fail?

In general you are right, that should only lead to an explosion in the number of candidates to test and hence possibly to worse computing performance. However, many candidates will be tested twice and since there is no duplicate removal yet, I don't know what the result would be (the stepper might abort on 0 step size?). The ODD and ITk grids are at least not meant to be used like this, to my understanding.

I am also not sure about the navigation cache sizes, yet. That depends on how we will implement the calculation of the maximal number of candidates that come from a grid of these detectors (i.e. just the maximal bin size?) or whether we set a global maximum size and let the navigator handle the cache exhaustion automatically (which I think would be more robust, and I started to implement a while ago but haven't found the time to finish). On the host, this won't be a problem, but on device this could lead to weird effects

Bottom line, for the moment at least it is safer not to do a neighbourhood lookup on grids that contain the entire neighbourhood already in a single bin.

Maybe you can set the window size in the examples only if the detector is named "toy_detector" or "wire_chamber"?

beomki-yeo:
Um... shouldn't be only one candiate picked up from the neighborhoods? I don't know why you are worrying about duplicate or explosion from this search window 🤔

niermann999:
> Um... shouldn't be only one candiate picked up from the neighborhoods? I don't know why you are worrying about duplicate or explosion from this search window 🤔

No, all surface that intersect successfully with the initial ray are put into the cache and then in subsequent calls the navigator goes essentially into guided navigation through these candidates. The grids in the ODD and ITk, however, register a single surface multiple times into surrounding bins, so that when all of those bins would get included into the lookup, the same surface will be tested multiple times and multiple copies of it will be put into the cache. And since an entire neighbourhood is registered in a single bin, the number of entries per bin is much higher than for the toy detector and wire chamber, making a neighbourhood lookup on the ACTS exported detectors more expensive

beomki-yeo:
OK I see. Thanks for the explanations!

PR #539, opened 2024-04-14
Improved error reporting for greedy solver

SylvainJoube:
A few tings:
- Disabled info messages from the greedy solver to avoid distracting the user from important warning and error messages.
- Added an error message when every measurement.measurement_id equals zero:

> ERROR: @greedy_ambiguity_resolution_algorithm: Measurements must have unique IDs. But here, each measurement has 0 as ID (measurement.measurement_id == 0). This may be solved by loading measurement_id from the appropriate file, or by assigning a unique ID to each new measurement during CCA.

- Added a warning message when too many measurements share the ID 0. It now says, for example:

> "WARNING @greedy_ambiguity_resolution_algorithm: 14% of input measurements have an ID equal to 0 (measurement.measurement_id == 0). This may be suspicious."

- Fixed a minor typo.

### DIFF
--- a/core/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cpp
+++ b/core/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cpp
@@ -32,24 +32,24 @@ namespace traccc {
 
 namespace {
 
-#define LOG_INFO(msg)                       \
-    if (_config.verbose_info) {             \
-        logger::info() << msg << std::endl; \
-    }
-
-#define LOG_FLOOD(msg)                       \
-    if (_config.verbose_flood) {             \
-        logger::flood() << msg << std::endl; \
-    }
-
 #define LOG_ERROR(msg)                       \
     if (_config.verbose_error) {             \
         logger::error() << msg << std::endl; \
     }
 
 #define LOG_WARN(msg)                       \
-    if (_config.verbose_error) {            \
+    if (_config.verbose_warning) {          \
         logger::warn() << msg << std::endl; \
+    }
+
+#define LOG_INFO(msg)                       \
+    if (_config.verbose_info) {             \
+        logger::info() << msg << std::endl; \
+    }
+
+#define LOG_DEBUG(msg)                       \
+    if (_config.verbose_debug) {             \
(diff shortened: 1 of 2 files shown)


core/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cpp
```diff
@@ -47,21 +47,31 @@ namespace {
...
-        std::cout << "@greedy_ambiguity_resolution_algorithm: ";
+        std::cout << "INFO: @greedy_ambiguity_resolution_algorithm: ";
         return std::cout;
     }
 
     static std::ostream& flood() {
-        std::cout << "@greedy_ambiguity_resolution_algorithm: ";
+        std::cout << "INFO: @greedy_ambiguity_resolution_algorithm: ";
         return std::cout;
     }
```
krasznaa, line 76: I'm generally not the biggest fan of all of these output messages. :thinking: But as long as we use them, why not name this as either a `DEBUG` or `VERBOSE` output level? To be a bit more in line with Athena / Python / etc. message logging conventions.

krasznaa (commented):
A little later I'll very much want to review / revise what we do for log messages in all of these algorithms. But for now I'm happy with printing at least fewer messages.

We should just think of whether "flood" messages, with `INFO:` message lines are the things that we want to go for in the meanwhile. :thinking:

core/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cpp
SylvainJoube: Thanks for your review :)

I made a mistake, I meant to say `FLOOD:` for "flood" messages. I completely agree with you, I think it's much better to have a common vocabulary and not confuse developers with too many different names referring to the same concept. I’ve changed `FLOOD` into `DEBUG` (since I only used these messages for debugging purposes). They are here to help me in case anything goes wrong and I need to debug the code, but I agree, the fewer messages we print on the terminal, the better, and there is no point in repeatedly saying "I'm doing this and that, everything works fine", better save space for the really important messages.

krasznaa (approved):
:+1:

Issue #540, opened 2024-04-14
Compilation with Double Precision Blocked


beomki-yeo:
Maybe we need to make detray configuration always float number?

```
[beomki@pc-54 traccc_build]$ cmake ../traccc -DTRACCC_CUSTOM_SCALARTYPE=double -DDETRAY_CUSTOM_SCALARTYPE=double
```

```
/home/beomki/projects/traccc/traccc/examples/run/cpu/truth_fitting_example.cpp: In function ‘int main(int, char**)’:
/home/beomki/projects/traccc/traccc/examples/run/cpu/truth_fitting_example.cpp:119:44: error: no match for ‘operator=’ (operand types are ‘detray::propagation::config<double>’ and ‘detray::propagation::config<float>’)
  119 |     fit_cfg.propagation = propagation_opts.config;
      |                                            ^~~~~~
In file included from /home/beomki/projects/traccc/traccc/core/include/traccc/fitting/fitting_config.hpp:12,
                 from /home/beomki/projects/traccc/traccc/core/include/traccc/fitting/fitting_algorithm.hpp:14,
                 from /home/beomki/projects/traccc/traccc/examples/run/cpu/truth_fitting_example.cpp:11:
/home/beomki/projects/traccc/traccc_build/_deps/detray-src/core/include/detray/propagator/propagation_config.hpp:18:8: note: candidate: ‘constexpr detray::propagation::config<double>& detray::propagation::config<double>::operator=(const detray::propagation::config<double>&)’
   18 | struct config {
      |        ^~~~~~
/home/beomki/projects/traccc/traccc_build/_deps/detray-src/core/include/detray/propagator/propagation_config.hpp:18:8: note:   no known conversion for argument 1 from ‘detray::propagation::config<float>’ to ‘const detray::propagation::config<double>&’
/home/beomki/projects/traccc/traccc_build/_deps/detray-src/core/include/detray/propagator/propagation_config.hpp:18:8: note: candidate: ‘constexpr detray::propagation::config<double>& detray::propagation::config<double>::operator=(detray::propagation::config<double>&&)’
/home/beomki/projects/traccc/traccc_build/_deps/detray-src/core/include/detray/propagator/propagation_config.hpp:18:8: note:   no known conversion for argument 1 from ‘detray::propagation::config<float>’ to ‘detray::propagation::config<double>&&’
[...]

PR #542, opened 2024-04-15
Double Precision Build Fix, main branch (2024.04.15.)

krasznaa:
Fixed the build of the project with double precision scalars. (Using `-DTRACCC_CUSTOM_SCALAR_TYPE=double` + `-DDETRAY_CUSTOM_SCALAR_TYPE=double`.)

This is to fix the issue described in #540. Which was introduced in #533.

It is only meant to be a temporary hack. The configuration types, much as @beomki-yeo proposed in #540, should become non-templated types in Detray in a future update. :thinking:


beomki-yeo (approved):
Thanks for the update. This can be rolled back once detray is fixed

PR #544, opened 2024-04-17
CUDA CI Update, main branch (2024.04.17.)

krasznaa:
This is an attempt at bringing the CUDA runtime tests back into a working order. I don't expect the first try to succeed (CUDA 11.6 seems just as broken for these tests as 11.5), but I have some further ideas as well...

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,39 +1,76 @@
 # TRACCC library, part of the ACTS project (R&D line)
 #
-# (c) 2022-2023 CERN for the benefit of the ACTS project
+# (c) 2022-2024 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
+# Stages of the CI build.
 stages:
   - build
   - test
 
+# Job-wide variables.
+variables:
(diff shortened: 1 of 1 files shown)


krasznaa:
@paulgessinger, are draft PRs not executed on the GitLab CI? :confused: I was hoping that I could keep this "cleanly" in a draft stage until I sorted out exactly how to fix the existing issue. :thinking:

paulgessinger:
@krasznaa No. This is intentional to manage the load on our CI VMs. With the current resources, I don't see a great way around this.

beomki-yeo:
Can we fix the CI soon? it is not nice to leave this unfixed for long time.

krasznaa:
I'll try to find a working setup finally...

krasznaa:
:frowning: Unfortunately the GPU CI machine is not set up to use CVMFS at the moment.

@paulgessinger, we'd need 2 things to make things "perfect":
  - The Acts GPU CI node would need to mount CVMFS into the containers that it starts;
  - We'd need to update the node to the latest NVIDIA GPU driver.

Let's maybe discuss offline what's the best way to make this happen. :thinking:

krasznaa:
Replaced by #616.

PR #545, opened 2024-04-17
Device Clusterization Reorganization, main branch (2024.04.17.)

krasznaa:
After #543, this is an even more ambitious PR for splitting `traccc::cuda::clusterization_algorithm` and `traccc::sycl::clusterization_algorithm` in two. (Re-)Introducing the `traccc::cuda::spacepoint_formation_algorithm` and `traccc::sycl::spacepoint_formation_algorithm` algorithms.

The biggest change is in `traccc::device::ccl_kernel`. Since now that function receives a resizable measurement buffer, it no longer deals with figuring out the index of the measurement that it should fill. It just uses vecmem's built-in ability for doing this sort of a thing.

I also:
  - Got rid of `traccc::cuda::experimental::clusterization_algorithm` and `traccc::sycl::experimental::clusterization_algorithm`. The updated algorithms now behave in the way that those "experimental" algorithms were meant to.
    * But I kept `traccc::cuda::experimental::spacepoint_formation` and `traccc::sycl::experimental::spacepoint_formation` around for now. Since the Detray usage in those is still something that I'll want to steal from. :wink:
  - Got rid of some minute duplications in the code. Introducing some new code that the CUDA and SYCL implementations would use from the same place.
  - Did some general code cleanup here and there.

With all of this, I simplified the device code in a few places a bit. So I wanted to see a little more in detail what happens to the performance of the code with these updates. Unfortunately the answer is a bit complicated... With the multi-threaded throughput tests, using 4 CPU threads, I see the following:

| | $\mu$ = 40 | $\mu$ = 140 | $\mu$ = 200 | $\mu$ = 300 |
|--|--|--|--|--|
| Old CUDA | 2285.54 | 1454.86 | 920.469 | 527.403 |
| New CUDA | 1685.72 | 1038.05 | 742.887 | 424.643 |
| Old SYCL | 1568.65 | 1084.27 | 746.129 | 431.996 |
| New SYCL | 1587.81 | 778.757 | 523.391 | 296.209 |

It indeed looks far from great. :frowning: However, when I do the same with the single-threaded throughput tests, I get:

| | $\mu$ = 40 | $\mu$ = 140 | $\mu$ = 200 | $\mu$ = 300 |
|--|--|--|--|--|
| Old CUDA | 1395.47 | 674.289 | 471.894 | 300.83 |
| New CUDA | 1407.7 | 674.373 | 478.964 | 300.309 |
| Old SYCL | 963.808 | 456.35 | 331.508 | 227.2 |
| New SYCL | 971.387 | 390.687 | 283.706 | 187.615 |

The CUDA numbers here look all good, but SYCL is still not great. :worried: But I don't fully understand why.

At first I was convinced that it would be the simplifications that I made in `traccc::device::ccl_kernel` that would be responsible for the slowdown. But not really. Even when I went back to the previous logic for how positions in the output container are chosen, the throughput numbers didn't change much.

I'll just give up for today... But let's discuss about this tomorrow...


krasznaa:
I'm still stumped. :frowning:

According to NSight Systems the new code does not execute kernels or memory copies any more slowly. If anything, it is even faster than the current code. :thinking:
  - With the current code I get this overall summary for running 1000 $\mu$=200 events on 4 CPU threads:

image

  - With the new code, for the same setup, I get:

image

Notice how the CCL kernel became faster with the updates. :thinking: Remember that, while I do introduce an additional algorithm with this re-shuffling, the number of kernels being executed remains the same!

So, since I could just not understand how the throughput test would become slower, while the kernels seemingly become faster, I ran the applications through VTune as well. And there is indeed a smoking gun there. :thinking: The current code achives the following CPU occupancy during the throughput test:

image

While this PR's code does this:

image

(The big spike at 1 thread is due to the initialization in both cases.)

However, I can't figure out what is causing this. :frowning: So if anybody has any good idea, I'm all ears. Since the "threading analysis" of VTune is just not revealing anything to me at the moment...

beomki-yeo:
I don't know why you get rid of the later part of clusterization which counts the measurement.  (I also don't know why you removed the sorting part. It will need to be rolled back if you encounter weird results from tracking finding once everything is in the single piece.)
How are you counting the number of measurements in the `spacepoint_formation`? Have you checked if you have redundant spacepoints during the seeding?

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
```diff
@@ -244,73 +252,23 @@ TRACCC_DEVICE inline void ccl_kernel(
...
-            atom(measurement_count);
-        outi = atom.fetch_add(outi);
-    }
-
-    barrier.blockBarrier();
-
-    /*
-     * Get the position to fill the measurements found in this thread group.
-     */
-    const unsigned int groupPos = outi;
```
stephenswat, line 284: Yeah I'm confused, where has this code gone?

krasznaa: This code is trying to be super fancy with how it
  - counts the total number of clusters;
  - figures out where to put the created measurement in the output container.

I decided that since resizable vecmem buffers can do all of this out of the box anyway, I'd not put the individual measurements into specific places in the output container. And would rather push them back into the output container in a random order. The final part of the code now does that.

It definitely works correctly, as all the tests are still happy with the output of the algorithm.

But indeed, if the ordering of the measurements was important, then this PR's code is not doing the same thing as the current `main` branch does. :thinking: What exact ordering do we require from the reconstructed measurements? Is it required that ones coming from the same module would be consecutive in the container? Because I'm not sure why that would be happening even with the current code? (Since the thread-blocks can operate across modules as-is.)

And yeah, removing all this code is what made the kernel a little faster. That it now only does the

```
    for (details::index_t tst = 0, cid; (cid = tst * blckDim + threadId) < size;
         ++tst) {
```

loop once, instead of twice. (Exchanging one extra loop, atomic operations on shared variables, and a few atomic operations on a global variable, for more atomic operations on the global variable.)

beomki-yeo: >  Is it required that ones coming from the same module would be consecutive in the container? 

Right. This is what is required from the track finding. The order iteself may not be very important.

beomki-yeo: > Since the thread-blocks can operate across modules as-is.

This is not point of sorting. Unless the measurements in the same module are consecutive, it will be very difficult to iterate over the measurements in the same module during the CKF.

stephenswat: Did some checking with the compute sanitizer, seems like the following is popping up frequently:

```
========= Invalid __global__ read of size 4 bytes
=========     at traccc::device::reduce_problem_cell(const vecmem::device_vector<const traccc::cell> &, unsigned short, unsigned int, unsigned int, unsigned char &, unsigned short *)+0x510 in /mnt/ssd1/sswatman/traccc/device/common/include/traccc/clusterization/device/impl/reduce_problem_cell.ipp:25
=========     by thread (28,0,0) in block (0,0,0)
=========     Address 0x3ca7a10 is out of bounds
=========     and is 139,622,790,563,312 bytes before the nearest allocation at 0x7efc7aa00000 of size 4 bytes
=========     Device Frame:void traccc::device::ccl_kernel<traccc::cuda::barrier>(unsigned short, unsigned short, unsigned int, vecmem::data::vector_view<const traccc::cell>, vecmem::data::vector_view<const traccc::cell_module>, unsigned short, unsigned short, unsigned int &, unsigned int &, unsigned int &, unsigned short *, unsigned short *, T1 &, vecmem::data::vector_view<traccc::measurement>, vecmem::data::vector_view<unsigned int>)+0x2fc0 in /mnt/ssd1/sswatman/traccc/device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp:226
=========     Device Frame:traccc::cuda::kernels::ccl_kernel(vecmem::data::vector_view<const traccc::cell>, vecmem::data::vector_view<const traccc::cell_module>, unsigned short, unsigned short, vecmem::data::vector_view<traccc::measurement>, vecmem::data::vector_view<unsigned int>)+0x15b0 in /mnt/ssd1/sswatman/traccc/device/cuda/src/clusterization/clusterization_algorithm.cu:40
```

krasznaa: There were two issues with the unit tests:
   - I didn't add correct host->device->host copies to the CUDA test when switching it to use `traccc::cuda::clusterization_algorithm`;
   - I switched the clusterization code to using `traccc::details::is_far_enough(...)` in an effort to increase the code sharing between the CPU and GPU codebases. But as it turns out, the existing function (which works fine for the host algorithm) was not working well for the FastSV algorithm. So now I switched it to a setup that does.

stephenswat: Great, looks like everything is in order functionality-wise now; thanks!

However, it also seems to bring some scary performance degradation. Before this PR (commit `3b587755`):

```
[2379250] traccc_throughput_mt_cuda@127.0.0.1
  traccc::cuda::kernels::ccl_kernel(vecmem::data::vector_view<const traccc::cell>, vecmem::data::vector_view<const traccc::cell_module>, unsigned short, unsigned short, vecmem::data::vector_view<traccc::measurement>, unsigned int &, vecmem::data::vector_view<unsigned int>) (314, 1, 1)x(128, 1, 1), Device 0, CC 8.6, Invocations 110
    Section: Command line profiler metrics
    ---------------------- ----------- ---------- ---------- ----------
    Metric Name            Metric Unit    Minimum    Maximum    Average
    ---------------------- ----------- ---------- ---------- ----------
    gpu__time_duration.sum     nsecond 367,424.00 377,664.00 371,650.91
    ---------------------- ----------- ---------- ---------- ---------- 
```

After this PR (commit `35f10fa2`):

```
  traccc::cuda::kernels::ccl_kernel(vecmem::data::vector_view<const traccc::cell>, vecmem::data::vector_view<const traccc::cell_module>, unsigned short, unsigned short, vecmem::data::vector_view<traccc::measurement>, vecmem::data::vector_view<unsigned int>) (314, 1, 1)x(128, 1, 1), Device 0, CC 8.6, Invocations 110
    Section: Command line profiler metrics
    ---------------------- ----------- ------------ ------------ ------------
    Metric Name            Metric Unit      Minimum      Maximum      Average
    ---------------------- ----------- ------------ ------------ ------------
    gpu__time_duration.sum     nsecond 1,490,784.00 1,508,992.00 1,499,022.55
    ---------------------- ----------- ------------ ------------ ------------
```

I.e. a slowdown of about 4 times. :(

beomki-yeo: Also if you want to rewind the memory of the measurement sorting, please take a look into the past PR:
https://github.com/acts-project/traccc/pull/467

beomki-yeo:
Out of curiosity have you checked if you are getting the same results from cpu and cuda?

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
beomki-yeo: Thanks for being aware of that!

stephenswat: I agree it's a bit mysterious... I think it might be worth investigating the contention on atomics in global memory. NSight Compute allows you to see why threads are stalling. If the number of long scoreboard stalls - which I think should include atomics - is higher in the updated version, that could explain it.

stephenswat: Suspiciously, the version in this PR (blue) simply performs way more instructions than the version before this PR (green). :thinking: Are we doing some duplicate work?

image

device/common/include/traccc/clusterization/device/impl/reduce_problem_cell.ipp
```diff
@@ -48,15 +39,16 @@ inline void reduce_problem_cell(
          * impossible for that cell to ever be adjacent to this one.
          * This is a small optimisation.
          */
-        if (cells[j].channel1 + 1 < c1 || cells[j].module_link != mod_id) {
```
stephenswat, line 51: I wonder if it is due to the fact that this if statement and the one below are no longer using the in-register information about the channels, but rather fetching it again from memory. :thinking:

krasznaa: :thinking: But all of these functions are declared `inline`. And on these specific calls we were accessing the `channel1` and `module_link` variables directly from global memory to begin with.

But I also don't have better ideas. We'll need to revert the changes one-by-one, and see what makes a difference.

krasznaa: At the same time, I was hoping that running a profiling on an Intel GPU may reveal something. I don't know, maybe VTune would have a more intuitive interface. :thinking: But at least on my laptop's iGPU I didn't find anything really interesting...

image

Unlike on our NVIDIA GPUs, the `ccl_kernel` is not even the slowest kernel on this hardware... :confused:

device/common/include/traccc/clusterization/device/impl/reduce_problem_cell.ipp
```diff
@@ -70,14 +62,15 @@ inline void reduce_problem_cell(
          * Note that this check now looks in the opposite direction! An
          * important difference.
          */
-        if (cells[j].channel1 > c1 + 1 || cells[j].module_link != mod_id) {
+        if (traccc::details::is_far_enough(reference_cell, cells.at(j))) {
```
krasznaa, line 65: I'll test it out tomorrow, but I'm pretty sure that this is the faulty line. 🤔 Cell order very much matters for this function call. And I believe the code currently uses the wrong one. Leading to the for-loop going on for much longer than it should.

Let's hope it will get most of the performance back at least...

krasznaa: With this we are back to the numbers that I first reported. With the current code working like this:

Screenshot from 2024-04-19 07-36-55

And this PR's code working like:

Screenshot from 2024-04-19 07-39-12

So `traccc::device::ccl_kernel(...)` is speeding up with this PR. But the seed finding is slowing down a little. Since the clusterization now reports to behave correctly, I can only imagine that the measurement/spacepoint order ending up differently would be the reason for this. But I'm not quite sure why. Why it would be such a large effect. Since it's not even the spacepoint binning that seems to slow down. :confused:

krasznaa:
So... fun! I now bumped into the same issue that I believe is giving us grief with clusterizing the ODD cells on a GPU. It seems that the code meant to order cells during CSV reading, is not doing its job. :frowning:

I was getting afraid that there would be no thing left to debug... :face_exhaling:

device/common/include/traccc/clusterization/device/impl/reduce_problem_cell.ipp
stephenswat: Perfect, great work!

krasznaa:
Never mind, we don't get two birds with just the one stone... :frowning: I just didn't set up the new assertion correctly in the code...

stephenswat:
It would be possible - although weird - that the same cell has multiple activations in the same input?

stephenswat:
Aha!

```
traccc/data/odd/muon100GeV-geant4 $ cat event000000000-cells.csv | cut -d ',' -f 1,3,4 | sort | uniq -c

...

2 2161728508332653312,219,0
```

So there are some duplicate cells in the input data!

krasznaa:
That was one of my guesses. But, the following is now also failing for me. :confused:

```diff
diff --git a/io/src/read_geometry.cpp b/io/src/read_geometry.cpp
index c88e8fe..f52ce52 100644
--- a/io/src/read_geometry.cpp
+++ b/io/src/read_geometry.cpp
@@ -19,6 +19,7 @@
 #include <vecmem/memory/host_memory_resource.hpp>
 
 // System include(s).
+#include <cassert>
 #include <stdexcept>
 
 namespace {
@@ -50,6 +51,7 @@ read_json_geometry(std::string_view filename) {
         barcode_map = std::make_unique<
             std::map<std::uint64_t, detray::geometry::barcode>>();
         for (const auto& surface : detector.surfaces()) {
+            assert(barcode_map->find(surface.source) == barcode_map->end());
             (*barcode_map)[surface.source] = surface.barcode();
         }
     }
```

I.e. the same surface identifier shows up multiple times from Detray. Could it be that we end up merging modules in some weird way?

Though this is probably just a red herring. At least for the clusterization...

krasznaa:
> In fact, some are repeated up to 12 times:
> 
> ```
>       7 2089670364539207936,226,0
>       7 2089670364539207936,229,0
>       7 2089670364539207936,230,0
>       7 2089670364539207936,231,0
>       7 2089670364539317760,156,0
>       7 2089670501978336256,132,0
>       8 1224979373522735104,328,703
>       8 1224979373522735104,329,704
>       8 1729382531788187392,177,42
>       8 1729382531788189440,12,27
>       8 1729382531788189440,13,27
>       8 1729382531788189440,14,27
>       8 1729382669227169536,6,87
>       8 1729382669227169536,7,87
>       8 2089670364539207936,227,0
>       8 2089670364539207936,228,0
>       8 2089670501977937664,38,0
>       8 2089670501977937664,39,0
>       8 2089670501977937664,43,0
>       8 2089670501977937664,44,0
>       8 2089670501978336256,129,0
>       9 1729382531788187392,178,42
>       9 1729382531788187392,179,42
>      10 2089670501977937664,40,0
>      11 2089670501977937664,42,0
>      12 2089670501977937664,41,0
> ```

Since you have some "shell magic" at the ready as it seems, could you remove the duplicates from the files and send me the updated TGZ? I'll upload it as a new version.

stephenswat:
Let me doctor up some deduplication code and produce some deduplicated files.

stephenswat:
It's also worth understanding how this happens, because I can't say it makes much sense to me. Is there any physical reason why we would get a read-out for one pixel twice? Or 12 times?

stephenswat:
Interestingly, the full TML dataset also seems to have some duplicated hits, although the number is _much_ smaller than it is for the ODD files. Could be that this just succeeded by chance, e.g. there was never a duplicated hit with 8 other hits in the neighbourhood, so there was some buffer in the 8-length array to compensate.

krasznaa:
There must be some issue in the Acts "digitization" code. :thinking: This behaviour sounds like a plausible outcome if some mistake is made there. (Remember, simulation just tells us how much energy was deposited exactly where in the detector volumes. We then need to turn those energy deposits into information that the real hardware would've read out as well. I.e. we need to "digitize" the simulated data, as we call it in ATLAS.)

Pinging @asalzburger for info. :wink:

stephenswat:
Okay, `pcadp04:/mnt/ssd1/sswatman/traccc/data/odd` has been deduplicated.

stephenswat:
```
$ build/bin/traccc_seq_example_cuda --input-directory=odd/muon100GeV-geant4/ --detector-file=geometries/odd/odd_geometry_detray.json --use-detray-detector --digitization-file=geometries/odd/odd-digi-geometric-config.json --input-events=10

Running Full Tracking Chain Using CUDA

>>> Detector Options <<<
  Detector file       : geometries/odd/odd_geometry_detray.json
  Material file       : 
  Surface rid file    : 
  Use detray::detector: yes
>>> Input Data Options <<<
  Input directory               : odd/muon100GeV-geant4/
  Number of input events        : 10
>>> Track Seeding Options <<<
  None
  Run performance checks: no
>>> Accelerator Options <<<
  Compare with CPU results: no

Detector check: OK
==> Statistics ... 
- read    792140 cells from 112034 modules
- created (cpu)  0 measurements     
- created (cpu)  0 spacepoints     
- created (cuda) 319346 spacepoints     
- created  (cpu) 0 seeds
- created (cuda) 42343 seeds
==>Elapsed times...
           File reading  (cpu)  1464 ms
         Clusterization (cuda)  42 ms
   Spacepoint formation (cuda)  24 ms
                Seeding (cuda)  47 ms
           Track params (cuda)  3 ms
                     Wall time  1587 ms
```

:smile:

stephenswat:
See `pcadp04:/mnt/ssd1/sswatman/traccc/data/traccc-data-v6.tar.gz`.

asalzburger:
Ouch, I guess we could see an issue of Geant4 versus Fatras simulation, while Fatras is guaranteed to have one step per module (i.e. the particle only intersects the module once), Geant4 physics can force to split a segment into several steps as some fancy physics processes may happen. e.g. delta rays.

If we don't catch them before the clusterizer, we might produce duplicate cells ?1?

krasznaa:
So... Stephen, since I'd like to sort this out, so that we could move on to doing all the rest that we still need to do: How do you feel about the changes?

The very last one, where I made the code go from using bare pointers to the shared memory blocks, to using vecmem vectors, did cause a small performance drop. (O(1%)) Which I personally am okay with, given the debugging benefits. (In debug mode the code will be more talkative about where it encountered an issue.) But I'm open to a discussion about this.

In any case, I'd want to sort out the addition of the new data file in a separate PR. And then of course I'll want to move on to looking at the full ODD tracking chain with CUDA. :wink:

stephenswat:
I think all the issues are resolved, I don't think a 1% performance drop warrants any debate, we can just get it in no problem.

stephenswat (approved):

Issue #527, opened 2024-03-13
Device Clusterization Crashes on ODD Cells (2024.03.13.)


krasznaa:
As discussed in #509, unfortunately the current code cannot run accelerated clusterization on the ODD files used by the project. :frowning: The following, for instance, crashes:

```
[bash][celeborn]:traccc > ./out/build/sycl/bin/traccc_seq_example_cuda --input-directory=odd/muon100GeV-geant4/ --detector-file=geometries/odd/odd_geometry_detray.json --use-detray-detector --digitization-config-file=geometries/odd/odd-digi-geometric-config.json --events=10

Running the full tracking chain using CUDA

>>> Common options <<<
  Input data format            : csv
  Input directory              : odd/muon100GeV-geant4/
  Events                       : 10
  Skipped events               : 0
  Target cells per partition   : 1024
  Check performance            : 0
  Perform ambiguity resolution : 1
>>> Detector options <<<
  Detector file        : geometries/odd/odd_geometry_detray.json
  Material file        : 
  Grid file            : 
  Use detray::detector : yes

>>> Full tracking chain options <<<
  Digitization configuration file: geometries/odd/odd-digi-geometric-config.json

WARNING: mask store has empty collection no. 2
WARNING: mask store has empty collection no. 3
WARNING: mask store has empty collection no. 6
WARNING: mask store has empty collection no. 7
WARNING: No material in detector
WARNING: acceleration data structures store has empty collection no. 1
WARNING: acceleration data structures store has empty collection no. 2
WARNING: acceleration data structures store has empty collection no. 3
WARNING: acceleration data structures store has empty collection no. 4
WARNING: No entries in volume finder
Detector check: OK
Modules are ordered: yes
Cells are ordered: yes
CUDAassert: an illegal memory access was encountered /home/krasznaa/ATLAS/projects/traccc/traccc/device/cuda/src/utils/stream.cpp 57
[bash][celeborn]:traccc >
```

@stephenswat had the very reasonable idea that the cells read in from the ODD CSV files may not be ordered "correctly". Since the device clusterization code has some deep-routed assumptions about the layout of the data that it receives.

So I added the following to the code, to see what's what:

```diff
diff --git a/examples/run/cuda/seq_example_cuda.cpp b/examples/run/cuda/seq_example_cuda.cpp
index f732942..3096058 100644
--- a/examples/run/cuda/seq_example_cuda.cpp
+++ b/examples/run/cuda/seq_example_cuda.cpp
[...]

PR #546, opened 2024-04-19
Deduplicate cells in CSV input

stephenswat:
This PR adds deduplication at the file reading level and prints a warning, which should fix the problems encountered in #527.

### DIFF
--- a/io/src/csv/read_cells.cpp
+++ b/io/src/csv/read_cells.cpp
@@ -13,6 +13,7 @@
 // System include(s).
 #include <algorithm>
 #include <cassert>
+#include <set>
 #include <stdexcept>
 #include <utility>
 #include <vector>
@@ -96,9 +97,26 @@ void read_cells(
     std::vector<std::pair<csv::cell, unsigned int>> allCells;
     allCells.reserve(50000);
 
+    // Keep track of duplicate cells
+    using cell_key_t =
+        std::tuple<decltype(std::declval<csv::cell>().geometry_id),
+                   decltype(std::declval<csv::cell>().channel0),
+                   decltype(std::declval<csv::cell>().channel1)>;
+
+    std::set<cell_key_t> seen;
+    std::size_t dedup = 0;
+
     // Read all cells from input file.
     csv::cell iocell;
     while (reader.read(iocell)) {
+        cell_key_t key = {iocell.geometry_id, iocell.channel0, iocell.channel1};
+
+        if (seen.find(key) != seen.end()) {
+            dedup++;
+            continue;
+        } else {
+            seen.insert(key);
+        }
 
         // Modify the geometry ID of the cell if a barcode map is provided.
         const std::uint64_t original_geometry_id = iocell.geometry_id;
@@ -133,6 +151,11 @@ void read_cells(
             allCells.push_back({iocell, pos});
             ++(cellCounts[pos]);
         }
+    }
+
+    if (dedup > 0) {
+        std::cout << "WARNING: @read_cells: Had to deduplicate " << dedup
+                  << " rows in input file \"" << filename << "\"" << std::endl;
     }
 
     // Transform the cellCounts vector into a prefix sum for accessing


stephenswat:
Okay, so we have some tests which _rely_ on the fact that cells are activated twice in one event. Can we deduplicate them by adding the activations together maybe?

io/src/csv/read_cells.cpp
```diff
@@ -96,9 +97,26 @@ void read_cells(
...
+    std::set<cell_key_t> seen;
+    std::size_t dedup = 0;
+
     // Read all cells from input file.
     csv::cell iocell;
     while (reader.read(iocell)) {
+        cell_key_t key = {iocell.geometry_id, iocell.channel0, iocell.channel1};
+
+        if (seen.find(key) != seen.end()) {
+            dedup++;
```
krasznaa, line 115: Let's add a check here that their activation values would be ~the same. If they are not, we're dealing with an actual error, and the code should throw an exception about it. But for "real duplicates" this can indeed be a good idea to suppress them with a warning.

krasznaa (commented):
The https://github.com/acts-project/traccc/blob/main/tests/io/mock_data/event000000000-cells.csv file is a weird one. The duplicate cells have a little different activation values. :confused:

I would vote for removing the duplicate from the file, and then updating the rest of the files / tests to make them successful with the modified cell file. :thinking:

io/src/csv/read_cells.cpp
stephenswat: So what I suspect is happening here is that we simulate particles, they cross some surface around the same spot, and a cell is recorded for each particle, possibly in the same pixel. It makes sense, then, that the activation values could be quite different. So I don't think we'll find my cases where the activation value is the same.

I'm not sure how this kind of detector response simulation should behave, e.g. do two particles depositing energy in a pixel simply add up? Or is there some more complex way to combine them?

@beomki-yeo you designed this test, perhaps you can weigh in?

stephenswat: The test seems to be designed specifically to test cases in which two particles activate the same cell, so I think simply deduplicating the input file will defeat the purpose of the test. :frowning:

krasznaa: But then the test is bad... :thinking: The real hardware will never report the same pixel multiple times, unless there's a bug somewhere in the data readout chain. So the simulation shouldn't behave differently either.

Yes, the deposited energies generally simply add up. Naively, you would first collect all energy deposits into some in-memory container, and then go through the pixels one by one, adding up all deposits that happened in the geometric volume of that pixel. With a simple addition. (Energy can be as simple as that. :wink:)

It's not really a debate whether duplicate cells in the CSV files are bad or not. :thinking: We can certainly discuss about the ordering of the cells (both in the files and in memory). But duplicates are definitely a no-no.

beomki-yeo: For energy deposit for multiple particles on the same cell we might need an expert.  But we can just  add up as you suggested for the moment.

> I think simply deduplicating the input file will defeat the purpose of the test

Right. I think we need to leave `event_map` (truth) as it is and just shrinking the cell readout data. This should be possible I guess?

beomki-yeo: > We can certainly discuss about the ordering of the cells (both in the files and in memory). But duplicates are definitely a no-no

I think we need to consider splitting cell data and particle-to-cell data (Like there is a dedicated map for measurement-to-hit data)

krasznaa: :confused: In the end, once every one of our files is set up correctly, we should have:
  - particles pointing at a list of cells that they "produced";
  - cells pointing at a list of particles that contributed to them.

A cell is still just a single thing, I don't see that changing. Our analysis model has to be able to deal with multiple particles contributing to the same cell. I can't imagine that the analysis done in Athena would happen in any other way. :thinking:

krasznaa:
Superseded by #547.

PR #550, opened 2024-04-22
(Host and CUDA) Measurement Sorting, main branch (2024.04.22.)

krasznaa:
As we discussed with @beomki-yeo earlier, track finding needs the measurements to be ordered. (Measurements coming from the same detector module need to be beside each other in the container.) But `traccc::cuda::clusterization_algorithm` and `traccc::sycl::clusterization_algorithm` does not ensure this. (`traccc::host::cuda_clusterization` does, because of how the algorithm works.)

While later on I believe it will be possible to make the GPU clusterization algorithms observe this ordering on their output, for now I decided that we should do the practical thing, and just add extra algorithms for ordering existing measurement collections. This is what `traccc::host::measurement_sorting_algorithm` and `traccc::cuda::measurement_sorting_algorithm` do. Simply by using [std::sort](https://en.cppreference.com/w/cpp/algorithm/sort) and [thrust::sort](https://github.com/NVIDIA/thrust/blob/main/thrust/sort.h) respectably.

At the same time I also added `traccc::cuda::measurement_sorting_algorithm` to `traccc_seq_example_cuda`, `traccc_throughput_st_cuda` and `traccc_throughput_mt_cuda`, to see what's what. Note though that I didn't add `traccc::host::measurement_sorting_algorithm` anywhere at the moment. The sequence example works as it should (results compared with the host algorithm remain the same), and the throughputs do the following:

| | TrackML ($\mu$=200) |
|-|-|
| Current | 371.545 |
| This PR | 300.678 |

(At first I wanted to make throughput tests with the ODD files as well. But forgot that that doesn't work just yet. We'll need to make those executables be able to read a Detray geometry first...)

I agree, this is not brilliant. :frowning: But right now I'm in result oriented mode, and adding this extra sorting step shall get us closer to running the full ODD chain with CUDA. :thinking:

### DIFF
--- a/examples/run/cuda/full_chain_algorithm.cpp
+++ b/examples/run/cuda/full_chain_algorithm.cpp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2022-2023 CERN for the benefit of the ACTS project
+ * (c) 2022-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -42,6 +42,7 @@ full_chain_algorithm::full_chain_algorithm(
       m_target_cells_per_partition(target_cells_per_partition),
       m_clusterization(memory_resource{*m_cached_device_mr, &m_host_mr}, m_copy,
                        m_stream, m_target_cells_per_partition),
+      m_measurement_sorting(m_copy, m_stream),
       m_spacepoint_formation(memory_resource{*m_cached_device_mr, &m_host_mr},
                              m_copy, m_stream),
       m_seeding(finder_config, grid_config, filter_config,
@@ -73,6 +74,7 @@ full_chain_algorithm::full_chain_algorithm(const full_chain_algorithm& parent)
       m_target_cells_per_partition(parent.m_target_cells_per_partition),
       m_clusterization(memory_resource{*m_cached_device_mr, &m_host_mr}, m_copy,
                        m_stream, m_target_cells_per_partition),
+      m_measurement_sorting(m_copy, m_stream),
       m_spacepoint_formation(memory_resource{*m_cached_device_mr, &m_host_mr},
                              m_copy, m_stream),
       m_seeding(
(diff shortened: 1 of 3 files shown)


krasznaa:
Note that I didn't bother with a `traccc::sycl::measurement_sorting_algorithm` at the moment because of the technical issues that we are still having with oneDNN. :thinking:

stephenswat:
No need for this right now, but if vecmem vectors got some kind of bulk resize, we could easily re-implement the behaviour we had before and bring back a lot of this performance.

beomki-yeo:
Sorting is good but I am not convinced of why it should be done as a dedicated algorithm. 
I think it would have been much easier by adding `thrust::sort` or `std::sort` at the last of `clusterization_algorithm`

It is also OK to put it at the front of CKF if you think that's better

beomki-yeo:
>  we are still having with oneDNN

And there is a typo: oneDNN -> oneDPL

krasznaa:
> Sorting is good but I am not convinced of why it should be done as a dedicated algorithm. I think it would have been much easier by adding `thrust::sort` or `std::sort` at the last of `clusterization_algorithm`
> 
> It is also OK to put it at the beginning of CKF if you think that's better

Putting it at the end of the clusterization algorithm would've been an option. But having it separately allows us to still do tests that don't need the sorting. It was purely a code design consideration behind this.

I would definitely not put the sorting at the beginning of the CKF algorithm. :thinking: We may not depend on it right now, but having the measurement and spacepoint containers being out of sync at the end of the event reconstruction, would seem like a bad idea. :thinking:

krasznaa:
> No need for this right now, but if vecmem vectors got some kind of bulk resize, we could easily re-implement the behaviour we had before and bring back a lot of this performance.

Not sure what you mean here. :confused: This PR's performance loss comes from an extra set of kernels being executed by Thrust. Plus the extra (at least one) synchronization point that the algorithm adds.

To make the clusterization produce ordered measurements, I fear we'll need to do some deeper re-design. We'd need to do a count + fill kernel design. :thinking: But I didn't think it fully through yet.

stephenswat:
> Not sure what you mean here. 😕 This PR's performance loss comes from an extra set of kernels being executed by Thrust. Plus the extra (at least one) synchronization point that the algorithm adds.

Well yeah, the previous kernel produced perfectly sorted measurements (if none of the modules had a massive number of hits) or very close-to-sorted arrays, which are more cheap to sort than ones with less sorted structure. So yes, the old algorithms didn't need sorting or they needed much less sorting. :confused:

krasznaa:
With all that being said, would anyone either ask for some concrete change, or approve this? I'll want to finally make `traccc_seq_example_cuda` go all the way tomorrow...

stephenswat (approved):
Sure, we can recover this performance later.

beomki-yeo (approved):

PR #551, opened 2024-04-22
Project Version Update, main branch (2024.04.22.)

krasznaa:
Update the project to version `0.9.0`, in preparation for the next tag to be made. (For @fredevb. :wink:)

PR #553, opened 2024-04-24
Project Version Update, main branch (2024.04.24.)

krasznaa:
Similar to #551, this is to "pre-load" the correct project version for the time that we'll tag `v0.10.0`.

Which I hope to do sometime next week, to mark the version of the code that we would execute our initial ODD performance studies with. :wink:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 
 # Set up the project.
 cmake_minimum_required( VERSION 3.9 )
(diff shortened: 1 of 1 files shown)


stephenswat (approved):
I still prefer incrementing this right before tagging so we can pivot to tagging v1.0.0 or v0.9.1 instead of v0.10.0, but let's go ahead with this.

PR #555, opened 2024-04-25
Make alpaka queue blocking, fix serial CPU running

StewMH:
Fix a crash seen running the CPU alpaka build. Unfortunately does not fix a similar looking crash from running the HIP version.


krasznaa (approved):
Sure, no objections. Of course eventually we'll want to come back to this question, once we start comparing the compute performance of the different implementations. :thinking:

PR #557, opened 2024-04-25
ref: Use angle from surface class instead of intersector

niermann999:
While debugging the detray update I found that in the ```seed_generator``` the cosine of the incidence angle on the intersection is quite often invalid, because the mask tolerance and overstepping tolerance that are given to the ```intersection_update``` call do not match what the navigator used to find the track parameters on surface. This leads to essentially a zero path through the material (thickness gets divided by the cosine). Because putting the navigation configuration in there to set all of the tolerances is a bit awkward, I used the angle calculation of the surface directly. I cleaned that interface up a bit in acts-project/detray#723

### DIFF
--- a/device/common/include/traccc/finding/device/impl/apply_interaction.ipp
+++ b/device/common/include/traccc/finding/device/impl/apply_interaction.ipp
@@ -54,13 +54,22 @@ TRACCC_DEVICE inline void apply_interaction(
     const cxt_t ctx{};
     const auto free_vec = sf.bound_to_free_vector(ctx, bound_param.vector());
 
+    using scalar_type = typename detector_t::scalar_type;
     const auto& mask_store = det.mask_store();
     intersection_type sfi;
     sfi.sf_desc = det.surface(bound_param.surface_link());
     sf.template visit_mask<
         detray::intersection_update<detray::ray_intersector>>(
         detray::detail::ray<transform3_type>(free_vec), sfi,
-        det.transform_store());
+        det.transform_store(),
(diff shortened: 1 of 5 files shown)


niermann999:
On the first commit I checked that the two angle calculations produce the same values

device/common/include/traccc/finding/device/impl/apply_interaction.ipp
```diff
@@ -49,25 +36,16 @@ TRACCC_DEVICE inline void apply_interaction(
...
+    const typename detector_t::geometry_context ctx{};
 
     // Apply interactor
     typename interactor_type::state interactor_state;
     interactor_type{}.update(
         bound_param, interactor_state,
         static_cast<int>(detray::navigation::direction::e_forward), sf,
-        sfi.cos_incidence_angle);
+        std::abs(
+            sf.cos_angle(ctx, bound_param.dir(), bound_param.bound_local())));
```
krasznaa, line 48: This should become [std::fabs](https://en.cppreference.com/w/cpp/numeric/math/fabs) for CUDA/HIP, and `cl::sycl::fabs` for SYCL. Though I don't know if we would have the infrastructure for this already. :thinking:

krasznaa (commented):
Apart from the (f)abs question/issue, I'm very happy with this. Any apparent code simplification is a good change. :smile:

Even the (f)abs issue I could imagine skipping in this PR, since we'll need to fix all Detray and traccc both for that still. :thinking: So one more is not much of a difference... (I'll just wait to see the opinion of others...)

krasznaa (approved):
Okay, we could just do this for now then. :thinking:

PR #558, opened 2024-04-26
Alpaka: Clustering + Throughput

CrossR:
Slightly delayed due to some other work and then needing to get my branch back up to date...but this covers basically all the current Alpaka work, outside of the start of my Track finding testing and @StewMH HIP work.

This includes some of the clusterisation work, as well as the throughput examples. In some of the examples, there is some commented out code in there still, which is basically the parts from the CUDA examples that don't have an equivalent in Alpaka yet. Not sure if you'd prefer it moved, or left?

I'll also highlight some of the bits that I think people will have opinions on below.

Once there has been a brief look, I can look at squashing this down to something more reasonable.

Still ongoing (though it sounded like not a blocker in the previous meeting):

 - Debugging / verifying the work with a CUDA back end. Everything runs, but I want to verify if the memory issues I was hitting are gone, and ensure the results compare to the native CUDA. Right now, I can sometimes get results that don't directly match the CUDA with the same input (and a fixed seed) so I think I have some form of issue still.


device/alpaka/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ include( traccc-alpaka-functions )
 include( traccc-compiler-options-cpp )
 
 set(PUBLIC_LIBRARIES traccc::core detray::core detray::utils vecmem::core covfie::core)
-set(PRIVATE_LIBRARIES traccc::device_common alpaka::alpaka)
+set(PRIVATE_LIBRARIES traccc::device_common alpaka::alpaka traccc::Thrust)
```
CrossR, line 12: Temporary, see below.

device/alpaka/src/clusterization/measurement_sorting_algorithm.cpp
```diff
@@ -0,0 +1,46 @@
...
+    // TODO: This obviously won't work for HIP, but we don't have a HIP
+    // equivalent of Thrust yet.
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+    thrust::sort(thrust::device, measurements_view.ptr(),
+                 measurements_view.ptr() + n_measurements,
+                 measurement_sort_comp());
+#elif !defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+    std::sort(measurements_view.ptr(), measurements_view.ptr() + n_measurements,
+              measurement_sort_comp());
+#endif
```
CrossR, line 40: This won't work for HIP / isn't very ideal.

@StewMH is going to be looking into getting `rocThrust` support, in which case this (and the later uses over in the track finding) should be much easier. Alpaka does have some thrust-like bits around this in https://github.com/alpaka-group/vikunja (transforms and reduces as far as I see), but not a direct replacement for a sort.

My understanding of PR #550 by @krasznaa was that this is basically a necessary evil, and eventually something like an algorithm re-design may mean it isn't needed, but for now I think this is "good enough".

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
```diff
@@ -65,7 +65,9 @@ TRACCC_DEVICE void fast_sv_1(
              ++tst) {
             const details::index_t cid = tst * blckDim + tid;
 
+#ifndef ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
             __builtin_assume(adjc[tst] <= 8);
+#endif
```
CrossR, line 70: Somewhat lucky we haven't hit this sort of thing before....

`gcc` doesn't support this, and we are a pretty long way off the support in the C++ standard being adopted everywhere (an equivalent is part of C++23 I believe?).

For now, I've ignored it like this, but any actually sensible suggestions welcome, rather than this weird hack.

examples/run/alpaka/seeding_example_alpaka.cpp
```diff
@@ -101,6 +117,10 @@ int seq_run(const traccc::opts::track_seeding& seeding_opts,
     // Performance writer
     traccc::seeding_performance_writer sd_performance_writer(
         traccc::seeding_performance_writer::config{});
+    // traccc::finding_performance_writer find_performance_writer(
+    //     traccc::finding_performance_writer::config{});
+    // traccc::fitting_performance_writer fit_performance_writer(
+    //     traccc::fitting_performance_writer::config{});
```
CrossR, line 123: Just kept around as it made comparing to the CUDA easier...happy to remove.

(Same for the rest of this file etc)

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
stephenswat: The best solution here would be for us to define an `ASSUME` macro that either does something if the compiler, or is a noop if not.

core/include/traccc/definitions/qualifiers.hpp
```diff
@@ -31,3 +31,16 @@
...
 #define TRACCC_ALIGN(x) alignas(x)
 #endif
+
+#if __has_builtin(__builtin_assume)
+#define TRACCC_ASSUME(cond) __builtin_assume(cond)
+#else
+#define TRACCC_ASSUME(cond)
+#endif
+
+#if defined(__CUDACC__) || defined(__HIP__) || defined(__clang__) || \
```
StewMH, line 41: think this should be `defined(_OPENMP)` rather than clang.

core/include/traccc/definitions/qualifiers.hpp
```diff
@@ -31,3 +31,16 @@
 #else
 #define TRACCC_ALIGN(x) alignas(x)
 #endif
+
+#if __has_builtin(__builtin_assume)
+#define TRACCC_ASSUME(cond) __builtin_assume(cond)
+#else
+#define TRACCC_ASSUME(cond)
+#endif
```
StewMH, line 39:
```suggestion
#if __GNUC__ < 10
#define TRACCC_ASSUME(cond)
#elif __has_builtin(__builtin_assume)
#define TRACCC_ASSUME(cond) __builtin_assume(cond)
#else
#define TRACCC_ASSUME(cond)
#endif
```

GCC < 10 is unaware of `__has_builtin`

stephenswat: The recommended syntax is `#if defined __has_builtin` as per https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fbuiltin.html

stephenswat: I'm sorry to be this anal, but this is not really a qualifier. Perhaps we could create a new file `builtins.hpp` or `hints.hpp` (open for discussion) for this?

StewMH: If it's `hints.hpp` then we can put the new pragma unroll in there as well.

stephenswat: Agreed!

examples/run/alpaka/CMakeLists.txt
```diff
@@ -4,21 +4,42 @@
...
+set(LIBRARIES vecmem::core traccc::io traccc::performance
+    traccc::core traccc::device_common traccc::alpaka alpaka::alpaka
+    traccc::options ${EXTRA_LIBS})
+
+traccc_add_executable( seq_example_alpaka "seq_example_alpaka.cpp"
+    LINK_LIBRARIES ${LIBRARIES} )
 traccc_add_executable( seeding_example_alpaka "seeding_example_alpaka.cpp"
     LINK_LIBRARIES ${LIBRARIES} )
 
+# #
```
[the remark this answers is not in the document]
CrossR: PR #572 added track fitting and finding to the throughput examples, which the Alpaka code base doesn't support yet, breaking that bit of this PR.

I can either remove all the code, or leave it as I have done here with that code not compiling and it can just be reactivated when I add the track finding and fitting bits (which is in progress and about 80% of the way there).

StewMH:
@krasznaa, @stephenswat, @beomki-yeo could we have a quick review? There aren't many changes outside the alpaka code.

device/alpaka/src/clusterization/measurement_sorting_algorithm.cpp
CrossR: Should be fixed with the recent rocThrust change, and I'm testing it more thoroughly with my current work.

beomki-yeo (commented):
Can we merge `qualifiers.hpp` and `hints.hpp` into a single file with the name of `directives.hpp`?

beomki-yeo (approved):
Nah. we can think of the file naming later

device/alpaka/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+struct CCLKernel {
+    template <typename TAcc>
+    ALPAKA_FN_ACC void operator()(
+        TAcc const& acc, const cell_collection_types::const_view cells_view,
+        const cell_module_collection_types::const_view modules_view,
+        const device::details::index_t max_cells_per_partition,
+        const device::details::index_t target_cells_per_partition,
+        measurement_collection_types::view measurements_view,
+        vecmem::data::vector_view<unsigned int> cell_links) const {
```
krasznaa, line 30: Generally, this setup is fine with me. I'm just thinking out loud here.
  - Would it not be nicer / more symmetric to call this something like `traccc::alpaka::kernels::ccl_kernel`? That's the sort of naming that we went with in `traccc::cuda`.
  - I think Alpaka allows such structs to carry member variables as well. Or is it not flexible enough for that? :thinking: I was just wondering, since you pass an actual object instance to `alpaka::exec` and not just the type of the struct, could some of these (many) arguments be delegated to member variables? Technically, the "configuration variables" would seem like appropriate members, while the data views indeed look more understandable as operator arguments. :thinking:

examples/run/alpaka/CMakeLists.txt
```diff
@@ -4,21 +4,42 @@
 #
 # Mozilla Public License Version 2.0
 
-set(LIBRARIES vecmem::core traccc::io traccc::performance
-    traccc::core traccc::device_common traccc::alpaka alpaka::alpaka
-    traccc::options)
+set(EXTRA_LIBS "")
```
krasznaa, line 7: I personally prefer the formalism:

```suggestion
set(EXTRA_LIBS)
```

Though technically this one is fine as well.

examples/run/alpaka/seeding_example_alpaka.cpp
krasznaa: Indeed, let's not add a whole bunch of commented lines. :thinking: You should absolutely keep this code in a commit in your own repository. But I'd prefer not to put it in like this into the main repo.

krasznaa (changes requested):
I'm very happy with how you added a bunch of `->ignore()` and `->wait()` calls on the various copy operations. However, since it's super easy to miss some of them, did you try to compile the code with `-DVECMEM_FAIL_ON_ASYNC_ERRORS=TRUE`? :thinking:

https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-options.cmake#L58-L60

I added that feature specifically to help with making sure that our code would only wait where it needs to. (Without that flag, if the user forgets to either wait for an event or ignore it, the code will wait for the event. Just to be safe.)

device/alpaka/src/clusterization/clusterization_algorithm.cpp
CrossR: > Would it not be nicer / more symmetric to call this something like traccc::alpaka::kernels::ccl_kernel? That's the sort of naming that we went with in traccc::cuda.

I'd originally tried to split them up, so they were obviously distinct, but having them mirror the CUDA / actual kernel names could be nicer too. Regardless, I'll look at fixing that up in the follow up finding / fitting PR, just so I can change them all in one go.

> I think Alpaka allows such structs to carry member variables as well. Or is it not flexible enough for that? 🤔 I was just wondering, since you pass an actual object instance to alpaka::exec and not just the type of the struct, could some of these (many) arguments be delegated to member variables? Technically, the "configuration variables" would seem like appropriate members, while the data views indeed look more understandable as operator arguments. 🤔

Possibly! I think the style shown here matches the Alpaka-recommended way at least, but it may be better for our needs to swap....I can do some testing soon and see, and when I look at renaming the kernels, I can see if swapping lots of args to member variables makes sense.

device/alpaka/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -0,0 +1,136 @@
...
+    // debugging.
+    //
+    vecmem::data::vector_buffer<unsigned int> cell_links(num_cells, m_mr.main);
+    m_copy.get().setup(cell_links)->ignore();
+
+    // Launch ccl kernel. Each thread will handle a single cell.
+    const device::details::ccl_kernel_helper helper{
+        m_target_cells_per_partition, num_cells};
+    auto workDiv =
+        makeWorkDiv<Acc>(helper.num_partitions, helper.threads_per_partition);
```
stephenswat, line 104: This will also need some updating with #595. I recognise that you have been working on this for a _long_ time, though, so I don't blame you if you want to get this in first.

device/alpaka/src/clusterization/measurement_sorting_algorithm.cpp
```diff
@@ -0,0 +1,46 @@
...
+    // Sort the measurements in place
+    // TODO: This obviously won't work for HIP, but we don't have a HIP
+    // equivalent of Thrust yet.
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+    thrust::sort(thrust::device, measurements_view.ptr(),
+                 measurements_view.ptr() + n_measurements,
+                 measurement_sort_comp());
+#elif !defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+    std::sort(measurements_view.ptr(), measurements_view.ptr() + n_measurements,
+              measurement_sort_comp());
```
stephenswat, line 39: Doesn't this assume that the code is host-accessible? What we could do is use #632 to implement a _very inefficient_ but functional sorting algorithm. We can do that later, though.

device/alpaka/src/clusterization/spacepoint_formation_algorithm.cpp
```diff
@@ -0,0 +1,80 @@
...
+    const traccc::memory_resource& mr, vecmem::copy& copy)
+    : m_mr(mr), m_copy(copy) {}
+
+spacepoint_formation_algorithm::output_type
+spacepoint_formation_algorithm::operator()(
+    const measurement_collection_types::const_view& measurements_view,
+    const cell_module_collection_types::const_view& modules_view) const {
+
+    // Setup alpaka
+    auto devHost = ::alpaka::getDevByIdx(::alpaka::Platform<Host>{}, 0u);
```
stephenswat, line 47: Same comment as above.

device/alpaka/src/clusterization/spacepoint_formation_algorithm.cpp
```diff
@@ -0,0 +1,80 @@
...
+    auto queue = Queue{devAcc};
+
+    // Get the number of measurements.
+    const measurement_collection_types::const_view::size_type num_measurements =
+        m_copy.get().get_size(measurements_view);
+
+    // Create the result buffer.
+    spacepoint_collection_types::buffer spacepoints(num_measurements,
+                                                    m_mr.main);
+    m_copy.get().setup(spacepoints)->ignore();
```
stephenswat, line 58: Does Alpaka use stream ordering? I.e. do we need to wait for this before launching the kernel?

device/alpaka/src/seeding/seed_finding.cpp
```diff
@@ -359,7 +360,7 @@ seed_finding::output_type seed_finding::operator()(
     seed_collection_types::buffer seed_buffer(
         pBufHost_counter->m_nTriplets, m_mr.main,
         vecmem::data::buffer_type::resizable);
-    m_copy.setup(seed_buffer);
+    m_copy.setup(seed_buffer)->ignore();
```
stephenswat, line 363: Same as above, do we need to wait on this queue before launching any kernels?

device/alpaka/src/utils/barrier.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+    ALPAKA_FN_ACC
+    bool blockOr(bool predicate) {
+        return ::alpaka::syncBlockThreadsPredicate<::alpaka::BlockOr>(
+            *m_acc, predicate);
+    }
+
+    private:
+    const TAcc* m_acc;
+};
```
stephenswat, line 34: Not a hard requirement, but could you perhaps add `blockAnd` and `blockCount` as defined in #602?

examples/run/alpaka/full_chain_algorithm.hpp
```diff
@@ -0,0 +1,133 @@
...
+    /// Device memory resource
+    vecmem::hip::device_memory_resource m_device_mr;
+    /// Memory copy object
+    vecmem::hip::copy m_copy;
+#else
+    /// Device memory resource
+    vecmem::memory_resource& m_device_mr;
+    /// Memory copy object
+    vecmem::copy m_copy;
+#endif
```
stephenswat, line 103: This is really not ideal :cry: but we can fix it later.

stephenswat (commented):
Looks good, but can we clarify the thing where we are ignoring events without synchronizing them?

Also, could you perhaps implement the CCA test harness in Alpaka? That would be useful for testing; but can also be done in a future PR.

device/alpaka/src/clusterization/measurement_sorting_algorithm.cpp
CrossR: It does, we do in general need a better way of dealing with this.

For CUDA, we should be able to use thrust, HIP -> rocThrust....outside of that, right now it should be a CPU and therefore be host accessible.

The test and validation with rocThrust is still ongoing and is likely going to need some extra work, though ideally that could exist outside of the scope of this PR, as its probably going to need a decent number of follow up PRs (mostly due to the pervasieness of thrust across traccc / detray etc making a "simple swap" less simple).

device/alpaka/src/clusterization/spacepoint_formation_algorithm.cpp
CrossR: Potentially...I don't believe there is any stream ordering (or, at least the vecmem operations aren't part of any alpaka queue at least).

examples/run/alpaka/full_chain_algorithm.hpp
CrossR: I think it is becoming more and more obvious that potentially a `vecmem::alpaka` or something similar that deals with this complication once rather than every time we import and every time we use `vecmem` is likely a better solution.

device/alpaka/CMakeLists.txt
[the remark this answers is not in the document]
CrossR: Somewhat...rocThrust support isn't really there yet, so this gets things working for CUDA.

Its temporary until we have an option to not use it for HIP, at which point we will only need it for CUDA.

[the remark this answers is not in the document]
stephenswat: Alright, as this is in the Alpaka branch and shouldn't affect the rest of traccc I'm okay with this. :slightly_smiling_face:

CrossR:
> Looks good, but can we clarify the thing where we are ignoring events without synchronizing them?

Yep, will follow up after this HLT workshop to check that.

> Also, could you perhaps implement the CCA test harness in Alpaka? That would be useful for testing; but can also be done in a future PR.

We've been speaking recently about how nice it would be to get the Alpaka work tested more thoroughly, so will follow up with this after this PR!

stephenswat:
> We've been speaking recently about how nice it would be to get the Alpaka work tested more thoroughly, so will follow up with this after this PR!

Alright! The test harness is designed to be platform independent, so _hopefully_ supporting Alpaka won't be more than, say, 50 LOC. We will see.

stephenswat:
That said @CrossR I think it would be excellent to get this in even if it is not entirely perfect yet; is it okay for you if I merge this?

CrossR:
Just debugging the `pragma` issue with GCC, then should be good to go.

The alpaka code is self contained enough, outside of the pragma change, that there shouldn't be any other interactions so the few missing bits here can be added as part of follow up PRS (plus I can get my recent work on track finding and fitting in, and get the Alpaka code closer to matching the CUDA status).

stephenswat:
@CrossR can you ping me when the issue is resolved?

CrossR:
@stephenswat I've disabled the `pragma` in GCC for now (since the compiler was the problem one in the first place annoyingly), so this should be good to merge.

PR #560, opened 2024-04-30
rocThrust for thrust functionality in HIP

StewMH:
Add test of rocThrust library so we can use thrust on AMD GPUs in the future. Conditional on new TRACCC_BUILD_HIP flag: we don't have a plain HIP implementation of any algorithms so it's slight overkill.

Note that test_thrust.hip is just a copy of test_thrust.cu. So we could achieve the same via ifdefs and conditional CMake, although the directory `test/cuda` would ideally change name.

Surprisingly the two thrust libraries can coexist since only NVidia thrust defines `THRUST_ENABLE_INSTALL_RULES` and therefore the CMake commands defined by both do not clash. This is more or less by accident, but hopefully OK for now.


StewMH:
cc @CrossR

StewMH:
Unclear to me why this is failing:
```
[ 12%] Creating directories for 'rocprim-download'
[ 25%] Performing download step (git clone) for 'rocprim-download'
-- rocprim-download download command succeeded.  See also /__w/traccc/traccc/build/rocprim-download/rocprim-download-prefix/src/rocprim-download-stamp/rocprim-download-download-*.log
[ 37%] No patch step for 'rocprim-download'
[ 50%] Performing configure step for 'rocprim-download'
CMake Error at /__w/traccc/traccc/build/rocprim-download/rocprim-download-prefix/src/rocprim-download-stamp/rocprim-download-configure-.cmake:49 (message):
  Command failed: 1

   '/usr/local/bin/cmake' '-DBUILD_TEST=OFF' '-DCMAKE_INSTALL_PREFIX=/__w/traccc/traccc/build/_deps/rocthrust-build/deps/rocprim' '-DCMAKE_PREFIX_PATH=/opt/rocm' '-GUnix Makefiles' '/__w/traccc/traccc/build/rocprim-src'

  See also

    /__w/traccc/traccc/build/rocprim-download/rocprim-download-prefix/src/rocprim-download-stamp/rocprim-download-configure-*.log


make[2]: *** [CMakeFiles/rocprim-download.dir/build.make:92: rocprim-download-prefix/src/rocprim-download-stamp/rocprim-download-configure] Error 1
make[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/rocprim-download.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
CMake Error at build/_deps/rocthrust-src/cmake/DownloadProject.cmake:168 (message):
  Build step for rocprim failed: 2
```
Locally I have rocPRIM, but I can make it download with `-DDOWNLOAD_ROCPRIM=TRUE`. However, it replicates.

StewMH:
OK I ran locally with the Docker container. It complains that:
```
CMake Error at cmake/VerifyCompiler.cmake:29 (message):
  On ROCm platform 'hipcc' or HIP-aware Clang must be used as C++ compiler.
```
So we need a new container with rocm + clang. Trying it out with 22.04

krasznaa:
:thinking: I'm not all too sight about this... :frowning: If rocThrust can't deal with HIP not being the thing used for the host code build, then we have a problem. :frowning: Since they don't just want "any version" of clang. :thinking: They want the clang version shipped with HIP.

(It's a longer story, that AMD's "HIP language support in CMake" makes use of the underlying `clang++` executable directly, they don't use the `hipcc` frontend. That setup is also a bit wonky with combining GCC + Clang for the project build. Though it does work somewhat at least.)

More than anything else, this looks like a shortcoming in their CMake configuration. I'm not at all convinced that the basic problem is on our side. Even if updating the Docker images might well be due.

StewMH:
We could potentially get round this by installing rocprim and maybe some other developer tools in advance. That's the setup I have locally.

StewMH:
No response on the issue above from AMD. However, there is probably some value in running a clang build in the CI even if we are doing it for the wrong reasons.

The CI looks OK now - @krasznaa, @stephenswat could I have a review?

.github/workflows/builds.yml
```diff
@@ -30,6 +30,9 @@ jobs:
           - name: CPU
             container: ghcr.io/acts-project/ubuntu2004:v30
             options: -DTRACCC_USE_ROOT=FALSE
+          - name: HIP 
+            container: ghcr.io/acts-project/ubuntu2204_rocm_clang
```
krasznaa, line 34: You need to put a particular version number here. We can't have the CI using the "latest" image. (Allowing the CI to break by somebody updating the image in some unforeseen way.)

CMakeLists.txt
```diff
@@ -118,6 +119,17 @@ set( TRACCC_THRUST_OPTIONS "" CACHE STRING
 mark_as_advanced( TRACCC_THRUST_OPTIONS )
 thrust_create_target( traccc::Thrust ${TRACCC_THRUST_OPTIONS} )
 
+option( TRACCC_USE_SYSTEM_ROCTHRUST
+   "Pick up an existing installation of rocThrust from the build environment"
+   ${TRACCC_USE_SYSTEM_LIBS} )
+if( TRACCC_BUILD_HIP )
```
krasznaa, line 125: You will need a separate cache variable here, not `TRACCC_BUILD_HIP`.

The logic with all of our externals is that we use 2 cache variables with all of them.
  - `TRACCC_SETUP_<FOO>`: Flag controlling whether traccc should do **anything** to set up the external in question;
  - `TRACCC_USE_SYSTEM_<FOO>`: Flag controlling whether to set up `<FOO>` through `find_package(...)` or by building it ourselves.

The first one allows us to build traccc as part of an even higher-level project. Either Acts, or Athena. (Both of which we exercise already.) And allow Acts / Athena to set up a rocThrust version that they want.

Long story short, please introduce `TRACCC_SETUP_ROCTHRUST`, and use that here. `TRACCC_BUILD_HIP` should just be used to decide whether the HIP code inside of traccc would be built. Not what externals would be set up.

cmake/traccc-compiler-options-hip.cmake
```diff
@@ -0,0 +1,12 @@
...
+#
+# (c) 2021-2024 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# Include the helper function(s).
+include( traccc-functions )
+
+#HIP requires position-independent executables for linking
+traccc_add_flag( CMAKE_HIP_FLAGS "-fPIE" )
```
krasznaa, line 11: I'm on board with adding a placeholder file like this, but `-fPIE` should not be added like this. Please use the following formalism instead:

https://github.com/acts-project/vecmem/blob/main/tests/hip/CMakeLists.txt#L30-L34

That is just more platform independent.

cmake/traccc-compiler-options-hip.cmake
```diff
@@ -0,0 +1,12 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2021-2024 CERN for the benefit of the ACTS project
```
krasznaa, line 3: In the R&D projects I prefer to date the files with a starting year that they were created in. So in this case I would write:

```suggestion
# (c) 2024 CERN for the benefit of the ACTS project
```

But this is purely a subjective thing.

extern/rocThrust/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+if( POLICY CMP0135 )
+   cmake_policy( SET CMP0135 NEW )
+endif()
+
+# Tell the user what's happening.
+message( STATUS "Building rocThrust as part of the TRACCC project" )
+
+# Declare where to get rocThrust from.
+set( TRACCC_ROCTHRUST_SOURCE
+   "GIT_REPOSITORY;https://github.com/ROCm/rocThrust;GIT_TAG;rocm-6.1.0"
```
krasznaa, line 21: Crud, rocThrust uses submodules just like Thrust does... 😦 My preference would be to download a TGZ file for it instead, but I guess you checked, and their TGZ files don't contain the code from the submodules of the repository?

But even if we stay with using git download, at the very least please do:

```suggestion
   "GIT_REPOSITORY;https://github.com/ROCm/rocThrust.git;GIT_TAG;rocm-6.1.0"
```

(Leaving off the `.git` postfix prevents this from working with older git versions.)

tests/hip/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+enable_language(HIP)
```
krasznaa, line 1: Please add a copyright statement to the new file.

extern/rocThrust/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+   CACHE STRING "Source for Thrust, when built as part of this project" )
+mark_as_advanced( TRACCC_ROCTHRUST_SOURCE )
+FetchContent_Declare( rocThrust ${TRACCC_ROCTHRUST_SOURCE} )
+
+# Get it into the current directory.
+FetchContent_MakeAvailable( rocthrust )
```
krasznaa, line 27: You'll have to surround this line with something analogous to:

https://github.com/acts-project/traccc/blob/main/extern/eigen3/CMakeLists.txt#L27-L46

We must turn off `BUILD_TESTING` for rocThrust, while still keeping it turned on for our own code. Doing this should actually resolve most of the issues that you've been seeing. Since we really only need the rocThrust build system to make the project's headers available to us. We don't need it to build anything itself!

krasznaa (changes requested):
A couple of things... 😉

.github/workflows/builds.yml
StewMH: My understanding is that we need a new tag of `machines` for this. I could change this to `ubuntu2204_rocm_clang:sha-3a6b0b2` for now, which matches latest. Or we can wait for a new machines tag.

krasznaa: Then we should make a new tag of it. Though again, I believe that once we turn off the unit test builds of rocThrust, we could possibly survive with something as old as `ghcr.io/acts-project/ubuntu1804_rocm:v11`. 🤔

cmake/traccc-compiler-options-hip.cmake
krasznaa: No. Rather, you should add

```cmake
set_target_properties( traccc_test_hip PROPERTIES
   POSITION_INDEPENDENT_CODE TRUE )
```

in the newly introduced HIP test `CMakeLists.txt`.

You see, shared libraries are **always** built with "position independent code". (Otherwise they would fundamentally not work.) It's static libraries and executables that usually are built with position **dependent** code. So we just need to tell a couple of them to "be independent". We shouldn't do this on a global scale at all.

.github/workflows/builds.yml
StewMH: Unfortunately not. It still insists on running its compiler checks with this change...

cmake/traccc-compiler-options-hip.cmake
StewMH: Ah, I misunderstood. Added this in the two executables that need it.

StewMH:
I followed up on the cmake warnings in https://github.com/ROCm/rocm-cmake/issues/154

krasznaa:
Stewart, how would you feel about me intervening here? :thinking: I'd like to put https://github.com/krasznaa/traccc/commit/1c26d464f6cc2ccb756b47748591cc2c690bec04 on top of your developments. Technically I can do it, but didn't want to modify your branch before running it by you.

Long story short, I spent some time yesterday and today to force rocThrust into submission. To force its CMake configuration to work with our existing ROCm/HIP setups. Mind you, it now works for me with both `HIP_PLATFORM=amd` **and** `HIP_PLATFORM=nvidia`. :smile:

Though there is still one glaring issue with all of this. We absolutely must depend on rocThrust privately only. As I don't see yet how we could export rocThrust properly with the rest of our code. (Their CMake kung-fu is not an amazing one.) :frowning:

StewMH:
Looks good to me. I did consider patching it but thought this might be getting a bit complicated

krasznaa:
So... We're on a good track, but still one issue to sort out... :thinking:

This setup now works with older versions of ROCm/HIP, and with HIP's NVIDIA backend. (Even when using a newer version of HIP.) But at least with HIP 6.0.2, which I tested just now on my home PC that has an AMD card in it, this current setup fails to build with:

```
[build] [ 68%] Linking HIP executable ../../bin/traccc_test_hip
[build] /usr/bin/ld: /opt/rocm-6.0.2/lib/libamdhip64.so: undefined reference to `hsa_amd_memory_copy_engine_status@ROCR_1'
[build] /usr/bin/ld: /opt/rocm-6.0.2/lib/libamdhip64.so: undefined reference to `hsa_amd_memory_async_copy_on_engine@ROCR_1'
[build] clang++: error: linker command failed with exit code 1 (use -v to see invocation)
[build] gmake[2]: *** [tests/hip/CMakeFiles/traccc_test_hip.dir/build.make:105: bin/traccc_test_hip] Error 1
[build] gmake[1]: *** [CMakeFiles/Makefile2:5021: tests/hip/CMakeFiles/traccc_test_hip.dir/all] Error 2
[build] gmake[1]: *** Waiting for unfinished jobs....
```

This is an interesting issue. The missing symbols come from `libhsa-runtime64.so`. Which is absolutely part of my ROCm/HIP installation. But so far, for the naive HIP tests that we did in these projects, I never needed that library. :thinking: So

https://github.com/acts-project/vecmem/blob/main/cmake/FindHIPToolkit.cmake

doesn't look for it, it only sets up `HIP::hiprt`. (Which is the only library that we needed to link against explicitly so far.)

Long story short, we'll need a new VecMem tag as well before this PR would be in a really good state. :frowning:

krasznaa (approved):
Turns out, that no. :thinking: It's rather a misconfiguration on my machine.

I have multiple different versions of ROCm installed here. With the "default" being version 5.4. (Since oneAPI's AMD plugin supports that version officially.) But in this build I was trying to rather use 6.0. But clearly my environment setting breaks down with that. :frowning:

As soon as I use version 5.4 (which has the same library setup), things work fine. So there is probably some need to make the VecMem code smarter later on, but this PR does not need to wait for that.

PR #563, opened 2024-05-04
Limit the number of tracks (or branches) per seed for GPU CKF

beomki-yeo:
GPU version of #488. Also added a dedicated test to validate the change

### DIFF
--- a/device/cuda/src/finding/finding_algorithm.cu
+++ b/device/cuda/src/finding/finding_algorithm.cu
@@ -147,6 +147,7 @@ __global__ void propagate_to_next_surface(
     vecmem::data::vector_view<unsigned int> param_to_link_view,
     vecmem::data::vector_view<typename candidate_link::link_index_type>
         tips_view,
+    vecmem::data::vector_view<unsigned int> n_tracks_per_seed_view,
     unsigned int& n_out_params) {
 
     int gid = threadIdx.x + blockIdx.x * blockDim.x;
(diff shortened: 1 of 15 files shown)


device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -219,14 +220,22 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
-    bound_track_parameters_collection_types::device seeds(seeds_buffer);
+    bound_track_parameters_collection_types::const_device seeds(seeds_buffer);
     thrust::copy(thrust::cuda::par.on(stream), seeds.begin(), seeds.end(),
                  in_params.begin());
 
+    // Number of tracks per seed
+    vecmem::data::vector_buffer<unsigned int> n_tracks_per_seed_buffer(
+        n_seeds, m_mr.main);
+    vecmem::device_vector<unsigned int> n_tracks_per_seed_device(
+        n_tracks_per_seed_buffer);
```
krasznaa, line 237: I'd prefer to not use a `vecmem::device_vector` here. As it could easily result in some unsafe function calls later on. :thinking:

I'd rather use `n_tracks_per_seed_buffer.ptr()` and `n_tracks_per_seed_buffer.ptr() + n_tracks_per_seed_buffer.size()` as the parameters of `thrust::fill`.

Or, now that I think about it, since you're "just" zeroing out the vector, let's not use `thrust::fill` at all. You should instead write:

```c++
m_copy.memset(n_tracks_per_seed_buffer, 0)->ignore();
```

When using `vecmem::cuda::async_copy`, this will be an asynchronous operation out of the box.

krasznaa (changes requested):
I don't claim to fully understand the code.

But apart from the technicality of how to "zero out" a 1D buffer, I'm fine with the rest.

device/cuda/src/finding/finding_algorithm.cu
beomki-yeo: Thanks for the useful syntax. Fixed now

krasznaa (approved):
Let's go with this. :thinking:

Issue #565, opened 2024-05-04
(Asynchronous) Error in CUDA Clusterization, main branch (2024.05.04.)


krasznaa:
The ODD ttbar simulation just keeps on giving...

While I can process the low-ish $\mu$ samples successfully, at and above $\mu$ = 100 I run into:

```
[bash][Legolas]:traccc > ./out/build/sycl/bin/traccc_seq_example_cuda --detector-file=geometries/odd/odd-detray_geometry_detray.json --grid-file=geometries/odd/odd-detray_surface_grids_detray.json --use-detray-detector --digitization-file=geometries/odd/odd-digi-geometric-config.json --input-directory=odd/geant4_ttbar_mu100/ --input-events=2

Running Full Tracking Chain Using CUDA

>>> Detector Options <<<
  Detector file       : geometries/odd/odd-detray_geometry_detray.json
  Material file       : 
  Surface rid file    : geometries/odd/odd-detray_surface_grids_detray.json
  Use detray::detector: yes
  Digitization file   : geometries/odd/odd-digi-geometric-config.json
>>> Input Data Options <<<
  Input data format             : csv
  Input directory               : odd/geant4_ttbar_mu100/
  Number of input events        : 2
  Number of input events to skip: 0
>>> Clusterization Options <<<
  Target cells per partition: 1024
>>> Track Seeding Options <<<
  None
>>> Track Finding Options <<<
  Track candidates range   : 3:100
  Minimum step length for the next surface: 0.5 [mm] 
  Maximum step counts for the next surface: 100
  Maximum Chi2             : 30
  Maximum branches per step: 4294967295
  Maximum number of skipped steps per candidates: 3
>>> Track Propagation Options <<<
  Constraint step size  : 3.40282e+38 [mm]
  Overstep tolerance    : -100 [um]
  Minimum mask tolerance: 1e-05 [mm]
  Maximum mask tolerance: 1 [mm]
  Search window         : 0 x 0
  Runge-Kutta tolerance : 0.0001
>>> Performance Measurement Options <<<
  Run performance checks: no
>>> Accelerator Options <<<
  Compare with CPU results: no

WARNING: No material in detector
WARNING: No entries in volume finder
Detector check: OK
WARNING: No material in detector
WARNING: No entries in volume finder
Detector check: OK
[...]

PR #561, opened 2024-05-03
Data V7, main branch (2024.05.03.)

krasznaa:
Switched to version 7 of the traccc data file.

It has 2 major updates:
  - The ODD geometry files have been updated to the "latest" ones generated by @asalzburger. (He since seems to have generated slightly newer ones, but I stuck with the ones that I tested the traccc code with already.)
  - Replaced the existing ODD Geant4 and Fatras simulation files with a new set of Geant4 files.

I did the latter with the code, and instructions from: https://github.com/acts-project/acts/pull/3169

As it is explained in that ticket, the current `main` branch of [acts](https://github.com/acts-project/acts) can run simulation on the ODD geometry out of the box. Which I did in the following way:
  - Cloned the HEAD of the Acts `main` branch;
    * Making sure that the ODD submodule of the repository would be set up correctly. (`git submodule init && git submodule update`)
  - Started a Docker container with the following command in the directory where I cloned Acts:

```
docker run -it --rm -v $PWD:$PWD -w $PWD ghcr.io/acts-project/ubuntu2204:v41
```

  - Inside the container, used the following BASH script to build the code, and run some specific simulations:

```sh
#!/bin/bash
#
# Script for running a set of ODD simulation jobs for traccc
# performance measurements.
#

# Stop on errors.
set -e

# Default script arguments.
ACTS_DIR="./acts"
BUILD_DIR="./build"

# Helper function for printing usage information for the script.
usage() {
    echo "Script for running ODD simulations for traccc using the"
    echo "ghcr.io/acts-project/ubuntu2204:v41 Docker image."
    echo ""
    echo "Usage: ${BASH_SOURCE[0]} [options]"
    echo "Options:"
    echo "  -h/--help:      Print this message."
    echo ""
    echo "  -a/--acts-dir:  Directory with the Acts sources"
    echo "                  [${ACTS_DIR}]"
    echo "  -b/--build-dir: Build directory for Acts"
    echo "                  [${BUILD_DIR}]"
    echo ""
}

# Parse the command line argument(s).
while [[ $# > 0 ]]
do
    case $1 in
	-a|--acts-dir)
	    ACTS_DIR=$2
	    shift
	    ;;
	-b|--build-dir)
	    BUILD_DIR=$2
	    shift
	    ;;
	-h|--help)
	    usage
	    exit 0
	    ;;
	*)
	    echo "ERROR: Unknown argument: $1"
	    echo ""
	    usage
	    exit 1
	    ;;
    esac
    shift
done

[...]

PR #566, opened 2024-05-04
ODD ttbar Clusterization Fix, main branch (2024.05.04.)

krasznaa:
Made high-$\mu$ ODD ttbar samples succeed with (device) clusterization.

As discussed in #565, with $\mu$ = 100 and higher, on the files that I generated as described in #561, I had to do this to avoid running into assertion errors from the clusterization code. :thinking:

I'm very open to discussions about why this is, but the code, with its current logic of how it splits the cells into partitions, does need this update for the highest pileup ODD simulations.


stephenswat (approved):
This is, of course, a perfectly fine change to make. Just note that the trade-off here is that increasing this parameter is expected to decrease occupancy and, as a result, performance. But that's something we can figure out later.

PR #570, opened 2024-05-04
Allow more branches in the ckf matching rate test

beomki-yeo:
This test is failing after #563. 

There seems to be almost identical twin tracks thanks to the poor random generator, therefore, the max number of branches per seed should increase in this specific test (10->500).

If one asks why it didn't fail before #563, max_num_branches_per_seed was the max value.


krasznaa (approved):
Looks good.

Issue #569, opened 2024-05-04
CUDA Reconstruction Stuck (2024.05.04.)


krasznaa:
Processing every ODD ttbar event that I made in #561, I have one that makes the reconstruction run forever. :confused:

```
[bash][Legolas]:traccc > ./out/build/sycl/bin/traccc_seq_example_cuda --detector-file=geometries/odd/odd-detray_geometry_detray.json --grid-file=geometries/odd/odd-detray_surface_grids_detray.json --use-detray-detector --digitization-file=geometries/odd/odd-digi-geometric-config.json --input-directory=odd/geant4_ttbar_mu200/ --input-events=1 --input-skip=2 

Running Full Tracking Chain Using CUDA

>>> Detector Options <<<
  Detector file       : geometries/odd/odd-detray_geometry_detray.json
  Material file       : 
  Surface rid file    : geometries/odd/odd-detray_surface_grids_detray.json
  Use detray::detector: yes
  Digitization file   : geometries/odd/odd-digi-geometric-config.json
>>> Input Data Options <<<
  Input data format             : csv
  Input directory               : odd/geant4_ttbar_mu200/
  Number of input events        : 1
  Number of input events to skip: 2
>>> Clusterization Options <<<
  Target cells per partition: 1024
>>> Track Seeding Options <<<
  None
>>> Track Finding Options <<<
  Track candidates range   : 3:100
  Minimum step length for the next surface: 0.5 [mm] 
  Maximum step counts for the next surface: 100
  Maximum Chi2             : 30
  Maximum branches per step: 4294967295
  Maximum number of skipped steps per candidates: 3
>>> Track Propagation Options <<<
  Constraint step size  : 3.40282e+38 [mm]
  Overstep tolerance    : -100 [um]
  Minimum mask tolerance: 1e-05 [mm]
  Maximum mask tolerance: 1 [mm]
  Search window         : 0 x 0
  Runge-Kutta tolerance : 0.0001
>>> Performance Measurement Options <<<
  Run performance checks: no
>>> Accelerator Options <<<
  Compare with CPU results: no

WARNING: No material in detector
WARNING: No entries in volume finder
Detector check: OK
WARNING: No material in detector
WARNING: No entries in volume finder
Detector check: OK
WARNING: @traccc::io::csv::read_cells: 17547 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/odd/geant4_ttbar_mu200/event000000002-cells.csv
```

[...]

PR #571, opened 2024-05-04
Kalman Fitter Step Aborter, main branch (2024.05.04.)

krasznaa:
In order to address / work around the issue described in #569, added a "step count aborter" to the track fitting.

It is to prevent the track fitting from going into an endless loop. This simple failsafe will abort the propagation of a track if it cannot find the next detector surface in a maximal number of steps. (100 by default.)

Note that this is yet another reminder that we'll have to add some message logging facilities into this project. :thinking: For now I reused the [VECMEM_DEBUG_MSG](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/utils/debug.hpp) macro. But this is far from perfect. :frowning: (We can't just use `printf(...)` here, since SYCL doesn't support such output.)

From what I've seen with the ODD simulation files, 100 as the maximal number of steps seems to be a reasonable value. But I'm open for input on this. (100 is also what the CKF uses currently.)

### DIFF
--- a/core/include/traccc/fitting/kalman_filter/kalman_step_aborter.hpp
+++ b/core/include/traccc/fitting/kalman_filter/kalman_step_aborter.hpp
@@ -1,0 +1,65 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/definitions/qualifiers.hpp"
+
+// Detray include(s).
+#include "detray/propagator/base_actor.hpp"
+
+// VecMem include(s).
+#include <vecmem/utils/debug.hpp>
+
+namespace traccc {
+
+/// Aborter making sure that propagation would not exceed a certain step count
+///
+/// It is used mostly as a failsafe during the fitting stage. As long as
+/// everything works well, this aborter should never trigger. It is here to
+/// serve as a failsafe in case of a bug in the track finding / fitting
+/// algorithm.
+///
+struct kalman_step_aborter : public detray::actor {
+
+    /// The state of the aborter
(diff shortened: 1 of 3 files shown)


beomki-yeo (approved):
Looks good to me

core/include/traccc/fitting/kalman_filter/kalman_step_aborter.hpp
```diff
@@ -0,0 +1,57 @@
...
+    template <typename propagator_state_t>
+    TRACCC_HOST_DEVICE void operator()(state& abrt_state,
+                                       propagator_state_t& prop_state) const {
+
+        // Abort if the step count exceeds the maximum allowed
+        if (++(abrt_state.step) > abrt_state.max_steps) {
+            VECMEM_DEBUG_MSG(1, "Kalman fitter step aborter triggered");
+            prop_state._heartbeat &= prop_state._navigation.abort();
+        }
+    }
```
beomki-yeo, line 53: I have to mention that it is desirable  to reset the step for every surface intersection. 100 of CKF is also for the surface-to-surface transport

It would be enough to add sth like:
```
auto& navigation = prop_state._navigation;
if (navigation.is_on_surface()){
    abrt_state.step = 0;
}
```

@niermann999 can you take a look if this command is correct? I don't remember how to check if the track is on surface

beomki-yeo (changes requested):
I actually this needs some changes...

core/include/traccc/fitting/kalman_filter/kalman_step_aborter.hpp
krasznaa: Sure, happy to take this on. This updated version of the code also makes the reconstruction finish on my GPU. :wink:

beomki-yeo (approved):
I meant to check is_on_any_surface rather than is_on_sensitive_surface but this won't make big difference.

PR #573, opened 2024-05-04
Fix the default value of maxmimum number of branches per seed

beomki-yeo:
This is important parameter for GPU CKF and it is not supposed to be the max of `unsigned int`. 

May not impact the performance  a lot now I think :thinking: But it is pretty error prone with current default value


krasznaa (approved):
:+1:

Let me just merge all these in, in order...

PR #574, opened 2024-05-05
Spacepoint OBJ Writing, main branch (2024.05.05.)

krasznaa:
I wanted to make the hacky thing that I've been doing over the last week, a bit more formal. Laying a foundation for saving even more complex meshes / paths / etc. later on for fancy visualization.

There's also a bit more of a fundamental thing that I'm introducing here. :thinking: Instead of having our code use `traccc::io::data_directory()` all over the place, I now introduced `traccc::io::get_absolute_path(...)`. The purpose of this function is to allow us to use absolute path names for our input/output files/directories if we so choose. Since doing this would've made my life a whole lot easier while experimenting with the ODD Geant4 simulation...

But I didn't want to go too crazy with this PR. Here I only allow `traccc::io::write(...)` to write to an absolute path, and not only under `$TRACCC_TEST_DATA_DIR`.

Finally, I taught `traccc_seq_example` to output its reconstructed spacepoints if you specify `--output-directory` among its arguments. (My idea is that later on we could output a whole bunch of other OBJ files as well if we choose.)

As a side-thing: It's now pretty clear to me that the "single muon" samples that I generated are far from being single muons. :confused: Since I tend to see 4 high-pt particles in each of those events.

image

(The two different colours are from two separate events in this picture.)

Not a big problem at the moment, but something to debug with @asalzburger later on. :thinking:

### DIFF
--- a/examples/options/include/traccc/options/output_data.hpp
+++ b/examples/options/include/traccc/options/output_data.hpp
@@ -13,6 +13,7 @@
 
 // System include(s).
 #include <string>
+#include <string_view>
 
 namespace traccc::opts {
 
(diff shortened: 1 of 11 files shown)


asalzburger:
Oh yea, I see the "4" in the particle config - will fix it asap.

krasznaa:
@beomki-yeo, could you help this one through? I'd like to get #576 merged in with relative haste...

beomki-yeo (approved):
Looks good to me

PR #575, opened 2024-05-05
Finding performance writer: comments + fake tracks

SylvainJoube:
While writing a ROOT documentation for the `finding` and `ambiguity resolution` algorithms, I thought it would be interesting to have a visualization for the `fake tracks` rates. The `greedy ambiguity resolution algorithm` successfully managed to remove every track duplicates while retaining reasonable efficiency, but I wanted to make sure that it would also greatly reduce the number of fake tracks, at least on the data generated by the toy detector (mostly 100GeV muons).

The `fake_tracks_plot_tool.hpp` file is a duplication of the code from the `duplication_plot_tool.hpp` file. I think it would be better to merge the two codes, so I will do it if you think it’s a good idea.

If you like the idea, I will write the corresponding documentation and append it to the `doc/root/README.md` file from [PR 535](https://github.com/acts-project/traccc/pull/535)

For example, with toy detector generated data and 100GeV muons, the mean value for “fake tracks” goes from 1.707 with the `finding algorithm` down to 0.0155 after the `ambiguity resolution algorithm` (a x110 decrease!):

After the `finding algorithm`:

After the `ambiguity resolution algorithm`:

I used the following code to generate the ROOT data with `traccc_seeding_example`:

```bash
# Generate toy detector data
./bin/traccc_simulate_toy_detector --gen-vertex-xyz-mm=0:0:0 --gen-vertex-xyz-std-mm=0:0:0 --gen-mom-gev=100:100 --gen-phi-degree=0:360 --gen-events=10 --gen-nparticles=2000 --output-directory=detray_simulation/toy_detector/n_particles_2000/ --gen-eta=-3:3 --constraint-step-size-mm=1

# …copy data were needed…

# Run the seeding example
./bin/traccc_seeding_example --input-directory=detray_simulation/toy_detector/n_particles_2000/ --detector-file=toy_detector_geometry.json --material-file=toy_detector_homogeneous_material.json --grid-file=toy_detector_surface_grids.json --input-event=1 --track-candidates-range=3:30 --constraint-step-size-mm=1000 --check-performance
```

### DIFF
--- a/performance/src/efficiency/finding_performance_writer.cpp
+++ b/performance/src/efficiency/finding_performance_writer.cpp
@@ -10,6 +10,7 @@
 
 #include "duplication_plot_tool.hpp"
 #include "eff_plot_tool.hpp"
+#include "fake_tracks_plot_tool.hpp"
 #include "track_classification.hpp"
 
 // ROOT include(s).
@@ -31,7 +32,8 @@ struct finding_performance_writer_data {
     finding_performance_writer_data(
         const finding_performance_writer::config& cfg)
         : m_eff_plot_tool({cfg.var_binning}),
-          m_duplication_plot_tool({cfg.var_binning}) {}
+          m_duplication_plot_tool({cfg.var_binning}),
+          m_fake_tracks_plot_tool({cfg.var_binning}) {}
 
     /// Plot tool for efficiency
     eff_plot_tool m_eff_plot_tool;
@@ -40,6 +42,10 @@ struct finding_performance_writer_data {
     /// Plot tool for duplication rate
     duplication_plot_tool m_duplication_plot_tool;
     duplication_plot_tool::duplication_plot_cache m_duplication_plot_cache;
+
+    // Plot tool for fake tracks monitoring
+    fake_tracks_plot_tool m_fake_tracks_plot_tool;
+    fake_tracks_plot_tool::fake_tracks_plot_cache m_fake_tracks_plot_cache;
 
     measurement_particle_map m_measurement_particle_map;
     particle_map m_particle_map;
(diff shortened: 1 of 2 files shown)


SylvainJoube:
If I’m not mistaken, we can do the following math.  
5296 tracks were found and fitted, 1780 remained after ambiguity resolution. So only 1 out of (about) 3 tracks were kept, whereas only 1 out of 110 fake tracks were kept. All duplicated tracks were removed, meaning the ambiguity resolution algorithm disproportionately affected fake and duplicated tracks. The efficiency graphs seem to be largely unaffected by ambiguity resolution (but I can’t compare the figures as they are not mentioned in the ROOT plots for some reason I don’t understand).

```
==> Statistics ... 
- read    7830 spacepoints
- read    7830 measurements
- created (cpu)  4794 seeds
- created (cpu)  5296 found tracks
- created (cpu)  5296 fitted tracks
- created (cpu)  1780 ambiguity free tracks
```

beomki-yeo:
Great I assigned myself for review

SylvainJoube:
Thanks a lot Beom Ki! :)

performance/src/efficiency/finding_performance_writer.cpp
```diff
@@ -126,22 +156,43 @@ void finding_performance_writer::write_common(
...
         std::vector<particle_hit_count> particle_hit_counts =
             identify_contributing_particles(measurements, evt_map.meas_ptc_map);
 
         if (particle_hit_counts.size() == 1) {
             auto pid = particle_hit_counts.at(0).ptc.particle_id;
             match_counter[pid]++;
         }
+
+        if (particle_hit_counts.size() > 1) {
+            for (particle_hit_count const& phc : particle_hit_counts) {
```
beomki-yeo, line 179: Can you explain the detail of logic? If there are more than one particle contributing to the track, it is considered to be a fake track?

According to this definition and the later changes below, I guess one single track can increase the fake count multiple times? (i.e. If there are three particles contributing to a track then the fake count increases by 3?)

beomki-yeo (commented):
Looks good to me but I might need to understand the logic clearly. Actually I don't know the exact definition of fake rate. Could you find its definition?

performance/src/efficiency/finding_performance_writer.cpp
SylvainJoube: Thanks for your review :)

I defined a `fake track` as “a track made with measurements from more than one single truth particle”.

For each `truth particle`, the `fake rate` tells how many `fake tracks` were reconstructed with at least one measurement belonging to this particle.

The ROOT plots show “the number of `fake tracks` per particle property/value”, so it’s the other way around if I understood you correctly. It’s: if 3 `fake tracks` share one common particle, then the `fake count` for this particle will be 3.

If a `fake track` is made of measurements belonging to 3 truth particles, then the `fake_counter` for each particle will be increased by 1.

beomki-yeo: Thanks for the explanation. Then the y-axis of the fake rate plots is the averaged number of fake tracks per particle?

SylvainJoube: Sorry for taking so long to reply. Yes, absolutely, I have just changed the plot title to `"Averaged number of fake tracks per particle"`, that's a much better and informative title, thanks!

beomki-yeo (approved):
Looks good to me. I will update the branch and merge

SylvainJoube:
Okay, thanks!

PR #577, opened 2024-05-07
README Update, main branch (2024.05.07.)

krasznaa:
Add the status of the Alpaka and Kokkos algorithms. Hopefully getting it correct. But others should speak up if not.

Somebody with time on their hands should also update the Mermaid graph. Pointing at all of the existing algorithm classes. :thinking: But that seemed like a bit too much tinkering for me for today...

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,16 +4,16 @@ Demonstrator tracking chain for accelerators.
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


krasznaa:
Actually, with all the talk about rocThrust recently, I think it's fair to label "track finding" as being under development with Alpaka as well... :thinking:

PR #581, opened 2024-05-08
ODD Strip Module Handling, main branch (2024.05.08.)

krasznaa:
As discussed many times recently, currently we are handling all modules / surfaces as pixels when reconstructing ODD data. Which leads to some funky behaviour in its short and long strips.

As discussed on Mattermost, we can use a sort of hacky way to determine which modules / surfaces are 1D and 2D, by looking at the variations assigned to their 2 dimensions. Since the digitization config always describes 2 dimensions for the surfaces, even for strip modules. (It's a long story...)

With the info added to `traccc::digitization_config` and then to `traccc::cell_module`, I could make the host and device measurement creation algorithms take this information into account.

It did not cause any significant changes in our host<->device agreement numbers, those just moved around a tiny bit. But also, this did not seem to break anything either... :wink:

### DIFF
--- a/core/include/traccc/clusterization/impl/measurement_creation.ipp
+++ b/core/include/traccc/clusterization/impl/measurement_creation.ipp
@@ -98,7 +98,7 @@ TRACCC_HOST_DEVICE inline void fill_measurement(
         if (module.pixel.dimension == 1) {
             m.meas_dim = 1;
             m.local[1] = 0.f;
-            m.variance[1] = 1000.f;
+            m.variance[1] = module.pixel.variance_y;
         }
     }
 }
--- a/core/include/traccc/clusterization/impl/measurement_creation.ipp
+++ b/core/include/traccc/clusterization/impl/measurement_creation.ipp
@@ -93,6 +93,13 @@ TRACCC_HOST_DEVICE inline void fill_measurement(
 
         // For the ambiguity resolution algorithm, give a unique measurement ID
         m.measurement_id = measurement_index;
+
+        // Adjust the measurement object for 1D surfaces.
+        if (module.pixel.dimension == 1) {
+            m.meas_dim = 1;
+            m.local[1] = 0.f;
+            m.variance[1] = 1000.f;
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


core/include/traccc/clusterization/impl/measurement_creation.ipp
```diff
@@ -93,6 +93,13 @@ TRACCC_HOST_DEVICE inline void fill_measurement(
 
         // For the ambiguity resolution algorithm, give a unique measurement ID
         m.measurement_id = measurement_index;
+
+        // Adjust the measurement object for 1D surfaces.
+        if (module.pixel.dimension == 1) {
+            m.meas_dim = 1;
+            m.local[1] = 0.f;
+            m.variance[1] = 1000.f;
```
krasznaa, line 101: Note that I set this variance sort of arbitrarily here. :thinking:

For the ODD the correct values, for the different modules, would be: `2024.39306640625`, `972.7376098632812` and `2029.76708984375`.

If you think that would be useful to have, I could add it as yet one more parameter to `traccc::cell_module`. :thinking:

krasznaa (commented):
Note that I also didn't touch the spacepoint creation algorithm in this PR. That one will need a bit more careful PR, since our device code currently assumes that it would create a fixed sized space point collection. Which would have to become a resizable one. :thinking:

So, in a future PR...

core/include/traccc/clusterization/impl/measurement_creation.ipp
krasznaa: Decided to make it "more correct" in the end.

With SoA containers hopefully the overhead of all of this will be minimal in the end. :thinking:

krasznaa:
I'll open a PR with "truth maching code updates" still today. But until then, somebody should have a look at this PR. Since the truth matching code that I'll propose, absolutely relies on these changes.

beomki-yeo (approved):
I am OK with the change for the moment. I hope that it is not the permanent solution as you implied in the description above.

krasznaa:
This sort of "detector description" information will have to remain in some traccc specific format, so that we could give it to the measurement and spacepoint creation algorithms conveniently. 

But the source of the information will not remain so hacky of course. 

I've been working for a while on replacing our current cell_module type completely. It was just not the highest priority recently. But I'll have a proposal soon for how we should handle this info in our algorithms in the long term. 😉

PR #583, opened 2024-05-11
Data Ignore Update, main branch (2024.05.11.)

krasznaa:
This has been bugging me since a while, so finally it's time to fix it.

Many of our unit tests are creating files/directories under `data/`. While that is a separate issue of its own, this just always leaves me with a bunch of directories in my local clone that git doesn't know about. Worse yet, VSCode really freaks out about the thousands of files that it all of a sudden tries to track.

The `*/` formalism seems to do what we need there. :thinking:


beomki-yeo (approved):
Looks good to me

PR #585, opened 2024-05-13
ROOT Build Fix, main branch (2024.05.13.)

krasznaa:
Made the [ROOT](https://root.cern/) build use "System VC".

Any modern version of [ROOT](https://root.cern/) comes with both [TBB](https://github.com/oneapi-src/oneTBB) and [Vc](https://github.com/VcDevel/Vc) included. So our project should not build either of those things, when using [ROOT](https://root.cern/).

I came across this a couple of times already, but working on #582 was the last straw to finally do something about it. :thinking:


beomki-yeo (approved):
Good

PR #586, opened 2024-05-13
ref: Update to detray version 0.65.0

niermann999:
Mainly removes a few instances of too strict checking for the material and the stepper, as well as removing the transform for concentric cylinders and setting their boundary values correctly.

The material maps should now run by just adding the material map file to the detector reader

### DIFF
--- a/extern/detray/CMakeLists.txt
+++ b/extern/detray/CMakeLists.txt
@@ -18,7 +18,7 @@ message( STATUS "Building Detray as part of the TRACCC project" )
 
 # Declare where to get Detray from.
 set( TRACCC_DETRAY_SOURCE
-"URL;https://github.com/acts-project/detray/archive/refs/tags/v0.64.1.tar.gz;URL_MD5;28b9a1a761a37536fe0b9ee3502a0aba"
+"URL;https://github.com/acts-project/detray/archive/refs/tags/v0.65.0.tar.gz;URL_MD5;4bb0e3a489fea8e460223def1fcd01be"
    CACHE STRING "Source for Detray, when built as part of this project" )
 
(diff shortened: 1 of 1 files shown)


beomki-yeo (approved):

niermann999:
> At least a sentence of a description for the update could be good. 😛

Yes, sorry, I had an appointment and just about managed to get the CI running before I had to leave :)  I am not too optimistic about this fixing the bug that came up in the ttbar sample though, but let's see

krasznaa:
It was absolutely worth a shot. But unfortunately the crash, while it changed a little, is still there. :frowning:

niermann999:
> It was absolutely worth a shot. But unfortunately the crash, while it changed a little, is still there. 😦

OK, I discovered a problem in a detray sample as well and I am trying to recreate that atm. Otherwise, can you give me a recipe on how to run the problematic event?

krasznaa:
> > It was absolutely worth a shot. But unfortunately the crash, while it changed a little, is still there. 😦
> 
> OK, I discovered a problem in a detray sample as well and I am trying to recreate that atm. Otherwise, can you give me a recipe on how to run the problematic event?

If you build the current `main` branch of [traccc](https://github.com/acts-project/traccc), you can reproduce the crash with:

```
./bin/traccc_seq_example --detector-file=geometries/odd/odd-detray_geometry_detray.json --grid-file=geometries/odd/odd-detray_surface_grids_detray.json --use-detray-detector --digitization-file=geometries/odd/odd-digi-geometric-config.json --input-directory=/afs/cern.ch/user/k/krasznaa/work/public/traccc-odd-crash --input-events=500 --input-skip 295
```

As long as you have access to AFS that is. (On `pcadp04` you do.)

PR #587, opened 2024-05-13
ODD Throughput Profiling, main branch (2024.05.13.)

krasznaa:
Made the (original, TML) script able to use the latest version of the throughput measurement executables, and introduced a script for running a set of ODD throughput measurements.

It is kept as a draft for now, since some code bugs still prevent me from doing larger scale testing...


krasznaa:
Let's revive this one. Since I was using this script during the last week to perform throughput measurements for CHEP 2024.

stephenswat (approved):
:+1:

PR #588, opened 2024-05-13
More (Wavefront) Obj Writing, main branch (2024.05.13.)

krasznaa:
Taught `traccc_seq_example` how to write [Wavefront Obj](https://en.wikipedia.org/wiki/Wavefront_.obj_file) files from seeds and track candidates as well.

Leading to results like:

image
image
image

The nice renders will come later (for my presentation on Thursday if things go well), for now these are just some screen grabs. :wink:

I'm very surprised of why the seeds look as they do. :thinking: Since we absolutely build spacepoints out of every measurement. So I'm not sure yet why seeds don't seem to be formed from the strip measurements. :confused:

But that just underlines that this PRs code will be useful not just for making pretty pictures, but for some debugging as well. :wink:

### DIFF
--- a/examples/run/cpu/seq_example.cpp
+++ b/examples/run/cpu/seq_example.cpp
@@ -245,6 +245,11 @@ int seq_run(const traccc::opts::input_data& input_opts,
                 traccc::performance::timer timer{"Seeding", elapsedTimes};
                 seeds = sa(spacepoints_per_event);
             }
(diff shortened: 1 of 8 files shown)


krasznaa:
Any comments? Eventually I'll want to add an output for the fitted tracks as well. :wink: (Not in this PR...)

stephenswat (approved):
Looks fine to me :)

Issue #564, opened 2024-05-04
Ambiguity Resolution Errors on ODD ttbar Simulation (2024.05.04.)


krasznaa:
Now that I finally started using the ttbar samples that I produced as described in #561, I came across an interesting type of error message from the track ambiguity resolution code. :thinking:

```
[bash][Legolas]:traccc > ./out/build/sycl/bin/traccc_seq_example --detector-file=geometries/odd/odd-detray_geometry_detray.json --grid-file=geometries/odd/odd-detray_surface_grids_detray.json --use-detray-detector --digitization-file=geometries/odd/odd-digi-geometric-config.json --input-directory=odd/geant4_ttbar_mu20/ --input-events=10

Running Full Tracking Chain on the Host

>>> Detector Options <<<
  Detector file       : geometries/odd/odd-detray_geometry_detray.json
  Material file       : 
  Surface rid file    : geometries/odd/odd-detray_surface_grids_detray.json
  Use detray::detector: yes
  Digitization file   : geometries/odd/odd-digi-geometric-config.json
>>> Input Data Options <<<
  Input data format             : csv
  Input directory               : odd/geant4_ttbar_mu20/
  Number of input events        : 10
  Number of input events to skip: 0
>>> Clusterization Options <<<
  Target cells per partition: 1024
>>> Track Seeding Options <<<
  None
>>> Track Finding Options <<<
  Track candidates range   : 3:100
  Minimum step length for the next surface: 0.5 [mm] 
  Maximum step counts for the next surface: 100
  Maximum Chi2             : 30
  Maximum branches per step: 4294967295
  Maximum number of skipped steps per candidates: 3
>>> Track Propagation Options <<<
  Constraint step size  : 3.40282e+38 [mm]
  Overstep tolerance    : -100 [um]
  Minimum mask tolerance: 1e-05 [mm]
  Maximum mask tolerance: 1 [mm]
  Search window         : 0 x 0
  Runge-Kutta tolerance : 0.0001
>>> Track Ambiguity Resolution Options <<<
  Run ambiguity resolution : yes
>>> Performance Measurement Options <<<
  Run performance checks: no

WARNING: No material in detector
WARNING: No entries in volume finder
Detector check: OK
WARNING: No material in detector
WARNING: No entries in volume finder
Detector check: OK
[...]

PR #590, opened 2024-05-14
Ambiguity Resolution: remove duplicate measurements for each track

SylvainJoube:
This PR ensures that the greedy solver deals with tracks that have unique measurement IDs, removing duplicates, if any.

I was able to replicate the issue with the `geant4_ttbar_mu20` data provided by Attila on PR [Data V7, main branch (2024.05.03.) #561](https://github.com/acts-project/traccc/pull/561) and the command line written at the beginning of issue [Ambiguity Resolution Errors on ODD ttbar Simulation (2024.05.04.) #564](https://github.com/acts-project/traccc/issues/564).

It turns out that I did not anticipate that a track could be made of the same `measurement_id` several times. This happened twice on the generated data from `geant4_ttbar_mu20`: for tracks 1176 and 720, that’s why the greedy solver was showing an error. The same “error” should exist on ACTS too, since the Greedy Solver code remains mostly unchanged. I have just taken a look at file [`Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.ipp` from ACTS](https://github.com/acts-project/acts/blob/67acd8ab6eb3f2b0d422e769211552f9d94bc579/Core/include/Acts/AmbiguityResolution/GreedyAmbiguityResolution.ipp#L44).

For example, in issue [Ambiguity Resolution Errors on ODD ttbar Simulation (2024.05.04.) #564](https://github.com/acts-project/traccc/issues/564), the track of index 720 has measurements: 2878 4167 5155 15837 15837, with 15837 twice.

I could have changed the `std::vector<std::size_t> measurements` into a `std::set<std::size_t>` but I'm not sure of the performance implications of such a change, and I'm a little short on time, so this is a quick fix!

### DIFF
--- a/core/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cpp
+++ b/core/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cpp
@@ -152,7 +152,8 @@ void greedy_ambiguity_resolution_algorithm::compute_initial_state(
 
         // Create the list of measurement_id of the current track
         std::vector<std::size_t> measurements;
-        std::set<std::size_t> already_added_mes;
+        std::unordered_map<std::size_t, std::size_t> already_added_mes;
+        bool duplicated_measurements = false;
 
         for (auto const& st : states) {
             std::size_t mid = st.get_measurement().measurement_id;
@@ -163,14 +164,40 @@ void greedy_ambiguity_resolution_algorithm::compute_initial_state(
 
             // If the same measurement is found multiple times in a single
             // track: remove duplicates.
-            if (already_added_mes.find(mid) != already_added_mes.end()) {
+            auto dm = already_added_mes.insert(
+                std::pair<std::size_t, std::size_t>(mid, 1));
+
+            // If the measurement was already present in already_added_mes
+            if (!dm.second) {
+                // Increment the count for this measurement_id
+                ++(dm.first->second);
+                duplicated_measurements = true;
                 LOG_DEBUG("(1/3) Track " << track_index
                                          << " has duplicated measurement "
                                          << mid << ".");
             } else {
-                already_added_mes.insert(mid);
                 measurements.push_back(mid);
             }
+        }
+
+        // If at least one measurement is found multiple times in this track:
+        // print warning message and display the track's measurement list
+        if (duplicated_measurements) {
+            std::stringstream ss;
+            ss << "Track " << track_index << " has duplicated measurement(s):";
+
+            for (const auto& m : already_added_mes) {
+                if (m.second != 1) {
+                    ss << " " << m.first << " (" << m.second << " times)";
+                }
+            }
+
+            ss << ". Measurement list:";
+            for (auto const& st : states) {
+                ss << " " << st.get_measurement().measurement_id;
+            }
+
+            LOG_WARN(ss.str());
         }
 
         // Add this track chi2 value
--- a/core/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cpp
+++ b/core/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cpp
@@ -152,13 +152,25 @@ void greedy_ambiguity_resolution_algorithm::compute_initial_state(
 
         // Create the list of measurement_id of the current track
         std::vector<std::size_t> measurements;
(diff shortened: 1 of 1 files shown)


krasznaa (changes requested):
It's excellent that you understood where the issues / error messages come from.

But I'm not convinced that this should be "fixed" in the ambiguity resolution algorithm. :thinking: Making the algorithm robust against such input is absolutely good. But the track finding should really not be attaching the same measurement multiple times to the same track.

@beomki-yeo, how would this happen? Looping tracks? :confused:

For now, this sort of a check seems good. (Just making the thing into a warning.) But eventually, once the track finding is not expected to produce such tracks, I would demote these checks into assertions. I.e. something like:

```c++
assert([&]() {
    // code that decides whether there are measurement duplicates on a track
    }() == false);
```

(I've been thinking recently about adding a few more of such assertions into our algorithms...)

SylvainJoube:
Thank you for your feedback Attila!

I’ve modified the code to clearly display a single warning message per track with duplicate measurements. It can handle multiple duplicate measurements without being too verbose. The output now displays:

```
WARNING: @greedy_ambiguity_resolution_algorithm: Track 1171 has duplicated measurement(s): 15195 (2 times). Measurement list: 3291 3475 4518 7396 8617 9632 9625 10484 15195 15195 15269 15268

WARNING: @greedy_ambiguity_resolution_algorithm: Track 720 has duplicated measurement(s): 15837 (2 times). Measurement list: 2878 4167 5155 15837 15837
```

krasznaa (approved):
Okay, let's go for this. :+1:

SylvainJoube:
@krasznaa @beomki-yeo ping 🦊

krasznaa:
Ahh, shoot... I thought with the approval in place you would be able to merge it yourself. Completely forgot about this... :frowning:

SylvainJoube:
Thank you Attila! No worries, I was just afraid that, as time went by, I would have complicated merge conflicts :’)

krasznaa:
Unfortunately we have a lot of traffic on the repository today, so my auto-merge will most likely not work anymore. But let's see... Since the CI takes a long time to run right now, we might as well wait for it to finish, and only rebase the PR if really needed.

SylvainJoube:
Okay, we'll wait then! The `CI Bridge` says `Pipeline refused` though, is it an issue or should we wait?

beomki-yeo:
I dont know how it would happen. But as @krasznaa mentioned, this is not something to be fixed from ambiguity solver. At least, this kind of things should be documented somewhere cleary (e.g. issues)

@SylvainJoube Does this issue cause a serious problem in the final result? Otherwise, I would suggest documenting this in the issue category instead of fixing it in the ambiguity solver. Then someone will be able to try to handle the issue by fiximg the CKF or whatever.

beomki-yeo (changes requested):
The multiple module id should be understood better rather than fixing it in the ambiguity solver. For example, I would try to increase the configurarion value of minimal distance to reach the next surface in the finding_config.

beomki-yeo:
I don't get why @krasznaa merged this PR after I requested the change.

krasznaa:
I was curious when the question would come.

This was because of auto-merge. You made your comment Beomki, but didn't de-activate the auto-merge that was active on the PR. So when the CI finally finished, and nobody else merged another PR to prevent this from getting merged (as happened once earlier that day), it finally went in.

So now we know that when there is one approval and one change request on a PR, GitHub considers the PR approved. 🤔

It's a bit unfortunate indeed, but I personally didn't consider this PR too hurtful. If you check the discussion on it, I agree with you that the ambiguity resolution is not the place to sort this issue out. At the same time, this looked like an okay intermediate relief for #564.

So now, it's time to start looking for a real fix. 😉

beomki-yeo:
I didn't know that the auto merge was activated. The PR itself is not very hurtful but I think we don't have to rush in merging PRs.

krasznaa:
This was an unfortunate series of events. But "rushing" is a strange way to describe it. I approved this PR weeks ago. It was not something brand new.

beomki-yeo:
Fine. Didn't even realize that i was pinged that long time ago - I was too busy with other affairs before the vacation.

Still, this duplicate ID problem needs to posted in the issue rather being left behind and forgotten eventually.

PR #591, opened 2024-05-15
oneAPI 2024.1.0 Support, main branch (2024.05.15.)

krasznaa:
This is a sibling of https://github.com/acts-project/detray/pull/734. The reason for making these changes is explained in that PR. 😉

Long story short, we need to handle floating point numbers with `fabs` in device code. 😉


beomki-yeo (approved):
Looks good to me

PR #592, opened 2024-05-23
Truth Particle Update, main branch (2024.05.23.)

krasznaa:
This is the first piece of #582.

It introduces `traccc::particle_container_types`, which I in turn want to use for efficiency measurements. The rename on the members of `traccc::particle` is not mandatory. I just didn't like the `pos` name at all, and since most other variables have "full length" names, I thought that `mom` would be better as `momentum`. But if you have a strong different opinion, I'm open for a discussion. :thinking:

Since I tested it with #582 already, I think the I/O code is okay. Though the unit test added in this PR is indeed pretty weak. :frowning:


krasznaa:
Anybody willing to give this one a look? I'd really like to progress on this front... :thinking:

@stephenswat, @beomki-yeo?

io/include/traccc/io/read_measurements.hpp
```diff
@@ -29,10 +31,14 @@ namespace traccc::io {
...
 /// @param format The format of the cell data files (to read)
+/// @param bardoce_map An object to perform barcode re-mapping with
+///                    (For Acts->Detray identifier re-mapping, if necessary)
 ///
 void read_measurements(measurement_reader_output& out, std::size_t event,
                        std::string_view directory,
-                       data_format format = data_format::csv);
+                       data_format format = data_format::csv,
+                       const std::map<std::uint64_t, detray::geometry::barcode>*
+                           barcode_map = nullptr);
```
stephenswat, line 41: Why not just `std::optional<std::map<...>>`?

io/include/traccc/io/read_measurements.hpp
```diff
@@ -41,9 +47,13 @@ void read_measurements(measurement_reader_output& out, std::size_t event,
...
 /// @param format The format of the cell data files (to read)
+/// @param bardoce_map An object to perform barcode re-mapping with
+///                    (For Acts->Detray identifier re-mapping, if necessary)
 ///
 void read_measurements(measurement_reader_output& out,
                        std::string_view filename,
-                       data_format format = data_format::csv);
+                       data_format format = data_format::csv,
+                       const std::map<std::uint64_t, detray::geometry::barcode>*
+                           barcode_map = nullptr);
```
stephenswat, line 57: Same as above.

stephenswat (changes requested):

io/include/traccc/io/read_measurements.hpp
stephenswat: Alright!

stephenswat (approved):

PR #594, opened 2024-05-27
ref: Update to detray version 66.1

niermann999:
feat: cuda navigation validation by @niermann999 in #732
oneAPI 2024.1.0 Support, main branch (2024.05.15.) by @krasznaa in #734
feat: Add simd intersectors and benchmark them by @niermann999 in #712
feat: Move material maps generation to factory, add material maps to new toy detector and remove old toy detector by @niermann999 in #709
ref: Allow printing of propagation config and remove templates by @niermann999 in #718
Update README.md by @beomki-yeo in #737
fix: navigation and make mask tolerance scaling configurable by @niermann999 in #738

### DIFF
--- a/examples/options/include/traccc/options/track_propagation.hpp
+++ b/examples/options/include/traccc/options/track_propagation.hpp
@@ -24,7 +24,7 @@ class track_propagation : public interface {
     /// @{
 
     /// Propagation configuration object
-    detray::propagation::config<float> config;
+    detray::propagation::config config;
 
     /// @}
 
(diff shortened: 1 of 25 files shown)


examples/options/include/traccc/options/track_propagation.hpp
```diff
@@ -41,12 +41,7 @@ class track_propagation : public interface {
     ///
     /// @param cfg The configuration object to fill
     ///
-    void setup(detray::propagation::config<float>& cfg) const;
-    /// Set up a configuration object based on the command line options
-    ///
-    /// @param cfg The configuration object to fill
-    ///
-    void setup(detray::propagation::config<double>& cfg) const;
+    void setup(detray::propagation::config& cfg) const;
```
krasznaa, line 44: You could actually simplify this code even a bit further. :thinking:

I introduced this `setup(...)` function in #542, to hack around the template parameter of `detray::propagation::config`. Now that the type is not templated, we could get rid of this function completely, undoing 752ae646538cbf48ca09bb5f3aafffccd642c24c. Would you be up for that? :smile:

krasznaa (approved):
Apart from the one comment, as long as the CI is happy, I'm happy with this update. Still, it could be nice to roll back #542 with this PR.

examples/options/include/traccc/options/track_propagation.hpp
niermann999: Done. I also let detray print the propagation options now. 

This PR needs an additional fix from detray to go in first:
https://github.com/acts-project/detray/pull/739

Issue #567, opened 2024-05-04
Increase robustness of clustering partitioning


stephenswat:
Writing this down here as a suggestion for any students or anyone else wanting to get started on traccc.

The clustering algorithm relies on being able to partition the hits into segments which are separated by at least one full row (or column) on a 2D pixel-like detector of zero-activation cells. This guarantees that there are no cross-partition clusters. The algorithm uses shared memory which is of limited side; the maximum partition size $n_\text{max}$ determines the amount of shared memory used and, as a result, the performance of the algorithm: as $n_\text{max}$ increases, performance decreases. However, this gives us an algorithm with a probabilistic success rate. For a hit density $d$, a module width or height $n$, the success probability for a given partition is approximated by $p = 1 - (1 - (1 - d)^n)^{\lfloor\frac{n_\text{max}}{dn}\rfloor+1}$. Although this chance is _tiny_, it still exists.

There are to projects here. First, the success probability can be increased by making the partition algorithm smarter. Second, there needs to be some mechanism to rescue the clustering in the unlikely event that a partition fails to be created.

Increasing the success probability can be done using the knowledge that a full empty row is actually a bit excessive; in reality, we only need to ensure that there is no cluster sharing between two adjacent rows. We can verify this by reifying adjacent rows and checking if they overlap under an 8-adjacency rule. This will lower performance, but probably not by much. Additional kudos if you can come up with a robust estimate of the success probability under this new rule.

Secondly, we need some logic to allocate memory in order to finish the clustering if we have an oversized cluster. This can be done fairly easily by allocating some scratch space from the device. You can allocate global memory in kernels using `malloc`; although this is not recommended for performance reasons, the overhead should be acceptable for this extremely rare edge case. The memory should be used to salvage the partitioning and then be deallocated.

stephenswat:
I believe that this issue has been resolved with the scratch space which can be atomically reserved by threads.

PR #595, opened 2024-05-28
Improve robustness and performance of CCL

stephenswat:
This commit partially addresses #567. In the past, the CCL kernel was unable to deal with extremely large partitions. Although this is very unlikely to happen, our ODD samples contain a few cases of partitions so large it crashes the code. This commit equips the CCL code with some scratch memory which it can reserve using a mutex. This allows it enough space to do its work in global memory. Although this is, of course, slower, it should happen very infrequently. Parameters can be tuned to determine that frequency. This commit also contains a few optimizations to the code which reduce the running time on a μ = 200 ODD ttbar event from about 1100 microseconds to 700 microseconds on an RTX A5000.


stephenswat:
This will need some updates to work with SYCL. Also, using a mutex is not the most efficient approach but it is good enough. I would have preferred to simple use `malloc` on the device, but I don't know if this works outside of CUDA.

stephenswat:
Note that this also makes the code more run-time configurable my removing some of the compile-time constexpr parameters.

krasznaa (changes requested):
I'm fully on board with the proposed logic. Just have a gazillion of technical comments... :stuck_out_tongue:

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
```diff
@@ -199,41 +205,73 @@ TRACCC_DEVICE inline void ccl_kernel(
...
+    // case. The first thread of the block will attempt to enter a critical
+    // section by obtaining a lock on a mutex in global memory. When this is
+    // obtained, we can use some memory in global memory instead of the shared
+    // memory. This can be done more efficiently, but this should be a very
+    // rare edge case.
+    if (size > max_cells_per_partition) {
+        if (threadId == 0) {
+            uint32_t false_int = 0;
+            while (backup_mutex.compare_exchange_strong(false_int, 1u))
+                ;
```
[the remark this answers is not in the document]
stephenswat: Speaking of which, vecmem could really do with a version of `compare_exchange_strong` that doesn't require a reference to its first argument so that we don't need strange hacks like this.

device/common/include/traccc/clusterization/device/ccl_kernel.hpp
```diff
@@ -23,33 +24,30 @@
...
 /// Helper struct for calculating some of the input parameters of @c ccl_kernel
 struct ccl_kernel_helper {
-
     /// Constructor setting the helper parameters
     ///
     /// @param[in] target_cells_per_partition Target average number of cells per
-    ///                                       thread block
+    /// thread block
+    /// @param[in] target_cells_per_thread Target average number of cells per
+    /// thread.
```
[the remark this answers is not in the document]
stephenswat: Sure, added.

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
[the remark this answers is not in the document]
krasznaa: I don't know a whole lot about this topic, why it's desired to have that exact API, but `vecmem::device_atomic_ref` is just following what [std::atomic_ref](https://en.cppreference.com/w/cpp/atomic/atomic_ref/compare_exchange), [cuda::atomic_ref](https://nvidia.github.io/cccl/libcudacxx/extended_api/synchronization_primitives/atomic_ref.html) and [sycl::atomic_ref](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:atomic-references) all have in their APIs. I think it would be bad to provide an API differing from those. :thinking:

device/common/include/traccc/clusterization/device/ccl_kernel.hpp
```diff
@@ -94,7 +92,12 @@ TRACCC_DEVICE inline void ccl_kernel(
     const details::index_t target_cells_per_partition,
     unsigned int& partition_start, unsigned int& partition_end,
     unsigned int& outi, vecmem::data::vector_view<details::index_t> f_view,
-    vecmem::data::vector_view<details::index_t> gf_view, barrier_t& barrier,
+    vecmem::data::vector_view<details::index_t> gf_view,
+    vecmem::data::vector_view<details::index_t> f_backup_view,
+    vecmem::data::vector_view<details::index_t> gf_backup_view,
+    vecmem::data::vector_view<unsigned char> adjc_backup_view,
+    vecmem::data::vector_view<details::index_t> adjv_backup_view,
+    vecmem::device_atomic_ref<uint32_t> backup_mutex, barrier_t& barrier,
```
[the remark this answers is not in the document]
stephenswat: Done.

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
```diff
@@ -199,41 +205,73 @@ TRACCC_DEVICE inline void ccl_kernel(
     barrier.blockBarrier();
 
     // Vector of indices of the adjacent cells
-    details::index_t adjv[details::MAX_CELLS_PER_THREAD][8];
+    details::index_t _adjv[details::MAX_CELLS_PER_THREAD * 8];
+    details::index_t* adjv = &_adjv[0];
```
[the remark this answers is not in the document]
stephenswat: It does, this is just left over from when it was a 2D array. :)

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -54,7 +60,14 @@ clusterization_algorithm::clusterization_algorithm(
...
       m_copy(copy),
       m_stream(str),
-      m_target_cells_per_partition(target_cells_per_partition) {}
+      m_target_cells_per_partition(target_cells_per_partition),
+      m_f_backup(256 * target_cells_per_partition, m_mr.main),
+      m_gf_backup(256 * target_cells_per_partition, m_mr.main),
+      m_adjc_backup(256 * target_cells_per_partition, m_mr.main),
+      m_adjv_backup(256 * 8 * target_cells_per_partition, m_mr.main),
+      m_backup_mutex(vecmem::make_unique_alloc<unsigned int>(m_mr.main)) {
+    cudaMemset(m_backup_mutex.get(), 0, 1);
```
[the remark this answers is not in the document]
krasznaa: :thinking: Not directly.

I'll have to think about this a little. Since unique pointers don't carry their sizes with them, it's not trivial to create a `vecmem::data::vector_view` from them. :frowning: And I really want to keep `vecmem::copy` operating on views. :thinking:

krasznaa:
What's happening with this? :thinking: It would make my life a whole lot easier with running profiling, if this was cleaned up / pushed in. (Right now this prevents me from "quickly" collecting a set of throughput numbers on a Hopper chip. :frowning:)

stephenswat:
This should work now.

core/include/traccc/clusterization/clustering_config.hpp
```diff
@@ -0,0 +1,24 @@
...
+    clustering_config(unsigned int _target_cells_per_partition = 2048,
+                      unsigned int _target_cells_per_thread = 8,
+                      unsigned int _backup_partition_size = 256 * 4096)
+        : target_cells_per_partition(_target_cells_per_partition),
+          target_cells_per_thread(_target_cells_per_thread),
+          backup_partition_size(_backup_partition_size) {}
+
+    unsigned int target_cells_per_partition;
+    unsigned int target_cells_per_thread;
+    unsigned int backup_partition_size;
```
krasznaa, line 22: Is there a benefit to writing so many lines? Why not simply:

```suggestion
    unsigned int target_cells_per_partition = 2048;
    unsigned int target_cells_per_thread = 8;
    unsigned int backup_partition_size = 256 * 4096;
```

?

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -42,19 +48,37 @@ __global__ void ccl_kernel(
...
+          m_mr.main),
+      m_gf_backup(
+          m_config.backup_partition_size * m_config.target_cells_per_partition,
+          m_mr.main),
+      m_adjc_backup(
+          m_config.backup_partition_size * m_config.target_cells_per_partition,
+          m_mr.main),
+      m_adjv_backup(m_config.backup_partition_size * 8 *
+                        m_config.target_cells_per_partition,
+                    m_mr.main),
```
krasznaa, line 76: Could the config struct provide these values with some helper functions? It's a pity to copy-paste these expressions between the CUDA and SYCL code. :thinking:

krasznaa (changes requested):
I'm very on board with the configuration changes. But as long as we do this, we should also update:
  - https://github.com/acts-project/traccc/blob/main/examples/options/include/traccc/options/clusterization.hpp
  - https://github.com/acts-project/traccc/blob/main/examples/options/src/clusterization.cpp

And of course also how it's used in all of our executables... :thinking:

stephenswat:
This now depends on https://github.com/acts-project/traccc/pull/607.

stephenswat:
This is now ready to review. I decided to move away from the existing "top-down" configuration, where the user specifies how many cells they want per partition to a "bottom-up" config where they specify the number of threads, and the number of cells per thread. Makes life a bit easier.

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
```diff
@@ -16,12 +16,10 @@ TRACCC_HOST_DEVICE
 inline void aggregate_cluster(
     const cell_collection_types::const_device& cells,
     const cell_module_collection_types::const_device& modules,
-    const vecmem::data::vector_view<const unsigned short>& f_view,
-    const unsigned int start, const unsigned int end, const unsigned short cid,
-    measurement& out, vecmem::data::vector_view<unsigned int> cell_links,
+    const vecmem::device_vector<details::index_t>& f, const unsigned int start,
```
krasznaa, line 19: Same comment here, about the const-ness.

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
```diff
@@ -33,13 +34,13 @@ namespace traccc::device {
...
-TRACCC_DEVICE void fast_sv_1(
-    vecmem::device_vector<details::index_t>& f,
-    vecmem::device_vector<details::index_t>& gf,
-    unsigned char adjc[details::MAX_CELLS_PER_THREAD],
-    details::index_t adjv[details::MAX_CELLS_PER_THREAD][8],
-    const details::index_t tid, const details::index_t blckDim,
-    barrier_t& barrier) {
+TRACCC_DEVICE void fast_sv_1(vecmem::device_vector<details::index_t>& f,
+                             vecmem::device_vector<details::index_t>& gf,
+                             unsigned char* adjc, details::index_t* adjv,
```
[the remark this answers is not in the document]
krasznaa: No comment on this? :confused:

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -23,38 +26,60 @@ namespace kernels {
...
+      m_f_backup(m_config.backup_size(), m_mr.main),
+      m_gf_backup(m_config.backup_size(), m_mr.main),
+      m_adjc_backup(m_config.backup_size(), m_mr.main),
+      m_adjv_backup(m_config.backup_size() * 8, m_mr.main),
+      m_backup_mutex(vecmem::make_unique_alloc<unsigned int>(m_mr.main)) {
+    m_copy.get().setup(m_f_backup)->ignore();
+    m_copy.get().setup(m_gf_backup)->ignore();
+    m_copy.get().setup(m_adjc_backup)->ignore();
+    m_copy.get().setup(m_adjv_backup)->ignore();
+    TRACCC_CUDA_ERROR_CHECK(cudaMemset(
```
krasznaa, line 79: This could be made into an asynchronous call, right? :thinking:

device/sycl/src/clusterization/clusterization_algorithm.sycl
```diff
@@ -27,11 +28,24 @@ class ccl_kernel;
...
+      m_adjc_backup(m_config.backup_size(), m_mr.main),
+      m_adjv_backup(m_config.backup_size() * 8, m_mr.main),
+      m_backup_mutex(vecmem::make_unique_alloc<unsigned int>(m_mr.main)) {
+    unsigned int zero = 0;
+    details::get_queue(m_queue)
+        .memcpy(
+            m_backup_mutex.get(), &zero,
+            sizeof(
+                std::remove_extent_t<decltype(m_backup_mutex)::element_type>))
+        .wait_and_throw();
```
krasznaa, line 47: You could use `cl::sycl::queue::memset`. It's only in very old SYCL compilers that it's not available. Which we shouldn't worry about with this project. (In VecMem we have an explicit test for whether that call is available or not.)

And unfortunately you are correct to use `wait_and_throw()` here. :frowning: Unlike CUDA, here we **do** need to wait for this operation to finish.

Finally: You forgot about the `vecmem::copy::setup` calls for all of the buffers. (Yes, wait for those as well...)

examples/options/include/traccc/options/clusterization.hpp
```diff
@@ -8,27 +8,31 @@
...
 
     private:
     /// @name Options
     /// @{
     /// The number of cells to merge in a partition
-    unsigned short target_cells_per_partition = 1024;
+    unsigned int threads_per_partition;
+    unsigned int max_cells_per_thread;
+    unsigned int target_cells_per_thread;
+    unsigned int backup_size_multiplier;
```
krasznaa, line 35: I prefer "my" setup here, with giving default values to the properties in the header. And then using those in the `->default_value(...)` calls in the source file.

examples/options/src/clusterization.cpp
```diff
@@ -15,20 +15,43 @@ namespace traccc::opts {
...
 
 std::ostream& clusterization::print_impl(std::ostream& out) const {
-
-    out << "  Target cells per partition: " << target_cells_per_partition;
+    out << "  Threads per partition:      " << threads_per_partition
+        << std::endl;
+    out << "  Target cells per thread:    " << target_cells_per_thread
+        << std::endl;
+    out << "  Max cells per thread:       " << max_cells_per_thread
+        << std::endl;
```
krasznaa, line 53: I'm a bit surprised that you went with 3 `std::endl` calls here. :confused: I prefer just using "`\n`" in such places. The user can decide when they want to flush the stream, themselves.

device/sycl/src/clusterization/clusterization_algorithm.sycl
```diff
@@ -76,42 +91,65 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
     m_copy.get().setup(cell_links)->wait();
     vecmem::data::vector_view<unsigned int> cell_links_view(cell_links);
 
+    vecmem::data::vector_view<device::details::index_t> f_backup_view(
+        m_f_backup);
+    vecmem::data::vector_view<device::details::index_t> gf_backup_view(
+        m_gf_backup);
+    vecmem::data::vector_view<unsigned char> adjc_backup_view(m_adjc_backup);
+    vecmem::data::vector_view<device::details::index_t> adjv_backup_view(
+        m_adjv_backup);
```
krasznaa, line 100: This code is fine, and you may even argue that this is the more readable way to do this. But a slightly more compact way for passing buffers to SYCL lambdas is like:

```c++
vecmem::data::vector_buffer<...> foo{...};
...
   h.parallel_for<...>(..., [..., foo_view = vecmem::get_data(foo), ...] {...});
```

This is more of an advertisement, since we don't really use that formalism all too much in our code right now. But at one point we could make the code slightly more compact like that. :thinking:

krasznaa (changes requested):
I still like the logic, and the general setup. Just some technical comments, as usual.

core/include/traccc/clusterization/clustering_config.hpp
```diff
@@ -0,0 +1,32 @@
...
+        return threads_per_partition * max_cells_per_thread;
+    }
+
+    constexpr std::size_t target_partition_size() const {
+        return threads_per_partition * target_cells_per_thread;
+    }
+
+    constexpr std::size_t backup_size() const {
+        return max_partition_size() * backup_size_multiplier;
+    }
```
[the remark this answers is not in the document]
krasznaa: And I just now remembered... Since you call some of these functions in device code, it would be nice to put `TRACCC_HOST_AND_DEVICE` on them. Yes, `constexpr` does make them work, but since CUDA's support for `constexpr` is still an "experimental feature", it would not hurt us to mark these functions explicitly as device functions. :thinking:

device/cuda/src/clusterization/clusterization_algorithm.cu
stephenswat: In principle yes, but this is executed one during the entire application during the setup, so I don't think the additional complexity is worth it!

krasznaa: Additional complexity? :confused: You already have the stream right there. All the `vecmem::copy::setup` calls are implicitly asynchronous themselves. Switching to [cudaMemsetAsync](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g7c9761e21d9f0999fd136c51e7b9b2a0) seems like a pretty small change here. :thinking:

device/common/include/traccc/clusterization/device/aggregate_cluster.hpp
```diff
@@ -32,8 +32,8 @@ TRACCC_HOST_DEVICE
 inline void aggregate_cluster(
     const cell_collection_types::const_device& cells,
     const cell_module_collection_types::const_device& modules,
-    const vecmem::data::vector_view<const unsigned short>& f_view,
-    unsigned int start, unsigned int end, unsigned short cid, measurement& out,
+    const vecmem::device_vector<details::index_t>& f_view, unsigned int start,
```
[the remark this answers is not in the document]
stephenswat: The compiler don't like that:

```
note: candidate function not viable: no known conversion from 'device_vector<details::index_t>' to 'const device_vector<const details::index_t>' for 3rd argument
```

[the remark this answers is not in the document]
stephenswat: Okay, managed to make this work with some extra temporary variables.

device/sycl/src/clusterization/clusterization_algorithm.sycl
stephenswat: Done!

core/include/traccc/clusterization/clusterization_algorithm.hpp
```diff
@@ -33,6 +34,8 @@ class clusterization_algorithm
           const cell_module_collection_types::const_view&)> {
 
     public:
+    using config_type = clustering_config;
+
```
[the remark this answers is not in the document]
stephenswat: Okay, replaced with `std::monostate`. :+1:

krasznaa:
Unfortunately after looking at the CCL code once again, I'm still not 100% sure... Shouldn't we explicitly memset the "helper buffers" to zero on every event?

But generally, there's even a deeper multi-threading issue here. :thinking: Right now, for simplicity, our MT tests make N copies of each algorithm. But I'd really like to make all our algorithms reentrant. But for that, the helper buffers would need to be created and re-created inside the execute operator for every event separately. :thinking: Which indeed is a significant overhead over the single allocation. But caching may just erase that overhead...?

Not to mention that all of a sudden I became a bit worried: All these shared memory buffers that we use. Should those not be "zeroed out" by the first threads of each thread block as well? Does CUDA, SYCL, etc. guarantee for us that the blocks would be 0-initialized?

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
```diff
@@ -199,48 +209,90 @@ TRACCC_DEVICE inline void ccl_kernel(
...
+    }
+
+    assert(size <= f->size());
+    assert(size <= gf->size());
+
+    details::index_t thread_cell_count = 0;
+    for (details::index_t _cid;
+         (_cid = thread_cell_count * blckDim + threadId) < size;
+         ++thread_cell_count) {
+    }
```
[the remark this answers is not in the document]
stephenswat: Doing it with floating point operations is certainly not the go, but there is a closed-form integer computation for this, that is true.

stephenswat:
> But generally, there's even a deeper multi-threading issue here. 🤔 Right now, for simplicity, our MT tests make N copies of each algorithm. But I'd really like to make all our algorithms reentrant. But for that, the helper buffers would need to be created and re-created inside the execute operator for every event separately. 🤔 Which indeed is a significant overhead over the single allocation. But caching may just erase that overhead...?

I don't see the problem here, this algorithm is fully reentrant; the helper buffers are all guarded by the mutex, so you could use this algorithm from many streams and it would work fine! Problems only start to pop up if you have multiple devices, but even that can be remedied later.

> Not to mention that all of a sudden I became a bit worried: All these shared memory buffers that we use. Should those not be "zeroed out" by the first threads of each thread block as well? Does CUDA, SYCL, etc. guarantee for us that the blocks would be 0-initialized?

Again, I don't think there is any reason to be worried. The CCA kernel uses exactly five shared variables, namely:

* `partition_start` which is unconditionally set before it is used in the `ccl_kernel` function.
* `partition_end` which is also unconditionally set before it is used in the `ccl_kernel` function.
* `outi` which is unconditionally zeroed before it is used  in the `ccl_kernel` function.
* `f` which is unconditionally set before it is used in the `ccl_kernel` function.
* `f` which is also unconditionally set before it is used in the `ccl_kernel` function.

krasznaa:
Okay, you convinced me about the reentrancy aspect. :thinking:

stephenswat:
Let's put this on hold for a hot second while we finish up #558.

core/include/traccc/clusterization/clusterization_algorithm.hpp
```diff
@@ -8,6 +8,7 @@
 #pragma once
 
 // Library include(s).
+#include "traccc/clusterization/clustering_config.hpp"
```
krasznaa, line 11: Optional: Maybe worth removing this include, as it's not needed in this specific header.

device/alpaka/include/traccc/alpaka/clusterization/clusterization_algorithm.hpp
```diff
@@ -57,12 +64,19 @@ class clusterization_algorithm
...
     /// The copy object to use
     std::reference_wrapper<vecmem::copy> m_copy;
+    /// Memory reserved for edge cases
+    vecmem::data::vector_buffer<device::details::index_t> m_f_backup,
+        m_gf_backup;
+    vecmem::data::vector_buffer<unsigned char> m_adjc_backup;
+    vecmem::data::vector_buffer<device::details::index_t> m_adjv_backup;
+    vecmem::unique_alloc_ptr<unsigned int> m_backup_mutex;
+    mutable std::mutex m_setup_mutex;
+    mutable bool m_mutex_is_setup = false;
```
krasznaa, line 79: The goal here is to do the setup strictly just once, even in a multi-threaded environment, right?

The standard way of doing that is with [std::call_once](https://en.cppreference.com/w/cpp/thread/call_once). I.e. put a mutable [std::once_flag](https://en.cppreference.com/w/cpp/thread/once_flag) member into the class, and then run the necessary kernel inside of an `std::call_once` call. :thinking: (This comes from having run into some issues in Athena when we did this sort of a thing without `std::call_once`...)

device/alpaka/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -74,6 +95,23 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
...
+    while (!m_mutex_is_setup) {
+        if (m_setup_mutex.try_lock()) {
+            if (!m_mutex_is_setup) {
+                m_mutex_is_setup = true;
+
+                auto workDiv = makeWorkDiv<Acc>(1, 1);
+
+                ::alpaka::exec<Acc>(queue, workDiv, ZeroMutexKernel{},
+                                    m_backup_mutex.get());
+                ::alpaka::wait(queue);
```
krasznaa, line 108: Remind me, why is a memset operation not an option here instead of running a kernel with a single thread? :confused:

krasznaa (changes requested):
I think the Alpaka code was just left with some outdated stuff. :thinking: Since the CUDA and SYCL code sets up the "mutex variable" much more simply than the Alpaka code does.

Other than cleaning up the Alpaka code, and resolving the merge conflict, I'm on board with the PR. :wink:

device/alpaka/src/clusterization/clusterization_algorithm.cpp
stephenswat: I think Alpaka's memset functionality only works with its own built-in buffers accessors; vecmem has the same restriction, e.g. it only works with buffers, etc. That left me with the option of either using preprocessor macros to differentiate CPU/CUDA/HIP/etc. compilation, or to launch a kernel to simply set the flag, which turned out to be a lot easier.

device/alpaka/include/traccc/alpaka/clusterization/clusterization_algorithm.hpp
stephenswat: That's a good idea.

core/include/traccc/clusterization/clusterization_algorithm.hpp
stephenswat: :+1:

krasznaa (approved):
Let's do it finally.

PR #596, opened 2024-05-29
Add helper class for thread and block identifiers

stephenswat:
The number of parameters in a lot of our kernels and device functions is becoming absurdly large. Part of this is that we are passing variables like thread identifiers, block identifiers and block sizes for each of these functions. Not only does this make code harder to read, it may also increase register pressure on the device causing potentially slow spilling. This commit introduces a new type that handles thread identifiers more compactly and efficiently.


device/common/include/traccc/device/thread_id.hpp
```diff
@@ -0,0 +1,86 @@
...
+#elif defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
+    TRACCC_DEVICE thread_id1(const ::sycl::nd_item<1>& item) : m_item(item) {}
+#else
+    TRACCC_DEVICE thread_id1(unsigned int threadIdx, unsigned int blockIdx,
+                             unsigned int blockDimx, unsigned int gridDimx)
+        : m_threadIdx(threadIdx),
+          m_blockIdx(blockIdx),
+          m_blockDimx(blockDimx),
+          m_gridDimx(gridDimx) {}
+#endif
```
krasznaa, line 26: I don't like this design. :frowning:

I'm on board with introducing a simple, project-specific struct that we would pass this info along with. But let's just make that a struct with a couple of integer members, and that's it. Let's not make the struct differ with SYCL and with CUDA. :thinking:

For simplicity we can make helper functions in `traccc::cuda`, `traccc:sycl`, etc. for constructing an object of that simple struct, based on the (intrinsic) variables available in CUDA, SYCL, etc. But the struct itself should just be a trivial one.

device/sycl/src/clusterization/clusterization_algorithm.sycl
```diff
@@ -103,15 +104,15 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
 
                     // Barrier used in the algorithm.
                     traccc::sycl::barrier barry_r(item);
+                    device::thread_id1 thread_id(item);
```
krasznaa, line 107: Eventually this should be declared `const` as well.

device/common/include/traccc/clusterization/device/ccl_kernel.hpp
```diff
@@ -87,7 +88,7 @@ struct ccl_kernel_helper {
 /// put into
 template <typename barrier_t>
 TRACCC_DEVICE inline void ccl_kernel(
-    details::index_t threadId, details::index_t blckDim, unsigned int blockId,
+    const thread_id1 thread_id,
```
krasznaa, line 91: Is it better to pass this object by value, or by reference? I'm not actually sure...

krasznaa (changes requested):
I'm on board with the general idea. In fact, I've wondered about something similar myself as well.

But let's do it in a cleaner (and I believe simpler) way.

device/common/include/traccc/device/thread_id.hpp
stephenswat: I disagree with you on this one. Storing the variables for all platforms requires more registers on CUDA platforms, where we could just use the (much more efficient) special registers. This kind of design decreases register pressure and has the potential to increase performance!

device/sycl/src/clusterization/clusterization_algorithm.sycl
stephenswat: True, fixed.

device/common/include/traccc/device/thread_id.hpp
krasznaa: You cannot argue that this would be good code here. :thinking: I understand your point, but this cannot be the solution.

In such a case we'd need to do something similar to what we do with barriers. That would at least result in reasonably clean / maintainable code. Though I'm not happy about that design for this use-case either to be honest.

Whatever the case, we are not putting such a macro-hell into `traccc::device_common`.

stephenswat: Yeah trust me I don't like having a pile of unhygienic C++ macros like this, but the alternative is to have more templates like for the barrier...

stephenswat:
I've split this into a CUDA and a SYCL class now, unified by - you guessed it - a concept.

stephenswat:
@CrossR could you check if the Alpaka code in this PR makes sense?

CrossR:
Looks good!

stephenswat:
@paulgessinger if you could take a look next week I'd really appreciate that. :smile:

device/sycl/include/traccc/sycl/utils/thread_id.hpp
```diff
@@ -0,0 +1,48 @@
...
+    auto inline TRACCC_DEVICE getBlockDimX() const {
+        return m_item.get_local_range(0);
+    }
+
+    auto inline TRACCC_DEVICE getGridDimX() const {
+        return m_item.get_global_range(0);
+    }
+
+    private:
+    const ::sycl::nd_item<1> m_item;
```
krasznaa, line 46: I wonder that as long as we "have to" use a reference inside of `traccc::alpaka::thread_id1` anyway, should we not use a const reference here as well? That could reduce the registry pressure a little more. Maybe...

krasznaa (approved):
I'm happy with this code setup. Apart from the one comment, as long as it is brought up to date, I'm happy with this update to go in.

device/sycl/include/traccc/sycl/utils/thread_id.hpp
stephenswat: :+1:

PR #598, opened 2024-05-30
Add missing ACTS JSON dependency

stephenswat:
As it stands, our CMake configuration imports ACTS without the JSON plugin component while our code depends on it. This can lead to some nasty build errors. This commit fixes the issue by correctly adding the JSON plugin dependency, allowing the build to fail early at configuration time.


krasznaa (approved):
Right. I never tried to build traccc against an Acts installation that would miss this plugin. :thinking:

The Acts developers may say otherwise, but I think the ship has still not sailed for renaming `ActsCore` to `Acts::Core`, `ActsPluginJson` to `Acts::PluginJson` and so on. As with that naming it would always result in a CMake configuration error if somebody tried to use a component that wasn't there.

Maybe something to bring up in a wider context @asalzburger?

PR #599, opened 2024-05-31
Add new CI platform for C++20 builds

stephenswat:
In light of some recent issues in Detray as well as some work that I want to put into traccc, this commit adds a new C++20 platform based on Ubuntu 24.04 that will test the compilation of (the host part of) traccc on the newer C++ standard.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -26,24 +26,39 @@ jobs:
         platform:
           - name: CPU
             container: ghcr.io/acts-project/ubuntu2004:v30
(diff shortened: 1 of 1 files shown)


stephenswat:
This will, at the very least, need https://github.com/acts-project/detray/pull/746 to be merged and integrated into traccc before it will pass.

stephenswat:
This is now ready to go.

stephenswat:
The SYCL CI briidge failure is unrelated (see #625)

paulgessinger (approved):
LGTM

PR #600, opened 2024-05-31
Remove naming conflicts with `module` keyword

stephenswat:
In C++20, `module` becomes a keyword. For obvious reasons, we use this as a variable name all over our code, but some compilers are not fans of this naming collision: they will emit warning or even errors because using keywords as variable names is poor practice. This commit replaces occurances of `module` with `mod` to clear up the ambiguity.


krasznaa (approved):
:+1:

PR #601, opened 2024-05-31
Remove use of deprecated `std::is_pod` type trait

stephenswat:
The `std::is_pod` type trait has been deprecated in C++20. This commit replaces that type trait with its recommended equivalent, namely a conjunction of `std::is_trivial` and `std::is_standard_layout`.


krasznaa (approved):
:+1:

PR #602, opened 2024-05-31
Add concepts for barrier-like types

stephenswat:
This commit continues my eternal crusade against unconstrained template types. It lays the groundwork for using concepts in traccc, and it adds a concept for barrier-like objects which are used in the CCL kernels as well as, possibly, other kernels in the future.

### DIFF
--- a/device/common/include/traccc/device/concepts/barrier.hpp
+++ b/device/common/include/traccc/device/concepts/barrier.hpp
@@ -1,0 +1,61 @@
+/**
+ * traccc library, part of the ACTS project (R&D line)
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include "traccc/definitions/concepts.hpp"
+
+#if __cpp_concepts >= 201907L
+#include <concepts>
+#endif
+
+namespace traccc::device::concepts {
+#if __cpp_concepts >= 201907L
+/**
+ * @brief Concept to ensure that a type is barrier-like, i.e. it can
+ * synchronize threads in a compute block.
+ *
+ * @tparam T The barrier-like type.
+ */
+template <typename T>
+concept barrier = requires(T& b) {
+    /*
+     * Check for the general, nulary barrier function which simply synchronizes
+     * threads without return value.
+     */
+    { b.blockBarrier() }
+    ->std::same_as<void>;
+
+    /*
+     * Check for the unary boolean-argument synchronization functions.
+     */
+    requires requires(bool p) {
+        /*
+         * `blockOr` should return true iff one or more of the threads in the
+         * block issues the call with a truthful argument.
+         */
+        { b.blockOr(p) }
+        ->std::same_as<bool>;
+
+        /*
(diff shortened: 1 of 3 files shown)


stephenswat:
This depends on #599 so it can be tested in the CI.

core/include/traccc/definitions/concepts.hpp
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#if __cpp_concepts >= 201907L
+#define CONSTRAINT(x) x
+#else
+#define CONSTRAINT(x) typename
```
krasznaa, line 14: Since we will likely have similar macros in the other projects as well, let's call this one `TRACCC_CONSTRAINT` instead. :thinking:

```suggestion
#define TRACCC_CONSTRAINT(x) x
#else
#define TRACCC_CONSTRAINT(x) typename
```

device/common/include/traccc/device/concepts/barrier.hpp
```diff
@@ -0,0 +1,27 @@
...
+namespace traccc::device::concepts {
+#if __cpp_concepts >= 201907L
+template <typename T>
+concept barrier = requires(T& b) {
+    { b.blockBarrier() }
+    ->std::same_as<void>;
+    { b.blockOr(std::declval<bool>()) }
+    ->std::same_as<bool>;
+};
+#endif
```
krasznaa, line 26: This is a bit subjective, but I'd rather have just a single check on the availability of concepts. I think the following should be enough:

```c++
/**
 * traccc library, part of the ACTS project (R&D line)
 *
 * (c) 2024 CERN for the benefit of the ACTS project
 *
 */

#pragma once

#if __cpp_concepts >= 201907L

// System include(s).
#include <concepts>

/// Namespace holding all C++20 concepts used in device code
namespace traccc::device::concepts {

/// Concept that language-specific "barrier types" must fullfil
template <typename T>
concept barrier = requires(T& b) {
    { b.blockBarrier() }
    ->std::same_as<void>;
    { b.blockOr(std::declval<bool>()) }
    ->std::same_as<bool>;
};

}  // namespace traccc::device::concepts

#endif 
```

I.e. please add some documentation to the code as well. :wink:

krasznaa (changes requested):
I am very on board with pushing forward with C\+\+20. Just minor technical comments on my side.

core/include/traccc/definitions/concepts.hpp
stephenswat: Good point!

device/common/include/traccc/device/concepts/barrier.hpp
stephenswat: Yeah this has ups and downs that I was on the fence about, too. On the one hand, your solution is cleaner. On the other hand, this way it depends on the preprocessor whether the `traccc::device::concepts` namespace exists. In my experience errors about non-existent namespaces are super confusing, so I preferred to have the namespace exist regardless and then to fill it with concepts only if the feature is enabled... Can go either way though.

stephenswat:
Let's see what the new C++20 CI says. :smile:

PR #603, opened 2024-05-31
GPU CI Updates, main branch (2024.05.31.)

krasznaa:
Switched the "GPU builds" to the latest Acts Docker images. At the same time I removed a CUDA build that I just didn't understand why we had. :confused: I just enabled `-DTRACCC_ENABLE_NVTX_PROFILING=TRUE` for our "standard" Release and Debug CUDA build instead.

I instead added a oneAPI based build with an Intel backend. Since until now we've only been testing the build for NVIDIA and AMD backends in the CI for some reason. :confused: While the Intel backend allows us to do both Release and Debug builds, which should be good for the CI's test coverage. :thinking: (The NVIDIA and AMD backends still have issues with `assert(...)` calls unfortunately. :frowning: Making Debug builds impossible with them.)

The big goal here is to finally revive #442, which should (hopefully...) start working with this new setup. :thinking:

### DIFF
--- a/.github/ci_setup.sh
+++ b/.github/ci_setup.sh
@@ -1,6 +1,6 @@
 # TRACCC library, part of the ACTS project (R&D line)
 #
-# (c) 2022-2023 CERN for the benefit of the ACTS project
+# (c) 2022-2024 CERN for the benefit of the ACTS project
 #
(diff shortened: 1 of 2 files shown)


stephenswat (approved):
Looks solid! :+1:

PR #607, opened 2024-06-07
Add a simple spinlock mutex type

stephenswat:
In order to facilitate some edge case handling in device code, this commit adds a very simple spinlock-type mutex using vecmem atomic references.

### DIFF
--- a/device/common/include/traccc/device/impl/mutex.ipp
+++ b/device/common/include/traccc/device/impl/mutex.ipp
@@ -1,0 +1,48 @@
+/**
+ * traccc library, part of the ACTS project (R&D line)
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <cassert>
+#include <vecmem/memory/device_atomic_ref.hpp>
+
+namespace traccc::device {
+template <typename T>
+mutex<T>::mutex(T& p) : m_atomic(p) {}
+
+template <typename T>
+mutex<T>::mutex(const vecmem::device_atomic_ref<T>& r) : m_atomic(r) {}
+
+template <typename T>
+void mutex<T>::lock() {
+    while (!try_lock())
+        ;
+}
+
+template <typename T>
+bool mutex<T>::try_lock() {
+    assert(!m_is_locked);
+
+    T false_v = static_cast<T>(false);
+    bool s = m_atomic.compare_exchange_strong(false_v, static_cast<T>(true),
+                                              vecmem::memory_order::acquire);
+
+#ifndef NDEBUG
+    m_is_locked |= s;
+#endif
+
+    return s;
+}
+
+template <typename T>
+void mutex<T>::unlock() {
+    assert(m_is_locked);
+
+    m_atomic.store(static_cast<T>(false), vecmem::memory_order::release);
+}
+}  // namespace traccc::device
--- a/device/common/include/traccc/device/mutex.hpp
+++ b/device/common/include/traccc/device/mutex.hpp
@@ -1,0 +1,77 @@
+/**
+ * traccc library, part of the ACTS project (R&D line)
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <vecmem/memory/device_atomic_ref.hpp>
+
+#include "traccc/definitions/qualifiers.hpp"
+
+namespace traccc::device {
+/*
+ * A mutex object over type T.
+ *
+ * @warning This class assumes that the value is written to _only_ by mutex
+ * class objects. Writing to the given pointers or atomic references in any
+ * other way is undefined behaviour. Furthermore, it is assumed that the
+ * initial value of the underlying pointer is false.
+ *
+ * @warning This is a spinlock. Do not use when more efficient implementations
+ * are available.
+ */
+template <typename T = uint32_t>
+class mutex {
+    public:
+    /*
+     * Construct a mutex from a pointer.
+     */
+    TRACCC_HOST_DEVICE
+    mutex(T &);
+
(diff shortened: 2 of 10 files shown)


stephenswat:
This will fail until https://github.com/acts-project/vecmem/pull/275 is merged.

stephenswat:
This is now ready for review.

device/common/include/traccc/device/mutex.hpp
```diff
@@ -0,0 +1,74 @@
...
+     * @warning Using this method on a mutex that is not locked is undefined
+     * behaviour.
+     */
+    TRACCC_HOST_DEVICE
+    void unlock();
+
+    private:
+    const vecmem::device_atomic_ref<T> m_atomic;
+
+    bool m_is_locked = false;
```
krasznaa, line 70: If we only need it for debugging, it would be good to hide it behind `#ifndef NDEBUG`. :thinking:

krasznaa (commented):
Is the templating actually needed? Could we not just choose one specific integer type that would need to be used with mutexes? This way we just make it easier for somebody to use an inappropriate type. :thinking: (With a bunch of complicated compiler errors to follow.)

Of course in that case the functions will all need to be explicitly declared `inline`. (They are currently implicitly that because of the templating.)

device/common/include/traccc/device/impl/mutex.ipp
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+#pragma once
+
+#include <cassert>
+#include <vecmem/memory/device_atomic_ref.hpp>
+
+namespace traccc::device {
+template <typename T>
+mutex<T>::mutex(T* p) : m_atomic(*p) {}
```
krasznaa, line 16: Also: Since we're de-referencing it right away (i.e. a `nullptr` is never an option), the interface should require a reference instead of a pointer. Putting the burden on the caller for not providing an invalid pointer.

krasznaa:
Also-also: Would it not be good to also provide a `traccc::device::lock_guard` type? Since it's usually not a good idea to call `lock()` and `unlock()` on a mutex manually. :thinking:

Unfortunately I don't think that we could use `std::lock_guard` with CUDA. (I'm pretty sure that SYCL would accept it though.) Since that would be the nicest, using our custom mutex with the general lock guard...

stephenswat:
Unfortunately, it's not that simple. :worried: Because RAII is implicitly executed by every thread in a warp, using a naive `lock_guard` structure would guarantee deadlock as 32 threads simultaneously try to lock the same mutex. The way out of this would be to implement `lock_guard`-like types for every platform, but then we start hiding which thread owns the lock from the user, which is in itself back if they want to do sub-group level work division... The only way I can image that would make this work is to incorporate cooperative groups into the mix, but we don't use those anywhere, and while SYCL has an equivalent semantic construct I am not sure if HIP, Alpaka and Kokkos do.

stephenswat:
> Is the templating actually needed? Could we not just choose one specific integer type that would need to be used with mutexes?

The danger here is that different platform might support lock-free operations on different types. And some types may be slower than others. Although I am pretty sure something like `uint32_t` is a sane choice on most platforms, I think it makes sense to leave the choice open to the user. A good default value is probably a nice idea, though.

device/common/include/traccc/device/mutex.hpp
stephenswat: Yeah I generally don't like changing the internal structure of types behind `ifdefs`, but if this is only used as a private header it's probably fine.

device/common/include/traccc/device/impl/mutex.ipp
stephenswat: Sure. :+1:

device/common/include/traccc/device/unique_lock.hpp
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <mutex>
+#include <vecmem/memory/device_atomic_ref.hpp>
```
krasznaa, line 12: This include should not be needed.

```suggestion
```

device/common/include/traccc/device/unique_lock.hpp
```diff
@@ -0,0 +1,76 @@
...
+    void unlock();
+
+    /*
+     * Check if the lock is locked by this object.
+     */
+    TRACCC_HOST_DEVICE
+    bool owns_lock() const;
+
+    private:
+    mutex_type* m_mutex_ptr = nullptr;
```
krasznaa, line 71: Would a reference not be better? These locks are not meant to be copiable/movable anyway, right?

krasznaa (changes requested):
Any chance of adding some SYCL tests as well...? Read: Please add SYCL tests as well.

device/common/include/traccc/device/unique_lock.hpp
stephenswat: This is a subset of `std::unique_lock`'s semantics, which also support the ability for a lock not to be associated with any mutex at all. If we ever want to extend `traccc::device::unique_lock` to support the same semantics, this would be a good start. `std::unique_lock` is movable.

stephenswat:
Sure, I can add some SYCL tests. But this kind of "write one piece of common code, have to write tests for all platforms" is going to quickly become unmanageable if we start doing the same for Kokkos, Alpaka, etc. :worried:

krasznaa:
I'm very open to suggestions. But we must test it somehow that the "common code" indeed works correctly on all backends that we want to use. If you prefer to set up just one test with Alpaka or Kokkos for all available backends, I'm fine with that.

stephenswat:
This is now ready, with SYCL tests.

krasznaa (approved):
The `lock` <-> `_lock` names are used in a bit confusing way in the SYCL tests, but since those are just tests, this is more of a nitpick. Other than that, this seems fine.

PR #609, opened 2024-06-11
Add sanity checks for contiguity and orderedness

stephenswat:
In some places, we rely on certain properties of our inputs. For one, we rely on our measurements being contiguous (with respect to the module identifiers) in the fitting code, and we rely on our cells being ordered in some way in the clustering code. I aim to emit properly contiguous elements from the CCL code in the future, but I want to make this process foolproof. To that end, I have added two new sanity checks which aim to find problems with input. One of them checks whether elements are contiguous (according to some projection function) and the other checks whether the elements are in some order. This commit also employs these checks for debug builds in some places.

### DIFF
--- a/core/include/traccc/finding/finding_algorithm.ipp
+++ b/core/include/traccc/finding/finding_algorithm.ipp
@@ -7,6 +7,8 @@
 
 // Project include(s).
 #include "traccc/finding/candidate_link.hpp"
+#include "traccc/sanity/contiguous_on.hpp"
+#include "traccc/utils/projections.hpp"
 
 // detray include(s).
 #include "detray/geometry/barcode.hpp"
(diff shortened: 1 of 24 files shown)


beomki-yeo:
Just to check: For being contiguous, is it required  that unique ID is ordered? (In the measurement vector for track finding, the order of module ID does not really matter)

stephenswat:
> Just to check: For being contiguous, is it required that unique ID is ordered? (In the measurement vector for track finding, the order of module ID does not really matter)

Nope! It requires the IDs to be contiguous but _not_ ordered. So an array $[5, 5, 5, 1, 1, 9, 9, 9, 9, 9, 4, 4]$ is considered contiguous, too.

beomki-yeo (commented):
Can you add a `is_contigous_on` function for CPU as well? 
CPU track finding also requires measurement vector to be contiguous w.r.t module IDs

stephenswat:
Sure, I'll add that.

stephenswat:
@beomki-yeo I've added the CPU equivalents.

beomki-yeo (approved):
good

PR #610, opened 2024-06-11
Add blockwise add and count in CUDA and SYCL

stephenswat:
This commit extends our barrier types with the `blockAnd` and `blockCount` methods, the latter of which is a ballot vote across a work group. Implemented in both CUDA and SYCL.

### DIFF
--- a/tests/cuda/test_barrier.cu
+++ b/tests/cuda/test_barrier.cu
@@ -1,0 +1,146 @@
+/**
+ * traccc library, part of the ACTS project (R&D line)
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include <gtest/gtest.h>
+
+#include <vecmem/memory/cuda/managed_memory_resource.hpp>
+#include <vecmem/memory/unique_ptr.hpp>
+
+#include "../../device/cuda/src/utils/barrier.hpp"
+
+__global__ void testBarrierAnd(bool* out) {
+    traccc::cuda::barrier bar;
+
+    bool v;
+
+    v = bar.blockAnd(false);
+    if (threadIdx.x == 0) {
+        out[0] = v;
+    }
+
+    v = bar.blockAnd(true);
+    if (threadIdx.x == 0) {
+        out[1] = v;
+    }
+
+    v = bar.blockAnd(threadIdx.x % 2 == 0);
+    if (threadIdx.x == 0) {
+        out[2] = v;
+    }
+
+    v = bar.blockAnd(threadIdx.x < 32);
+    if (threadIdx.x == 0) {
+        out[3] = v;
+    }
+}
+
+TEST(CUDABarrier, BarrierAnd) {
+    vecmem::cuda::managed_memory_resource mr;
+    constexpr std::size_t n_bools = 4;
+
+    vecmem::unique_alloc_ptr<bool[]> out =
+        vecmem::make_unique_alloc<bool[]>(mr, n_bools);
+
+    testBarrierAnd<<<1, 1024>>>(out.get());
+
+    ASSERT_EQ(cudaGetLastError(), cudaSuccess);
+    ASSERT_EQ(cudaDeviceSynchronize(), cudaSuccess);
+
+    EXPECT_FALSE(out.get()[0]);
+    EXPECT_TRUE(out.get()[1]);
+    EXPECT_FALSE(out.get()[2]);
+    EXPECT_FALSE(out.get()[3]);
+}
+
+__global__ void testBarrierOr(bool* out) {
+    traccc::cuda::barrier bar;
+
+    bool v;
+
+    v = bar.blockOr(false);
+    if (threadIdx.x == 0) {
+        out[0] = v;
+    }
+
+    v = bar.blockOr(true);
+    if (threadIdx.x == 0) {
+        out[1] = v;
+    }
+
+    v = bar.blockOr(threadIdx.x % 2 == 0);
+    if (threadIdx.x == 0) {
+        out[2] = v;
+    }
+
+    v = bar.blockOr(threadIdx.x < 32);
+    if (threadIdx.x == 0) {
+        out[3] = v;
+    }
+}
+
+TEST(CUDABarrier, BarrierOr) {
+    vecmem::cuda::managed_memory_resource mr;
+    constexpr std::size_t n_bools = 4;
+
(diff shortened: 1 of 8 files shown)


krasznaa (approved):
Nothing against this, I just fear that without any actual usage for the new functions in this PR, we wouldn't know if they work as they should or not. :thinking: I.e. I would've added these new functions in the PR in which they first get used.

tests/sycl/test_barrier.sycl
```diff
@@ -0,0 +1,53 @@
...
+ * Mozilla Public License Version 2.0
+ */
+
+#include <gtest/gtest.h>
+
+#include <CL/sycl.hpp>
+
+#include "../../device/sycl/src/utils/barrier.hpp"
+
+TEST(CUDABarrier, BarrierAnd) {
```
krasznaa, line 15:
```suggestion
TEST(SYCLBarrier, BarrierAnd) {
```

:stuck_out_tongue:

tests/sycl/test_barrier.sycl
```diff
@@ -0,0 +1,53 @@
...
+    ::sycl::queue queue;
+    constexpr std::size_t n_bools = 4;
+
+    bool *out = malloc_shared<bool>(n_bools, queue);
+
+    queue.submit([&](cl::sycl::handler &h) {
+        h.parallel_for(cl::sycl::range<1>(128), [=](::sycl::nd_item<1> item) {
```
krasznaa, line 22:
```suggestion
        h.parallel_for<class BarrierOrTest>(cl::sycl::range<1>(128), [=](::sycl::nd_item<1> item) {
```

The "class-less" formalism is an Intel extension. It's a little more robust to specify a class on this call that would identify the kernel.

tests/sycl/test_barrier.sycl
```diff
@@ -0,0 +1,53 @@
...
+
+#include <CL/sycl.hpp>
+
+#include "../../device/sycl/src/utils/barrier.hpp"
+
+TEST(CUDABarrier, BarrierAnd) {
+    ::sycl::queue queue;
+    constexpr std::size_t n_bools = 4;
+
+    bool *out = malloc_shared<bool>(n_bools, queue);
```
krasznaa, line 19:
```suggestion
    bool *out = cl::sycl::malloc_shared<bool>(n_bools, queue);
```

Though using a vecmem construct would ensure error checking here as well...

tests/sycl/test_barrier.sycl
```diff
@@ -0,0 +1,53 @@
...
+            if (threadIdx.x == 0) {
+                out[2] = v;
+            }
+
+            v = bar.blockAnd(threadIdx.x < 32);
+            if (threadIdx.x == 0) {
+                out[3] = v;
+            }
+        });
+    });
```
krasznaa, line 47:
```suggestion
    }).wait_and_throw();
```

tests/sycl/test_barrier.sycl
```diff
@@ -0,0 +1,53 @@
...
+
+    ASSERT_FALSE(out[0]);
+    ASSERT_TRUE(out[1]);
+    ASSERT_FALSE(out[2]);
+    ASSERT_FALSE(out[3]);
```
krasznaa, line 52:
```suggestion
    EXPECT_FALSE(out[0]);
    EXPECT_TRUE(out[1]);
    EXPECT_FALSE(out[2]);
    EXPECT_FALSE(out[3]);
```

So that all failures would be reported by the test.

tests/sycl/test_barrier.sycl
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+#include <gtest/gtest.h>
+
+#include <CL/sycl.hpp>
+
+#include "../../device/sycl/src/utils/barrier.hpp"
+
+TEST(CUDABarrier, BarrierAnd) {
+    ::sycl::queue queue;
```
krasznaa, line 16:
```suggestion
    cl::sycl::queue queue;
```

The convention is that if you include `<CL/sycl.hpp>`, you would get the types in the `cl::sycl::` namespace. With `<sycl/sycl.hpp>` you would get them in the `sycl::` one.

Intel allows you to mix these, but for instance [AdaptiveCpp](https://github.com/AdaptiveCpp/AdaptiveCpp) doesn't.

And yes, since `<CL/sycl.hpp>` is an "older standard", I've been using that everywhere so far.

tests/cuda/test_barrier.cu
```diff
@@ -0,0 +1,158 @@
...
+TEST(CUDABarrier, BarrierCount) {
+    constexpr std::size_t n_ints = 4;
+    int* out;
+    int host_out[n_ints];
+
+    ASSERT_EQ(cudaMalloc(&out, n_ints * sizeof(int)), cudaSuccess);
+
+    testBarrierCount<<<1, 1024>>>(out);
+
+    ASSERT_EQ(cudaPeekAtLastError(), cudaSuccess);
```
krasznaa, line 146: On a synchronous kernel execution you might as well do:

```suggestion
    ASSERT_EQ(cudaDeviceSynchronize(), cudaSuccess);
```

Maybe in addition to your call. Though for "synchronous kernels" I've rather been using `cudaGetLastError()` instead. :thinking: For instance:

https://github.com/acts-project/vecmem/blob/main/tests/cuda/test_cuda_containers_kernels.cu#L59-L60

tests/cuda/test_barrier.cu
```diff
@@ -0,0 +1,158 @@
...
+
+    ASSERT_EQ(
+        cudaMemcpy(host_out, out, n_ints * sizeof(int), cudaMemcpyDeviceToHost),
+        cudaSuccess);
+
+    ASSERT_EQ(out[0], 0);
+    ASSERT_EQ(out[1], 1024);
+    ASSERT_EQ(out[2], 512);
+    ASSERT_EQ(out[3], 32);
```
krasznaa, line 155: These should also be made into `EXPECT_` calls instead. :thinking:

krasznaa (changes requested):
I like the tests in general, just have a lot of technical comments about them. :stuck_out_tongue:

stephenswat:
Oh sorry I only pushed these so I could run them through act on pcadp04,
I'll turn this into a draft and finalize

stephenswat:
Okay, this is ready to go now. :+1:

krasznaa (approved):
I'm overall very happy with how this looks.

The one thing that could be good to do for consistency is to move `traccc::cuda::barrier` into a private header as well. Since I very much agree with how you moved the SYCL header.

PR #612, opened 2024-06-13
Bump vecmem version to 1.6.0

stephenswat:
This commit bumps the vecmem version to 1.6.0, which features new bulk insert operations as well as fixes to atomic references.


krasznaa (approved):
:+1: 

Let's hope the CI agrees. :crossed_fingers:

PR #613, opened 2024-06-13
Bump clang-format to version 18

stephenswat:
The version of clang-format that we use in this project is quite old (version 10) and it is no longer able to properly format all of the new stuff that has come into C++. See, for example, the formatting nightmares in #609 and #602. This commit bumps the formatter up to clang-format 18.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   format:
     runs-on: ubuntu-latest
-    container: ghcr.io/acts-project/format10:v11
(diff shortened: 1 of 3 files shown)


stephenswat:
This depends on https://github.com/acts-project/machines/pull/106.

beomki-yeo:
I am fine with v18

AJPfleger:
I think we should also update the following two files:
- CI/check_format_local
- docs/contribution/run_formatting.md

stephenswat:
This is obsolete now that we use pre-commit.

PR #614, opened 2024-06-14
Fix broken CUDA unit test after #605

stephenswat:
The expected values in this unit tests were consistenly a value of 0.5 off since the merging of #605. This commit updates the expected values, and also renames the unit test to be more in line with our other tests.


beomki-yeo (approved):
Looks good to me

PR #616, opened 2024-06-18
GitLab CI Update, main branch (2024.06.18.)

krasznaa:
Updated the GitLab CI builds to the latest Docker images, used by the GitHub CI as well. (`ghcr.io/acts-project/ubuntu2004_cuda:47` and `ghcr.io/acts-project/ubuntu2004_cuda_oneapi:47`)

At the same time introduced SYCL builds/tests with an NVIDIA backend as well.

I'm quite curious how successful this will be... :thinking:

### DIFF
--- a/tests/cuda/test_spacepoint_formation.cpp
+++ b/tests/cuda/test_spacepoint_formation.cpp
@@ -25,7 +25,7 @@
 
 using namespace traccc;
 
-TEST(spacepoint_formation, cuda) {
+TEST(CUDASpacepointFormation, cuda) {
 
     // Memory resource used by the EDM.
     vecmem::cuda::managed_memory_resource mng_mr;
--- a/tests/cuda/test_thrust.cu
+++ b/tests/cuda/test_thrust.cu
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2023 CERN for the benefit of the ACTS project
+ * (c) 2023-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -32,7 +32,7 @@ vecmem::cuda::device_memory_resource device_resource;
 
 }  // namespace
 
-TEST(thrust, sort) {
+TEST(CUDAThrust, sort) {
 
     vecmem::vector<unsigned int> host_vector{{3, 2, 1, 8, 4}, &host_resource};
 
@@ -53,7 +53,7 @@ TEST(thrust, sort) {
     ASSERT_EQ(host_vector[4], 8);
 }
 
-TEST(thrust, scan) {
+TEST(CUDAThrust, scan) {
 
     vecmem::vector<unsigned int> host_vector{{3, 2, 1, 8, 4}, &host_resource};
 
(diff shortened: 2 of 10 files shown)


paulgessinger:
Changes look good to me, let's see what the CI says.

btw it doesn't hurt to throw an `nvidia-smi` in the test jobs, this helps debugging in case something goes wrong with the container runtime

krasznaa:
Indeed. I forgot about `nvidia-smi` by mistake. :thinking:

Plus now I also have some SYCL runtime errors to hunt down. :stuck_out_tongue:

krasznaa:
As it turns out, the SYCL tests were failing because after building binaries **only** for an NVIDIA backend, the test was trying to execute the tests on the Xeon CPU of the test machine. :stuck_out_tongue:

But emboldened by the possibility of testing the "Intel backend" actively as well, I added some corresponding configuration to the GitLab CI configuration. But on this, I'm interested in your opinion. Since the CPU backend test is not really different from what we do on GitHub as well. :thinking: So I can easily be convinced of turning that off. (Once we make some Intel GPUs available for testing, this would change of course.)

The SYCL clusterization failure is "real", but @stephenswat promised to fix that in a separate PR. As it should be the same sort of issue as the one he fixed in #614.

krasznaa:
One could also revive the discussion about strictly only running the CUDA and SYCL tests in the GitLab CI. :thinking: (Currently all CPU tests are also being run.)

Unfortunately that would need some slightly different kind of code in the YAML file. (To ask for different test executables from the CUDA and SYCL tests...)

paulgessinger:
@krasznaa How long do you expect the extra tests to last? If there's no reason to run them on GitLab (i.e. cvmfs), I'd lean towards sticking to GitHub resources for these for now.

krasznaa:
As it turned out, we don't actually run any SYCL tests on CPUs in GitHub at the moment. (Otherwise we would've noticed #618 even sooner.) Also, just because our own machine is a well understood hardware, I think it might be a bit safer to run those CPU tests on our own machine. :thinking:

The PR now tries to make sure that it would only run the actual CUDA and SYCL tests using GitLab. Skipping the (non-SYCL) CPU tests, since those are run on GitHub already.

Finally, I **finally** found a good way for downloading the CodePlay oneAPI plugins in a script. Since for actually running the tests with the CUDA backend, we need the appropriate plugin to be installed. Which I left out of https://github.com/acts-project/machines/pull/102, as I just didn't know how to do this correctly.

Once a new tag of the Docker images include the necessary plugins, the script introduced by this PR can be removed. :thinking:

tests/cuda/test_spacepoint_formation.cpp
```diff
@@ -25,7 +25,7 @@
 
 using namespace traccc;
 
-TEST(spacepoint_formation, cuda) {
+TEST(CUDASpacepointFormation, cuda) {
```
stephenswat, line 28: Nit: This is not a very descriptive test name (the "cuda" part) and does not match camelcase naming.

tests/cuda/test_thrust.cu
```diff
@@ -32,7 +32,7 @@ vecmem::cuda::device_memory_resource device_resource;
 
 }  // namespace
 
-TEST(thrust, sort) {
+TEST(CUDAThrust, sort) {
```
stephenswat, line 35: Nit: Tests in this file do not match camelcase naming.

stephenswat:
Since time is a valuable resource on this machine, let's disable the build of the examples library for these tests. Also, we should consider making a dedicated docker image for this with the dependencies as well as the data directory pre-installed.

stephenswat (approved):

PR #618, opened 2024-06-18
Fix SYCL clustering test after #605

stephenswat:
This is the SYCL equivalent of #614.


stephenswat:
I hadn't picked up on this before since I didn't have a working SYCL build on my laptop, but I do now!

stephenswat:
Rebased, will need re-approval.

PR #620, opened 2024-06-20
Fix search window syntax

AnkithVarri:
Current example fails with the confusing error message due to the failed syntax.


stephenswat (approved):
Makes sense. Welcome to the project! Are you participating in our bi-weekly meetings at all, and are you in our Mattermost channel?

PR #621, opened 2024-06-20
Removed implicit conversion warnings

fredevb:
Fixed some implicit conversion warnings.

### DIFF
--- a/core/include/traccc/seeding/spacepoint_binning_helper.hpp
+++ b/core/include/traccc/seeding/spacepoint_binning_helper.hpp
@@ -116,8 +116,8 @@ inline TRACCC_HOST_DEVICE size_t is_valid_sp(const seedfinder_config& config,
     if (spPhi > config.phiMax || spPhi < config.phiMin) {
         return detray::detail::invalid_value<size_t>();
     }
-    size_t r_index = getter::perp(
-        vector2{sp.x() - config.beamPos[0], sp.y() - config.beamPos[1]});
+    size_t r_index = static_cast<size_t>(getter::perp(
+        vector2{sp.x() - config.beamPos[0], sp.y() - config.beamPos[1]}));
 
     if (r_index < config.get_num_rbins()) {
         return r_index;
--- a/core/include/traccc/seeding/spacepoint_binning_helper.hpp
+++ b/core/include/traccc/seeding/spacepoint_binning_helper.hpp
@@ -82,7 +82,7 @@ inline std::pair<detray::axis2::circular<>, detray::axis2::regular<>> get_axes(
 
         // divide 2pi by angle delta to get number of phi-bins
         // size is always 2pi even for regions of interest
-        phiBins = static_cast<detray::index>(std::ceil(2 * M_PI / deltaPhi));
+        phiBins = std::llround(2 * M_PI / deltaPhi + 0.5);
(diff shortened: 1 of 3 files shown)


core/include/traccc/seeding/spacepoint_binning_helper.hpp
```diff
@@ -82,7 +82,7 @@ inline std::pair<detray::axis2::circular<>, detray::axis2::regular<>> get_axes(
 
         // divide 2pi by angle delta to get number of phi-bins
         // size is always 2pi even for regions of interest
-        phiBins = std::ceil(2 * M_PI / deltaPhi);
+        phiBins = static_cast<detray::index>(std::ceil(2 * M_PI / deltaPhi));
```
stephenswat, line 85: This should be `std::llround(2 * M_PI / deltaPhi + 0.5)`.

core/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cpp
```diff
@@ -245,7 +245,7 @@ void greedy_ambiguity_resolution_algorithm::compute_initial_state(
             double ratio = static_cast<float>(mcount_idzero) /
                            static_cast<float>(mcount_all);
             if (ratio > warning_threshold) {
-                std::size_t percent = ratio * 100;
+                std::size_t percent = static_cast<std::size_t>(ratio * 100);
```
stephenswat, line 248: Let's just render this as a floating point number. E.g.:

```cpp
std::stringstream stream;
stream << std::fixed << std::setprecision(2) << (ratio * 100.);
std::string s = stream.str();
```

core/include/traccc/seeding/spacepoint_binning_helper.hpp
```diff
@@ -116,8 +116,8 @@ inline TRACCC_HOST_DEVICE size_t is_valid_sp(const seedfinder_config& config,
     if (spPhi > config.phiMax || spPhi < config.phiMin) {
         return detray::detail::invalid_value<size_t>();
     }
-    size_t r_index = getter::perp(
-        vector2{sp.x() - config.beamPos[0], sp.y() - config.beamPos[1]});
+    size_t r_index = static_cast<size_t>(getter::perp(
+        vector2{sp.x() - config.beamPos[0], sp.y() - config.beamPos[1]}));
```
stephenswat, line 120: Casting this directly to an integer seems wrong, unless the bin size is exactly a unit of length 1... Are you sure this is correct?

stephenswat:
Also, please work on more descriptive PR titles and descriptions.

core/include/traccc/seeding/spacepoint_binning_helper.hpp
fredevb: I was just fixing the warning:
`implicit conversion turns floating-point number into integer: 'float' to 'size_t' (aka 'unsigned long') [-Werror,-Wfloat-conversion] 
size_t r_index = getter::perp(`

stephenswat: I'm suspicious of this line... I would have expected a division by the size of each r-bin in order to get an actual index. Again, this code only works if the bin size is exactly one. But I can open an issue regarding this.

PR #622, opened 2024-06-20
build: include tbb as a system library and add compile flag `-Wfloat-conversion`

fredevb:
These changes are needed for traccc to be included in ACTS (without the changes the clang-tidy CI fails because of the warnings in tbb). The code added to tbb is based on the code in the eigen cmake. Furthermore, I added the compiler flag "-Wfloat-conversion" to align the compiler flags with ACTS.

### DIFF
--- a/cmake/traccc-compiler-options-cpp.cmake
+++ b/cmake/traccc-compiler-options-cpp.cmake
@@ -23,6 +23,7 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wshadow" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wunused-local-typedefs" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-pedantic" )
+   traccc_add_fkag( CMAKE_CXX_FLAGS "-Wfloat-conversion" )
 
    # Fail on warnings, if asked for that behaviour.
    if( TRACCC_FAIL_ON_WARNINGS )
--- a/cmake/traccc-compiler-options-cpp.cmake
+++ b/cmake/traccc-compiler-options-cpp.cmake
@@ -23,7 +23,7 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wshadow" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wunused-local-typedefs" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-pedantic" )
-   traccc_add_fkag( CMAKE_CXX_FLAGS "-Wfloat-conversion" )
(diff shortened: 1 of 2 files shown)


cmake/traccc-compiler-options-cpp.cmake
```diff
@@ -23,6 +23,7 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wshadow" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wunused-local-typedefs" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-pedantic" )
+   traccc_add_fkag( CMAKE_CXX_FLAGS "-Wfloat-conversion" )
```
krasznaa, line 26: Typo...

extern/tbb/CMakeLists.txt
```diff
@@ -27,5 +27,13 @@ FetchContent_Declare( TBB ${TRACCC_TBB_SOURCE} )
 set( TBB_TEST FALSE CACHE BOOL "Turn off the TBB tests" )
 set( TBB_STRICT FALSE CACHE BOOL "Do not throw errors on compiler warnings" )
 
+# Treat the TBB headers as "system headers", to avoid getting warnings from
+# them.
+get_target_property( _incDirs tbb INTERFACE_INCLUDE_DIRECTORIES )
+target_include_directories( tbb
+   SYSTEM INTERFACE ${_incDirs} )
+unset( _incDirs )
```
krasznaa, line 35: :thinking: TBB is actually a `SHARED` library, not an `INTERFACE` one. (I didn't think about this in the office.) So you'll likely rather need to do:

```suggestion
get_target_property( _incDirs tbb INCLUDE_DIRECTORIES )
target_include_directories( tbb
   SYSTEM PUBLIC ${_incDirs} )
unset( _incDirs )
```

krasznaa (changes requested):
Please check what happens with your build after all of this. For instance, you could do:

```
VERBOSE=1 make traccc_tbb_task_example
```

, and see how the TBB include directory/directories appear(s) in the compilation commands. They should **only** appear with `-isystem` for this PR to be successful.

krasznaa:
And I very much agree with @stephenswat. You should write a little more details into the PRs. For instance in this case you could've documented that you wrote the code based on the setup for Eigen. :thinking:

extern/tbb/CMakeLists.txt
```diff
@@ -27,5 +27,13 @@ FetchContent_Declare( TBB ${TRACCC_TBB_SOURCE} )
 set( TBB_TEST FALSE CACHE BOOL "Turn off the TBB tests" )
 set( TBB_STRICT FALSE CACHE BOOL "Do not throw errors on compiler warnings" )
 
+# Treat the TBB headers as "system headers", to avoid getting warnings from
+# them.
+get_target_property( _incDirs tbb INCLUDE_DIRECTORIES )
+target_include_directories( tbb
+   SYSTEM PUBLIC ${_incDirs} )
+unset( _incDirs )
```
stephenswat, line 35: This needs to go _after_ the call to `FetchContent`.

PR #623, opened 2024-06-20
Fix the commands in README.md

beomki-yeo:
- `--events` is replaced by `--gen-events` or `--input-events`

- Fix the unfixed search window syntaxes


krasznaa (approved):
:+1:

PR #625, opened 2024-06-21
Update OneAPI CUDA and ROCm images

stephenswat:
The CI has started failing on the CI bridge because we are exceeding the maximum number of daily downloads from the CodePlay website. In order to fix this, we can update the images for these builds to version 48 which comes with the plugins installed. In addition, this commit removes the downloading script for those plugins as it is no longer necessary.


krasznaa (approved):
This looks correct. As long as the CI is happy, I'm happy as well.

PR #626, opened 2024-06-21
Add CLI options to configuration type conversion

stephenswat:
As I am finalizing #595, I am noticing that it is really quite a task to update the configuration of our algorithms. Not only do you need to update the configuration type and the ways your algorithm uses it, but you also need to update the way the CLI options are translated into those configuration files across _many_ executables.

In this commit I am trying to make this process a little easier by specifying the `config_provider` mixin for command line options classes. This allows us to specify only once how options should be translated to configuration types, making the process easier and less prone to bugs.

### DIFF
--- a/examples/run/cpu/truth_fitting_example.cpp
+++ b/examples/run/cpu/truth_fitting_example.cpp
@@ -116,7 +116,7 @@ int main(int argc, char* argv[]) {
 
     // Fitting algorithm object
     typename traccc::fitting_algorithm<host_fitter_type>::config_type fit_cfg;
-    fit_cfg.propagation = propagation_opts.config;
+    fit_cfg.propagation = propagation_opts;
 
     traccc::fitting_algorithm<host_fitter_type> host_fitting(fit_cfg);
 
--- a/examples/run/cuda/truth_fitting_example_cuda.cpp
+++ b/examples/run/cuda/truth_fitting_example_cuda.cpp
@@ -158,7 +158,7 @@ int main(int argc, char* argv[]) {
 
     // Fitting algorithm object
     typename traccc::fitting_algorithm<host_fitter_type>::config_type fit_cfg;
-    fit_cfg.propagation = propagation_opts.config;
+    fit_cfg.propagation = propagation_opts;
 
     traccc::fitting_algorithm<host_fitter_type> host_fitting(fit_cfg);
     traccc::cuda::fitting_algorithm<device_fitter_type> device_fitting(
--- a/examples/options/include/traccc/options/clusterization.hpp
+++ b/examples/options/include/traccc/options/clusterization.hpp
@@ -8,26 +8,29 @@
 #pragma once
 
 // Local include(s).
+#include "traccc/options/details/config_provider.hpp"
 #include "traccc/options/details/interface.hpp"
 
 namespace traccc::opts {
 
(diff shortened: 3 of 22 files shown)


krasznaa (changes requested):
:thinking: It's a neat idea...

But I would leave `traccc::opts::clusterization` alone in this PR. :thinking: I think in its current form it just confuses things more than it helps.

We've talked about this with @beomki-yeo in passing in the past. That generally **every** algorithm should come with a configuration struct eventually. Instead of some of them having such structs, and some of them just receiving some primitive variables.

In such a setup, where there would be `traccc::<algorithm>::config` types for everything, these auto-converter operators would result in a very nice code layout indeed. But for just handling a single `unsigned short`, doing this change (in my opinion) just makes the code harder to read.

The changes to `traccc::opts::track_finding` and `traccc::opts::track_propagation` are very interesting on the other hand. :+1:

stephenswat:
I'm glad you like it. I agree that doing this for a single short is silly, but #595 will expand the number of options to the clustering algorithm. Using this configuration provider pattern for the clustering will make it trivially easy to implement the new config in #595, although you are correct that making the change later is not much more difficult.

So if you prefer I can revert the change for clustering in this PR and move that to #595; otherwise we can leave it here. No real different either way.

examples/run/cpu/truth_fitting_example.cpp
```diff
@@ -114,9 +114,12 @@ int main(int argc, char* argv[]) {
         0.01 / detray::unit<scalar>::GeV,
         1 * detray::unit<scalar>::ns};
 
+    // Propagation configuration
+    detray::propagation::config propagation_config(propagation_opts);
+
     // Fitting algorithm object
     typename traccc::fitting_algorithm<host_fitter_type>::config_type fit_cfg;
-    fit_cfg.propagation = propagation_opts.config;
+    fit_cfg.propagation = propagation_config;
```
krasznaa, line 122: Since in this example `detray::propagation::config` is only needed once, creating a separate `propagation_config` object is not really needed here, no? 🤔

examples/run/cuda/truth_fitting_example_cuda.cpp
```diff
@@ -156,9 +156,12 @@ int main(int argc, char* argv[]) {
         0.01 / detray::unit<scalar>::GeV,
         1 * detray::unit<scalar>::ns};
 
+    // Propagation config
+    detray::propagation::config propagation_config(propagation_opts);
+
     // Fitting algorithm object
     typename traccc::fitting_algorithm<host_fitter_type>::config_type fit_cfg;
-    fit_cfg.propagation = propagation_opts.config;
+    fit_cfg.propagation = propagation_config;
```
krasznaa, line 164: Same here. Couldn't we just

```suggestion
    fit_cfg.propagation = propagation_opts;
```

?

krasznaa (changes requested):
Okay, let's update `traccc::ops::clusterization` in this PR already then. I'm now just wondering about possibly shortening the example executable code even further. 🤔

I.e. is it necessary in any of the cases to create a separate `detray::propagation::config` object? I'm not actually sure...

stephenswat:
Updated 6 instances where a config object was created and used only once.

krasznaa (approved):
:+1:

PR #629, opened 2024-06-22
Improve numerical stability of CCA variance

stephenswat:
I noticed that, at some point, a factor of $\frac{1}{12}$ was added to the variance of measurements and this slipped through because the tolerance on the variance test was extremely large, i.e. no smaller than 0.1. This is an unacceptably high tolerance, and so I decided that the variance computation was in need of an update. I decided to adopt two strategies to do this. The first is the implementation of Welford's online algorithm, which relies on the following recurrence relation:

$$\sigma^2_n = \left(1 - \frac{w_n}{W_n}\right) \sigma^2_{n-1} + \frac{w_n}{W_n} (x_n - \mu_n) (x_n - \mu_{n-1})$$

This is significantly less prone to catastrophic cancellation. Second, I shifted the entire computation by the position of the first cell, which brings the computation closer to zero where floating point computation is more accurate. This depends on two equivalences:

$$\mu(x_1, \ldots, x_n) = \mu(x_1 - C, \ldots, x_n - C) + C$$

and

$$\sigma^2(x_1, \ldots, x_n) = \sigma^2(x_1 - C, \ldots, x_n - C)$$

Combined, these factors allow me to drop the tolerance in the tests from a _minimum_ of 0.1 to a fixed value of 0.0001.

### DIFF
--- a/core/include/traccc/clusterization/impl/measurement_creation.ipp
+++ b/core/include/traccc/clusterization/impl/measurement_creation.ipp
@@ -7,6 +7,8 @@
 
 #pragma once
 
+#include <cassert>
+
 namespace traccc::details {
 
 TRACCC_HOST_DEVICE
@@ -28,6 +30,8 @@ inline vector2 position_from_cell(const cell& cell, const cell_module& mod) {
 TRACCC_HOST_DEVICE inline void calc_cluster_properties(
     const cell_collection_types::const_device& cluster, const cell_module& mod,
     point2& mean, point2& var, scalar& totalWeight) {
+    point2 offset{0., 0.};
+    bool first_processed = false;
 
     // Loop over the cells of the cluster.
     for (const cell& cell : cluster) {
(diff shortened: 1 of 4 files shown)


stephenswat:
I also had to update the CPU measurement creation algorithm which, as I discovered in this PR, wasn't actually computing the variance _at all_ and just defaulting to $\frac{1}{12}$. 🤭

core/include/traccc/clusterization/impl/measurement_creation.ipp
```diff
@@ -83,14 +96,12 @@ TRACCC_HOST_DEVICE inline void fill_measurement(
         m.surface_link = mod.surface_link;
         // normalize the cell position
         m.local = mean;
-        // normalize the variance
-        m.variance[0] = var[0] / totalWeight;
-        m.variance[1] = var[1] / totalWeight;
```
krasznaa, line 88: I can't comment on that line unfortunately, but as long as we're removing the division by `totalWeight`, let's remove the whole `if (totalWeight > 0.)` check (from line 77) as well. Since that thing was just a menace so far. :frowning: Basically a cluster with a total weight of `0.` should've been checked for with an assertion, not an `if` statement. Since such clusters would've caused havoc in other ways as well. :thinking: You also removed the equivalent check from the GPU code anyway. :wink:

(We might still want to add `assert(totalWeight > 0.)`, as that should really be an assertion at this point. Since cells below some limit will always be discarded.)

tests/common/tests/cca_test.hpp
```diff
@@ -112,15 +114,16 @@ class ConnectedComponentAnalysisTests
...
         cca_truth_hit io_truth;
         while (truth_reader.read(io_truth)) {
             ASSERT_TRUE(result.find(io_truth.geometry_id) != result.end());
 
             const vecmem::vector<traccc::measurement> &meas =
                 result.at(io_truth.geometry_id);
 
-            traccc::scalar tol = std::max(
-                0.1, 0.0001 * std::max(io_truth.channel0, io_truth.channel1));
+            traccc::scalar tol = 0.0001;
```
krasznaa, line 126: :thinking: This is just a subjective comment, but you could make all the new variables `const`. Plus you should probably use `f` on all the literal constants.

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
```diff
@@ -67,14 +93,26 @@ inline void aggregate_cluster(
...
+                    first_processed = true;
+                }
+
+                cell_position = cell_position - offset;
+
+                const point2 diff_old = cell_position - mean;
+                mean = mean + diff_old * weight_factor;
+                const point2 diff_new = cell_position - mean;
+
+#pragma unroll
```
krasznaa, line 112: You should of course use the macro introduced for Alpaka. But I wonder... is it even worth to do all this @#$^ for this tiny/trivial loop? Why not just write out the two statements with a copy-paste and be done with it? :thinking:

krasznaa (changes requested):
I'm very supportive overall. :+1:

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
krasznaa: What I wonder is: Vectorization is not going to be a thing here, right? (Or could it be?) If not, just writing 2 statements instead of the for-loop would seem to be the way to go.

tests/common/tests/cca_test.hpp
stephenswat: Fair point, that will save poor @fredevb some work later. :slightly_smiling_face:

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
stephenswat: I think the nicest solution would be to use the `point2` semantics, but I don't think it supports element-wise multiplication. :slightly_frowning_face:

stephenswat:
Updated. :+1:

krasznaa (approved):

PR #630, opened 2024-06-22
Add SYCL tests for CCA code

stephenswat:
Currently, we implement CCA tests only for CPU code and for CUDA code, but we have an easily reusable test harness for this, so we might as well test our SYCL code, too. This commit adds a new set of tests to the SYCL suite which tests the functionality and accuracy of our CCL code.


stephenswat:
This may fail until #629 is merged.

krasznaa (approved):
This should receive the higher precision check with #629 as well. So, since the CI does not complain, we might as well merge in this one first. :thinking:

PR #631, opened 2024-06-26
Add infrastructure for working with concepts

stephenswat:
This commit adds the `TRACCC_CONSTRAINT` macro which acts as a concept in C++20 builds, and simply acts as the `typename` keyword in builds with older standards. It also adds the `TRACCC_ENFORCE_CONCEPTS` CMake flag which will throw an error if concepts are not available.

### DIFF
--- a/core/include/traccc/definitions/concepts.hpp
+++ b/core/include/traccc/definitions/concepts.hpp
@@ -1,0 +1,18 @@
+/**
+ * traccc library, part of the ACTS project (R&D line)
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#if __cpp_concepts >= 201907L
+#define TRACCC_CONSTRAINT(...) __VA_ARGS__
+#elif defined(TRACCC_ENFORCE_CONCEPTS)
+#error \
+    "`TRACCC_ENFORCE_CONCEPTS` is set, but concepts are not available. This constitutes a fatal error."
+#else
+#define TRACCC_CONSTRAINT(...) typename
+#endif
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -35,7 +35,7 @@ jobs:
           - name: CPU
             container: ghcr.io/acts-project/ubuntu2404:48
             cxx_standard: "20"
-            options: -DTRACCC_USE_ROOT=FALSE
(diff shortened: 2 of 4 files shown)


stephenswat:
Okay it took some faffing about because of CMake but this is ready to go.

core/include/traccc/definitions/concepts.hpp
```diff
@@ -0,0 +1,18 @@
...
+
+#pragma once
+
+#if __cpp_concepts >= 201907L
+#define TRACCC_CONSTRAINT(...) __VA_ARGS__
+#elif defined(TRACCC_ENFORCE_CONCEPTS)
+#error \
+    "`TRACCC_ENFORCE_CONCEPTS` is set, but concepts are not available. This constitutes a fatal error."
+#else
+#define TRACCC_CONSTRAINT(...) __VA_ARGS__
```
krasznaa, line 17: Shouldn't this be

```suggestion
#define TRACCC_CONSTRAINT(...)
```

? :confused:

krasznaa (changes requested):
I do not like the change to `ci_setup.sh`. It should not be in the business of making decisions about compiler flags like that. :frowning:

I remember that you found that older CMake versions don't respect `CMAKE_CUDA_STANDARD` as they should. But then, let's do something similar to what we settled on in https://github.com/acts-project/acts/pull/3318 with @paulgessinger. That the CI configuration YAML would specify some flags explicitly, using `CMAKE_CUDA_FLAGS`.

Yes, it's a bit fragile, but what's going on in this PR is quite fragile as well.

core/include/traccc/definitions/concepts.hpp
stephenswat: No, not quite, bit it should be `typename`. :laughing:

stephenswat:
Setting `CMAKE_CUDA_FLAGS` is high-key bad because it causes CMake to – silently, of course – ignore whatever is set in the `CUDAFLAGS` environment variable though... It's strictly more fragile than this approach as far as I am concerned but okay, I'll change it.

stephenswat:
Updated!

krasznaa (approved):
Unfortunately you are correct that setting `CMAKE_CUDA_FLAGS` makes the configuration ignore the `$CUDAFLAGS` environment variable. :frowning: Which can indeed cause some problems. :frowning:

I believe that in the single job in which we need to force `-std=c++20` on CUDA, it's fine to have `$CUDAFLAGS` ignored. So, just since this code setup is simpler, let's go with this anyway.

P.S. We will need to test C\+\+20 with SYCL as well later on, but it's fine not to worry about that in this PR.

PR #632, opened 2024-06-26
Implement block-wide odd-even sort

stephenswat:
Sorting small arrays is a relatively common problem in GPGPU programming. Many useful algorithms exist, and some are provided by libraries like CUB. An algorithm close to my heart is odd-even sort because it is exceedingly simply, relatively efficient for small arrays and, importantly, it uses $\mathcal{O}(1)$ space. This commit adds new implementations of this sorting algorithm for block-wide odd-even sort in a portable way.

### DIFF
--- a/device/common/include/traccc/device/sort.hpp
+++ b/device/common/include/traccc/device/sort.hpp
@@ -1,0 +1,86 @@
+/**
+ * traccc library, part of the ACTS project (R&D line)
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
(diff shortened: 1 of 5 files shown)


stephenswat:
Depends on #631, #602, and #596. Replaces #347.

stephenswat:
This is now ready for review.

krasznaa (approved):
I don't claim to understand the algorithm. :frowning:

But "looks-wise" I like it. And the CI is happy with it, so: :+1:

PR #634, opened 2024-06-26
Download data files after build in CI

stephenswat:
As it stands right now, we download all the data files before we ever even build the traccc project. This is a bit wasteful, as a failed build will render all that downloading useless. This commit moves the data downloading step to after the build step, and also makes it conditional on whether the tests will actually be run.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -82,8 +82,6 @@ jobs:
         shell: bash
     steps:
       - uses: actions/checkout@v3
-      - name: Download data files
-        run: data/traccc_data_get_files.sh
       - name: Configure
         run: |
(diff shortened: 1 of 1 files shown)


stephenswat:
Note: this used to work fine when the data files were several megabytes; now they are gigabytes!

krasznaa (approved):
Sure.

Though, as long as we're doing this, does it make sense to have the download as a separate step from the "Test" step? It might be easier to just run the download script right before executing `ctest`. :thinking:

paulgessinger (approved):

stephenswat:
> Sure.
> 
> Though, as long as we're doing this, does it make sense to have the download as a separate step from the "Test" step? It might be easier to just run the download script right before executing `ctest`. 🤔

Both work, I'd struggle to make any reasonable argument one way or the other.

PR #636, opened 2024-07-03
Enable CI tests for CPU-based SYCL builds

stephenswat:
As it stands, we only test our SYCL code on the CI bridge, and that only runs on GPU-like platforms. While it is good that we test there, it is also important for us to test this code on CPU platforms. Indeed, such platforms behave differently in terms of lockstep execution and memory allocation.


krasznaa (approved):
Sure.

PR #637, opened 2024-07-03
Zero output in mutex and unique lock tests

stephenswat:
Turns out that on the platforms we test for, memory allocations are zeroed out implicitly. If you try running the code on architectures that don't do this, though, the results are incorrect. To correct for this, this commit adds explicit zeroing for the output variables of the mutex and unique lock tests.


beomki-yeo (approved):
Looks simple enough

PR #640, opened 2024-07-10
Add debug output to clustering algorithms

stephenswat:
In #595, I equipped the CCA code with some edge case handling which allows it to handle oversized partitions. Although this makes sure the algorithm works, it also risks to slow down execution. In order to better understand how much performance we might be losing, this commit adds the ability for the SYCL and CUDA algorithms to print some warnings if they ever encounter this edge case.

### DIFF
--- a/core/include/traccc/clusterization/clustering_config.hpp
+++ b/core/include/traccc/clusterization/clustering_config.hpp
@@ -55,6 +55,13 @@ struct clustering_config {
     unsigned int backup_size_multiplier;
 
     /**
+     * @brief Flag to enforce debug output.
+     *
+     * @warning This will slown down the clustering algorithm.
+     */
+    bool enable_debug_output;
+
+    /**
      * @brief The maximum number of cells per partition.
      */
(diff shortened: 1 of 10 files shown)


krasznaa:
:thinking: I think this opens a much bigger question that I've been putting off all this time...

We will need some "monitoring functionality" for the reconstruction algorithms. I only realised this a few months ago when an Allen developer talked about how they optimised the monitoring functionality in their algorithms.

You see, for debugging it's fine to transfer intermediate data products back to the host, and debug the data there. But in production running, even though we're not actively debugging the code at that point, we'll still need to be able to produce "monitoring histograms" about what the reconstruction is doing. (Number of spacepoints, their positions, this sort of stuff.) We need this to be able to detect errors in our data quickly/reliably.

This sort of information, about how often the reconstruction needed to take a slow route, falls right into this type of monitoring information in my mind. So rather than adding something specifically just for this, I think we should rather start designing a general monitoring functionality for our code. :thinking:

PR #642, opened 2024-07-16
Add missing wait statements to CCA testing code

stephenswat:
The current CCA testing code for CUDA and SYCL is missing some necessary wait statements, which are causing me issues in #609. This commit fixes the problem by adding the missing waits.

### DIFF
--- a/tests/cuda/test_cca.cpp
+++ b/tests/cuda/test_cca.cpp
@@ -34,15 +34,15 @@ cca_function_t f = [](const traccc::cell_collection_types::host& cells,
         static_cast<traccc::cell_collection_types::buffer::size_type>(
             cells.size()),
         device_mr};
-    copy.setup(cells_buffer);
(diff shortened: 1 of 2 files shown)


krasznaa (changes requested):
I'm very surprised about your conclusion on this one. :thinking: Why should we need to wait on these operations? If anything, I would conclude that the `setup(...)` calls were missing `ignore()` calls. :thinking:

These copies **should** be fine to be put into the same stream as the kernels that get launched after them. Why do we need synchronization points on them? :confused:

krasznaa (approved):
Ahh, never mind. The SYCL versions do need to be synchronized I guess. Since `sycl::queue`-s don't work quite in the same way as a CUDA stream. (Events are not necessarily ordered in them.) So the CUDA test could technically would just fine without these changes. It's only the SYCL test that is more finnicky. :thinking:

But since we're only talking about some unit tests here, uniformity is probably more important to have.

PR #643, opened 2024-07-17
Enable thread backend for all Alpaka builds

stephenswat:
Currently, the serial CPU backend is enabled in Alpaka only if you download Alpaka using FetchContent. When using a preinstalled version of Alpaka, the default backend is not correctly set. In this PR, I move the default backend configuration from a file which is only loaded for FetchContent usage to the main CMakeLists.txt file, allowing it to be used in all builds.


stephenswat:
@CrossR I can't seem to ask you for a review, but could you please give a green light?

StewMH (approved):
Looks good to me.

paulgessinger (approved):

PR #644, opened 2024-07-17
Require multiple threads for Alpaka CCL code

stephenswat:
The CCL code currently does not support running with a single thread per block; it has to be run with multiple threads per block by design. In order to require this, I have added a new function that asserts this about a device (note that in Alpaka 1.2.0 there is `IsMultiThreadAcc`, but we cannot use this yet). I have also added a static assertion for this.


stephenswat:
Depends on #643.

stephenswat:
This is now ready to go; @CrossR or @StewMH could you take a look?

CrossR:
Also looks good to me!

paulgessinger (approved):

PR #645, opened 2024-07-17
Add `wait` or `ignore` for all async copies

stephenswat:
Using https://github.com/acts-project/vecmem/pull/286 I was able to find a few more cases where we do not properly wait for or ignore events. In this PR I opted to add those missing statements according to a simple heuristic, i.e. I added waits for all SYCL and share code and I added ignores for all CUDA code. Reason for this is that CUDA events are stream-ordered, so you can't really shoot yourself in the foot too much. SYCL events, on the other hand, are unordered, so they need stricted checking.


stephenswat:
@krasznaa Just so you're also aware of this. Will rebase.

stephenswat:
I believe this is obsolete.

PR #647, opened 2024-07-22
Make IO test mock data path absolute

stephenswat:
The current IO mock data path is relative, which breaks down if data is not stored in the `data/` directory. This commit makes the path absolute via some CMake code to ensure that it is treated as an absolute path.


beomki-yeo (approved):
Looks good to me

PR #648, opened 2024-07-22
Raise minimum C++ standard to 20

stephenswat:
Now that Acts has a minimum C++ version of C++20, we can follow suit. This commit removes the preprocessor macros that ensure backwards compatibility on concepts and increases the C++ version requirement to
20. Also updates the CI.


device/common/include/traccc/seeding/device/impl/select_seeds.ipp
```diff
@@ -153,13 +153,13 @@ inline void select_seeds(
                 scalar seed1_sum = 0;
                 scalar seed2_sum = 0;
 
-                const internal_spacepoint<spacepoint>& ispB1 =
```
niermann999, line 156: Why is this being copied?

niermann999 (approved):

device/common/include/traccc/seeding/device/impl/select_seeds.ipp
stephenswat: This avoids a (possibly false-positive) compiler warning.

PR #649, opened 2024-07-22
Make Kokkos code compatible with C++20

stephenswat:
The Kokkos code has two problems when it comes to C++20 compatibility. Firstly, the current version simply does not work in C++20 mode; compilation requires Kokkos 4.3.1 or later. There are also some warnings that prevent compilation. This commit fixes these issues by updating to Kokkos 4.3.1 and by adding some hacks to fix the compiler warnings.

### DIFF
--- a/device/kokkos/src/seeding/spacepoint_binning.cpp
+++ b/device/kokkos/src/seeding/spacepoint_binning.cpp
@@ -46,6 +46,10 @@ spacepoint_binning::output_type spacepoint_binning::operator()(
     const unsigned int num_threads = 32 * 8;
     const unsigned int num_blocks = (sp_size + num_threads - 1) / num_threads;
 
+    // Hack to avoid warnings thrown by C++20
+    seedfinder_config config = m_config;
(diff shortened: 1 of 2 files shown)


stephenswat:
@storrealbac @Yhatoh what's the status on the Kokkos code? Do you ever plan on expanding it? Also; can you please let me know if this change is okay for you?

Yhatoh:
Dear @stephenswat,

First, the change that you propose I agree with them. 
The status of the update I think is the same as the last one I presented to you, that is because I had to stop working on, due that I'm doing my PhD and had conflict with my project (and my university has some strict rule with it). Sorry for that, I think I forgot to mention about.

Also, the other student that was working on was @uwo-o, he is Hugo Campos (Sebastian had to leave the profe time ago).

And, yet I don't know who is continuing my work, you can find my last updates in my fork.

Best Regards,
Gabriel Carmona

stephenswat:
Thanks for the update @Yhatoh

paulgessinger (approved):

PR #650, opened 2024-07-23
Add Acts integration build CI job

stephenswat:
This commit adds a new CI job that tries to build Acts with the traccc plugin as it exists in the tag for which the CI job is launched. This will help ensure compatibility with Acts until the integration takes place.

### DIFF
--- a/.github/workflows/acts_integration.yml
+++ b/.github/workflows/acts_integration.yml
@@ -1,0 +1,58 @@
+# TRACCC library, part of the ACTS project (R&D line)
+#
+# (c) 2024 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+name: Acts Integration Builds
+
+on:
+  push:
+  pull_request:
+    branches:
+      - main
+
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}-acts_integration
+  cancel-in-progress: true
+
+jobs:
+  acts_build:
+    name: ${{ matrix.build }}-${{ matrix.tag }}
+    runs-on: ubuntu-latest
+
+    container: ghcr.io/acts-project/ubuntu2404:53
+
+    strategy:
+      matrix:
+        build:
+          - Release
+          - Debug
+        tag:
+          - main
+
+    steps:
+      - name: Checkout traccc
+        uses: actions/checkout@v4
+        with:
+          path: traccc/
+      - name: Checkout Acts
+        run:
+          git clone --depth 1 --branch ${{ matrix.tag }} https://github.com/acts-project/acts.git
+      - name: Configure Acts
+        run: |
+          cmake \
+            -DCMAKE_CXX_STANDARD=20 \
+            -DCMAKE_CUDA_STANDARD=20 \
+            -DCMAKE_HIP_STANDARD=20 \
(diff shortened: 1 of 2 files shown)


stephenswat:
Pinging @fredevb; check out the change to the CMakeLists.txt file that allows this to compile.

.github/workflows/acts_integration.yml
```diff
@@ -0,0 +1,63 @@
...
+            -DACTS_SETUP_TRACCC=ON \
+            -DACTS_SETUP_VECMEM=ON \
+            -DACTS_SETUP_DETRAY=ON \
+            -DACTS_SETUP_COVFIE=ON \
+            -DTRACCC_SETUP_THRUST=ON \
+            -DACTS_DETRAY_SOURCE="URL;https://github.com/acts-project/detray/archive/refs/tags/v0.69.1.tar.gz;URL_MD5;38a0f8e786ac6a61fabcd3a28d7bb2d6" \
+            -DACTS_SETUP_ALGEBRAPLUGINS=ON \
+            -DACTS_TRACCC_SOURCE="SOURCE_DIR;${GITHUB_WORKSPACE}/traccc" \
+            -S ${GITHUB_WORKSPACE}/acts \
+            -B build_acts
```
stephenswat, line 60: @fredevb Ideally, the setup would be much simpler so these flags would not be necessary:

```suggestion
            -DCMAKE_BUILD_TYPE=${{ matrix.build }} \
            -DCMAKE_CXX_STANDARD=20 \
            -DCMAKE_CUDA_STANDARD=20 \
            -DCMAKE_HIP_STANDARD=20 \
            -DCMAKE_SYCL_STANDARD=20 \
            -DACTS_BUILD_PLUGIN_TRACCC=ON \
            -DACTS_SETUP_TRACCC=ON \
            -DACTS_TRACCC_SOURCE="SOURCE_DIR;${GITHUB_WORKSPACE}/traccc" \
            -S ${GITHUB_WORKSPACE}/acts \
            -B build_acts
```

.github/workflows/acts_integration.yml
```diff
@@ -0,0 +1,58 @@
...
+    name: ${{ matrix.build }}-${{ matrix.tag }}
+    runs-on: ubuntu-latest
+
+    container: ghcr.io/acts-project/ubuntu2404:53
+
+    strategy:
+      matrix:
+        build:
+          - Release
+          - Debug
```
paulgessinger, line 30: Do we need to test both? I imagine they should either both work or both fail, right?

.github/workflows/acts_integration.yml
```diff
@@ -0,0 +1,58 @@
...
+        tag:
+          - main
```
paulgessinger, line 32: Do we expect random breakage due to changes in the ACTS repo that would pop up here?

krasznaa (changes requested):
We should talk about it in person... The setup of this will need to be more complicated than this. I'm on board with this type of a check, but as we know from the similar jobs that we have for Acts and Gaudi (and their integration into Athena), in this simple setup the job would just fail pretty often.

We have mechanisms that we use for Acts and Gaudi to be able to handle interface changes. If we want to introduce a test like this, we'll have to use the same mechanism between Acts and traccc.

PR #651, opened 2024-07-24
feat: make building of tracccc::io optional (but on by default)

asalzburger:
In order to include traccc (temporarily) as a plugin to ACTS, I suggest to add the CMake flag

`TRACCC_BUILD_IO` 

When switched off, neither the IO module and dependent examples/testing/performance modules will be built, but none of those are needed in the ACTS integration effort, where event data and geometry are  translated in memory.

This also breaks (some of?) the circular dependency for the digitisation IO.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,7 @@ option( TRACCC_BUILD_KOKKOS "Build the Kokkos sources included in traccc"
    FALSE )
 option( TRACCC_BUILD_ALPAKA "Build the Alpaka sources included in traccc"
    FALSE )
+option( TRACCC_BUILD_IO "Build the IO module (needed by examples, performance, testing)" TRUE )
 option( TRACCC_BUILD_TESTING "Build the (unit) tests of traccc" TRUE )
 option( TRACCC_BUILD_EXAMPLES "Build the examples of traccc" TRUE )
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -299,21 +300,33 @@ endif()
...
 if ( TRACCC_BUILD_EXAMPLES )
    # Find Boost.
    find_package( Boost REQUIRED COMPONENTS program_options filesystem)
-
-   add_subdirectory( examples )
+   if ( TRACCC_BUILD_IO )
+      add_subdirectory( examples )
+   else()
+      message(STATUS "traccc::io not built, traccc::examples are forcefully switched off.")
+   endif()
```
stephenswat, line 320: I'd prefer if this was a bit more vocal and explicit, i.e. I don't think it's a good idea for there to be any possibility for the user to specify `TRACCC_BUILD_EXAMPLES` and then for the build system to decide not to do it. I would suggest the following:

```suggestion
   if ( NOT TRACCC_BUILD_IO )
      message(FATAL_ERROR "traccc::io is disabled, but it is required to build the examples.")
   endif()
   add_subdirectory( examples )
```

CMakeLists.txt
```diff
@@ -299,21 +300,33 @@ endif()
...
 endif()
 
 # Set up the test(s).
 if( BUILD_TESTING AND TRACCC_BUILD_TESTING )
-   add_subdirectory( tests )
+   if ( TRACCC_BUILD_IO )
+      add_subdirectory( tests )
+   else()
+      message(STATUS "traccc::io not built, traccc::tests are forcefully switched off.")
+   endif()
```
stephenswat, line 329:
```suggestion
   if ( NOT TRACCC_BUILD_IO )
      message(FATAL_ERROR "traccc::io is disabled, but it is required to build the tests.")
   endif()
   add_subdirectory( tests )
```

stephenswat (changes requested):

CMakeLists.txt
asalzburger: You have a good point here!

stephenswat (approved):

PR #652, opened 2024-07-24
Archive the Futhark and Kokkos code

stephenswat:
Our Futhark and Kokkos codebases are not being actively developed right now, so I suggest we archive them into an archive directory.


krasznaa (changes requested):
It's an ancient PR at this point, but...

It's an absolute no to Kokkos. That thing we shall still revive at one point.

As for Futhark, the code is pretty small, so I guess we could move it like this. But why not just remove it completely? Version control is exactly for this, no?

PR #653, opened 2024-07-25
Enable tests for Alpaka builds

stephenswat:
Now that we have some Alpaka tests, we are ready to actually run them in the CI.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -49,11 +49,6 @@ jobs:
             cxx_standard: "20"
             options: -DTRACCC_BUILD_KOKKOS=TRUE
             run_tests: false
-          - name: ALPAKA
-            container: ghcr.io/acts-project/ubuntu2404:56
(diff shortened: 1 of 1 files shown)


stephenswat:
Interestingly, debug builds make Alpaka code so slow that it times out the CI. :thinking:

krasznaa:
Two options we have here:
  - Increase the timeout from 1500 seconds to something larger.
  - Only test Alpaka in Release mode.

I'm not even sure which of these two bad options is the better one. But probably I'd go for the timeout increase. :thinking:

stephenswat:
It's really interesting why the Alpaka code is so much slower; we could also try adding an `-O2` to the Alpaka build; also far from ideal, though.

StewMH:
I think disabling for debug is fine.

stephenswat:
Okay, I have updated this to run the Alpaka tests for release builds only.

paulgessinger (approved):

PR #663, opened 2024-08-01
Increase minimum CMake version to 3.22

stephenswat:
Earlier versions of CMake were unable to properly configure GCC for C++20 compilation, opting to add the `-std=c++2a` flag instead of the `-std=c++20` flag which broke the build. CMake 3.22 correctly sets the C++20 flag, and as such we are setting this as our new minimum version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@
 # Mozilla Public License Version 2.0
 
 # Set up the project.
-cmake_minimum_required( VERSION 3.9 )
+cmake_minimum_required( VERSION 3.22 )
 project( traccc VERSION 0.13.0 LANGUAGES CXX )
 
 # Set up the used C++ standard(s).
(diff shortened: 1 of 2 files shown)


krasznaa (changes requested):
Unfortunately this is not enough. :frowning: 3.22 might be enough for the host code compilation, but not for CUDA. :frowning:

https://gitlab.kitware.com/cmake/cmake/-/blob/v3.22.0/Modules/Compiler/NVIDIA-CUDA.cmake

We need to go all the way to 3.26 for C\+\+20 support with CUDA. :scream:

https://gitlab.kitware.com/cmake/cmake/-/blob/v3.26.0/Modules/Compiler/NVIDIA-CUDA.cmake

I.e., this PR should request 3.26 in: https://github.com/acts-project/traccc/blob/main/device/cuda/CMakeLists.txt

stephenswat:
I added a CMake 3.26 requirement for the CUDA library.

krasznaa (approved):
:+1:

PR #665, opened 2024-08-02
Move shared CCL variables into single struct

stephenswat:
The current CCL kernels have so many parameters that it's a real pain in the rear to maintain them and to make changes to them. This commit reduces the number of parameters a little bit by taking all statically-known shared memory data and unifying it into a single struct which can be passed around more easily.


beomki-yeo (approved):
looks good to me

krasznaa (commented):
You should add/update Doxygen documentation for the code. But other than that, yeah, no objections.

PR #552, opened 2024-04-23
ODD CUDA Reconstruction, main branch (2024.04.23.)

krasznaa:
After the last ~2 weeks of developments, full-chain ODD reconstruction with CUDA is finally here. :partying_face:

Unfortunately it's not quite as perfect just yet as it should be. :frowning:

In order to get the code running, I had to generalize `traccc::cuda::finding_algorithm` a little. It was not exercised with a resizable measurement buffer so far, so it was doing an invalid access when I gave it a resizable container. (I had to deal with exactly this in #545 and #550 already, so it didn't take long to find the issue...)

Other than that, I "just" had to update `seq_example_cuda.cpp` to add track finding and track fitting to it. :thinking: With this PR's code, I now get:

```
[bash][atspot01]:traccc > ./out/build/sycl/bin/traccc_seq_example_cuda --detector-file=geometries/odd/odd_geometry_detray.json --use-detray-detector --digitization-file=geometries/odd/odd-digi-geometric-config.json --input-directory=odd/muon100GeV-geant4/ --input-events=3 --compare-with-cpu

Running Full Tracking Chain Using CUDA

>>> Detector Options <<<
  Detector file       : geometries/odd/odd_geometry_detray.json
  Material file       : 
  Surface rid file    : 
  Use detray::detector: yes
  Digitization file   : geometries/odd/odd-digi-geometric-config.json
>>> Input Data Options <<<
  Input data format             : csv
  Input directory               : odd/muon100GeV-geant4/
  Number of input events        : 3
  Number of input events to skip: 0
>>> Clusterization Options <<<
  Target cells per partition: 1024
>>> Track Seeding Options <<<
  None
>>> Track Finding Options <<<
  Track candidates range   : 3:100
  Maximum Chi2             : 30
  Maximum branches per step: 4294967295
>>> Track Propagation Options <<<
  Constraint step size : 3.40282e+38 [mm]
  Overstep tolerance   : -100 [um]
  Mask tolerance       : 15 [um]
  Search window        : 0 x 0
  Runge-Kutta tolerance: 0.0001
>>> Performance Measurement Options <<<
  Run performance checks: no
>>> Accelerator Options <<<
  Compare with CPU results: yes

WARNING: No material in detector
WARNING: No entries in volume finder
Detector check: OK
WARNING: No material in detector
WARNING: No entries in volume finder
Detector check: OK
[...]

PR #666, opened 2024-08-02
Ensure that CCL output is contiguous on modules

stephenswat:
Right now we are using a sorting algorithm to ensure that the entire array of measurements is contiguous in global memory, but this isn't strictly necessary. This commit alters the CCL algorithm slightly to guarantee that the output is always contiguous.

### DIFF
--- a/device/cuda/src/clusterization/clusterization_algorithm.cu
+++ b/device/cuda/src/clusterization/clusterization_algorithm.cu
@@ -42,8 +42,7 @@ __global__ void ccl_kernel(
     vecmem::data::vector_view<device::details::index_t> adjv_backup_view,
     unsigned int* backup_mutex_ptr) {
 
-    __shared__ std::size_t partition_start, partition_end;
-    __shared__ std::size_t outi;
+    __shared__ device::details::ccl_kernel_static_smem_parcel shared;
     extern __shared__ device::details::index_t shared_v[];
     vecmem::device_atomic_ref<unsigned int> backup_mutex(*backup_mutex_ptr);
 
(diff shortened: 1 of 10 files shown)


stephenswat:
Depends on #552, #665. I will do some performance testing on this.

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -140,6 +138,12 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
         m_gf_backup, m_adjc_backup, m_adjv_backup, m_backup_mutex.get());
     TRACCC_CUDA_ERROR_CHECK(cudaGetLastError());
 
+#ifndef NDEBUG
+    TRACCC_CUDA_ERROR_CHECK(cudaStreamSynchronize(stream));
+    assert(is_contiguous_on(measurement_module_projection(), m_mr.main, m_copy,
+                            m_stream, measurements));
+#endif
```
krasznaa, line 145: This is really not an important point, but would it not be possible to express this with something like:

```c++
assert([&]() -> bool {
   TRACCC_CUDA_ERROR_CHECK(cudaStreamSynchronize(stream));
   return is_contiguous_on(measurement_module_projection(), m_mr.main, m_copy,
                            m_stream, measurements); }() == true);
```

? I'd just prefer to use `NDEBUG` directly if I can help it...

krasznaa (commented):
With the actual CCL code I'll go with your judgement here. Nothing jumps out to me as code that I wouldn't like, but I also didn't try to understand exactly what is being done. :frowning:

It doesn't have to be this PR, but I guess we'll want to remove the sorting algorithm(s) from the example applications then. I'm fine with that happening in a separate PR, though you should definitely test at least locally that track finding/fitting would take this sorting well. :thinking:

stephenswat:
> It doesn't have to be this PR, but I guess we'll want to remove the sorting algorithm(s) from the example applications then. I'm fine with that happening in a separate PR, though you should definitely test at least locally that track finding/fitting would take this sorting well. 🤔

Indeed, we should disable the sorting application. I was thinking of adding some boolean flag to the output to indicate whether or not the output is contiguous, but I am not sure if it's worth the hassle.

krasznaa:
No, there's no need to represent this in the EDM. As long as the documentation on the algorithms is clear, it's fine to just know whether extra sorting is needed on something or not.

beomki-yeo:
Once this works. could you also remove the `measurement_sorting_algorithm`?

krasznaa:
> Once this works. could you also remove the `measurement_sorting_algorithm`?

I would **not** want to do that! Yes, we will need to remove that / those algorithm(s) from the standard algorithm chain. But having an algorithm in the repository that is not actively used at a specific time (but is otherwise functional), is not a bad thing. I want to keep those algorithms around. We may make use of them in some other way later on, who knows. :thinking:

beomki-yeo:
OK that makes sense

PR #668, opened 2024-08-06
Use traccc scalar type in seeding config

stephenswat:
The maximum scattering angle was being calculated as a floating point values, which was causing errors in the Acts integration. This minor change updates this to use the generic scalar type, which can be synchronised with the Acts scalar type.


beomki-yeo (approved):

krasznaa:
Curious. 🤔 For "configuration types" I generally prefer to just use `float` directly. What's the exact issue that we're solving in Acts with this?

PR #669, opened 2024-08-08
Add a CI build for double precision

beomki-yeo:
We need to make sure that the traccc is compiled normally for the double precision

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -59,6 +59,13 @@ jobs:
           - Debug
         include:
           - platform:
+              name: CUDA
+              container: ghcr.io/acts-project/ubuntu2404_cuda:56
+              cxx_standard: "20"
+              options: -DTRACCC_CUSTOM_SCALARTYPE=double -DDETRAY_CUSTOM_SCALARTYPE=double -DTRACCC_BUILD_CUDA=TRUE -DCMAKE_CUDA_FLAGS="-std=c++20"
+              run_tests: false
+            build: Release
+          - platform:
               name: "SYCL"
               container: ghcr.io/acts-project/ubuntu2404_cuda_oneapi:56
               cxx_standard: "20"
--- a/simulation/include/traccc/simulation/measurement_smearer.hpp
+++ b/simulation/include/traccc/simulation/measurement_smearer.hpp
@@ -96,24 +96,26 @@ struct measurement_smearer {
             if constexpr (std::is_same_v<
                               typename mask_t::local_frame_type,
                               detray::line2D<traccc::default_algebra>>) {
-                iomeas.local0 =
+                iomeas.local0 = static_cast<float>(
                     std::max(std::abs(matrix_operator().element(meas, 0u, 0u)) +
                                  offset[0],
-                             static_cast<scalar_type>(0.f));
+                             scalar_type(0.f)));
             } else if constexpr (std::is_same_v<typename mask_t::shape,
                                                 detray::annulus2D>) {
-                iomeas.local1 =
-                    matrix_operator().element(meas, 0u, 0u) + offset[0];
+                iomeas.local1 = static_cast<float>(
+                    matrix_operator().element(meas, 0u, 0u) + offset[0]);
             } else {
-                iomeas.local0 =
-                    matrix_operator().element(meas, 0u, 0u) + offset[0];
+                iomeas.local0 = static_cast<float>(
+                    matrix_operator().element(meas, 0u, 0u) + offset[0]);
             }
         } else if (meas_dim == 2u) {
             const auto proj = subs.projector<2u>();
             matrix_type<2u, 1u> meas = proj * bound_params.vector();
 
-            iomeas.local0 = matrix_operator().element(meas, 0u, 0u) + offset[0];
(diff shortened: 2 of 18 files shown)


.github/workflows/builds.yml
```diff
@@ -39,6 +39,11 @@ jobs:
             cxx_standard: "20"
             options: -DTRACCC_BUILD_CUDA=TRUE -DCMAKE_CUDA_FLAGS="-std=c++20"
             run_tests: false
+          - name: CUDA-double
```
krasznaa, line 42: As inconvenient it is, the name should be `CUDA`, not `CUDA-double`. :thinking: That name is something that we can make some decisions about. Even if we currently don't. (This is why all of the SYCL tests are named `SYCL` as well.)

krasznaa (changes requested):
Do we need both Release and Debug builds with this setting?

I am absolutely on board with adding this. If hardware resources were not a concern, I could imagine just doubling the number of our tests outright. Adding one more "axis" to our test matrix.

Actually, should we try that? @paulgessinger, I guess you may have the most experience here. With how slow our CI may become if we throw so much work at it. :thinking:

beomki-yeo:
> Do we need both Release and Debug builds with this setting?

No, running it  for Debug only would be ideal for the current setting. I just could not find a elegant way to exclude the Release build

And I think it is going to be too much if we double the amount of tests. I am here assuming that successful tests with the FP32 also guarantee the success with the FP64, hopefully

.github/workflows/builds.yml
beomki-yeo: OK Fixed now

stephenswat:
> I just could not find a elegant way to exclude the Release build

The easy way to do this would be using the `matrix.include` method, as seen in https://github.com/acts-project/traccc/blob/main/.github/workflows/builds.yml#L60.

In short, you'd add something like this:

```yaml
        include:
          - platform:
            - name: CUDA
              container: ghcr.io/acts-project/ubuntu2404_cuda:56
              cxx_standard: "20"
              options: -DTRACCC_CUSTOM_SCALARTYPE=double -DDETRAY_CUSTOM_SCALARTYPE=double -DTRACCC_BUILD_CUDA=TRUE -DCMAKE_CUDA_FLAGS="-std=c++20"
              run_tests: false
            build: Release
```

paulgessinger:
@krasznaa Is this on the github or the gitlab side?

stephenswat:
> @krasznaa Is this on the github or the gitlab side?

This is on the 'Hub.

krasznaa:
Yepp, Pandora's box... :frowning:

Because of the issues, it could still make sense to set up `double` versions of pretty much all of the device backends. Otherwise there will always be some code that our tests don't cover. :thinking:

It doesn't need to be this PR though. You'll have plenty of work with just making the core (and CUDA) code work correctly like this. We can take care of fixing up everything else in follow-up PRs.

beomki-yeo:
While making this work, I also removed the awkward scalar type dependency in the `seeding_config` and `finding_config`

simulation/include/traccc/simulation/measurement_smearer.hpp
```diff
@@ -96,24 +96,26 @@ struct measurement_smearer {
             if constexpr (std::is_same_v<
                               typename mask_t::local_frame_type,
                               detray::line2D<traccc::default_algebra>>) {
-                iomeas.local0 =
+                iomeas.local0 = static_cast<float>(
```
krasznaa, line 99: Shouldn't we rather `static_cast` to `traccc::scalar` here instead? The EDM does use `traccc::scalar` and not `float` for a reason. I think here we should be just a bit more careful. :thinking:

simulation/include/traccc/simulation/smearing_writer.hpp
```diff
@@ -74,14 +74,14 @@ struct smearing_writer : detray::actor {
...
             particle.particle_id = particle_id;
-            particle.vx = pos[0];
-            particle.vy = pos[1];
-            particle.vz = pos[2];
-            particle.vt = track.time();
-            particle.px = mom[0];
-            particle.py = mom[1];
-            particle.pz = mom[2];
-            particle.q = track.charge();
+            particle.vx = static_cast<float>(pos[0]);
```
krasznaa, line 77: Also here. Or is the "EDM" not using `traccc::scalar` here? :confused: If the "EDM" uses `float`, then of course you're right, and this is how it should go...

krasznaa (commented):
I'm just a bit uncertain about the simulation changes. Everything else I'm very much on board with. :+1:

simulation/include/traccc/simulation/measurement_smearer.hpp
beomki-yeo: core EDM uses `traccc::scalar` but IO EDM uses float - I don't recall why the IO EDM happen to be float though

krasznaa (approved):
Okay, let's go then! :smile:

simulation/include/traccc/simulation/measurement_smearer.hpp
beomki-yeo: Right. I also think most of EDM do not need to be flexible in scalar type as they are now in traccc :thinking: I am also thinking of removing scalar template type for some classes in detray

PR #672, opened 2024-08-13
Disable Default CUDA Build, main branch (2024.08.13.)

krasznaa:
This is a bit of a subjective proposal, so I'm interested in the feedback...

To make my life just a little easier, I'd like to stop the project from automatically enabling the build of its CUDA code when it thinks that it found a functional CUDA version. And so be consistent with how the SYCL, Alpaka, etc. sources need to be "turned on" for the build.

My reason is a bit complicated. It stems from the issue described under: https://discourse.cmake.org/t/checklanguage-does-not-take-the-c-standard-into-account When I set up the latest version of oneAPI with the latest version of CUDA, right now I both have to explicitly turn on the build of the SYCL code (for the NVIDIA backend, that's why I also have CUDA set up), and turn off the build of the CUDA code. Unfortunately the `sycl` preset in `CMakePresets.json` doesn't do this at the moment. So I can't use VSCode for SYCL code development with the latest oneAPI version at the moment, without modifying the CMake code a little.

Instead of doing that, I thought that this specific update could be acceptable by everyone. This way I could easily use the `sycl` preset (with VSCode) with the latest oneAPI version, and the `cuda` preset in a different environment in which I do not set up oneAPI at all.

But this does mean that if one of you didn't use any presets, and didn't set `TRACCC_BUILD_CUDA` so far when doing CUDA developments, you'd need to have a little longer configuration command from now on. :thinking:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,17 +44,8 @@ set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY
 list( APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake" )
 include( traccc-functions )
 
-# Check if CUDA is available.
-include( CheckLanguage )
-check_language( CUDA )
-set( TRACCC_BUILD_CUDA_DEFAULT FALSE )
-if( CMAKE_CUDA_COMPILER )
-   set( TRACCC_BUILD_CUDA_DEFAULT TRUE )
-endif()
-
 # Flags controlling which parts of traccc to build.
-option( TRACCC_BUILD_CUDA "Build the CUDA sources included in traccc"
-   ${TRACCC_BUILD_CUDA_DEFAULT} )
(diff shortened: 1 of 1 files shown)


krasznaa:
As a little palate cleanser, what about this proposal?

@paulgessinger, @beomki-yeo?

beomki-yeo:
I don't use preset but do old-school cmake command like ancient people

krasznaa:
For that case this change means that you'd need to add `-DTRACCC_BUILD_CUDA=TRUE` explicitly when you want to turn on building the CUDA code. Would that be acceptable?

Right now I need to edit `CMakePresets.json` locally to do SYCL builds with the latest oneAPI version, in VSCode. This is not the end of the world (I could also add a `CMakeUserPresets.json` file locally if I were less lazy), but this proposed change can also be looked at as harmonizing the CUDA build with everything else. :wink:

beomki-yeo (approved):
Yeah it is not a big deal

PR #673, opened 2024-08-14
Externals Update, main branch (2024.08.14.)

krasznaa:
This is mainly to update to the latest VecMem version, to help with a build issue discovered in #627. (Because of that, it's relatively urgent.)

But while I was at it, I also updated to the latest version of GoogleTest. Which then also necessitated adding an `<iomanip>` include, which was missing so far from one of the tests.


beomki-yeo (approved):
LGTM

PR #674, opened 2024-08-14
Add CPU and CUDA benchmark for toy detector reconstruction

beomki-yeo:
This is for benchmarking the general speed of the track reconstruction in the toy geometry. (There will be a follow-up PR for monitoring)

Because the geometry and simulation data is not very forward compatible, the simulation is also done in the benchmark suite before the reconstruction. Of course, the simulation time is not included in the timing measurement.

Unfortunately, the simulation does not support the digitization, the performance of clusterization is not included in this benchmark.

Following is the output of CPU (+OpenMP) and CUDA benchmark with the single precision on my laptop:

### CPU
```
[beomki@device-11 traccc_build]$ ./bin/traccc_benchmark_cpu 
2024-08-14T10:31:38+02:00
Running ./bin/traccc_benchmark_cpu
Run on (12 X 4367.4 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 12288 KiB (x1)
Load Average: 3.16, 3.19, 2.06

Detector check: OK
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
ToyDetectorBenchmark/CPU 1.4495e+10 ns   1.3887e+10 ns            1
```

### CUDA
```
[beomki@device-11 traccc_build]$ ./bin/traccc_benchmark_cuda
2024-08-14T10:33:49+02:00
Running ./bin/traccc_benchmark_cuda
Run on (12 X 4344.84 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 12288 KiB (x1)
Load Average: 2.94, 3.16, 2.20

Detector check: OK
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
ToyDetectorBenchmark/CUDA 4734637722 ns   4719869245 ns            1
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,7 @@ option( TRACCC_BUILD_ALPAKA "Build the Alpaka sources included in traccc"
    FALSE )
 option( TRACCC_BUILD_IO "Build the IO module (needed by examples, performance, testing)" TRUE )
 option( TRACCC_BUILD_TESTING "Build the (unit) tests of traccc" TRUE )
+option( TRACCC_BUILD_BENCHMARKS "Build the benchmarks of traccc" TRUE )
 option( TRACCC_BUILD_EXAMPLES "Build the examples of traccc" TRUE )
 
 # Flags controlling what traccc should use.
@@ -291,6 +292,20 @@ if( TRACCC_SETUP_GOOGLETEST )
    endif()
 endif()
 
+# Set up Google Benchmark
+option( TRACCC_SETUP_BENCHMARKS
+   "Set up the Google Benchmark target(s) explicitly" TRUE )
+option( TRACCC_USE_SYSTEM_BENCHMARKS
+   "Pick up an existing installation of Google Benchmark from the build environment"
+   ${TRACCC_USE_SYSTEM_LIBS} )
+if( TRACCC_SETUP_BENCHMARKS )
+   if( TRACCC_USE_SYSTEM_BENCHMARKS )
+      find_package( benchmark REQUIRED )
+   else()
+      add_subdirectory( extern/benchmark )
+   endif()
+endif()
+
 option( TRACCC_ENABLE_NVTX_PROFILING
         "Use instrument functions to enable fine grained profiling" FALSE )
 
@@ -339,6 +354,16 @@ if( BUILD_TESTING AND TRACCC_BUILD_TESTING )
       message(FATAL_ERROR "traccc::io is disabled, but it is required to build the tests.")
    endif()
    add_subdirectory( tests )
+endif()
+
+# Set up the benchmark(s).
+if( TRACCC_BUILD_BENCHMARKS )
+   # Find Boost.
+   find_package( Boost REQUIRED COMPONENTS filesystem)
+   if ( NOT TRACCC_BUILD_IO )
+      message(FATAL_ERROR "traccc::io is disabled, but it is required to build the tests.")
+   endif()
+   add_subdirectory( benchmarks )
 endif()
 
 if(TRACCC_BUILD_FUTHARK)
--- a/benchmarks/common/benchmarks/toy_detector_benchmark.hpp
+++ b/benchmarks/common/benchmarks/toy_detector_benchmark.hpp
@@ -1,0 +1,186 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+// Traccc include(s).
+#include "traccc/definitions/common.hpp"
+#include "traccc/finding/finding_algorithm.hpp"
+#include "traccc/fitting/fitting_algorithm.hpp"
+#include "traccc/io/utils.hpp"
+#include "traccc/seeding/seeding_algorithm.hpp"
+#include "traccc/seeding/track_params_estimation.hpp"
+#include "traccc/simulation/measurement_smearer.hpp"
+#include "traccc/simulation/simulator.hpp"
+#include "traccc/simulation/smearing_writer.hpp"
+
+// Detray include(s).
+#include "detray/core/detector.hpp"
+#include "detray/definitions/units.hpp"
+#include "detray/detectors/bfield.hpp"
+#include "detray/detectors/build_toy_detector.hpp"
+#include "detray/geometry/mask.hpp"
+#include "detray/geometry/shapes/rectangle2D.hpp"
+#include "detray/io/frontend/detector_reader.hpp"
+#include "detray/io/frontend/detector_writer.hpp"
+#include "detray/navigation/detail/ray.hpp"
+#include "detray/navigation/navigator.hpp"
+#include "detray/propagator/propagator.hpp"
+#include "detray/propagator/rk_stepper.hpp"
+#include "detray/simulation/event_generator/track_generators.hpp"
+
+// VecMem include(s).
+#include <vecmem/memory/host_memory_resource.hpp>
+
+// Boost include(s).
+#include <boost/filesystem.hpp>
+
+// Google Benchmark include(s).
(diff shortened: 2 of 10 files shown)


CMakeLists.txt
```diff
@@ -324,7 +339,7 @@ endif()
 
 add_subdirectory( plugins )
 
-if ( TRACCC_BUILD_EXAMPLES )
+if ( TRACCC_BUILD_BENCHMARK OR TRACCC_BUILD_EXAMPLES )
```
krasznaa, line 342: From what I can see, you are only turning on the build of the examples when `TRACCC_BUILD_BENCHMARK` is `TRUE`, to re-use the `find_package(Boost)` call from this block. :confused:

You should just call `find_package(Boost)` in the `if(TRACCC_BUILD_BENCHMARK)` block as well. Let's not turn on the build of unnecessary things. :thinking:

CMakeLists.txt
```diff
@@ -341,6 +356,11 @@ if( BUILD_TESTING AND TRACCC_BUILD_TESTING )
    add_subdirectory( tests )
 endif()
 
+# Set up the benchmark(s).
+if( TRACCC_BUILD_BENCHMARK )
+   add_subdirectory( benchmarks )
```
krasznaa, line 361: Similar to what is done for the examples and the tests, you should add a check for `TRACCC_BUILD_IO` here as well.

benchmarks/common/benchmarks/toy_detector_benchmark.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+        // Type declarations
+        using writer_type = traccc::smearing_writer<
+            traccc::measurement_smearer<traccc::default_algebra>>;
+
+        // Writer config
+        typename writer_type::config smearer_writer_cfg{meas_smearer};
+
+        // Run simulator
+        const std::string full_path = traccc::io::data_directory() + sim_dir;
```
krasznaa, line 123: We will really need to find a better procedure for using temporary directories. :thinking:

But since that will be necessary in other places as well, we can do that in a future PR...

benchmarks/cpu/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+# Build the benchmark executable.
+traccc_add_executable(benchmark_cpu
+    "toy_detector_cpu.cpp"
+    LINK_LIBRARIES benchmark::benchmark benchmark::benchmark_main 
+    traccc::core traccc_benchmarks_common 
+    detray::core detray::utils vecmem::core)
+
+# Optimize the cpu benchmark (e.g. auto-vectorization)
+target_compile_options(traccc_benchmark_cpu PRIVATE
+    "-march=native" "-ftree-vectorize")
```
krasznaa, line 19: No, we should not be doing this. :thinking:

Even if logically we'd like to set things up like this, we would need to explicitly check first whether the selected host compiler supports these flags. It might not.

But generally, such optimization flags are not meant to be specified by the project itself. But by the building/packaging system that builds the project.

In practice, you should set the `$CXXFLAGS` environment variable to the optimization flags that you'd like, and build the code like that. This way you could try out different optimization settings without changing the code at all.

Finally: I believe `-ftree-vectorize` is enabled at `-O3` by default. So just by building with `-DCMAKE_BUILD_TYPE=Release`, that flag should already be there. :thinking:

benchmarks/cpu/toy_detector_cpu.cpp
```diff
@@ -0,0 +1,97 @@
...
+            auto track_candidates =
+                host_finding(det, field, measurements_per_event, params);
+
+            // Track fitting with KF
+            auto track_states = host_fitting(det, field, track_candidates);
+        }
+    }
+}
+
+BENCHMARK_MAIN();
```
krasznaa, line 97: The `traccc_benchmarks_common` library is already linked against `benchmark::benchmark_main`. You shouldn't put `BENCHMARK_MAIN()` into any of the source files.

benchmarks/cuda/toy_detector_cuda.cpp
```diff
@@ -0,0 +1,198 @@
...
+                     track_states_host_buffer.headers,
+                     vecmem::copy::type::device_to_host);
+            vecmem::copy::event_type item_event = copy(
+                track_states_cuda_buffer.items, track_states_host_buffer.items,
+                vecmem::copy::type::device_to_host);
+        }
+    }
+}
+
+BENCHMARK_MAIN();
```
krasznaa, line 198: Same here. No need for `BENCHMARK_MAIN()`.

benchmarks/common/benchmarks/toy_detector_benchmark.hpp
```diff
@@ -0,0 +1,184 @@
...
+// VecMem include(s).
+#include <vecmem/memory/host_memory_resource.hpp>
+
+// Boost include(s).
+#include <boost/filesystem.hpp>
+
+// Google Benchmark include(s).
+#include <benchmark/benchmark.h>
+
+namespace {}  // namespace
```
krasznaa, line 43: :question:

CMakeLists.txt
```diff
@@ -65,6 +65,7 @@ option( TRACCC_BUILD_ALPAKA "Build the Alpaka sources included in traccc"
    FALSE )
 option( TRACCC_BUILD_IO "Build the IO module (needed by examples, performance, testing)" TRUE )
 option( TRACCC_BUILD_TESTING "Build the (unit) tests of traccc" TRUE )
+option( TRACCC_BUILD_BENCHMARK "Build the benchmarks of traccc" TRUE )
```
krasznaa, line 68: Super subjective comment, but I myself would call it `TRACCC_BUILD_BENCHMARKS` (plural) instead. :thinking:

krasznaa (changes requested):
I'm absolutely on board with the addition of these benchmarks. Just have a number of technical comments...

CMakeLists.txt
beomki-yeo: IT was BENCHMARKS at the beginning but I changed to BENCHMARK to follow the detray standard...

beomki-yeo: No actually I was referring other variables :p will fix it to BENCHMARKS

CMakeLists.txt
beomki-yeo: OK

benchmarks/cpu/CMakeLists.txt
beomki-yeo: OK...

> In practice, you should set the $CXXFLAGS environment variable to the optimization flags that you'd like, and build the code like that. This way you could try out different optimization settings without changing the code at all.

How can I do this exactly in CMakeListsts.txt? Or do you mean this should be set by the shell command

benchmarks/cuda/toy_detector_cuda.cpp
beomki-yeo: OK

benchmarks/common/benchmarks/toy_detector_benchmark.hpp
beomki-yeo: Will remove

CMakeLists.txt
beomki-yeo: Right.. Will fix it

benchmarks/cpu/CMakeLists.txt
beomki-yeo: OK I removed them now from the CMakeLists.txt

krasznaa (approved):
Once you rebase it, it should be good to go. :+1:

PR #675, opened 2024-08-14
Fix a typo(?) in thrust CMake setting

beomki-yeo:
Not sure if this is a real typo or not


krasznaa (approved):
If our code doesn't need `libcudacxx`, then it's indeed better to switch it off. Let's see what the CI says.

stephenswat:
Unfortunately this line was necessary for the Acts integration. We can
discuss next week.

PR #676, opened 2024-08-14
build: Only apply `-Wfloat-conversion` if traccc is top-level

paulgessinger:
Also changes `-pedantic` to `-Wpedantic` which I think was a typo (they're different options)

### DIFF
--- a/cmake/traccc-compiler-options-cpp.cmake
+++ b/cmake/traccc-compiler-options-cpp.cmake
@@ -22,7 +22,7 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wextra" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wshadow" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wunused-local-typedefs" )
-   traccc_add_flag( CMAKE_CXX_FLAGS "-Wpedantic" )
+   traccc_add_flag( CMAKE_CXX_FLAGS "-pedantic" )
    if(PROJECT_IS_TOP_LEVEL)
      traccc_add_flag( CMAKE_CXX_FLAGS "-Wfloat-conversion" )
    endif()
--- a/cmake/traccc-compiler-options-cpp.cmake
+++ b/cmake/traccc-compiler-options-cpp.cmake
@@ -22,8 +22,10 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wextra" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wshadow" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wunused-local-typedefs" )
-   traccc_add_flag( CMAKE_CXX_FLAGS "-pedantic" )
-   traccc_add_flag( CMAKE_CXX_FLAGS "-Wfloat-conversion" )
+   traccc_add_flag( CMAKE_CXX_FLAGS "-Wpedantic" )
+   if(PROJECT_IS_TOP_LEVEL)
+     traccc_add_flag( CMAKE_CXX_FLAGS "-Wfloat-conversion" )
+   endif()
 
    # Fail on warnings, if asked for that behaviour.
    if( TRACCC_FAIL_ON_WARNINGS )


cmake/traccc-compiler-options-cpp.cmake
```diff
@@ -22,8 +22,10 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wextra" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wshadow" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wunused-local-typedefs" )
-   traccc_add_flag( CMAKE_CXX_FLAGS "-pedantic" )
-   traccc_add_flag( CMAKE_CXX_FLAGS "-Wfloat-conversion" )
+   traccc_add_flag( CMAKE_CXX_FLAGS "-Wpedantic" )
```
krasznaa, line 25: :confused: No, it was not a typo. `-pedantic` is the correct flag for GCC and LLVM/Clang.

cmake/traccc-compiler-options-cpp.cmake
```diff
@@ -22,8 +22,10 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
...
+   if(PROJECT_IS_TOP_LEVEL)
```
krasznaa, line 26: Where is `PROJECT_IS_TOP_LEVEL` meant to come from?

cmake/traccc-compiler-options-cpp.cmake
```diff
@@ -22,8 +22,10 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
...
+     traccc_add_flag( CMAKE_CXX_FLAGS "-Wfloat-conversion" )
```
krasznaa, line 27: Curious. How does this flag cause warnings when building traccc as part of another project?

This line is only changing the compilation flags for the traccc source files themselves. It should have no effect on any client of the traccc headers. So, how does it produce warnings in Acts? :confused:

krasznaa (changes requested):
I'm fairly confused...

cmake/traccc-compiler-options-cpp.cmake
paulgessinger: I was under the impression `-pedantic` is a C thing, but ok. `-Wpedantic` is still much more common.

cmake/traccc-compiler-options-cpp.cmake
paulgessinger: It's built in as of CMake 3.21 (https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html), and we already require 3.22 at minimum.

cmake/traccc-compiler-options-cpp.cmake
paulgessinger: traccc in ACTS includes ACTS headers, which then inherit traccc's compile flags. I'm working on making ACTS compile with `-Wfloat-conversion` but this will take a while.

krasznaa: Ahh, I see. Yes, this back-and-forth dependence between the two projects makes things a bit hard to track...

cmake/traccc-compiler-options-cpp.cmake
krasznaa: Okay. I didn't use it so far, but since we need s pretty new CMake version for this code anyway...

cmake/traccc-compiler-options-cpp.cmake
krasznaa: I **think** `-pedantic` is "more pedantic" than `-Wpedantic`.

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

We definitely use `-pedantic` in all of the ATLAS builds since forever. :thinking:

paulgessinger: ok

paulgessinger:
```suggestion
```

paulgessinger:
Revertded the `-pedantic` change @krasznaa

krasznaa (approved):
Okay, let's do this until the warnings are removed from the Acts headers. :+1:

krasznaa:
Don't forget to rebase the PR! Other than that, I'm good with this.

PR #677, opened 2024-08-14
macOS Fixes, main branch (2024.08.14.)

krasznaa:
Triggered by @paulgessinger pointing out that the latest version of the code doesn't work on macOS, this is what I had to do to fix that.

As it happens, the `_LIBCUDACXX_STRING_H_HAS_CONST_OVERLOADS` definition makes the code, or at least the part of it that we use, work on macOS. So I added a bit of code that would add `_LIBCUDACXX_STRING_H_HAS_CONST_OVERLOADS` to the `_Thrust_Thrust` target, if it seems that it's needed. I decided to do this only for the case where we build Thrust / CCCL ourselves. Since on macOS we would always do this. (With CUDA not actually supporting macOS.) Plus we can really only vouch for the behaviour of `_LIBCUDACXX_STRING_H_HAS_CONST_OVERLOADS` on the exact version of CCCL that we download ourselves. On any other version all bets are off with how it would work on macOS. :thinking:

The fix in `clustering_config.hpp` is a much more straight forward affair on the other hand. :wink:

@beomki-yeo and @paulgessinger: I'm happy for one of you to set up a macOS CI for the repository. I was not keen on bothering with that myself. (We only need Boost on macOS that may not be available right away on the CI machines.)

Note: It might be worth noting that it's the following code that `_LIBCUDACXX_STRING_H_HAS_CONST_OVERLOADS` affects:

https://github.com/NVIDIA/cccl/blob/v2.5.0/libcudacxx/include/cuda/std/detail/libcxx/include/string.h#L73

Deep, **deep** inside of the CCCL code...

### DIFF
--- a/extern/cccl/CMakeLists.txt
+++ b/extern/cccl/CMakeLists.txt
@@ -44,3 +44,24 @@ set ( CUB_ENABLE_HEADER_TESTING OFF CACHE BOOL "Disable CUB header testing" )
 
 # Get it into the current directory.
 FetchContent_MakeAvailable( CCCL )
+
+# Check if <thrust/pair.h> works out of the box.
+include( CheckCXXSourceCompiles )
+set( CMAKE_REQUIRED_INCLUDES
+   "${CCCL_SOURCE_DIR}/thrust"
+   "${CCCL_SOURCE_DIR}/libcudacxx/include" )
+set( _THRUST_TEST_SOURCE "#include <thrust/pair.h>\nint main() { return 0; }" )
+check_cxx_source_compiles( "${_THRUST_TEST_SOURCE}" TRACCC_THRUST_WORKS )
+# If not, check if the _LIBCUDACXX_STRING_H_HAS_CONST_OVERLOADS flag makes it
+# work.
+if( NOT TRACCC_THRUST_WORKS )
+   set( CMAKE_REQUIRED_DEFINITIONS -D_LIBCUDACXX_STRING_H_HAS_CONST_OVERLOADS )
+   check_cxx_source_compiles( "${_THRUST_TEST_SOURCE}"
+      TRACCC_THRUST_WORKS_WITH_PATCH )
+   if( TRACCC_THRUST_WORKS_WITH_PATCH )
+      target_compile_definitions( _Thrust_Thrust
+         INTERFACE _LIBCUDACXX_STRING_H_HAS_CONST_OVERLOADS )
+   else()
+      message( WARNING "Thrust does not seem to work. The build may fail." )
+   endif()
+endif()
--- a/core/include/traccc/clusterization/clustering_config.hpp
+++ b/core/include/traccc/clusterization/clustering_config.hpp
@@ -8,7 +8,7 @@
 
 #pragma once
 
-#include <cstdint>
+#include <cstddef>
 
 #include "traccc/definitions/qualifiers.hpp"
 


paulgessinger (approved):
Pretty hacky but I guess it gets the job done.

extern/cccl/CMakeLists.txt
```diff
@@ -44,3 +44,24 @@ set ( CUB_ENABLE_HEADER_TESTING OFF CACHE BOOL "Disable CUB header testing" )
...
+# Check if <thrust/pair.h> works out of the box.
+include( CheckCXXSourceCompiles )
+set( CMAKE_REQUIRED_INCLUDES
+   "${CCCL_SOURCE_DIR}/thrust"
+   "${CCCL_SOURCE_DIR}/libcudacxx/include" )
+set( _THRUST_TEST_SOURCE "#include <thrust/pair.h>\nint main() { return 0; }" )
+check_cxx_source_compiles( "${_THRUST_TEST_SOURCE}" TRACCC_THRUST_WORKS )
+# If not, check if the _LIBCUDACXX_STRING_H_HAS_CONST_OVERLOADS flag makes it
+# work.
+if( NOT THRUST_PAIR_WORKS )
```
krasznaa, line 57: Dang... Some issues were left behind from the copilot-recommended code... :frowning:

krasznaa:
Curiously, while a "Makefile based" build works fine for me (latest macOS + Xcode) with these changes

Screenshot 2024-08-14 at 4 34 35 PM

, I can not generate an Xcode project for the code as it seems.

```
...
-- Building Acts as part of the TRACCC project
-- Are C++20 concepts supported
-- Are C++20 concepts supported - no
-- Found Boost: /Users/krasznaa/Software/Boost/1.83.0/lib/cmake/Boost-1.83.0/BoostConfig.cmake (found suitable version "1.83.0", minimum required is "1.71.0")  
-- Looking for C++ include filesystem
-- Looking for C++ include filesystem - found
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED - Failed
-- Performing Test CXX_FILESYSTEM_STDCPPFS_NEEDED
-- Performing Test CXX_FILESYSTEM_STDCPPFS_NEEDED - Failed
-- Performing Test CXX_FILESYSTEM_CPPFS_NEEDED
-- Performing Test CXX_FILESYSTEM_CPPFS_NEEDED - Failed
CMake Error at /Users/krasznaa/ATLAS/projects/traccc/build-xcode/_deps/acts-src/cmake/FindFilesystem.cmake:248 (message):
  Cannot run simple program using std::filesystem
  /Users/krasznaa/ATLAS/projects/traccc/build-xcode/_deps/acts-src/CMakeLists.txt:241 (find_package)


-- Configuring incomplete, errors occurred!
See also "/Users/krasznaa/ATLAS/projects/traccc/build-xcode/CMakeFiles/CMakeOutput.log".
See also "/Users/krasznaa/ATLAS/projects/traccc/build-xcode/CMakeFiles/CMakeError.log".
```

Is this tested in the Acts CI? That `-G Xcode` would work...?

paulgessinger:
> Is this tested in the Acts CI? That `-G Xcode` would work...?

No, we don't support Xcode on macOS at all.

krasznaa:
> No, we don't support Xcode on macOS at all.

Which is just as well. Since it seems to have some nasty issues. :frowning: As you can see in the log, before failing on `std::filesystem`, it also cannot make use of C\+\+20 concepts.

But I'm not willing to debug that any further.

PR #678, opened 2024-08-15
Project Version Update, main branch (2024.08.15.)

krasznaa:
Update project to version `0.15.0`, in preparation for the next tag.

@paulgessinger, @beomki-yeo, let's keep attention to this. I don't like it when a tagged release advertises itself as a different version after being built. :thinking:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 
 # Set up the project.
 cmake_minimum_required( VERSION 3.22 )
-project( traccc VERSION 0.13.0 LANGUAGES CXX )
+project( traccc VERSION 0.15.0 LANGUAGES CXX )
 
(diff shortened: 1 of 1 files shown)


beomki-yeo (approved):
I agreee with you but this bothers me a lot at the same time. I wish the version number is updated automatically...

krasznaa:
This setup (where we pre-setup the version number) is the least work-prone procedure in my mind. :thinking:

Indeed, projects often have a "release procedure" that involves setting the project version as part of the procedure. Some of that could even be automated I guess. But as far as I know, none of that comes for free. Even in the best of cases we'd need to write some helper scripts that could be used during the release making procedure.

Long story short, I think this is the least painful of all the options for this project. I'm open to suggestions though...

PR #679, opened 2024-08-15
Add the workflow for continuous benchmark

beomki-yeo:
Follow-up PR after #674. (Thanks to @paulgessinger for advices)

This will run the benchmark code for every merged PR through [traccc-benchmark](https://gitlab.cern.ch/acts/traccc-benchmark) and show the event throughput plots at README

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,17 @@
+name: Benchmarks
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


krasznaa:
Paul will have to give the okay for the implementation details here. But generally, I'm absolutely on board with this. :wink:

paulgessinger (approved):

PR #682, opened 2024-08-16
Add `-Wold-style-cast` and fix related issues

paulgessinger:
This PR enables `-Wold-style-cast`. TBB unfortunately has lots of these, and some of the examples binaries include TBB headers. To solve this, the builtin TBB target is modified to include directories with `-isystem`.


cmake/traccc-functions.cmake
```diff
@@ -146,3 +146,35 @@ function( traccc_add_flag name value )
...
+    else()
+      message("Warning: ${lib} doesn't set INTERFACE_INCLUDE_DIRECTORIES. No include_directories set.")
+    endif()
+    if(scope)
+      target_link_libraries(${target} ${scope} ${lib})
+    else()
+      target_link_libraries(${target} ${lib})
+    endif()
+  endforeach()
+endfunction(acts_target_link_libraries_system)
```
krasznaa, line 180: This is an interesting way of doing it, but I'd do it in a different way... :thinking:

For a number of externals we already had to do a bit of massaging to have them appear as system headers in our build. Even though we build them as part of this project. See:
  - https://github.com/acts-project/traccc/blob/main/extern/eigen3/CMakeLists.txt#L48-L53
  - https://github.com/acts-project/algebra-plugins/blob/main/extern/vc/CMakeLists.txt#L48-L50
  - etc.

I'm not actually sure why we don't use the more compact formalism from algebra-plugins for Eigen as well. :confused: Probably that's my bad...

Point being, with that setup the "main CMake code" should not have to do anything fancy. When we take an external "from the system" with `find_package(...)`, its headers will be treated as system headers with older CMake versions as well. And when we build the external ourselves, this massaging still allows the CMake code of the main libraries to be none the wiser. :thinking:

krasznaa (changes requested):
With my recommendation you'd also need to change fewer files. :wink:

cmake/traccc-functions.cmake
paulgessinger: I don't think it's a good idea at all to modify targets from `FetchContent` at all, this blurs the line between what the child project does and what the parent project does. 

Adding it as an explicit argument when linking against the library in question is cleaner and conveys + documents what's going on.

paulgessinger: Anyway, not a hill I'm willing to die on, so I'll change it.

krasznaa (approved):
For better or worse, as long as this hides all the warnings from the TBB code, this is the setup that I prefer. (Since this modification of the `tbb` library target per definition doesn't change anything about the compilation commands of the TBB library itself.)

paulgessinger:
Ok, so this fails in the SYCL build, because in that case we take TBB from the global installation in `/opt/intel`, so bypass the `SYSTEM` fix I added, and therefore compile without `-isystem`.

```
In file included from /__w/traccc/traccc/examples/run/cpu/throughput_mt.cpp:9:
In file included from /__w/traccc/traccc/examples/run/cpu/../common/throughput_mt.hpp:37:
In file included from /__w/traccc/traccc/examples/run/cpu/../common/throughput_mt.ipp:40:
In file included from /opt/intel/oneapi/tbb/2021.13/env/../include/tbb/global_control.h:17:
/opt/intel/oneapi/tbb/2021.13/env/../include/tbb/../oneapi/tbb/global_control.h:92:48: error: use of old-style cast [-Werror,-Wold-style-cast]
   92 |         return r1::global_control_active_value((int)p);
      |                                                ^    ~
1 error generated.
gmake[2]: *** [examples/run/cpu/CMakeFiles/traccc_throughput_mt.dir/build.make:76: examples/run/cpu/CMakeFiles/traccc_throughput_mt.dir/throughput_mt.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:4455: examples/run/cpu/CMakeFiles/traccc_throughput_mt.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs...
```

Any ideas @krasznaa?

krasznaa:
This is a pretty evil error. :frowning:

Since we now give the local TBB headers as system headers to the oneAPI compiler, it decides that heck, it has "better system headers" for TBB than what we have. And since oneAPI's headers are now included "implicitly", the whole `-isystem` thing doesn't mean anything for it. :frowning: Since we only declare the directory inside the build directory with `-isystem`.

Let me think a bit. :thinking: I may need to call on some Intel developers about this in the end...

stephenswat:
What I'd try is this (not sure if I can suggest patches for unchanged files). In `.github/ci_setup.sh` change:

```bash
# Set up the correct environment for the SYCL tests.
if [ "${PLATFORM_NAME}" = "SYCL" ]; then
   if [ -f "/opt/intel/oneapi/setvars.sh" ]; then
      source /opt/intel/oneapi/setvars.sh --include-intel-llvm
   fi
fi
```

to

```bash
# Set up the correct environment for the SYCL tests.
if [ "${PLATFORM_NAME}" = "SYCL" ]; then
   if [ -f "/opt/intel/oneapi/setvars.sh" ]; then
      source /opt/intel/oneapi/setvars.sh --include-intel-llvm
      export CPATH=$(echo $CPATH | sed -e 's/\(^\|:\)[^:]*oneapi\/tbb[^:]*\($\|:\)/:/')
   fi
fi
```

That should prevent Intel's code from explicitly picking up its built-in headers.

stephenswat:
I thought about this, but I'm not sure how important `dev-utilities` is... :thinking:

krasznaa:
I think we're okay without it.

```
[bash][celeborn]:~ > ls -l /home/krasznaa/software/intel/oneapi-2024.2.1/dev-utilities/2024.2/include
total 12
-rw-r--r-- 1 krasznaa krasznaa 1102 May 24 13:04 dpc_common.hpp
-rw-r--r-- 1 krasznaa krasznaa  581 May 24 13:04 dpc_common_README.md
drwxr-xr-x 2 krasznaa krasznaa 4096 Aug  8 17:02 stb
[bash][celeborn]:~ > more /home/krasznaa/software/intel/oneapi-2024.2.1/dev-utilities/2024.2/include/dpc_common_README.md 
# oneAPI-samples/Common/


## `dpc_common.hpp`

Many oneAPI code samples share common include files.  These include files are installed locally with the product installation and 
can be located at `%ONEAPI_ROOT%\dev-utilities\latest\include` on your development system. These same include files are provided h
ere for your reference.

| Common files/directory            | Description
|:---                               |:---
| dpc_common.hpp                    | Common macros and functions for oneAPI samples
| STB                               | Image rendering library
[bash][celeborn]:~ >
```

Also: https://github-wiki-see.page/m/oneapi-src/oneAPI-samples/wiki/dpc_common.hpp

stephenswat:
Right, then I'd do something like this:

```bash
# Set up the correct environment for the SYCL tests.
if [ "${PLATFORM_NAME}" = "SYCL" ]; then
   if [ -f "/opt/intel/oneapi/setvars.sh" ]; then
      __OLD_CPATH=${CPATH}
      source /opt/intel/oneapi/setvars.sh --include-intel-llvm
      export CPATH=${__OLD_CPATH}
   fi
fi
```

stephenswat (approved):
Unfortunate that we had to add this semi-ugly CI fix, but that's Intel's fault not ours.

PR #684, opened 2024-08-16
Makee track param estimation more accurate

beomki-yeo:
Current track parameter estimation is based on approximation and this PR removes such approximation.
I believe the gain from making it more accurate outweights the loss from slightly longer computation time

Following https://github.com/acts-project/acts/pull/2982

### DIFF
--- a/core/include/traccc/seeding/track_params_estimation_helper.hpp
+++ b/core/include/traccc/seeding/track_params_estimation_helper.hpp
@@ -82,20 +82,16 @@ inline TRACCC_HOST_DEVICE bound_vector seed_to_bound_vector(
     scalar A = (uv2[1] - uv1[1]) / (uv2[0] - uv1[0]);
     scalar B = uv2[1] - A * uv2[0];
 
-    // Curvature (with a sign) estimate
-    scalar rho = -2.0f * B / getter::perp(vector2{1., A});
-    // The projection of the top space point on the transverse plane of
-    // the new frame
-    scalar rn = local2[0] * local2[0] + local2[1] * local2[1];
+    // Radius (with a sign)
+    scalar R = -getter::perp(vector2{1.f, A}) / (2.f * B);
     // The (1/tanTheta) of momentum in the new frame
-    static constexpr scalar G = static_cast<scalar>(1.f / 24.f);
     scalar invTanTheta =
-        local2[2] * std::sqrt(1.f / rn) / (1.f + G * rho * rho * rn);
+        local2[2] / (2.f * R * math::asin(getter::perp(local2) / (2.f * R)));
 
     // The momentum direction in the new frame (the center of the circle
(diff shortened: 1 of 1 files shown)


core/include/traccc/seeding/track_params_estimation_helper.hpp
```diff
@@ -82,15 +82,11 @@ inline TRACCC_HOST_DEVICE bound_vector seed_to_bound_vector(
     scalar A = (uv2[1] - uv1[1]) / (uv2[0] - uv1[0]);
     scalar B = uv2[1] - A * uv2[0];
 
-    // Curvature (with a sign) estimate
-    scalar rho = -2.0f * B / getter::perp(vector2{1., A});
-    // The projection of the top space point on the transverse plane of
-    // the new frame
-    scalar rn = local2[0] * local2[0] + local2[1] * local2[1];
+    // Radius (with a sign)
+    scalar R = -getter::perp(vector2{1., A}) / (2.f * B);
```
krasznaa, line 86: This is a bit inconsistent with the numeric literals. Both should be declared with `.f`. The `vector2` type then upconverts them to `double` if needed. :thinking:

krasznaa (changes requested):
This code is a miniscule part of our execution time. If this change provides an improvement for the code's physics performance, I'm okay with a few more floating-point calculations. :+1:

core/include/traccc/seeding/track_params_estimation_helper.hpp
beomki-yeo: Right. Fixed now

krasznaa (approved):
:+1:

core/include/traccc/seeding/track_params_estimation_helper.hpp
niermann999: Generally, smth like ```static_cast<double>(0.1f)``` vs ```0.1``` does not result in the same accuracy though, or does it? The representation error on the former is kept even if upcast into double precision later, while the latter represents the value actually in double precision. In this particular example for ```1.f``` and ```2.f``` it does not really matter though...

krasznaa: I was indeed thinking about proposing something with `static_cast`. But
  - as you say, for `1.0` and `2.0` this should all not really matter;
  - let's avoid unnecessary code where possible;
  - at the same time I wanted to avoid float conversion warnings popping up in the future.

I believe the current code that Beomki is proposing, should be pretty safe against warnings. :thinking:

niermann999: Yes, correct! The current code is fine and in general, I think, it is better for us to have some single precision literals in a double precision build, instead of having double precision operations sneak into a single precision build. I just wanted to make sure, that we are aware of the implications when switching configs/units to single precision permanently, since this recently came up again :)

PR #685, opened 2024-08-20
Update detray version to v0.73.0

beomki-yeo:
As of detray v0.73.0, we need to handle the particle hypothesis appropriately. 
Free covariance matrix is also removed as it turned out to be unnecessary.

It is not relevant with the detray version update, but the PR make the test for the track parameter estimation cover the positive charge as well.

### DIFF
--- a/core/include/traccc/finding/finding_config.hpp
+++ b/core/include/traccc/finding/finding_config.hpp
@@ -7,7 +7,11 @@
 
 #pragma once
 
+// traccc include(s).
+#include "traccc/definitions/primitives.hpp"
+
 // detray include(s).
+#include "detray/definitions/pdg_particle.hpp"
 #include "detray/definitions/units.hpp"
 #include "detray/propagator/propagation_config.hpp"
 
@@ -40,6 +44,10 @@ struct finding_config {
 
     /// Propagation configuration
     detray::propagation::config propagation{};
+
+    /// Particle hypothesis
+    detray::pdg_particle<traccc::scalar> ptc_hypothesis =
+        detray::muon<traccc::scalar>();
 
     /****************************
      *  GPU-specfic parameters
--- a/core/include/traccc/utils/particle.hpp
+++ b/core/include/traccc/utils/particle.hpp
@@ -1,0 +1,102 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// traccc include(s).
+#include "traccc/definitions/primitives.hpp"
+#include "traccc/definitions/qualifiers.hpp"
+#include "traccc/edm/track_parameters.hpp"
+
+// detray include(s).
+#include "detray/definitions/pdg_particle.hpp"
+
+// System include(s).
+#include <stdexcept>
+
+namespace traccc {
+
+namespace detail {
+
+template <typename scalar_t>
+TRACCC_HOST_DEVICE inline detray::pdg_particle<scalar_t>
+particle_from_pdg_number(const int pdg_num) {
+
+    switch (pdg_num) {
+        case 11:
+            return detray::electron<scalar_t>();
+        case -11:
+            return detray::positron<scalar_t>();
+        case 13:
+            return detray::muon<scalar_t>();
+        case -13:
+            return detray::antimuon<scalar_t>();
+        case 211:
+            return detray::pion_plus<scalar_t>();
+        case -211:
+            return detray::pion_minus<scalar_t>();
+    }
+
+    return detray::muon<scalar_t>();
+}
+
+// Apply the charge operator to return the antimatter
+template <typename scalar_t>
+TRACCC_HOST_DEVICE inline detray::pdg_particle<scalar_t> charge_conjugation(
+    const detray::pdg_particle<scalar_t>& ptc) {
+
+    const auto pdg_num = ptc.pdg_num();
+
+    switch (pdg_num) {
+        case 11:
+            return detray::positron<scalar_t>();
+        case -11:
+            return detray::electron<scalar_t>();
+        case 13:
+            return detray::antimuon<scalar_t>();
+        case -13:
+            return detray::muon<scalar_t>();
+        case 211:
+            return detray::pion_minus<scalar_t>();
+        case -211:
+            return detray::pion_plus<scalar_t>();
+    }
+
+    return detray::muon<scalar_t>();
+}
+
+// Return the consistent particle type based on the particle hypothesis and the
+// charge of the track parameters
+template <typename scalar_t>
+TRACCC_HOST_DEVICE inline detray::pdg_particle<scalar_t>
+correct_particle_hypothesis(
+    const detray::pdg_particle<scalar_t>& ptc_hypothesis,
+    const bound_track_parameters& params) {
+
+    if (ptc_hypothesis.charge() * params.qop() > 0.f) {
+        return ptc_hypothesis;
+    } else {
+        return charge_conjugation(ptc_hypothesis);
+    }
+}
+
+template <typename scalar_t>
+TRACCC_HOST_DEVICE inline detray::pdg_particle<scalar_t>
+correct_particle_hypothesis(
+    const detray::pdg_particle<scalar_t>& ptc_hypothesis,
+    const free_track_parameters& params) {
+
+    if (ptc_hypothesis.charge() * params.qop() > 0.f) {
+        return ptc_hypothesis;
+    } else {
+        return charge_conjugation(ptc_hypothesis);
+    }
+}
+
+}  // namespace detail
+
+}  // namespace traccc
--- a/device/cuda/src/finding/finding_algorithm.cu
+++ b/device/cuda/src/finding/finding_algorithm.cu
@@ -65,12 +65,14 @@ __global__ void make_barcode_sequence(
 /// CUDA kernel for running @c traccc::device::apply_interaction
 template <typename detector_t>
 __global__ void apply_interaction(
-    typename detector_t::view_type det_data, const int n_params,
+    typename detector_t::view_type det_data, const finding_config cfg,
+    const int n_params,
     bound_track_parameters_collection_types::view params_view) {
 
     int gid = threadIdx.x + blockIdx.x * blockDim.x;
 
-    device::apply_interaction<detector_t>(gid, det_data, n_params, params_view);
+    device::apply_interaction<detector_t>(gid, cfg, det_data, n_params,
+                                          params_view);
 }
 
 /// CUDA kernel for running @c traccc::device::count_measurements
@@ -351,7 +353,7 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
         nThreads = m_warp_size * 2;
         nBlocks = (n_in_params + nThreads - 1) / nThreads;
         kernels::apply_interaction<detector_type>
-            <<<nBlocks, nThreads, 0, stream>>>(det_view, n_in_params,
+            <<<nBlocks, nThreads, 0, stream>>>(det_view, m_cfg, n_in_params,
                                                in_params_buffer);
         TRACCC_CUDA_ERROR_CHECK(cudaGetLastError());
 
--- a/examples/options/include/traccc/options/generation.hpp
+++ b/examples/options/include/traccc/options/generation.hpp
@@ -8,8 +8,12 @@
 #pragma once
 
 // Project include(s).
+#include "traccc/definitions/primitives.hpp"
 #include "traccc/options/details/interface.hpp"
 #include "traccc/options/details/value_array.hpp"
+
+// detray include(s).
+#include "detray/definitions/pdg_particle.hpp"
 
 // System include(s).
 #include <cstddef>
@@ -38,8 +42,8 @@ class generation : public interface {
     opts::value_array<float, 2> phi_range{-180.f, 180.f};
     /// Range of eta
     opts::value_array<float, 2> eta_range{-2.f, 2.f};
-    /// Charge of particles
-    float charge{-1.f};
+    /// PDG number for particle type (Default: muon)
+    int pdg_number = 13;
 
     /// @}
 
@@ -48,6 +52,9 @@ class generation : public interface {
 
     /// Range of theta [rad]
     opts::value_array<float, 2> theta_range{0.f, 0.f};
+    /// Particle type
+    detray::pdg_particle<traccc::scalar> ptc_type =
+        detray::muon<traccc::scalar>();
 
     /// @}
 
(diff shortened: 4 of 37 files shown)


core/include/traccc/finding/finding_config.hpp
```diff
@@ -41,6 +45,10 @@ struct finding_config {
     /// Propagation configuration
     detray::propagation::config propagation{};
 
+    /// Particle hypothesis
+    detray::pdg_particle<traccc::scalar> ptc_hypothesis =
+        detray::muon<traccc::scalar>();
```
krasznaa, line 50: Should the default hypothesis be a muon for the track finding? Would it not be better to have it as a pion? :thinking:

krasznaa (commented):
No fundamental issues of course, just some technical questions.

Would we not need to teach `traccc::options` about the extension in `traccc::finding_config`? (And even the one in the simulation config I suppose.) We'd like to be able to set the new particle hypothesis value from the command line as well I guess. :thinking:

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -65,12 +65,14 @@ __global__ void make_barcode_sequence(
 /// CUDA kernel for running @c traccc::device::apply_interaction
 template <typename detector_t>
 __global__ void apply_interaction(
-    typename detector_t::view_type det_data, const int n_params,
+    typename detector_t::view_type det_data, const finding_config& cfg,
```
niermann999, line 68: Can you pass the config as a reference to a kernel?

krasznaa: In fact that's a very good catch!

```suggestion
    typename detector_t::view_type det_data, const finding_config cfg,
```

beomki-yeo: you mean pass the config as a value?

niermann999: Yes, you need to get the data in the ```finding_config``` to the device somehow, right?

krasznaa: :confused: I gave a code change suggestion.

You cannot pass a const reference to a kernel. That's exactly the same as passing a const pointer. Which is pointing at an object that sits on the stack. **On the host.** You need to pass config objects like this by value to the kernel.

From the kernel then you can pass objects forward using references. It's only on the host<->device boundary that you need to be careful.

beomki-yeo: Bro I said that I am ignorant of this.. You can't just say "I gave a code to you".. 

Thanks for the explanation though

niermann999: Ah, sorry, I thought you just overlooked this and that it would be clear if I highlight the line, my bad

beomki-yeo: ? what are you apologizing for lol

beomki-yeo: It is fixed now

beomki-yeo:
> Would we not need to teach traccc::options about the extension in traccc::finding_config? (And even the one in the simulation config I suppose.) We'd like to be able to set the new particle hypothesis value from the command line as well I guess

This is done now. As there was no function for casting particle type from double to float, I just used traccc::scalar everywhere but this can be fixed in the future

beomki-yeo:
One issue appeared was that the particle hypothesis may have different charge with the reconstructed track. For example, the track fitting starts with muon hypothesis but there can be tracks with positive charge. In that case, the particle hypothesis is changed to its anti-matter.

core/include/traccc/utils/particle.hpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// Apply the charge operator to return the antimatter
+template <typename scalar_t>
+TRACCC_HOST_DEVICE inline detray::pdg_particle<scalar_t> charge_operator(
+    const detray::pdg_particle<scalar_t>& ptc) {
+
+    const auto pdg_num = ptc.pdg_num();
+
+    if (pdg_num == 11) {
```
niermann999, line 53: Here is a stupid question: Apart from the nicer naming, can't we just multiply the charge with the sign of the pdg number and avoid the branching?

core/include/traccc/utils/particle.hpp
```diff
@@ -0,0 +1,98 @@
...
+    } else if (pdg_num == -211) {
+        return detray::pion_minus<scalar_t>();
+    }
+
+    return detray::muon<scalar_t>();
+}
+
+// Apply the charge operator to return the antimatter
+template <typename scalar_t>
+TRACCC_HOST_DEVICE inline detray::pdg_particle<scalar_t> charge_operator(
```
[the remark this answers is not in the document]
beomki-yeo: ok will fixt it

core/include/traccc/utils/particle.hpp
```diff
@@ -0,0 +1,98 @@
...
+        return detray::pion_plus<scalar_t>();
+    }
+
+    return detray::muon<scalar_t>();
+}
+
+// Return the consistent particle type based on the particle hypothesis and the
+// charge of the track parameters
+template <typename scalar_t>
+TRACCC_HOST_DEVICE inline detray::pdg_particle<scalar_t> consistent_particle(
```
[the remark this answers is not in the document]
beomki-yeo: ok makes sense

core/include/traccc/utils/particle.hpp
beomki-yeo: Isn't it rather problematic for majorana particle like gamma

core/include/traccc/utils/particle.hpp
```diff
@@ -0,0 +1,98 @@
...
+    const detray::pdg_particle<scalar_t>& ptc_hypothesis,
+    const bound_track_parameters& params) {
+
+    if (ptc_hypothesis.charge() * params.qop() > 0.f) {
```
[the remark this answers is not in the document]
beomki-yeo: This PR does not setup the hypothesis for the individual tracks but it is event-wise.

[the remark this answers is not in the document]
beomki-yeo: > Would we not actually need to monitor updates to the curvature of the track with regard to the B-field or am I confused?

I don't get this question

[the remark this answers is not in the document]
niermann999: Ah, Ok, because it seemed to call ```consistent_particle``` for every new step of the CKF and the material interaction, but I probably just don't know the code well enough

core/include/traccc/utils/particle.hpp
niermann999: Aren't those positive sign numbers? So nothing would change?

niermann999: Also can the charge in out core right now actually be zero? Would that not kill ```qop```?

beomki-yeo: Can you just explain with an example. multiplying the charge with the sign of pdg number sounds confusing to me.

core/include/traccc/utils/particle.hpp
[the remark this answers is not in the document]
beomki-yeo: If we don't like this, we may have to put the particle hypothesis into the bound track parameters, but lets not do that.

core/include/traccc/utils/particle.hpp
niermann999: Ah, never mind, on the pion the charge already does not correspond to the pdg sign anymore. But can't we then check for neutral particles (which I am not sure the code can handle anyway?) an otherwise just flip the charge?

beomki-yeo: If it just stays as electron with positve charge, it will go through electron bremss instead of positron bremss which does not make sense to me

niermann999: Ah, Ok, makes sense. But I thought on a code level ```electron``` and ```positron``` are not different types, they just have the sign of the pdg number and charge flipped, or is this where I am missing the point? So, I was hoping we could get rid of all this branching here and just flip the respective signs? But maybe not

niermann999: I don't know, e.g.:
```
if (ptc.charge() != 0) {
  return detray::pdg_particle<scalar_t>{-ptc.pdg_num(), ptc.mass(), -ptc.charge()};
} else {
  return ptc;
}
```
but then again, I probably just don't see the full picture here

beomki-yeo: This makes sense but are you sure that there is not any possibility that multiplying negative sign will cause a rounding error? I need to make sure that `(detray::pdg_particle(-electron.pdg_num(), electron.mass(), -electron.charge() == positron())`

niermann999: I was under the impression that this should just flip the signbit

beomki-yeo: OK there is not problem in your suggestion. But let me keep this at the moment. One single mistake in detray pdg definition can make things go south. Probably I will migrate this function to detray and  improve the detray test to make sure everything is OK.

Also don't worry about the branching. As mentioned earlier, the hypothesis is algorithm-wide and it is not likely tracks will be set to totally different types of particle

niermann999: Well, if we are calling this for every CKF step/material interaction etc, the code still has to do some branch checking on the GPU, especially if the default (muon) is not the first hit.

If you are worried just about the equality operator failing, is it not enough to compare the pdg numbers (the material interaction seems to be doing just that actually). These are integers, no rounding will happen there

beomki-yeo: checking if integer is 11 or 13 does not take long time.. this is something that we need to seriously worry about

beomki-yeo:
Any approval?

core/include/traccc/utils/particle.hpp
niermann999: No, the check does not take time and for the CPU this will not be a problem. But even in the examples where we randomize the charge, the warp could very well hit different branches in this code, or not? We need to see at some point what this compiles into, I guess

niermann999 (approved):

core/include/traccc/utils/particle.hpp
beomki-yeo: Yeah but that will account for 0.000001 % performance degradation of entire chain probably.

Let me come back to this issue after I make a safe guard in detray OK?

niermann999: OK, sure! Also, if somebody has a particle hypothesis that is not part of this list, say gamma which you mentioned earlier, will their particle just be switched to a muon silently?

examples/options/include/traccc/options/generation.hpp
```diff
@@ -38,8 +42,8 @@ class generation : public interface {
     opts::value_array<float, 2> phi_range{-180.f, 180.f};
     /// Range of eta
     opts::value_array<float, 2> eta_range{-2.f, 2.f};
-    /// Charge of particles
-    float charge{-1.f};
+    /// PDG number for particle type (Default: muon)
+    unsigned int pdg_number = 13;
```
krasznaa, line 46: Why would this be `unsigned int`? The PDG number is signed no?

krasznaa (changes requested):
Joana is right to worry about all the "if statements". 🤔 We'll have to see exactly what this does to the GPU code, but even when most of the threads will take one path, having any new branching point cannot be good for the code's performance. The question is only how bad it is. Not whether it's bad.

That being said, if we need to live with this, then we do. Still, any "trick" that allows us to perform an operation without an `if` or `switch` statement, is a good trick. 😉

beomki-yeo:
I think @krasznaa is misunderstanding sth. The branching only happens inside the function. It is not like it will create entire new branching for the next stages.

core/include/traccc/utils/particle.hpp
```diff
@@ -0,0 +1,100 @@
...
+// System include(s).
+#include <stdexcept>
+
+namespace traccc {
+
+namespace detail {
+
+template <typename scalar_t>
+TRACCC_HOST_DEVICE inline detray::pdg_particle<scalar_t>
+particle_from_pdg_number(const std::int32_t pdg_num) {
```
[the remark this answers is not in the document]
beomki-yeo: That was what with the detray definition. particle type is used in io - see the `io/particle.hpp`
I will change the `int` of particle type to `std::int32_t`

examples/options/include/traccc/options/generation.hpp
beomki-yeo: Good catch. will fix

beomki-yeo:
I have never said that it is free. I just said it is going to be "tiny".
Welp that's why I added the benchmark monitoring. So don't worry too much

niermann999:
> I have never said that it is free. I just said it is going to be "tiny". Welp that's why I added the benchmark monitoring. So don't worry too much

Yeah, its probably benign. It just seemed avoidable

core/include/traccc/utils/particle.hpp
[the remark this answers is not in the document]
krasznaa: This is definitely a subjective comment here. But I/O from/to a text file does not care about the exact bit-width of the type. So `int` is just fine when doing CSV file I/O. The advantage of `std::int32_t` comes if you'd want to write a binary file on Computer A, and then read it on Computer B. In case they define `int` differently (which they would be allowed to do), this could cause an issue. (The same when streaming the variable in a simple binary form over the network.) But text file I/O is basically never a concern with variable types... :thinking:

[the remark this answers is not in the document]
beomki-yeo: Hmm ok then I will just use int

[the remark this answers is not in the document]
beomki-yeo: Fixed now

core/include/traccc/utils/particle.hpp
```diff
@@ -0,0 +1,100 @@
...
+        return detray::electron<scalar_t>();
+    } else if (pdg_num == 13) {
+        return detray::antimuon<scalar_t>();
+    } else if (pdg_num == -13) {
+        return detray::muon<scalar_t>();
+    } else if (pdg_num == 211) {
+        return detray::pion_minus<scalar_t>();
+    } else if (pdg_num == -211) {
+        return detray::pion_plus<scalar_t>();
+    }
```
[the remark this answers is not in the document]
beomki-yeo: Changed to switch

core/include/traccc/utils/particle.hpp
```diff
@@ -0,0 +1,100 @@
...
+        return detray::positron<scalar_t>();
+    } else if (pdg_num == 13) {
+        return detray::muon<scalar_t>();
+    } else if (pdg_num == -13) {
+        return detray::antimuon<scalar_t>();
+    } else if (pdg_num == 211) {
+        return detray::pion_plus<scalar_t>();
+    } else if (pdg_num == -211) {
+        return detray::pion_minus<scalar_t>();
+    }
```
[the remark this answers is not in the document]
beomki-yeo: Fixed now

examples/options/include/traccc/options/generation.hpp
beomki-yeo: Fixed to int

krasznaa (approved):
:+1:

PR #686, opened 2024-08-20
Simplify ODD Strip Module Handling, main branch (2024.08.20.)

krasznaa:
Removed the `traccc::pixel_data::variance_y` variable. As discussed in https://github.com/acts-project/traccc/pull/627#discussion_r1649794653, there was no reason to keep this variable in memory. The values coming from the digitization files were the same as what we get with the formula implemented in the measurement creation code.

At the same time made the code for detecting 1D modules a lot more robust. Now the decision is made based on the number of "bins" in a module's segmentation.

I tested on our multi-muon ODD simulations that these changes don't affect the output of `traccc_seq_example`. (While also testing that for instance setting all modules to be 2D would have a visible effect.)

I wanted to put this in before making the same simplification in #627 as well. :thinking:


stephenswat (approved):

beomki-yeo (approved):
This also looks good to me

PR #692, opened 2024-08-31
Refactor the event map

beomki-yeo:
<details>

<summary>Outdated description</summary>

I will go through a massive cleanup in the near future, but let me mess it up one more time..
This PR also removes the annoying clusterization in the event map

### UPDATE

I compared the local positions of measurements from traccc CPU CCA against the local positions of `measurements.csv` (data directory: `odd/geant4_1muon_100GeV`). 
And I found that they have the same local positions (the below table shows the measurements of the first event). 
I thought the local positions of `measurements.csv` are the truth value, so there must be gap between them. 

<google-sheets-html-origin>
From CCA | From measurements.csv
-- | --
(-6.775,32.735) | (-6.775,32.735)
(-5.825,9.025) | (-5.825,9.025)
(-5.075,-10.7355) | (-5.075,-10.7355)
(-1.325,-28.7803) | (-1.325,-28.7803)
(3.825,2.49362) | (3.825,2.49362)
(6.475,24.1524) | (6.475,24.1524)
(-6.42745,25.8361) | (-6.42745,25.8361)
(-3.075,5.45458) | (-3.075,5.45458)
(-4.725,6.13291) | (-4.725,6.13291)
(7.875,7.82025) | (7.875,7.82025)
(-5.60696,5.13736) | (-5.60696,5.13736)
(0.125,-9.88765) | (0.125,-9.88766)
(4.13824,-30.5525) | (4.13824,-30.5525)
(6.21763,-5.97566) | (6.21763,-5.97566)
(-7.73042,-9.33519) | (-7.73042,-9.33519)
(1.825,-33.641) | (1.825,-33.641)
(2.31866,33.1868) | (2.31866,33.1868)
(2.625,25.8589) | (2.625,25.8589)
(-0.825,-4.29537) | (-0.825,-4.29537)
(-6.825,-34.2) | (-6.825,-34.2)
(-32.325,40.2) | (-32.325,40.2)
(-28.125,-29.4) | (-28.125,-29.4)
(-4.125,52.2) | (-4.125,52.2)
(-3.375,-7.8) | (-3.375,-7.8)
(17.475,-23.4) | (17.475,-23.4)
(-2.775,-57) | (-2.775,-57)
(3.6,-34.2) | (3.6,-34.2)
(12.225,28.8) | (12.225,28.8)
(10.2,-21) | (10.2,-21)
(-11.475,-25.8) | (-11.475,-25.8)
(-15.975,-49.8) | (-15.975,-49.8)
(0.450002,0.600002) | (0.450001,0.600002)
(-10.35,-25.8) | (-10.35,-25.8)
(6.675,-47.4) | (6.675,-47.4)
(7.575,17.4) | (7.575,17.4)
(-3.575,51.8) | (-3.575,51.8)
(-2.325,-31.2) | (-2.325,-31.2)
(-5.025,29.4) | (-5.025,29.4)
(17.25,0) | (17.25,0)
(15.25,0) | (15.25,0)
(14.75,0) | (14.75,0)
(12.75,0) | (12.75,0)
(-6.75,0) | (-6.75,0)
(-6.125,0) | (-6.125,0)
(34.125,0) | (34.125,0)
(34.125,0) | (34.125,0)
(42.875,0) | (42.875,0)
(43,0) | (43,0)
(-38.375,0) | (-38.375,0)
(46,0) | (46,0)
(42.625,0) | (42.625,0)
(-37.875,0) | (-37.875,0)
(-35.375,0) | (-35.375,0)

Perhaps, `measurements.csv` is not for storing the truth value but for storing CCA value in ACTS simulation? (But how does ACTS simulation perform the CCA?) Maybe I am just doing something stupid. @asalzburger @krasznaa  Do you have some ideas on this?

</details>

### PR Description

I decided to refactor the `event_map` along with the measurement creation validation as I found it makes things easier. Followings are the major changes:

1. `event_map` and `event_map` are removed and unified into `event_data` which can support both acts simulation data and detray simulation data.

2. Validation suites for the measurement creation are added. As expected, the qualities of the pull values of `d0` and `z0` are not great, thus, the conditions are quite loose.

3. csv geometry is deprecated as we are not supposed to use it anymore. Also the the ACTS cell data format has replaced `hit_id` column (csv tml) with `measurement_id` column (new format), and it is quite painful to deal with different data formats.

### Physics Performance of seq_example (double precision)

The pull value from the measurement creation is not great but can be tolerated for the moment.
The pull value from Kalman fit is definitely bad for unknown reasons - It will be followed up in the near future 

```
./bin/traccc_seq_example --detector-file=geometries/odd/odd-detray_geometry_detray.json --digitization-file=geometries/odd/odd-digi-geometric-config.json --grid-file=geometries/odd/odd-detray_surface_grids_detray.json --input-directory=odd/geant4_10muon_100GeV --check-performance --input-events=50
```
image
image
image

### Physics Performance of seeding_example (double precision)

```
./bin/traccc_seeding_example --detector-file=geometries/odd/odd-detray_geometry_detray.json --digitization-file=geometries/odd/odd-digi-geometric-config.json --grid-file=geometries/odd/odd-detray_surface_grids_detray.json --input-directory=odd/geant4_10muon_100GeV --check-performance --input-events=50
```

image
image


beomki-yeo:
Indeed [DigitizationAlgorithm](https://github.com/acts-project/acts/blob/main/Examples/Algorithms/Digitization/src/DigitizationAlgorithm.cpp) of ACTS write measurements.csv with the same measurement creation algorithm. I will use the truth hit information and do the coordinate transformation instead

stephenswat:
I believe that this is due to the fact that ACTS digitization is a rather simple heuristic that does not have any uncertainty, and that also ensures that the weighted centroid of the hits matches exactly with the measurement; i.e., the traccc CCA is the perfect inverse function of the ACTS digitization.

beomki-yeo:
> the traccc CCA is the perfect inverse function of the ACTS digitization.

Are you sure about this? The digitization seems more complicated than that :thinking:

beomki-yeo:
Maybe this one?
https://github.com/acts-project/acts/blob/main/Examples/Algorithms/Digitization/src/ModuleClusters.cpp

Not sure if this is an inversion of CCA.

beomki-yeo:
The PR won't be able to pass the test unless the EOS data is updated.
New data file is at: https://cernbox.cern.ch/s/Wyr4TKwFpENhO9H

beomki-yeo:
@krasznaa @stephenswat 

I removed the event maps and replace them with the `event_data`.
What it does is more or less the same.. hopefully more readable and makes sense.

Especially the measurement information is handled in a more sophisticated way - there are three types of them in `event_data`

1. `meas`: from `measurements.csv` of detray/acts simulation. It should be noted that they are not the truth measurements. For example, `measurements.csv` of acts simulation is a result of measurement_creation on the **truth clusters**, which means they have some uncertainties from the algorithm.
   
2. `truth_meas`: global-to-local transformation of `hits.csv`, so they are truth information

3. `found_meas`: Similar to `meas`, but results of measurement_creation on the **found clusters**. They are not filled during the construction of event_data. Users need to fill out them later after completing the CCA.

I have not checked the outputs of `--check_performance` as the current example executables are not stable but there will be follow-up PRs soon to fix this.

io/include/traccc/io/csv/cell.hpp
```diff
@@ -30,4 +33,14 @@ struct cell {
...
 
+inline traccc::cell make_cell_edm(const traccc::io::csv::cell& csv_cell) {
+
+    traccc::cell ce;
+    ce.channel0 = csv_cell.channel0;
+    ce.channel1 = csv_cell.channel1;
+    ce.activation = csv_cell.value;
+    ce.time = csv_cell.timestamp;
+    return ce;
+}
```
krasznaa, line 44: This should not be in this header file. :confused: You use this function only in the `traccc::io` library's implementation. So the function not only shouldn't be implemented `inline` in the header, it shouldn't even be declared in a public header. :thinking:

io/include/traccc/io/csv/measurement.hpp
```diff
@@ -38,4 +42,64 @@ struct measurement {
                    var_theta, var_time);
 };
 
+inline traccc::measurement make_measurement_edm(
+    const traccc::io::csv::measurement& csv_meas,
+    const silicon_detector_description::host* dd) {
```
krasznaa, line 47: Ditto.

io/include/traccc/io/csv/particle.hpp
```diff
@@ -35,4 +38,18 @@ struct particle {
...
+    point3 pos{csv_ptc.vx, csv_ptc.vy, csv_ptc.vz};
+    vector3 mom{csv_ptc.px, csv_ptc.py, csv_ptc.pz};
+
+    traccc::particle ptc{csv_ptc.particle_id, csv_ptc.particle_type,
+                         csv_ptc.process,     pos,
+                         csv_ptc.vt,          mom,
+                         csv_ptc.m,           csv_ptc.q};
+
+    return ptc;
+}
```
krasznaa, line 53: And here as well. I don't want any of these in public headers...

io/include/traccc/io/event_data.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+    track_candidate_container_types::host generate_truth_candidates(
+        seed_generator<detector_type>& sg, vecmem::memory_resource& resource);
+
+    // IO EDM containers
+    std::vector<io::csv::cell> m_cells;
+    std::vector<io::csv::hit> m_hits;
+    std::vector<io::csv::measurement> m_measurements;
+    std::vector<io::csv::measurement_hit_id> m_meas_hit_ids;
+    std::vector<io::csv::particle> m_particles;
```
krasznaa, line 63: I don't like using the CSV types like this. :frowning: This way the code can literally only work with a CSV input. Could you not implement the logic with only the EDM types that sit in `traccc::core`?

Which of course also brings up the question of why this class is not rather sitting in `traccc::performance` instead of `traccc::io`. :thinking:

io/include/traccc/io/event_data.hpp
```diff
@@ -0,0 +1,96 @@
...
+    // Particle to its Measurements map
+    std::map<particle, std::vector<measurement>> m_ptc_to_meas_map;
+    // Measurement to its track parameter map
+    std::map<measurement, std::pair<point3, point3>> m_meas_to_param_map;
+    // Modules per event
+    cell_module_collection_types::host m_modules;
+    // cluster to truth measurement map
+    std::map<vecmem::vector<cell>, measurement> m_cluster_to_truth_meas_map;
+    // CCA measurement to cluster map
+    std::map<vecmem::vector<cell>, measurement> m_cluster_to_found_meas_map;
```
krasznaa, line 83: I am getting **very** worried about all of these different maps. :frowning: Especially ones keyed on vectors of cells. That seems like the most fragile thing ever. Where cell order changes could fully upset the logic.

Also, what about `m_modules`? That thing will be removed very-very soon. The reconstruction algorithms are already not using it. Why use it here? (Searching through the changes, I don't see `m_modules` used **anywhere**. Which is at least good, you should be able to completely remove it.)

In fact I'm pretty sure that some of these maps are not currently used by any client code. Please look that through, and only keep the ones that are actually needed. Because this interface looks very unmaintainable right now. :frowning:

examples/options/include/traccc/options/detector.hpp
```diff
@@ -30,7 +30,8 @@ class detector : public interface {
     std::string grid_file;
     /// Use detray::detector for the geometry handling
     bool use_detray_detector = false;
-
+    // Use detray geometry id (No ACTS geometry conversion)
+    bool use_detray_simulation = false;
```
krasznaa, line 34: :thinking: This is meant for the input data, no? That when we read simulation files produced using Detray, we shouldn't change the IDs associated with measurements during reading... right?

I would propose to then rather put it into `traccc::opts::input_data`. :thinking:

examples/options/src/detector.cpp
```diff
@@ -35,6 +35,9 @@ detector::detector() : interface("Detector Options") {
         "digitization-file",
         po::value(&digitization_file)->default_value(digitization_file),
         "Digitization file");
+    m_desc.add_options()("use-detray-simulation",
+                         po::bool_switch(&use_detray_simulation),
```
krasznaa, line 39: Don't forget the `->default_value(...)` part. :wink:

krasznaa (changes requested):
I like that you make use of the detector description object. :smile: Though I don't fully understand how... Could you explain a bit more how `use_detray_simulation` is meant to be used/understood exactly?

examples/options/include/traccc/options/detector.hpp
beomki-yeo: What you understand is correct. It is just for not changing the IDs during the reading. 

I will change the location as suggested

beomki-yeo: I changed its name to `use-acts-geom-source` and move it to `input_data.hpp`

examples/options/src/detector.cpp
beomki-yeo: OK. . It is supposed to be `true` considering that the acts geometry is going to be the standard, but I set it to `false` as `use-detray-detector` is set to `false` as well. We can change this once we fully deprecate the csv geometry in the near future

io/include/traccc/io/csv/particle.hpp
beomki-yeo: Made a header and source for `make_measurement_edm`

io/include/traccc/io/csv/measurement.hpp
beomki-yeo: Made a header and source for make_measurement_edm

io/include/traccc/io/csv/cell.hpp
beomki-yeo: Made a header and source for make_measurement_edm

beomki-yeo:
Based on #729 #733 #734 #735

io/include/traccc/io/event_data.hpp
beomki-yeo: OK. I removed the maps that are not being used in this PR.

io/include/traccc/io/read_measurements.hpp
```diff
@@ -28,12 +28,14 @@ namespace traccc::io {
...
-/// @param[in]  dd        The detector description to point the measurements at
+/// @param[in]  use_acts_geom_source  Use acts geometry source
+/// @param[in]  detector  detray detector
 /// @param[in]  format    The format of the measurement data files (to read)
 ///
 void read_measurements(measurement_collection_types::host& measurements,
                        std::size_t event, std::string_view directory,
-                       const silicon_detector_description::host* dd = nullptr,
+                       bool use_acts_geom_source = true,
+                       const traccc::default_detector::host* detector = nullptr,
```
krasznaa, line 38: 🤔 I think the move here is to replace `silicon_detector_description::host` with `default_detector::host`, and that's it. The additional boolean is completely unnecessary as far as I can see.
  - When it's set to `true`, there must be a valid `detector` pointer;
  - When it's set to `false`, we don't care about the `detector` pointer.

So we should just use the same logic that's there at the moment.
  - When `detector` is specified, it's used to re-map the geometry IDs;
  - When it's left as `nullptr`, no re-mapping is done.

io/include/traccc/io/read_particles.hpp
```diff
@@ -57,13 +57,15 @@ void read_particles(particle_collection_types::host &particles,
...
+/// @param[in]  detector  detray detector
 /// @param[in]  filename_postfix Postfix for the particle file name(s)
 ///
 void read_particles(particle_container_types::host &particles,
                     std::size_t event, std::string_view directory,
+                    bool use_acts_geom_source = true,
+                    const traccc::default_detector::host *detector = nullptr,
                     data_format format = data_format::csv,
-                    const silicon_detector_description::host *dd = nullptr,
                     std::string_view filename_postfix = "-particles_initial");
```
krasznaa, line 69: In all of these cases you should just replace the `silicon_detector_description` pointer with a `default_detector` one in place. This would reduce the number of places that you need to update. (All the ones that didn't touch the `dd` pointer before.)

io/src/csv/make_measurement_edm.cpp
```diff
@@ -0,0 +1,59 @@
...
+                    indices[meas.meas_dim++] = ipar;
+                }; break;
+                case e_bound_loc1: {
+                    meas.local[1] = csv_meas.local1;
+                    meas.variance[1] = csv_meas.var_local1;
+                    indices[meas.meas_dim++] = ipar;
+                }; break;
+            }
+        }
+    }
```
krasznaa, line 46: I don't love this. 🤔 Since we're literally just doing almost the same thing for `0` and `1`, I think it would be more readable to just spell it out explicitly.

```suggestion
    // Local key is a 8 bit char and first and last bit are dummy value. 2 -
    // 7th bits are for 6 bound track parameters.
    // Ex1) 0000010 or 2 -> meas dim = 1 and [loc0] active -> strip or wire
    // Ex2) 0000110 or 6 -> meas dim = 2 and [loc0, loc1] active -> pixel
    // Ex3) 0000100 or 4 -> meas dim = 1 and [loc1] active -> annulus
    if (csv_meas.local_key & 0x2) {
        meas.local[0] = csv_meas.local0;
        meas.variance[0] = csv_meas.var_local0;
        indices[meas.meas_dim++] = 0;
    }
    if(csv_meas.local_key & 0x4) {
        meas.local[1] = csv_meas.local1;
        meas.variance[1] = csv_meas.var_local1;
        indices[meas.meas_dim++] = 1;
    }
```

io/include/traccc/io/csv/make_measurement_edm.hpp
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+namespace traccc::io::csv {
+
+/// Make measurement EDM from csv measurement
+///
+/// @param[in] csv_meas input csv measurement
+/// @param[in] acts_to_detray_id Map for acts-to-detray geometry ID converision
+///
+traccc::measurement make_measurement_edm(
```
krasznaa, line 24: Does this need to be a public header? If this function is only used as part of this library, let's move this header under `src/`. Then I can be a lot more forgiving about using an obscure `std::map` argument in its interface. 🤔

io/include/traccc/io/csv/make_measurement_edm.hpp
```diff
@@ -0,0 +1,28 @@
...
+    const traccc::io::csv::measurement& csv_meas,
+    std::map<geometry_id, geometry_id>* acts_to_detray_id);
```
krasznaa, line 26: But even as a private header, this should still be a const pointer.

krasznaa (changes requested):
On one hand it would be nice to get all of this into `main` before CHEP, but I feel relatively strongly about some of the design points that I highlighted. So I'd still rather spend the time to set this up correctly, for our future selves.

io/include/traccc/io/read_particles.hpp
beomki-yeo: This is subjective thing but let's just go this way because I wanted to put detector at the front. It's already quite a work to change this back. Anyway the most of lines should also be changed to set  `use_acts_geom_source` to `false`

krasznaa: You misunderstood me. I'm not saying that the command line option should not be there. I'm saying that there is no boolean argument needed in this function.

When calling this function, one would rather do something like:

```c++
traccc::io::read_particles(..., (input_opts.use_acts_geom_source ? &detector : nullptr), ...);
```

Then, in cases where we fundamentally don't have access to a `detray::detector` object, we can just write `nullptr`, and be done with it. Without having to write both `false` and `nullptr`.

My point remains that **in the I/O functions** the two arguments that you introduced are fully coupled. The boolean argument serves no function.

beomki-yeo: Please check my above comment: https://github.com/acts-project/traccc/pull/692#discussion_r1801163133

The boolean argument is required for a good reason.

io/include/traccc/io/read_measurements.hpp
beomki-yeo: > One can still argue that we can try to check whether there is an ACTS geom source or not in the detray file

Actually this is not the case because  we should be able to run the detray simulation on arbitrary geometry file. Yeah.. so I think `use_acts_geom_source` is necessary in any case

io/include/traccc/io/read_particles.hpp
beomki-yeo: > we can just write nullptr, and be done with it. Without having to write both false and nullptr.

For detray simulation, the arguments are set to `false` and `detector`.  Just for clarification:

1. `false or true`, `nullptr` -> tml csv (which is going to be deprecated at some point)
2. `false`, `detector` -> detray simulation
3. `true`, `detector` -> acts simulation

So what is going to be removed actually is the detector argument. `use_acts_geom_source` should sit there always

krasznaa:
Beomki, see me at one point today, as we need to talk this out live.
  - The **application** needs to have an additional boolean flag;
  - The **I/O code** doesn't.

I'll explain in more detail face to face.

io/include/traccc/io/csv/make_measurement_edm.hpp
beomki-yeo: yes

io/src/csv/make_measurement_edm.cpp
beomki-yeo: Not sure what your suggestion does. This code has existed (copied from ACTS) and I just moved it under `make_measurement_edm`
Let's do this in the other PR.

io/include/traccc/io/event_data.hpp
beomki-yeo: event_data is moved to performance directory. The csv containers are removed
But I am still using `io::csv::cell` in many places instead of struggling with SoA data type (even with the help of proxy it was a nightmare for me), where I need your mercy..

beomki-yeo:
It is good to go from my side

io/include/traccc/io/read_particles.hpp
beomki-yeo: As discussed in person, Attila's suggestion is taken

io/include/traccc/io/read_measurements.hpp
beomki-yeo: Same here

beomki-yeo:
@krasznaa The IO input is fixed now.
I don't know why build CI remains unfinished but I think that can be ignored

performance/CMakeLists.txt
```diff
@@ -64,7 +66,7 @@ traccc_add_library( traccc_performance performance TYPE SHARED
    "include/traccc/performance/throughput.hpp"
    "src/performance/throughput.cpp" )
 target_link_libraries( traccc_performance
-   PUBLIC traccc::core traccc::io covfie::core detray::test_utils )
+   PUBLIC traccc::core traccc::io covfie::core detray::test_utils dfelibs::dfelibs)
```
krasznaa, line 69: 🤔 The problem is not here. You should not add `dfelibs::dfelibs` here.

The problem is rather that `traccc::io` exposes some CSV specific classes in its public interface. Yet it links against `dfelibs::dfelibs` privately. You should change that to a public dependency. Here:

https://github.com/acts-project/traccc/blob/main/io/CMakeLists.txt#L84-L85

After you change that file, this list of libraries should not be changed anymore.

krasznaa (changes requested):
I'm fine with the I/O code's general setup at this point. The main part of the code, the `event_data` class, I can also make my peace with. (The goal sanctifies the means...) Let's just fix up some of the smaller issues, and then finally push this in today.

tests/io/mock_data/event000000000-cells.csv
```diff
@@ -2,8 +2,6 @@ geometry_id,measurement_id,channel0,channel1,timestamp,value
 1224979236083738112,0,1,0,0,0.0041470062
 1224979236083738112,0,0,1,0,0.00306466641
 1224979236083738112,0,1,1,0,0.00868905429
-1224979236083738112,1,1,1,0,0.00886478275
-1224979236083738112,1,1,2,0,0.00580428448
```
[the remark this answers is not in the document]
krasznaa: While I'm pointing this out here, I don't actually indent to hold up the PR because of this change.

io/include/traccc/io/csv/measurement.hpp
```diff
@@ -7,6 +7,10 @@
 
 #pragma once
 
+// Project include(s).
+#include "traccc/edm/measurement.hpp"
+#include "traccc/geometry/silicon_detector_description.hpp"
```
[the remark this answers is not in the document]
beomki-yeo: Right..

performance/CMakeLists.txt
beomki-yeo: OK

tests/io/mock_data/event000000000-cells.csv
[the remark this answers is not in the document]
beomki-yeo: I rolled back the changes.

krasznaa (approved):
Let's get this thing finally in. 👍

PR #693, opened 2024-09-02
Add benchmark formatting to CI

stephenswat:
The formatting CI currently doesn't check the (relatively new) benchmark directory. This commit ensures that benchmarks are correctly formatted.


krasznaa (approved):
Surprising that this was the only update necessary. 🤔

PR #694, opened 2024-09-02
Fix integer comparison bugs in device CKF

stephenswat:
This commit fixes a few bugs in the device CKF to do with the integral representation of the previous step number. Firstly, the previous step of step 0 (e.g. some sort of sentinel value) was being inconsistently defined as either the maximum value of a signed int or an unsigned int, and this was stored in an integer, leading to unexpected behaviour. Furthermore, the track building kernel was not correctly checking for the sentinel value, but rather looking for values smaller than 0.

### DIFF
--- a/device/common/include/traccc/finding/device/impl/add_links_for_holes.ipp
+++ b/device/common/include/traccc/finding/device/impl/add_links_for_holes.ipp
@@ -7,6 +7,8 @@
 
 #pragma once
 
+#include <utility>
+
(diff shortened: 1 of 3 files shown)


stephenswat:
@krasznaa updated as you requested.

beomki-yeo (approved):
Ah, finding the bug at the track building kernel is a good catch

krasznaa (approved):
👍

PR #695, opened 2024-09-03
Fuse inner loop kernels in device CKF

stephenswat:
The inner loop of the device CKF consists of five loops: material interaction application, measurement counting, candidate finding, hole writing, and propagation. I believe that the middle three can be easily merged into a single kernel, reducing the amount of work we have to do on the host and simplifying the code a lot. This commit makes that change.


device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -357,99 +328,38 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
-            (*global_counter_device).n_measurements_sum);
-        TRACCC_CUDA_ERROR_CHECK(cudaGetLastError());
-
-        // Global counter object: Device -> Host
-        TRACCC_CUDA_ERROR_CHECK(
-            cudaMemcpyAsync(&global_counter_host, global_counter_device.get(),
-                            sizeof(device::finding_global_counter),
-                            cudaMemcpyDeviceToHost, stream));
-
         m_stream.synchronize();
```
krasznaa, line 331: Is this synchronization point needed? I think it was only here for the `cudaMemcpyAsync(...)` call, which is now removed. 🤔

device/common/include/traccc/finding/device/find_tracks.hpp
```diff
@@ -43,22 +45,24 @@ namespace traccc::device {
 /// @param[out] n_total_candidates  The number of total candidates for the
```
krasznaa, line 45: Please update the description of the parameters.

krasznaa (changes requested):
While not understanding most of the changes in the logic, I like the general direction quite a bit. As long as the code still works after this in the same way as it did before, I'm very much on board. 😉

device/common/include/traccc/finding/device/impl/find_tracks.ipp
```diff
@@ -8,144 +8,276 @@
...
-
-    vecmem::device_vector<const unsigned int>::iterator lo1;
-
-    for (unsigned int i_meas = 0; i_meas < cfg.n_measurements_per_thread;
-         i_meas++) {
-        const unsigned int idx = stride + i_meas;
-
-        if (idx >= n_measurements_sum) {
-            break;
+    unsigned int in_param_id = thread_id.getGlobalThreadIdX();
```
beomki-yeo, line 80: maybe add `const`?

beomki-yeo (commented):
Reducing the number of kernels is definitely good. I think the CPU finding algorithm will also need to remove `orig_param_id` and `skip_counter` to keep the symmetry but this can be done later.

Have you checked the performance of this?

device/common/include/traccc/finding/device/impl/find_tracks.ipp
stephenswat: Done, thanks.

device/cuda/src/finding/finding_algorithm.cu
stephenswat: Indeed it is not. Removed.

device/common/include/traccc/finding/device/find_tracks.hpp
stephenswat: Done.

stephenswat:
PR updated.

> Have you checked the performance of this?

It's essentially identical:

**Before**:

```
Reconstructed track parameters: 3061231
Time totals:
                  File reading  80234 ms
            Warm-up processing  984 ms
              Event processing  8227 ms
Throughput:
            Warm-up processing  98.4611 ms/event, 10.1563 events/s
              Event processing  82.2799 ms/event, 12.1536 events/s
```

**After:**

```
Reconstructed track parameters: 3063909
Time totals:
                  File reading  79725 ms
            Warm-up processing  27888 ms
              Event processing  8229 ms
Throughput:
            Warm-up processing  2788.83 ms/event, 0.358573 events/s
              Event processing  82.2956 ms/event, 12.1513 events/s
```

beomki-yeo (approved):

PR #696, opened 2024-09-05
Update to detray version v0.74.2

niermann999:
Update to detray v0.74.2
Main changes:
- Remove distinction between free vector and free track parameters and refactor track param class interface a bit
- Move the transform index onto the barcode
- fixed window navigation (will remove the navigation buffer, as the navigator can now deal with a static fixed size memory)

### DIFF
--- a/device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
+++ b/device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
@@ -14,8 +14,6 @@ TRACCC_DEVICE inline void propagate_to_next_surface(
     std::size_t globalIndex, const config_t cfg,
     typename propagator_t::detector_type::view_type det_data,
     bfield_t field_data,
-    vecmem::data::jagged_vector_view<typename propagator_t::intersection_type>
-        nav_candidates_buffer,
     bound_track_parameters_collection_types::const_view in_params_view,
     vecmem::data::vector_view<const candidate_link> links_view,
     const unsigned int step, const unsigned int& n_in_params,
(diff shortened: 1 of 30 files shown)


niermann999:
I will need to remove an assertion from detray still (the seed generator smears the time, so that it can indeed be negative), but otherwise this should be ready

device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
```diff
@@ -40,18 +38,13 @@ TRACCC_DEVICE inline void propagate_to_next_surface(
...
-
     // Count the number of tracks per seed
     vecmem::device_atomic_ref<unsigned int> num_tracks_per_seed(
         n_tracks_per_seed.at(orig_param_id));
 
     const unsigned int s_pos = num_tracks_per_seed.fetch_add(1);
 
-    if (s_pos >= cfg.max_num_branches_per_seed ||
-        globalIndex >= nav_candidates.size()) {
+    if (s_pos >= cfg.max_num_branches_per_seed || globalIndex >= n_in_params) {
```
stephenswat, line 47: I don't believe this is correct; need to look into this.

stephenswat (changes requested):
I have a concern about one of the changes. Did you check the physics performance of these changes at all? Otherwise I can have a look into it.

device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
niermann999: Removed

stephenswat (approved):
After looking into it with Joana I am happy to approve (as long as the CI is also happy).

PR #697, opened 2024-09-09
Add missing files to CMakeLists.txt

beomki-yeo:
Neza has reported that `fatal error: traccc/clusterization/device/ccl_kernel_definitions.hpp: No such file or directory` appears in the client build.

I think it is because the corresponding files are missing in the `CMakeLists.txt` (not 100% sure though)


beomki-yeo:
It turns out that it does not solve the issue of Neza 🫠

stephenswat (approved):
Generally, adding header files to a target like this doesn't actually do anything. I believe Attila insists on this because he uses an IDE which relies on this information, but it doesn't change the compilation at all. Happy to approve, though.

PR #699, opened 2024-09-12
Replace hit_id column of cell with measurement_id

beomki-yeo:
Part of #692 

`hit_id` is not used anymore in ACTS csv data format so I suggest replacing them with right one.

It is not directly related with the purpose of this PR, but I also changed the `geometry_id` of mock_data to a value compatible with ODD. This will be needed in the next PRs with event_map refactoring


beomki-yeo:
Any review?

stephenswat (approved):

PR #701, opened 2024-09-16
Issue warnings for non-Release builds

stephenswat:
A common pitfall when using traccc seems to be that people build it in non-Release mode and then get poor performance. This commit adds a bunch of warning prints to make sure that the code will inform you if you try to run it in non-Release mode.

### DIFF
--- a/examples/run/common/optimization_warning.hpp
+++ b/examples/run/common/optimization_warning.hpp
@@ -1,0 +1,27 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <iostream>
+
+#if !defined(TRACCC_BUILD_TYPE_IS_RELEASE) || !defined(NDEBUG) || \
+    defined(_DEBUG)
+#define TRACCC_OPTIMIZATION_WARNING()                                         \
+    do {                                                                      \
+        std::cout                                                             \
+            << "WARNING: traccc was built without Release mode, without the " \
+               "`NDEBUG` flag, or (on MSVC) with the `_DEBUG` flag. "         \
+               "Performance is guaranteed to be much lower and compute "      \
+               "performance results should be considered unreliable!"         \
+            << std::endl;                                                     \
+    } while (false)
+#else
+#define TRACCC_OPTIMIZATION_WARNING() \
+    do {                              \
+    } while (false)
+#endif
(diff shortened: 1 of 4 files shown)


krasznaa:
So, you pretty much add this printout to every single executable as I can see it. :thinking: Would it not be easier to just print a warning when `traccc::core` is loaded? Then we wouldn't need to remember to put this macro into all the executables. Not to mention that I'm not even that big of a fan of how the macro is set up...

What I'm thinking of is a file called `optimization_warning.cxx` in `core/src`, with something like:

```c++
int checkOptimization() {
#if ...
    std::cerr << ...;
#endif
}
static const int dummy = checkOptimization();
```

It's a bit of a hack, but I've sort of enjoyed how [RooFit](https://root.cern/manual/roofit/) has been printing a greeting message to its users with the same "trick" since forever. :thinking:

stephenswat:
Updated to affect only the throughput examples.

examples/run/common/optimization_warning.hpp
```diff
@@ -0,0 +1,27 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022 CERN for the benefit of the ACTS project
```
krasznaa, line 3:
```suggestion
 * (c) 2024 CERN for the benefit of the ACTS project
```

krasznaa (changes requested):
Shoot, wanted to make all comments part of the review...

You sure want users to notice these messages. :laughing: 3 printouts per job. :open_mouth: (I'm on board though.)

examples/run/common/optimization_warning.hpp
```diff
@@ -0,0 +1,27 @@
...
+               "`NDEBUG` flag, or (on MSVC) with the `_DEBUG` flag. "         \
+               "Performance is guaranteed to be much lower and compute "      \
+               "performance results should be considered unreliable!"         \
+            << std::endl;                                                     \
+    } while (false)
+#else
+#define TRACCC_OPTIMIZATION_WARNING() \
+    do {                              \
+    } while (false)
+#endif
```
krasznaa, line 27: Coming back to this... Why a macro at this point? A `void optimization_warning()` function would seem perfectly adequate here. The extra function call is not really an issue, it's all happening during the initialization of the throughput applications. :thinking:

krasznaa: Okay, I have no idea what the GitHub UI is doing now... :confused:

krasznaa:
Ping! It would be nice to put this in after some further changes.

PR #706, opened 2024-09-20
Access track parameters in sorted order for CKF propagation

beomki-yeo:
CKF update part of #689, for easier review


device/common/include/traccc/edm/device/sort_key.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+// Project include(s).
+#include "traccc/definitions/primitives.hpp"
+#include "traccc/edm/track_parameters.hpp"
+
+namespace traccc::device {
+
+struct sort_key {
+    traccc::scalar key;
+};
```
krasznaa, line 18: Do we expect the key to become something more complicated than a float value? ("Yes" is absolutely a believable answer.)

If not, it would be easier to just do:

```suggestion
using sort_key = traccc::scalar;
```

This would allow us to not have to define `sort_key_comp`. :thinking:

device/common/include/traccc/edm/device/sort_key.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+TRACCC_HOST_DEVICE
+inline sort_key get_sort_key(const bound_track_parameters& params) {
+    // key = |theta - pi/2|
+    return sort_key{math::abs(params.theta() - constant<traccc::scalar>::pi_2)};
+}
```
krasznaa, line 24: Though if we keep `sort_key` as-is, couldn't this become a constructor on that class? :thinking: That would give the code a more C\+\+-ie interface.

device/common/include/traccc/finding/device/get_sort_key_value.hpp
```diff
@@ -0,0 +1,32 @@
...
+namespace traccc::device {
+
+/// Function used for fitting a track for a given track candidates
+///
+/// @param[in] globalIndex   The index of the current thread
+/// @param[in] params_view   The input parameters
+/// @param[out] keys_view    The key values
+/// @param[out] ids_view     The param ids
+///
+TRACCC_HOST_DEVICE inline void get_sort_key_value(
```
krasznaa, line 23: I don't like this function name. :frowning:  This is absolutely subjective, true, but how about `fill_sort_keys`? Just something with `fill` would sound better to me than `get`. :thinking:

device/common/include/traccc/finding/device/propagate_to_next_surface.hpp
```diff
@@ -41,6 +42,7 @@ TRACCC_DEVICE inline void propagate_to_next_surface(
     typename propagator_t::detector_type::view_type det_data,
     bfield_t field_data,
     bound_track_parameters_collection_types::const_view in_params_view,
+    vecmem::data::vector_view<const unsigned int> param_ids_view,
```
krasznaa, line 45:
```suggestion
    const vecmem::data::vector_view<const unsigned int>& param_ids_view,
```

I'm very happy that you correctly used the `const unsigned int` template parameter. Now we should make the variable itself `const` as well. :wink:

This should generally be cleaned up in all of these functions a bit, but I'm happy with just setting this one variable correctly as a start as well. :smile:

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -108,13 +110,25 @@ __global__ void find_tracks(
...
 
+/// CUDA kernel for running @c traccc::device::get_sort_key_value
+__global__ void get_sort_key_value(
+    bound_track_parameters_collection_types::const_view params_view,
+    vecmem::data::vector_view<device::sort_key> keys_view,
+    vecmem::data::vector_view<unsigned int> ids_view) {
+
+    int gid = threadIdx.x + blockIdx.x * blockDim.x;
+
+    device::get_sort_key_value(gid, params_view, keys_view, ids_view);
```
krasznaa, line 121: I would do:

```suggestion
    device::get_sort_key_value(threadIdx.x + blockIdx.x * blockDim.x, params_view, keys_view, ids_view);
```

The new function expects an `std::size_t` value for the "global ID". But here you produce currently a signed integer instead. Worse yet, you produce the signed integer out of `unsigned int` values. (The CUDA types are `unsigned int` I believe...)

Long story short, it seems better to just not use an explicit type here, and just feed the value to the function with this formalism. :thinking:

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -365,7 +379,33 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
+         *****************************************************************/
+
+        vecmem::data::vector_buffer<device::sort_key> keys_buffer(
+            global_counter_host.n_candidates, m_mr.main);
+        vecmem::data::vector_buffer<unsigned int> param_ids_buffer(
+            global_counter_host.n_candidates, m_mr.main);
+
+        nThreads = m_warp_size * 2;
+
+        if (global_counter_host.n_candidates > 0) {
```
krasznaa, line 392: :thinking: Should we not abort "more explicitly" when we don't have candidates anymore? It seems silly to still call `thrust::sort_by_key` and whatnot on empty containers.

krasznaa (changes requested):
To be clear, as I said on Friday, I'm absolutely happy with this development. It's just some of the technicalities that I have comments on. :wink:

device/cuda/src/finding/finding_algorithm.cu
stephenswat: The CUDA values are actually signed integers, so the `int` signature is correct. :cry:

beomki-yeo: Idk, I will just put the id to the function directly as suggested. Probably other functions will also need to be fixed

device/cuda/src/finding/finding_algorithm.cu
beomki-yeo: I will put the sorting part into the block then

device/common/include/traccc/finding/device/propagate_to_next_surface.hpp
beomki-yeo: OK

device/common/include/traccc/finding/device/get_sort_key_value.hpp
beomki-yeo: OK I used `fill_sort_keys`

device/common/include/traccc/edm/device/sort_key.hpp
beomki-yeo: OK I will just use `scalar`. Let's think about making struct if we ever need to do that in the future.

device/common/include/traccc/finding/device/fill_sort_keys.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#pragma once
+
+// Project include(s).
+#include "traccc/edm/device/sort_key.hpp"
+#include "traccc/edm/track_candidate.hpp"
+
+namespace traccc::device {
+
+/// Function used for fitting a track for a given track candidates
```
krasznaa, line 16: The description is a bit misleading unfortunately...

krasznaa (approved):
The comment about the documentation you may or may not fix. I don't feel strongly enough about it to hold up the PR just because of it. :+1:

device/common/include/traccc/finding/device/fill_sort_keys.hpp
beomki-yeo: Good catch. Fixed

beomki-yeo:
CI seems stuck again - let me just push this thru...

PR #708, opened 2024-09-21
Clusterization Bugfix, main branch (2024.09.21.)

krasznaa:
This PR is just here to help in the explanation of a bug that I'll open an issue about shortly.

### DIFF
--- a/device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
+++ b/device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
@@ -259,7 +259,7 @@ TRACCC_DEVICE inline void ccl_kernel(
          * on any cells that have been claimed by the previous block (if
          * any).
          */
-        while (start != 0 &&
+        while (start != 0 && start < num_cells &&
                cells_device.at(start - 1).module_link ==
                    cells_device.at(start).module_link &&
                cells_device.at(start).channel1 <=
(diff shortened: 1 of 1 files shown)


stephenswat:
Looks like you found it. :+1:

stephenswat (approved):

krasznaa:
Indeed. No thanks to `cuda-gdb` with that one... :frowning: Even though in `Debug` mode we don't ask for any optimizations from `nvcc`, the compiled code kept behaving very weirdly still... :confused:

It was the debug SYCL build, ran on the host CPU, that let me finally understand the issue. As in that one `gdb-oneapi` was actually giving me meaningful values for the variables in the problematic thread. :thinking:

So we may not want to get rid of SYCL all too soon, after everything that's been said recently.

stephenswat:
Closes #709.

stephenswat:
> Indeed. No thanks to `cuda-gdb` with that one... 😦 Even though in `Debug` mode we don't ask for any optimizations from `nvcc`, the compiled code kept behaving very weirdly still... 😕

Yes I have been seeing this too, `cuda-gdb` (and to a lesser extent `compute-sanitizer`) have been very unhelpful lately. I wonder if some update on NVIDIA's end broke the tools or if we're somehow using the wrong compiler flags? :thinking:

krasznaa:
Maybe we need to use `-O0` explicitly now? :thinking: Since CMake doesn't do that automatically.

Host compilers of course don't do any optimizations unless asked for it, but maybe `nvcc` now adopted `icpx`-es behaviour, that if you don't ask for anything, it tries to use some aggressive optimization to "help you"... :confused:

PR #712, opened 2024-09-21
Cell EDM Rewrite, main branch (2024.09.18.)

### LINKED ISSUE #691: Remove `module_link` and `cell_module`
They should be replaced by detray detector geometry id and surface.

krasznaa:
This is the next monster PR... Exchanging `traccc::cell_collection_types` and `traccc::cluster_container_types` with SoA versions.

To jump right to the chase: It doesn't bring any performance improvement. :frowning: This EDM change of course only affects clusterization. Which is already one of the fastest things that we run. Still if anything, I see an O(1%) performance drop during the TML $\mu$=200 throughput measurements with this update applied. :thinking:

On my RTX3080 I get the following with the current `main` branch:

```
[bash][Legolas]:traccc > ./build-orig/bin/traccc_throughput_st_cuda --input-directory=tml_full/ttbar_mu200/ --input-events=10 --cold-run-events=100 --processed-events=1000
...
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Reconstructed track parameters: 0
Time totals:
                  File reading  5475 ms
            Warm-up processing  359 ms
              Event processing  2814 ms
Throughput:
            Warm-up processing  3.59883 ms/event, 277.868 events/s
              Event processing  2.81497 ms/event, 355.244 events/s
[bash][Legolas]:traccc >
```

While this PR produces the following:

```
[bash][Legolas]:traccc > ./out/build/cuda/bin/traccc_throughput_st_cuda --input-directory=tml_full/ttbar_mu200/ --input-events=10 --cold-run-events=100 --processed-events=1000
...
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Reconstructed track parameters: 0
Time totals:
                  File reading  5471 ms
            Warm-up processing  305 ms
              Event processing  2830 ms
Throughput:
            Warm-up processing  3.0542 ms/event, 327.418 events/s
              Event processing  2.83068 ms/event, 353.272 events/s
[bash][Legolas]:traccc >
```

(There is variation on these numbers, but the "new" code is always just a little slower. :frowning:)

About the code:
  - I chose `traccc::edm::silicon_cell_collection` and `traccc::edm::silicon_cluster_collection` as the name of these containers. But I'm not too fond of these names either. So I'm open to suggestions.
  - Most of the changes of the PR are pretty trivial. Since the new containers follow the naming of the current `traccc::cell` closely.
  - The larger changes are:
    * The clusterization functions shared between the CPU and GPU algorithms have a little different interfaces now.
    * Some logic changes in the "mapping code". There I had to make some larger changes. And keep in mind that the fundamental setup of that code does not work well with SoA containers. So as we make more of our EDM into SoAs, the less it will be possible to maintain that code. :thinking:
    * The `traccc::edm::silicon_cluster_collection` type, only used in the host code, is now a jagged vector of cell indices. As such, `traccc::host::measurement_creation_algorithm` had to change its interface slightly.
  * As @stephenswat would've heard from me in the last days, I modified all of the "sanity code" that he wrote earlier, to work on either AoS or SoA containers. I'm fairly happy with what I did there, but I'm very open to suggestions with that code as well.

We'll have to do some profiling, but I suspect that the small performance drop comes from the fact that the PR's code always reads the cell data from global memory, whenever it needs it. Just loading some of the info into local registers in a couple of places will hopefully take us back to the previous performance. I just didn't want to complicate the code even further in this PR. :thinking:

This PR also closes #691.


core/include/traccc/clusterization/details/sparse_ccl.hpp
```diff
@@ -42,25 +42,29 @@ TRACCC_HOST_DEVICE inline unsigned int make_union(
...
+/// @param i     The index of the first cell
+/// @param j     The index of the second cell
 ///
 /// @return boolan to indicate 8-cell connectivity
 ///
-TRACCC_HOST_DEVICE inline bool is_adjacent(const traccc::cell& a,
-                                           const traccc::cell& b);
+TRACCC_HOST_DEVICE inline bool is_adjacent(
+    const edm::silicon_cell_collection::const_device& cells, unsigned int i,
+    unsigned int j);
```
stephenswat, line 53: This is a nice example of a place where a proxy object would be _really_ nice; complicating this function signature and making it so much less general is a bit unfortunate.

core/include/traccc/clusterization/impl/measurement_creation.ipp
```diff
@@ -19,41 +19,48 @@ inline scalar signal_cell_modelling(
...
 
     // Retrieve the specific values based on module idx
-    const cell::link_type module_link = cell.module_link;
-    return {
-        det_descr.reference_x().at(module_link) +
-            (scalar{0.5} + cell.channel0) * det_descr.pitch_x().at(module_link),
-        det_descr.reference_y().at(module_link) +
-            (scalar{0.5} + cell.channel1) *
-                det_descr.pitch_y().at(module_link)};
+    const unsigned int module_idx = cells.module_index().at(cell_idx);
```
stephenswat, line 27: I generally don't like that we don't have the cell type-defs anymore; can we put these in a helper struct or something?

core/include/traccc/clusterization/impl/measurement_creation.ipp
```diff
@@ -92,38 +100,50 @@ TRACCC_HOST_DEVICE inline void fill_measurement(
...
+    assert([&]() {
+        const unsigned int module_idx =
+            cells.module_index().at(clusters.cell_indices()[index].front());
+        for (const unsigned int cell_idx : clusters.cell_indices()[index]) {
+            if (cells.module_index().at(cell_idx) != module_idx) {
+                return false;
+            }
+        }
+        return true;
+    }() == true);
```
stephenswat, line 114:
```suggestion
    assert(std::all_of(clusters.cell_indices()[index].cbegin(), clusters.cell_indices()[index].cend(), [module_idx=cells.module_index().at(clusters.cell_indices()[index].front())](module_idx_type & m){ m == module_idx}));
```

core/include/traccc/clusterization/impl/sparse_ccl.ipp
```diff
@@ -41,23 +41,31 @@ TRACCC_HOST_DEVICE inline unsigned int make_union(
...
+    const edm::silicon_cell_collection::const_device& cells, unsigned int i,
+    unsigned int j) {
+
+    return ((cells.channel0()[i] - cells.channel0()[j]) *
+                (cells.channel0()[i] - cells.channel0()[j]) <=
+            1) &&
+           ((cells.channel1()[i] - cells.channel1()[j]) *
+                (cells.channel1()[i] - cells.channel1()[j]) <=
+            1) &&
+           (cells.module_index()[i] == cells.module_index()[j]);
```
stephenswat, line 54: Yeah I don't love how much less readable (and general) this is now... :cry:

core/include/traccc/sanity/contiguous_on.hpp
```diff
@@ -8,59 +8,52 @@
...
-requires std::regular_invocable<P, T> bool is_contiguous_on(
-    P&& projection, vecmem::data::vector_view<T> vector) {
-    // Grab the number of elements in our vector.
-    uint32_t n = vector.size();
+template <std::semiregular P, typename CONTAINER>
+requires std::regular_invocable<P, CONTAINER, std::size_t> bool
+is_contiguous_on(P&& projection, const CONTAINER& in) {
+
+    // Grab the number of elements in our container.
+    std::size_t n = in.size();
```
stephenswat, line 42: No, I don't like this. This could stay the same with a proxy type; let's not go around making code less functional and harder to read to facilitate this change if it can be avoided.

core/include/traccc/sanity/ordered_on.hpp
```diff
@@ -41,17 +41,16 @@ namespace traccc::host {
  * @return true If the vector is ordered on `R`.
  * @return false Otherwise.
  */
-template <std::semiregular R, typename T>
-requires std::relation<R, T, T> bool is_ordered_on(
-    R relation, vecmem::data::vector_view<T> vector) {
-    // Grab the number of elements in our vector.
-    uint32_t n = vector.size();
+template <std::semiregular R, typename CONTAINER>
```
stephenswat, line 44: As before, this code doesn't have to be changed at all.

core/include/traccc/seeding/track_params_estimation.hpp
```diff
@@ -1,14 +1,13 @@
...
- * (c) 2021-2022 CERN for the benefit of the ACTS project
+ * (c) 2021-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
 
 #pragma once
 
 // Library include(s).
-#include "traccc/edm/cell.hpp"
```
stephenswat, line 11: This file does not need the new, updated cell EDM?

core/include/traccc/utils/projections.hpp
```diff
@@ -8,21 +8,24 @@
...
 }
 }
 ;
 
 struct [[maybe_unused]] measurement_module_projection{
-    TRACCC_HOST_DEVICE auto operator()(const traccc::measurement& m)
-        const {return m.module_link;
+    template <typename MEASUREMENT_VECTOR> TRACCC_HOST_DEVICE auto operator()(
+        const MEASUREMENT_VECTOR& m, std::size_t i)
+        const {return m.at(i).module_link;
```
stephenswat, line 28: See comments upstairs.

io/src/csv/read_cells.cpp
```diff
@@ -25,22 +25,22 @@ namespace {
 /// Comparator used for sorting cells. This sorting is one of the assumptions
 /// made in the clusterization algorithm
 struct cell_order {
-    bool operator()(const traccc::cell& lhs, const traccc::cell& rhs) const {
-        if (lhs.module_link != rhs.module_link) {
-            return lhs.module_link < rhs.module_link;
-        } else if (lhs.channel1 != rhs.channel1) {
+    bool operator()(const traccc::io::csv::cell& lhs,
+                    const traccc::io::csv::cell& rhs) const {
+        if (lhs.channel1 != rhs.channel1) {
```
stephenswat, line 30: This is only equivalent if the cells are guaranteed to be on the same module; did you ensure this?

io/src/csv/read_cells.cpp
```diff
@@ -150,8 +148,13 @@ void read_cells(cell_collection_types::host& cells, std::string_view filename,
...
         for (auto& cell : cellz) {
-            cells.push_back(cell);
-            cells.back().module_link = ddIndex;
+            const std::size_t cellIndex = cells.size();
+            cells.resize(cellIndex + 1);
+            cells.channel0().at(cellIndex) = cell.channel0;
+            cells.channel1().at(cellIndex) = cell.channel1;
+            cells.activation().at(cellIndex) = cell.value;
+            cells.time().at(cellIndex) = cell.timestamp;
+            cells.module_index().at(cellIndex) = ddIndex;
```
stephenswat, line 157: Strongly consider adding an emplace method to the EDM object; this is too verbose and error-prone.

io/src/mapper.cpp
```diff
@@ -168,31 +170,48 @@ hit_cell_map generate_hit_cell_map(std::size_t event,
...
-                 iocell.timestamp, link});
+        auto pair = result.insert({hmap[iocell.measurement_id], {resource}});
+        edm::silicon_cell_collection::host& cells = pair.first->second;
+        const std::size_t cellIndex = cells.size();
+        cells.resize(cellIndex + 1);
+        cells.channel0().at(cellIndex) = iocell.channel0;
+        cells.channel1().at(cellIndex) = iocell.channel1;
+        cells.activation().at(cellIndex) = iocell.value;
+        cells.time().at(cellIndex) = iocell.timestamp;
+        cells.module_index().at(cellIndex) = link;
```
stephenswat, line 181: Same as above.

io/src/mapper.cpp
```diff
@@ -251,12 +289,34 @@ measurement_particle_map generate_measurement_particle_map(
...
+                         1e-6f) &&
+                        (std::abs(mCells.time().at(mCellIndex) -
+                                  pCells.time().at(pCellIndex)) < 1e-6f) &&
+                        (mCells.module_index().at(mCellIndex) ==
+                         pCells.module_index().at(pCellIndex))) {
+
+                        result[meas][ptc]++;
+                    }
+                }
+            }
```
stephenswat, line 319: Question: what is happening here? :worried:

stephenswat (changes requested):
I'm generally on board, but in its current state this makes the code much less readable and generic; I'd strongly recommend looking into a proxy object for this as it would make the code more readable _and_ require much fewer code line changes.

See, e.g. https://github.com/acts-project/acts/blob/main/Examples/Framework/include/ActsExamples/EventData/Measurement.hpp and https://github.com/acts-project/traccc/blob/main/core/include/traccc/utils/array_wrapper.hpp for inspiration.

core/include/traccc/seeding/track_params_estimation.hpp
krasznaa: Yepp. A couple of files didn't need this include at all.

io/src/csv/read_cells.cpp
krasznaa: Very good catch! :smile: I did this on purpose. Because the CSV reading code indeed, by construction, never compares cells coming from different modules using this function. So the `module_link` comparison earlier was not actually necessary.

At the same time, I was pondering about this myself as well. Since while the current code is absolutely correct like this, the very negligible performance loss incurred by checking the module ID could possibly be worth it for avoiding future problems, when the file reading logic may change. :thinking:

io/src/csv/read_cells.cpp
krasznaa: This would be tied to the proxy idea discussed above. Without a proxy type there's nothing better that we can do at the moment. :frowning:

stephenswat:
As mentioned, please increase the minimum vecmem version to 1.8.0.

io/src/mapper.cpp
krasznaa: :laughing: I also hate this...

What happens here is that we have two maps. One associating cells with all the "truth measurements" in the event. The other associating cells with all the "truth particles" in the event. And here we try to figure out which "truth measurements" belong to which "truth particles".

Since in our data cells can appear multiple times, when particles hit the same part of a surface, I had to require identity between every variable of the cells here. At first I was only checking for their integer identifiers. And that made the unit tests fail. That's why this ugly comparison was made.

Without a question this could also be expressed a lot more nicely with proxy types...

beomki-yeo:
Hmm. are you going to go through this for every EDM class?

beomki-yeo:
This will also have lots of conflictions with #692 :crying_cat_face:

stephenswat:
> This will also have lots of conflictions with #692 😿

The good news is that when the proxy objects are implemented, most of the code will remain unchanged.

krasznaa:
> > This will also have lots of conflictions with #692 😿
> 
> The good news is that when the proxy objects are implemented, most of the code will remain unchanged.

Some further developments are indeed underway...

core/include/traccc/clusterization/details/sparse_ccl.hpp
krasznaa: Have a look if you like the latest version of the code. :wink: I only made the clusterization code make use of `vecmem::edm::proxy` for now, just to see how everyone likes it.

krasznaa:
All of you, hold onto your hats. :smile: If/once we settle on https://github.com/acts-project/vecmem/pull/296, these are the types of updates that we will need to do to switch from the current AoS to a new SoA EDM:

image

At the same time I looked at profiles of the throughput application a little as well. This was very educational. As it turns out, the small slowdown is not due to the kernels. It seems to be due to the code spending a little more time on memory copies. :thinking:

That's not great news, as apparently the `vecmem::edm` code is not quite as efficient wrt. CPU usage as I hoped. But at least the SoA layout doesn't seem to have much of an impact on clusterization after all. (Remember, even with the current AoS layout, since `traccc::cell` is tiny, the memory access pattern of clusterization is pretty efficient already.)

core/include/traccc/clusterization/impl/measurement_creation.ipp
krasznaa: I'm afraid of using `std::all_of` here... :thinking: This function advertises itself as working in device code. And the STL algorithms don't have a terrific track record with that.

But I very much agree, I'm not in love with this assertion at the moment either. :frowning:

core/include/traccc/sanity/ordered_on.hpp
krasznaa: Have a look at your functions now! They should be a lot more agreeable now that they use the `vecmem::edm::proxy` type...

io/src/csv/read_cells.cpp
krasznaa: Have a look! I think the code is much more visually appealing now. :wink:

device/cuda/src/sanity/contiguous_on.cuh
```diff
@@ -24,117 +25,127 @@
...
+        S v1 = projection(in.at(tid - 1));
+        S v2 = projection(in.at(tid));
 
         if (v1 != v2) {
-            out[atomicAdd(out_size, 1u)] = v2;
+            out.push_back(v2);
         }
     } else if (tid == 0) {
-        out[atomicAdd(out_size, 1u)] = projection(in.at(tid));
+        out.push_back(projection(in.at(tid)));
```
krasznaa, line 54: @stephenswat take note that I changed this code a little. Since you were literally implementing a resizable 1D vector by hand previously, I decided to rather just use the vecmem types here that were designed for exactly this use case. :thinking:

krasznaa:
The good news is that once the code finally starts working on all platforms with all compilers, this very latest version is finally delivering on the performance front. :smile:

```
[bash][Legolas]:traccc > ./out/build/cuda/bin/traccc_throughput_st_cuda --input-directory=tml_full/ttbar_mu200/ --input-events=10 --cold-run-events=100 --processed-events=1000

Running Single-threaded CUDA GPU throughput tests

>>> Detector Options <<<
  Detector file       : tml_detector/trackml-detector.csv
  Material file       : 
  Surface grid file   : 
  Use detray::detector: no
  Digitization file   : tml_detector/default-geometric-config-generic.json
>>> Input Data Options <<<
  Input directory               : tml_full/ttbar_mu200/
  Number of input events        : 10
  Threads per partition:      256
  Target cells per thread:    8
  Max cells per thread:       16
  Scratch space size mult.:   256
>>> Track Seeding Options <<<
  None
>>> Track Finding Options <<<
  Max number of branches per seed: 10
  Max number of branches per surface: 10
  Minimum step length for the next surface: 0.5 [mm] 
  Maximum step counts for the next surface: 100
  Maximum Chi2             : 30
  Maximum branches per step: 10
  Maximum number of skipped steps per candidates: 3
  PDG Number: 13
Navigation
----------------------------
  Min. mask tolerance   : 1e-05 [mm]
  Max. mask tolerance   : 1 [mm]
  Mask tolerance scalor : 0.05
  Path tolerance        : 1 [um]
  Overstep tolerance    : -100 [um]
  Search window         : 0 x 0

Parameter Transport
----------------------------
  Min. Stepsize         : 0.0001 [mm]
  Runge-Kutta tolerance : 0.0001 [mm]
  Max. step updates     : 10000
  Stepsize  constraint  : 3.40282e+38 [mm]
  Path limit            : 5 [m]
  Use Bethe energy loss : true
  Do cov. transport     : true
  Use eloss gradient    : false
  Use B-field gradient  : false


>>> Throughput Measurement Options <<<
  Cold run event(s) : 100
  Processed event(s): 1000
  Log file          : 

WARNING: @traccc::io::csv::read_cells: 251 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/tml_full/ttbar_mu200/event000000000-cells.csv
WARNING: @traccc::io::csv::read_cells: 305 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/tml_full/ttbar_mu200/event000000001-cells.csv
WARNING: @traccc::io::csv::read_cells: 176 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/tml_full/ttbar_mu200/event000000002-cells.csv
WARNING: @traccc::io::csv::read_cells: 200 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/tml_full/ttbar_mu200/event000000003-cells.csv
WARNING: @traccc::io::csv::read_cells: 224 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/tml_full/ttbar_mu200/event000000004-cells.csv
WARNING: @traccc::io::csv::read_cells: 170 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/tml_full/ttbar_mu200/event000000005-cells.csv
WARNING: @traccc::io::csv::read_cells: 321 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/tml_full/ttbar_mu200/event000000006-cells.csv
WARNING: @traccc::io::csv::read_cells: 322 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/tml_full/ttbar_mu200/event000000007-cells.csv
WARNING: @traccc::io::csv::read_cells: 222 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/tml_full/ttbar_mu200/event000000008-cells.csv
WARNING: @traccc::io::csv::read_cells: 118 duplicate cells found in /data/ssd-1tb/projects/traccc/traccc/data/tml_full/ttbar_mu200/event000000009-cells.csv
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Reconstructed track parameters: 0
Time totals:
                  File reading  4968 ms
            Warm-up processing  358 ms
              Event processing  2715 ms
Throughput:
            Warm-up processing  3.58551 ms/event, 278.9 events/s
              Event processing  2.71537 ms/event, 368.274 events/s
[bash][Legolas]:traccc >
```

Though I **am** a little bit afraid of this possibly being artificial. Since the previous result was on `x86_64-ubuntu2204-gcc11-opt`, and these latest numbers are now on `x86_64-ubuntu2404-gcc13-opt`. (I upgraded my home PC during the weekend... :stuck_out_tongue:) Still, at least the hardware is still the same... :thinking:

beomki-yeo (approved):
I am good as long as @stephenswat is OK

core/include/traccc/clusterization/details/sparse_ccl.hpp
```diff
@@ -47,8 +47,9 @@ TRACCC_HOST_DEVICE inline unsigned int make_union(
 ///
 /// @return boolan to indicate 8-cell connectivity
 ///
-TRACCC_HOST_DEVICE inline bool is_adjacent(const traccc::cell& a,
-                                           const traccc::cell& b);
+template <typename T1, typename T2>
+TRACCC_HOST_DEVICE inline bool is_adjacent(const edm::silicon_cell<T1>& a,
+                                           const edm::silicon_cell<T2>& b);
```
stephenswat, line 52: I'd advocate for simply having two universal reference arguments, the `const edm::silicon_cell<T1>` syntax is highly confusing; it's not clear at all to me what `T1` and `T2` are even supposed to be.

It would be better to implement a concept for silicon cells and use that instead:

```suggestion
template <silicon_cell_like T1, silicon_cell_like T2>
TRACCC_HOST_DEVICE inline bool is_adjacent(T1 && a, T2 && b);
```

core/include/traccc/clusterization/details/sparse_ccl.hpp
```diff
@@ -59,8 +60,9 @@ TRACCC_HOST_DEVICE inline bool is_adjacent(const traccc::cell& a,
 ///
 /// @return boolan to indicate !8-cell connectivity
 ///
-TRACCC_HOST_DEVICE inline bool is_far_enough(const traccc::cell& a,
-                                             const traccc::cell& b);
+template <typename T1, typename T2>
+TRACCC_HOST_DEVICE inline bool is_far_enough(const edm::silicon_cell<T1>& a,
+                                             const edm::silicon_cell<T2>& b);
```
stephenswat, line 65: Same as above.

core/include/traccc/clusterization/impl/measurement_creation.ipp
```diff
@@ -17,37 +17,43 @@ inline scalar signal_cell_modelling(
     return signal_in;
 }
 
-TRACCC_HOST_DEVICE
-inline vector2 position_from_cell(
-    const cell& cell,
+template <typename T>
+TRACCC_HOST_DEVICE inline vector2 position_from_cell(
+    const edm::silicon_cell<T>& cell,
```
stephenswat, line 22: And here, this templating is super confusing.

core/include/traccc/clusterization/impl/measurement_creation.ipp
stephenswat: Perhaps we could implement out own `all_of`? That we can guarantee will run on device?

core/include/traccc/clusterization/impl/measurement_creation.ipp
```diff
@@ -92,38 +99,53 @@ TRACCC_HOST_DEVICE inline void fill_measurement(
...
+        }
+        return true;
+    }() == true);
 
     // Calculate the cluster properties
     scalar totalWeight = 0.f;
     point2 mean{0.f, 0.f}, var{0.f, 0.f};
-    calc_cluster_properties(cluster, det_descr, mean, var, totalWeight);
+    calc_cluster_properties(index, cells, clusters, det_descr, mean, var,
+                            totalWeight);
```
stephenswat, line 119: Does this still need to take an index now that we have proxies?

core/include/traccc/clusterization/impl/measurement_creation.ipp
```diff
@@ -92,38 +99,53 @@ TRACCC_HOST_DEVICE inline void fill_measurement(
...
     // Access the measurement in question.
     measurement& m = measurements[index];
 
     // The index of the module the cluster is on.
-    const cell::link_type module_link = cluster.at(0).module_link;
-
-    m.module_link = module_link;
-    m.surface_link = det_descr.geometry_id().at(module_link);
+    const unsigned int module_idx =
+        cells.module_index().at(clusters.cell_indices().at(index).front());
```
stephenswat, line 128: Please use the proxy-based syntax which is more natural to read:

```suggestion
    const unsigned int module_idx =
        cells.at(clusters.at(index).cell_indices().front()).module_index();
```

core/include/traccc/clusterization/impl/sparse_ccl.ipp
```diff
@@ -41,23 +41,27 @@ TRACCC_HOST_DEVICE inline unsigned int make_union(
...
+template <typename T1, typename T2>
+TRACCC_HOST_DEVICE inline bool is_far_enough(const edm::silicon_cell<T1>& a,
+                                             const edm::silicon_cell<T2>& b) {
 
-    assert((a.channel1 >= b.channel1) || (a.module_link != b.module_link));
-    return (a.channel1 > (b.channel1 + 1)) || (a.module_link != b.module_link);
+    assert((a.channel1() >= b.channel1()) ||
+           (a.module_index() != b.module_index()));
+    return (a.channel1() > (b.channel1() + 1)) ||
+           (a.module_index() != b.module_index());
```
stephenswat, line 60: Same story again, these should be universal references to cell-like objects.

core/include/traccc/edm/silicon_cell_collection.hpp
```diff
@@ -0,0 +1,203 @@
...
+        if (module_index() != other.module_index()) {
+            return (module_index() < other.module_index());
+        } else {
+            if (channel1() != other.channel1()) {
+                return (channel1() < other.channel1());
+            } else {
+                return (channel0() < other.channel0());
+            }
+        }
+    }
```
stephenswat, line 189: These can be unified with an `operator<=>` to reduce boilerplate.

core/include/traccc/sanity/contiguous_on.hpp
```diff
@@ -8,52 +8,48 @@
...
  * @return false Otherwise.
  */
-template <std::semiregular P, std::equality_comparable T>
-requires std::regular_invocable<P, T> bool is_contiguous_on(
-    P&& projection, vecmem::data::vector_view<T> vector) {
-    // Grab the number of elements in our vector.
-    uint32_t n = vector.size();
+template <std::semiregular P, typename CONTAINER>
+requires std::regular_invocable<P,
+                                decltype(std::declval<CONTAINER>().at(0))> bool
```
stephenswat, line 40:
```suggestion
                                typename CONTAINER::value_type> bool
```

core/include/traccc/sanity/ordered_on.hpp
stephenswat: Much better, thanks!

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
```diff
@@ -54,10 +54,10 @@ inline void aggregate_cluster(
     scalar totalWeight = 0.;
     point2 mean{0., 0.}, var{0., 0.}, offset{0., 0.};
 
-    const auto module_link = cells[cid + start].module_link;
+    const unsigned int module_idx = cells.module_index().at(cid + start);
```
stephenswat, line 57:
```suggestion
    const unsigned int module_idx = cells.at(cid + start).module_index();
```

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
```diff
@@ -54,10 +54,10 @@ inline void aggregate_cluster(
...
     const unsigned short partition_size = end - start;
-    const scalar pitch_x = det_descr.pitch_x().at(module_link);
-    const scalar pitch_y = det_descr.pitch_y().at(module_link);
+    const scalar pitch_x = det_descr.pitch_x().at(module_idx);
+    const scalar pitch_y = det_descr.pitch_y().at(module_idx);
```
stephenswat, line 60:
```suggestion
    const scalar pitch_x = det_descr.at(module_idx).pitch_x();
    const scalar pitch_y = det_descr.at(module_idx).pitch_y();
```

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
```diff
@@ -70,32 +70,30 @@ inline void aggregate_cluster(
          * Terminate the process earlier if we have reached a cell sufficiently
          * in a different module.
          */
-        if (cells[pos].module_link != module_link) {
+        if (cells.module_index().at(pos) != module_idx) {
```
stephenswat, line 73:
```suggestion
        if (cells.at(pos).module_index() != module_idx) {
```

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
```diff
@@ -70,32 +70,30 @@ inline void aggregate_cluster(
...
          * If the value of this cell is equal to our, that means it
          * is part of our cluster. In that case, we take its values
          * for position and add them to our accumulators.
          */
         if (f.at(j) == cid) {
 
-            if (this_cell.channel1 > maxChannel1) {
-                maxChannel1 = this_cell.channel1;
+            if (cells.channel1().at(pos) > maxChannel1) {
+                maxChannel1 = cells.channel1().at(pos);
```
stephenswat, line 85: General comment: please replace all the use of array indexing in this file with proxy syntax.

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
```diff
@@ -260,10 +262,10 @@ TRACCC_DEVICE inline void ccl_kernel(
          * any).
          */
         while (start != 0 && start < num_cells &&
-               cells_device.at(start - 1).module_link ==
-                   cells_device.at(start).module_link &&
-               cells_device.at(start).channel1 <=
-                   cells_device.at(start - 1).channel1 + 1) {
+               cells_device.module_index().at(start - 1) ==
+                   cells_device.module_index().at(start) &&
```
stephenswat, line 266: In this file: please use proxy syntax.

device/common/include/traccc/clusterization/device/impl/reduce_problem_cell.ipp
```diff
@@ -28,7 +28,7 @@ inline void reduce_problem_cell(
     const unsigned int pos = cid + start;
 
     // Load the "reference cell" into a local variable.
-    const cell reference_cell = cells.at(pos);
+    const auto reference_cell = cells.at(pos);
```
stephenswat, line 31:
```suggestion
    const auto& reference_cell = cells.at(pos);
```

device/sycl/src/sanity/contiguous_on.hpp
```diff
@@ -26,122 +23,155 @@
...
-
-    // Copy the total number of squashed elements, e.g. the size of the
-    // resulting vector.
-    uint32_t host_iout_size;
-
     queue
-        .memcpy(&host_iout_size, iout_size.get(), sizeof(uint32_t),
-                {kernel1_evt})
+        .submit([&](cl::sycl::handler& h) {
+            h.parallel_for<kernels::is_contiguous_on_compress_adjacent<
```
stephenswat, line 144: General comment about these SYCL kernels, I notice you've removed a lot of the dependencies between events. Did you ensure the sequencing of computations is correct?

io/src/csv/read_cells.cpp
```diff
@@ -148,10 +146,20 @@ void read_cells(cell_collection_types::host& cells, std::string_view filename,
...
         }
 
+        // Helper lambda for setting up SoA cells.
+        auto assign = [ddIndex](auto soa_cell, const cell& iocell) {
+            soa_cell.channel0() = iocell.channel0;
+            soa_cell.channel1() = iocell.channel1;
+            soa_cell.activation() = iocell.value;
+            soa_cell.time() = iocell.timestamp;
+            soa_cell.module_index() = ddIndex;
+        };
```
stephenswat, line 156: Did you investigate the possibility of adding an emplace function? Because this is still far from ideal, unfortunately. The main concern is that any change in the structure of the cell, esp. the addition of new fields, will go unnoticed by the compiler; this is a big bug surface.

core/include/traccc/clusterization/details/sparse_ccl.hpp
stephenswat: I'm also okay with some typedefs thrown around, e.g. having some typedef for host cells and one for device cells..

stephenswat (changes requested):
It's getting there. :+1:

core/include/traccc/clusterization/details/sparse_ccl.hpp
krasznaa: I really don't agree on this one. You want to wrote even more code, when the C\+\+11 templating I use here is already forcing the user to use a specific type of class in the argument list. :thinking:

I did not spell it out in the PR, but the separate `T1` and `T2` template arguments are needed to allow const and non-const, device and host cell proxies to be used in a single function, in any combination.

Honestly, you should get used to this syntax, as I feel pretty strongly about it being the correct way to go. :thinking:

core/include/traccc/clusterization/details/sparse_ccl.hpp
krasznaa: Ditto.

core/include/traccc/clusterization/impl/measurement_creation.ipp
krasznaa: As written above, the templating is needed because `T` can be 4 different types at a minimum.

core/include/traccc/clusterization/impl/measurement_creation.ipp
krasznaa: I guess. Though as long as we only have a single use, that seems like a waste of effort. I would introduce that function once we have a second client. :thinking:

core/include/traccc/clusterization/impl/measurement_creation.ipp
krasznaa: See: https://github.com/acts-project/vecmem/pull/298 Proxying clusters doesn't work with vecmem `v1.9.0`. :frowning:

core/include/traccc/clusterization/impl/measurement_creation.ipp
krasznaa: I really disagree. :thinking: When we only need a single variable, I much prefer to use the "container interface". Since that's the whole point of an SoA layout, that we could do this. Without creating temporary proxy objects all around the place.

core/include/traccc/clusterization/impl/sparse_ccl.ipp
krasznaa: You made the comment on the same function (the declaration and the implementation) twice...?

core/include/traccc/edm/silicon_cell_collection.hpp
krasznaa: That's an interesting idea. :thinking: I'm mainly afraid that these operators are already defined in a bit of a weird way. In order to keep the current beaviour, the spaceship operator would need to be defined in a fairly weird way.

Still, it's an interesting idea. :thinking:

core/include/traccc/sanity/contiguous_on.hpp
krasznaa: No, that's not a thing. :confused: Not on SoA containers.

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
krasznaa: I really do not see the point. :confused: At that point we're trying to make the life of the compiler harder on purpose. And seeing how oneAPI freaked out about this code, I really don't want to complicate things unnecessarily.

Plus, the whole point of an SoA for me is that we can load a single variable when we just need a single variable. :thinking:

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
krasznaa: Here I could be convinced. :thinking: Though at that point I'd probably write:

```suggestion
    const auto module_descr = det_descr.at(module_idx);
    const scalar pitch_x = module_descr.pitch_x();
    const scalar pitch_y = module_descr.pitch_y();
```

Sure, this I can do...

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
krasznaa: See all my comments above.

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
krasznaa: No. I really-really don't intend to. :thinking:

device/common/include/traccc/clusterization/device/impl/reduce_problem_cell.ipp
krasznaa: Proxies are not members of the SoA containers. They are returned by value by `at(...)`. So using a const reference is at best confusing, at worst it may lead to problems. (Though from what I understand, the compiler would still generate valid code with that. It would just confuse the reader I believe.)

Maybe this is the source of our disagreement? :confused: Every time the code creates a proxy, it's not for free. We can hope that the compiler does a good job with optimizing the code. But if the compiler is not smart enough, `container.at(i).foo()` will be slower than `container.foo().at(i)`.

device/sycl/src/sanity/contiguous_on.hpp
krasznaa: I did. I had to, because vecmem copies do not return `sycl::event` objects that we could use like this.

It did make this code a little less efficient. But since it's only used for debugging, I was not too worried.

io/src/csv/read_cells.cpp
krasznaa: I'm open to a concrete code suggestion. Unfortunately I couldn't come up with a solution to this yet.

Note that with xAODs we've never been able to do it. Efficiently at least. Because with the xAOD code in order to do this, we have to create a temporary "auxiliary store" for the standalone object that we would then use to emplace back into the container. I.e. the object that we give to `emplace_back(...)` still has to point to some container. (At least with the xAOD infrastructure it does.)

It may not be impossible to make this all happen. :thinking: But the idea I have right now, will need a fair amount of development. (I'd need to create a whole new type based on `vecmem::edm::schema`, which would be the argument of `vecmem::edm::host<>::emplace_back(...)`.)

Something to keep in mind. But we can't stop this PR based on this, as it will never finish then. :thinking:

core/include/traccc/clusterization/impl/measurement_creation.ipp
stephenswat: Okay, that's fair. :+1:

core/include/traccc/clusterization/details/sparse_ccl.hpp
stephenswat: Okay, I can't say I agree but it's surely nor worth arguing over. :smile:

core/include/traccc/sanity/contiguous_on.hpp
stephenswat: Oh, that's... unfortunate :frowning_face:

krasznaa:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=acts-project_traccc&pullRequest=712) **Quality Gate failed**
> Failed conditions  [2 New Bugs](https://sonarcloud.io/dashboard?id=acts-project_traccc&pullRequest=712) (required ≤ 0)  [C Reliability Rating on New Code](https://sonarcloud.io/dashboard?id=acts-project_traccc&pullRequest=712) (required ≥ A)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=acts-project_traccc&pullRequest=712)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarLint](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

Huhh... :thinking: What's your take on these errors @stephenswat?

stephenswat:
> Huhh... 🤔 What's your take on these errors @stephenswat?

SonarCloud actually makes a really valid point here about constraining universal references; I'd suggest we go ahead and implement them.

krasznaa:
> > Huhh... 🤔 What's your take on these errors @stephenswat?
> 
> SonarCloud actually makes a really valid point here about constraining universal references; I'd suggest we go ahead and implement them.

As long as you have a concrete idea of how to go about it, I'm happy to let you propose the improvement. :wink:

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
krasznaa: So... mea culpa. Indeed, in this function the same cell is taken for practically all of its info. So now I switched (back) to accessing a proxy for a given cell, and then using that proxy pretty much everywhere. On top of doing the "`module_descr` change" that I teased earlier.

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
krasznaa: I ended up doing sort of this.

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
krasznaa: I ended up doing sort of this.

device/sycl/src/sanity/contiguous_on.hpp
stephenswat: Okay, as long as it is still correct I am happy!

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
krasznaa: However in this file I didn't see much benefit to using a proxy syntax, even now. :thinking:

stephenswat:
Okay, I guess we need to get vecmem 1.10.0 and then we can put this in, right?

core/include/traccc/clusterization/impl/measurement_creation.ipp
krasznaa: Done.

core/include/traccc/edm/silicon_cell_collection.hpp
krasznaa: Unfortunately the code didn't become much simpler, since `==` is not provided by `<=>`. Still, switching `operator<` to `operator<=>` will make the class a bit more usable.

core/include/traccc/edm/impl/silicon_cell_collection.ipp
```diff
@@ -0,0 +1,66 @@
...
+template <typename BASE>
+template <typename T>
+TRACCC_HOST_DEVICE bool silicon_cell<BASE>::operator==(
+    const silicon_cell<T>& other) const {
+
+    return (channel0() == other.channel0()) &&
+           (channel1() == other.channel1()) &&
+           (math::fabs(activation() - other.activation()) < 1e-6f) &&
+           (math::fabs(time() - other.time()) < 1e-6f) &&
+           (module_index() == other.module_index());
```
stephenswat, line 49:
```suggestion
    return operator<=>(other) == std::weak_ordering::equivalent;
```

krasznaa: For the ordering, we are fine with weak ordering. Our algorithms don't care whether cells coming from different truth particles are in one order during reconstruction or in some other.

However, the equality operator is used in our efficiency measurement code extensively. Where we do care about which cell belongs to which truth particle. So for the equality comparison we need strong ordering. But I didn't want to slow down our reconstruction code with strong ordering.

Long story short, I believe the current code is the way to go.

stephenswat: Okay this is actually a fair point. :+1:

stephenswat (approved):
Okay, I think this can go in now.

PR #714, opened 2024-09-24
ref: benchmark toy detector config

niermann999:
Allow the toy detector grids to find all neighboring surfaces, this should increase the number of tracks that can be found and fitted. Furthermore, I restricted the simulation to an eta region, where the tracks should hit a decent number of sensitive surface. Apart from that, the detector is now copied to device instead of using managed memory.

Edit: I also remove the copy of the track states back to host containers and allowed openMP to use dynamic scheduling for better load balancing

### DIFF
--- a/benchmarks/common/benchmarks/toy_detector_benchmark.hpp
+++ b/benchmarks/common/benchmarks/toy_detector_benchmark.hpp
@@ -53,13 +53,12 @@ class ToyDetectorBenchmark : public benchmark::Fixture {
     traccc::seedfinder_config seeding_cfg;
     traccc::seedfilter_config filter_cfg;
     traccc::spacepoint_grid_config grid_cfg{seeding_cfg};
-    traccc::finding_config finding_cfg;
+    traccc::finding_config finding_cfg = get_trk_finding_config();
     traccc::fitting_config fitting_cfg;
 
     static constexpr std::array<float, 2> phi_range{
         -traccc::constant<float>::pi, traccc::constant<float>::pi};
-    static constexpr std::array<float, 2> theta_range{
-        0.f, traccc::constant<float>::pi};
+    static constexpr std::array<float, 2> eta_range{-3, 3};
     static constexpr std::array<float, 2> mom_range{
         10.f * traccc::unit<float>::GeV, 100.f * traccc::unit<float>::GeV};
 
@@ -101,7 +100,7 @@ class ToyDetectorBenchmark : public benchmark::Fixture {
         generator_type::configuration gen_cfg{};
         gen_cfg.n_tracks(n_tracks);
         gen_cfg.phi_range(phi_range);
-        gen_cfg.theta_range(theta_range);
+        gen_cfg.eta_range(eta_range);
         gen_cfg.mom_range(mom_range);
         generator_type generator(gen_cfg);
 
@@ -130,6 +129,8 @@ class ToyDetectorBenchmark : public benchmark::Fixture {
         // Set constrained step size to 1 mm
         sim.get_config().propagation.stepping.step_constraint =
             1.f * detray::unit<float>::mm;
+        // Otherwise same propagation configuration for sim and reco
+        sim.get_config().propagation = finding_cfg.propagation;
 
         sim.run();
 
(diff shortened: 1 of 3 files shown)


niermann999:
As soon as I let the grid actually search for neighbouring surfaces, the perfromance gets really bad...
```
Running ./bin/traccc_benchmark_cpu
Run on (48 X 2551.74 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x24)
  L1 Instruction 32 KiB (x24)
  L2 Unified 512 KiB (x24)
  L3 Unified 32768 KiB (x4)
Load Average: 0.92, 0.66, 1.04

Detector check: OK
-----------------------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------
ToyDetectorBenchmark/CPU 1.8804e+10 ns   1.3231e+10 ns            1 event_throughput_Hz=7.55819/s
```
Running ./bin/traccc_benchmark_cuda
Run on (48 X 2270.35 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x24)
  L1 Instruction 32 KiB (x24)
  L2 Unified 512 KiB (x24)
  L3 Unified 32768 KiB (x4)
Load Average: 1.80, 2.18, 1.64

Detector check: OK
------------------------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations UserCounters...
------------------------------------------------------------------------------------
ToyDetectorBenchmark/CUDA 1.3547e+10 ns   1.3471e+10 ns            1 event_throughput_Hz=7.42362/s
```

File: io/src/read_detector.cpp [hunk 1]
```diff
@@ -20,7 +20,8 @@
 namespace {
 
 /// Common implementation for constructing a detector from a set of input files
-template <typename detector_t>
+/// @tparam CAP Grid bin capacity 0: dynamic bin capacity
+template <typename detector_t, unsigned int CAP = 0u>
```
krasznaa, line 24: I expected that the PR would already be making use of this. 🤔

Unfortunately this is a pretty inconvenient API. 😦 Normally you don't need to explicitly define the `detector_t` type, as the compiler can deduce it from your call. But now, if you want to override the second value, you need to define `detector_t` explicitly as well. 😦

Not to mention that this is sitting in the `.cpp` file. Users will not be able to call this thing anyway.

So in this PR, it would be better to not introduce a new template parameter to this helper function, but just to hardcode the `0u` value on line 41. 🤔

benchmarks/cuda/toy_detector_cuda.cpp
```diff
@@ -84,8 +82,10 @@ BENCHMARK_F(ToyDetectorBenchmark, CUDA)(benchmark::State& state) {
     traccc::cuda::fitting_algorithm<device_fitter_type> device_fitting(
         fitting_cfg, mr, async_copy, stream);
 
+    // Copy detector to device
+    auto det_fixed_buff = detray::get_buffer(det, device_mr, copy);
```
krasznaa, line 86: Not sure why it's `det_fixed_buff`. I kept looking for what you "fixed" in it. 🤔 Just using `det_buff` would be a bit better of a name I think.

krasznaa (changes requested):
You'll think twice about asking for my review on the next PR. 😛

benchmarks/cuda/toy_detector_cuda.cpp
niermann999: I think I just copied that from somewhere where it meant "fixed size buffer". I will rename it

benchmarks/common/benchmarks/toy_detector_benchmark.hpp
```diff
@@ -155,6 +160,20 @@ class ToyDetectorBenchmark : public benchmark::Fixture {
...
+        // Configure the propagation for the toy detector
+        cfg.propagation.navigation.search_window = {3, 3};
+        cfg.propagation.navigation.overstep_tolerance =
+            -300.f * detray::unit<float>::um;
+        cfg.propagation.navigation.min_mask_tolerance =
+            1e-5f * detray::unit<float>::mm;
+        cfg.propagation.navigation.max_mask_tolerance =
+            3.f * detray::unit<float>::mm;
+        cfg.propagation.navigation.mask_tolerance_scalor = 0.05f;
+    }
```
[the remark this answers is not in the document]
niermann999: Indeed

krasznaa (approved):
Looks technically correct.

beomki-yeo:
This PR seems to stop the continuous benchmark.

PR #715, opened 2024-09-27
Update detray to v0.76.1

beomki-yeo:


stephenswat:
I see that detray 0.75.3 and 0.76.0 have both been tagged without updating
the version in the CMake file. Can you please get into the habit of
updating that?

beomki-yeo:
I am sorry :cry:

stephenswat (approved):

PR #716, opened 2024-09-27
Skip to fill the resolution histogram if the idx is out of range

beomki-yeo:
Sometimes the eta has a value out of the range defined in histogram which causes an error.


beomki-yeo:
As I am not seeing this error right now..

PR #719, opened 2024-09-30
Refactor the spacepoint formation

beomki-yeo:
This are the changes that I have been suggesting for long time.

1. `spacepoint_formation` takes detray detector instead of detector description. Direct use of `point_to_global` with `transform_matrix` should be avoided. Most importantly, the local-to-global transformation of annulus shape (Polar coordinate) does not work with `point_to_global` which only works for Cartesian coordinate.

2. As detector description is not required (and should not be used), I moved its directory to `seeding` where the `experimental::speacepoing_formation` is located. This makes sense because spacepoints are technically only used for seeding. 

~~+ Spacepoint formation is only applied to the 2D measurements as we do not support 1D measurements sp formation. Supporting 1D measurement will need to be followed up by further studies~~

Let me have 2D condition in the next PR (to test the performance without changing the actual results...)


device/alpaka/src/seeding/spacepoint_formation_algorithm.cpp
```diff
@@ -67,13 +70,15 @@ spacepoint_formation_algorithm::operator()(
...
+    ::alpaka::exec<Acc>(queue, workDiv, FormSpacepointsKernel<detector_t>{},
+                        det_view, measurements_view, num_measurements,
                         vecmem::get_data(spacepoints));
     ::alpaka::wait(queue);
 
     // Return the reconstructed spacepoints.
     return spacepoints;
 }
+// Explicit template instantiation
+template class spacepoint_formation_algorithm<default_detector::device>;
```
krasznaa, line 82: :smile: I'm very happy that you made use of the typedefs that I added a while ago. As this was exactly my idea with them.

However, while this way of writing the code does work, I think the user interface provided like this is not the easiest to understand. :frowning: I think we should provide explicit typedefs in the header files for the template specializations that we would actually provide in the different libraries. I.e. I'd put something like the following into the algorithm headers:

```c++
using default_spacepoint_formation_algorithm = spacepoint_formation_algorithm<default_detector::device>;
```

So that as long as the user only used existing typedefs, they could be sure that the algorithm would be available in the library. :thinking:

device/common/include/traccc/seeding/device/impl/form_spacepoints.ipp
```diff
@@ -28,17 +27,18 @@ inline void form_spacepoints(
...
     assert(measurements.size() == measurement_count);
-    const silicon_detector_description::const_device det_descr(det_descr_view);
     spacepoint_collection_types::device spacepoints(spacepoints_view);
-    assert(spacepoints.size() == measurement_count);
 
     // Fill the spacepoint using the common function.
-    details::fill_spacepoint(spacepoints.at(globalIndex),
-                             measurements.at(globalIndex), det_descr);
+    spacepoints.push_back(
+        details::create_spacepoint(det, measurements.at(globalIndex)));
```
krasznaa: It's a good idea to make the spacepoint container resizable. Since we don't know how many spacepoints we'll actually create from all of the measurements. But for this, we need to:
  - Make sure that all algorithms would create resizable spacepoint buffers. (Make use of `vecmem::data::buffer_type::resizable`!)
  - Make sure in this function that we would skip the non-2D measurements. Right now I believe the code will call the `create_spacepoint(...)` function for strips as well. :thinking:

device/cuda/src/seeding/spacepoint_formation_algorithm.cu
```diff
@@ -60,12 +64,17 @@ spacepoint_formation_algorithm::operator()(
...
     TRACCC_CUDA_ERROR_CHECK(cudaGetLastError());
 
     // Return the reconstructed spacepoints.
     return spacepoints;
 }
 
+// Explicit template instantiation
+template class spacepoint_formation_algorithm<default_detector::device>;
+template class spacepoint_formation_algorithm<telescope_detector::device>;
+template class spacepoint_formation_algorithm<toy_detector::device>;
```
krasznaa, line 78: Ahh, I didn't even see this when I made my earlier comment. This is exactly why we should have explicit typedefs in the headers. To make it clear which specializations exist for which library...

krasznaa (changes requested):
It's not the greatest that we still have a lot of "if Detray detector then ..." checks in the code, but I think you're right. Let's be pragmatic, and just make the code work for now. It will indeed be better to get rid of the TML setup in follow-up steps.

Generally I'm happy with how you went about this, just had a couple of comments on a quick look. :thinking:

device/common/include/traccc/seeding/device/impl/form_spacepoints.ipp
beomki-yeo: Sure. A couple of tests (`test_spacepoint_formation` for CPU/CUDA/SYCL) already make sure that they fail without the `resizable` tag.

I decided to drop the 2D condition in this PR and add it into the next one

beomki-yeo:
The performance remains the same

Main

```
Reconstructed track parameters: 3299715
Time totals:
                  File reading  617 ms
            Warm-up processing  147130 ms
              Event processing  4809 ms
Throughput:
            Warm-up processing  14713 ms/event, 0.067967 events/s
              Event processing  48.0923 ms/event, 20.7934 events/s
```

PR

```
Reconstructed track parameters: 3299774
Time totals:
                  File reading  618 ms
            Warm-up processing  938 ms
              Event processing  4811 ms
Throughput:
            Warm-up processing  93.8457 ms/event, 10.6558 events/s
              Event processing  48.1174 ms/event, 20.7825 events/s
```

device/alpaka/src/seeding/spacepoint_formation_algorithm.cpp
beomki-yeo: Do you mean making a new header file for these kind of typedefs?
```
using default_spacepoint_formation_algorithm = spacepoint_formation_algorithm<default_detector::device>;
using default_finding_algorithm = finding_algorithm<default_stepper_type, default_navigator_type>;
... etc.
```

I like the idea but let's do this in another PR. Doing this for cpu, cuda, sycl, and alpaka is already a huge task :skull:

krasznaa: No, not a separate header.

I'm trying to figure out the best setup myself as well... :thinking: I have an idea right now, but it will need some experimentation. So, let's just go with this type of templating for now, and I'll propose a slightly different code layout in a follow-up.

device/common/include/traccc/seeding/device/impl/form_spacepoints.ipp
krasznaa: It's a bit of a pity to switch to resizable buffers, for no benefit right now. :thinking: Since filling resizable buffers is always slower. But if we don't know their size, that's of course what we need to accept. :thinking:

In any case: You forgot to make the buffer resizable in the Alpaka algorithm. At least that needs to be fixed!

beomki-yeo: Right - fixed now

krasznaa (approved):
Okay, let's go like this for now. As I'm on board with the overall direction.

PR #720, opened 2024-10-01
Lower the values of qop stddevs

beomki-yeo:
The current stddev of qop is high which can change the sign of qop. This PR lowers the values of qop stddevs.


krasznaa:
Sure, no objections here. Let me know once it goes out of draft status.

beomki-yeo:
Yup, but it seems there is a fundamental problem deeper inside - investigating..

beomki-yeo:
When the qop of initial track parameter gets very small (~0.001) by the high stddev, the sign of qop can change in the kalman updater. 

While the exact reason is unknown, I guess that the high deviation from truth value causes kalman fitter to make large numerical uncertainties.

beomki-yeo:
Actually the stddev of q/p need to be given as something like 10% of truth q/p instead of a hard coded value. (Current examples will still break for >1000 GeV/c particles) But let's do that in the later PR

beomki-yeo:
@krasznaa review?

PR #722, opened 2024-10-01
De-Template Track Finding, main branch (2024.10.01.)

krasznaa:
So... this is how I've been imagining for a while now, how we should provide the reconstruction algorithms/tools. Not exposing any template based configurability in the APIs of the algorithms. :thinking:

I wanted to demonstrate this on the track finding, because:
  - it might just be our most elaborate piece of code;
  - contrary to how it looks, we actually only ever used a single template specialization of it so far.

I moved everything from `finding_algorithm.ipp` into a private header under `core/src/finding/`. That code is still implemented as a templated function, with the same level of configurability that `traccc::finding_algorithm` had. But now it will be up to us to create well documented user APIs to that code, in the form of algorithms that perform one specific type of track finding each. :thinking:

On the naming front I'm very open to good ideas. The name I chose now (`traccc::host::default_detector_const_field_ckf_algorithm`) is just meant to demonstrate that we may need to make the algorithm names quite descriptive in the end. Depending on how many different setups we may come up with.

I believe this code organization could be used for all algorithms that use `detray::detector`. What do you lot think?

P.S. I also tweaked the algorithmic code a little. Replacing some simple for-loops with some STL algorithms. I don't think it would change the code's performance, but I agree with @stephenswat that it's generally good practice to use STL algorithms wherever we can.


core/include/traccc/finding/default_detector_const_field_ckf_algorithm.hpp
```diff
@@ -0,0 +1,63 @@
...
+        const default_detector::host& det,
+        const detray::bfield::const_field_t::view_t& field,
+        const measurement_collection_types::const_view& measurements,
+        const bound_track_parameters_collection_types::const_view& seeds) const;
+
+    private:
+    /// Algorithm configuration
+    config_type m_config;
+
+};  // class default_detector_ckf_algorithm
```
stephenswat, line 61: I'm mostly on board, but why can't this just be a specialization of the old template type? Surely that would reduce the amount of boilerplate if we're going to end up with many different versions of this code (e.g. one constant field, one non-constant field, one combinatorial, one non-combinatorial, etc.)?

krasznaa: It's a question of UI for me. :thinking: Currently I find it super hard to even just instantiate a functional finding algorithm. I have to basically copy-paste a configuration that somebody else wrote. And having users blindly copy-paste stuff is not a good setup. :frowning:

My first idea was of course just to rename `finding_algorithm.ipp` to `finding_algorithm.cpp`, instantiate a couple of concrete types in it, and be done. (Like what we do with the CUDA version at the moment.) But I just don't like that UI... :thinking:

stephenswat: Okay, that's fine for me. I am happy to accept changes to these host algorithms as they will be deprecated in the not-so-distant future anyway.

But for the CUDA (and in the future SYCL) code I would like to come up with a solution that doesn't require so much boilerplate. :thinking:

beomki-yeo (changes requested):
Didn't read the full PR - but I don't think I like this
Can't you just make a header file that defines the types that we want

```c++
// algorithms.hpp
using host_default_detector_const_field_ckf_algorithm = host::finding_algorithm<default_navigator_type, default_stepper_type>
using cuda_default_detector_const_field_ckf_algorithm =
cuda::finding_algorithm<default_navigator_type, default_stepper_type>
.. etc
```

core/include/traccc/finding/default_detector_const_field_ckf_algorithm.hpp
beomki-yeo: > as they will be deprecated in the not-so-distant future anyway

Not related to the discussion, but why are you going to deprecate the host algorithms? That's kinda scary

stephenswat: I was under the impression that we were all on board of deprecating the traccc CPU algorithms in favour of the Acts CPU algorithms... Do you not agree? :confused:

beomki-yeo: Not at all. You don't have to use them in ACTS or other places, but please don't delete them

stephenswat: Could you motivate your point of view a bit? I'm happy to accept that you want to keep the algorithms around, sure, but I'm trying to understand why. We will need to keep this in mind when we integrate traccc into ACTS.

beomki-yeo: They simply do their roles by:
1. Maintaining the agreement between CPU and GPU in traccc
2. Debugging any physical issue, common to CPU and GPU, in an easier way (Surprisingly, compiling and testing host algorithms is easier than doing it for GPU
3. Optimizing the tracking performance in an easier way

Most importantly they are not 100% identical to ACTS CPU algorithms, so the above things will get super difficult if they are gone.

beomki-yeo: I am on board with adding a CMake parameter to turn off building host algorithms , if that can meet your needs

krasznaa: :confused: The project's host algorithms are not going anywhere until:
  - The common CPU/GPU code sits in `ActsCore`;
  - `ActsCore` provides an equivalent of what each GPU algorithm is doing.

The design that we have in this project, where we share code between the CPU and GPU codebase, has to survive when we integrate the code into Acts. I see that as a hard requirement.

Yes, eventually the separate traccc host algorithms will disappear. But I don't see that as anything happening soon... :thinking:

krasznaa:
> Didn't read the full PR - but I don't think I like this Can't you just make a header file that defines the types that we want
> 
> ```c++
> // algorithms.hpp
> using host_default_detector_const_field_ckf_algorithm = host::finding_algorithm<default_navigator_type, default_stepper_type>
> using cuda_default_detector_const_field_ckf_algorithm =
> cuda::finding_algorithm<default_navigator_type, default_stepper_type>
> .. etc
> ```

I'm pondering about this still. :thinking: I worry about the understandability of the API. Let me tweak the proposal a bit more...

core/include/traccc/finding/default_detector_const_field_ckf_algorithm.hpp
stephenswat: Okay, it's becoming clear to me that it would be beneficial for us all to sit down and talk about the future of traccc during either one of our bi-weekly meetings or during the ACTS developer's workshop, because I don't believe adding traccc to `ActsCore` is really on anybody's radar right now. We'll have to discuss this to produce a clear working plan.

krasznaa:
You guys might like this version even less, but here it goes... How about using some good old function overloading? :thinking:

Again, I'd really like to avoid having templating "at the algorithm level". :thinking:

krasznaa:
Once we sort out #713, I'll use that to characterize these changes in a bit more detail. 🤔

krasznaa:
So... Now that #713 can help with characterizing the changes of this PR.
  - In the current code, to get track finding into `traccc_seq_example`, the build of [seq_example.cpp](https://github.com/acts-project/traccc/blob/main/examples/run/cpu/seq_example.cpp) does everything. And it uses 1.55 GB of memory to do so. (With GCC 11.)
  - With this PR's code, bringing track finding to life in `traccc_seq_example` is split among the following source files:
    * `ckf_algorithm.cpp`: 350 MB
    * `ckf_algorithm_teldet_cfield.cpp`: 560 MB
    * `ckf_algorithm_defdet_cfield.cpp`: 810 MB
    * `seq_example.cpp`: 1.43 GB

So, we win a little more than 100 MB on compiling `seq_example.cpp`. While adding 3 more source files, with large, but semi-acceptable memory needs each.

As I discussed it with @stephenswat before lunch, I expect that after de-templating spacepoint creation and track fitting, `seq_example.cpp` should go down below 1 GB of memory. But I have no hard proof that it will actually happen.

As an interesting tidbit: I didn't do an exhaustive search yet for the biggest memory users, but [read_detector.cpp](https://github.com/acts-project/traccc/blob/main/io/src/read_detector.cpp), with 1.8 GB, seems to be a surprisingly large one. Clearly, constructing a geometry from JSON files is not an easy task for the compiler. 🤔

For @paulgessinger: Linking `libActsCore.so` takes a little more than 1 GB of memory. Underlining my previous statements that it's good to keep an eye on the linking steps as well. 🤔

krasznaa:
@stephenswat, @beomki-yeo, have a look again. How about organizing the code like this?

paulgessinger:
> For @paulgessinger: Linking `libActsCore.so` takes a little more than 1 GB of memory. Underlining my previous statements that it's good to keep an eye on the linking steps as well. 🤔

I've been looking into using CMake launchers for this (COMPILER, LINKER), that avoids having to abuse `CTEST_USE_LAUNCHERS` for this purpose, and is non-intrusive.

krasznaa:
Okay, taking an executive decision here... 🤔

PR #723, opened 2024-10-01
Update detray to v0.78.0

beomki-yeo:


beomki-yeo:
@niermann999 

Probably we need to fix this in detray for HIP compilation
```
/__w/traccc/traccc/build/_deps/detray-src/core/include/detray/utils/bit_encoder.hpp:48:30: error: cannot take address of consteval function 'extract_shift' outside of an immediate invocation
        return (v & mask) >> extract_shift(mask);
                             ^
/__w/traccc/traccc/build/_deps/detray-src/core/include/detray/utils/bit_encoder.hpp:65:26: note: declared here
    static consteval int extract_shift(value_t mask) noexcept {
                         ^
/__w/traccc/traccc/build/_deps/detray-src/core/include/detray/utils/bit_encoder.hpp:57:35: error: cannot take address of consteval function 'extract_shift' outside of an immediate invocation
        v = (v & ~mask) | ((id << extract_shift(mask)) & mask);
                                  ^
/__w/traccc/traccc/build/_deps/detray-src/core/include/detray/utils/bit_encoder.hpp:65:26: note: declared here
    static consteval int extract_shift(value_t mask) noexcept {
```                         ^

stephenswat:
Given that we have exactly zero (0) useful HIP code, we should probably just drop support for it at this point rather than spending valuable developer time on it, to be honest.

beomki-yeo:
The detray version is changed to v78.

There was an error in `event_map2` test (occurred from v77) which I don't understand pretty well. Given that the used event file in `/data/detray_simulation` is too old to be used and the `event_map2` will be replaced in the near future, I decided to remove the test

PR #726, opened 2024-10-03
Not using detector description for partial chain examples

beomki-yeo:
Currently, the examples of partial chains make an error when the geometry json and digitizatoin config json are pointing to different geometry.

Problem is some geometries (e.g. ITk) that we are using do not have digitization config yet so it will always clash with the default one from TML detector. 

This PR disables reading the digitization config file, which is not used in the algorithm anyway.

### DIFF
--- a/examples/run/alpaka/seeding_example_alpaka.cpp
+++ b/examples/run/alpaka/seeding_example_alpaka.cpp
@@ -120,14 +120,6 @@ int seq_run(const traccc::opts::track_seeding& seeding_opts,
      * Build a geometry
      *****************************/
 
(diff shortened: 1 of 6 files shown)


krasznaa (approved):
I believe this is correct. In these cases the CSV reading code only needs `traccc::silicon_detector_description` to set the "module index" correctly on the spacepoints, measurements, etc. But none of these executables need "module index" to be set to a valid value anyway. :thinking:

beomki-yeo:
Yeah I also tried to find which algorithms use `module_link` after the clusterization, but it seems none of them do :p

PR #727, opened 2024-10-05
Apply spacepoint formation to the 2D pixel detector only

beomki-yeo:
To avoid the repeat the condition of `if (meas.meas_dim == 2u)` in host and device code, I changed the function of `create_spacepoint` to `fill_spacepoint`


core/include/traccc/seeding/detail/spacepoint_formation.hpp
```diff
@@ -16,12 +16,14 @@ namespace traccc::details {
...
+/// @param[in]   measurement  The input measurement
+/// @param[out]  spacepoints  The spacepoint vector to be filled
 ///
-template <typename detector_t>
-TRACCC_HOST_DEVICE inline spacepoint create_spacepoint(const detector_t& det,
-                                                       const measurement& meas);
+template <typename detector_t, template <typename...> class vector_t>
+TRACCC_HOST_DEVICE inline void fill_spacepoint(
+    const detector_t& det, const measurement& meas,
+    vector_t<spacepoint>& spacepoints);
```
stephenswat, line 26:
```suggestion
template <typename detector_t, typename sp_vector_t>
TRACCC_HOST_DEVICE inline void fill_spacepoint(
    const detector_t& det, const measurement& meas,
    sp_vector_t& spacepoints);
```

stephenswat (changes requested):
Looks good, but let's not unnecessarily complicate the templating.

core/include/traccc/seeding/detail/spacepoint_formation.hpp
krasznaa: Why does it need to be templated code at all? 🤔 With the current EDM I would write this function as:

```suggestion
template <typename detector_t>
TRACCC_HOST_DEVICE inline bool is_valid_spacepoint(
    const detector_t& det, const measurement& meas);
template <typename detector_t>
TRACCC_HOST_DEVICE inline void fill_spacepoint(
    const detector_t& det, const measurement& meas,
    spacepoint& spacepoint);
```

And then leave it up to the caller to push new default spacepoints into their containers. Note that this formalism will work with eventual SoA containers as well.

I personally am not a big fan of either of the templating methods, as the implementation will still be AoS or SoA specific. So we don't win much by being generic in the interface.

krasznaa (commented):
At the same time I don't feel super strongly about this point. Not enough to veto either of your proposals...

core/include/traccc/seeding/detail/spacepoint_formation.hpp
beomki-yeo: I just added a function of `is_valid_measurement` and leave `create_spacepoint()` as it is for now.

krasznaa (approved):
:+1:

When doing the SoA migration, I can imagine adding some additional assertions here and there, but this is good with me as-is.

PR #728, opened 2024-10-07
Restructure device CKF for readability

stephenswat:
This commit doesn't change the CKF much functionally, but improves its readability a bit by moving variables closer to where they are used, and by adding smaller scopes. Also increases the clarity of the double buffering solution.

May decrease memory consumption slightly as it encourages the compiler to free memory earlier.


device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -286,76 +252,111 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
-        // Set the number of input parameters
-        const unsigned int n_in_params = (step == 0)
-                                             ? in_params_buffer.size()
-                                             : global_counter_host.n_out_params;
+    // Create a map for parameter ID to link ID
+    std::map<unsigned int, vecmem::data::vector_buffer<unsigned int>>
+        param_to_link_map;
 
-        // Terminate if there is no parameter to process.
-        if (n_in_params == 0) {
```
beomki-yeo, line 320: Is it OK to remove this or did you move it to somewhere

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -286,76 +252,111 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
          * Kernel2: Apply material interaction
          ****************************************************************/
 
-        nThreads = m_warp_size * 2;
-        nBlocks = (n_in_params + nThreads - 1) / nThreads;
-        kernels::apply_interaction<detector_type>
-            <<<nBlocks, nThreads, 0, stream>>>(det_view, m_cfg, n_in_params,
-                                               in_params_buffer);
-        TRACCC_CUDA_ERROR_CHECK(cudaGetLastError());
+        {
```
beomki-yeo, line 315: I suppose theses parentheses are for readibility?

beomki-yeo (approved):
Looks good to me. Just a couple of questions

device/cuda/src/finding/finding_algorithm.cu
stephenswat: It's in the for loop now:

```cpp
for (unsigned int step = 0;
         step < m_cfg.max_track_candidates_per_track && n_in_params > 0;
         step++)
```

device/cuda/src/finding/finding_algorithm.cu
stephenswat: Yes, and also for scopes, so we can redeclare `nThreads` and `nBlocks` for the next algorithms.

device/cuda/src/finding/finding_algorithm.cu
beomki-yeo: OK

device/cuda/src/finding/finding_algorithm.cu
beomki-yeo: OK this makes sense

device/common/include/traccc/finding/device/build_tracks.hpp
```diff
@@ -39,9 +39,9 @@ TRACCC_DEVICE inline void build_tracks(
         tips_view,
     track_candidate_container_types::view track_candidates_view,
     vecmem::data::vector_view<unsigned int> valid_indices_view,
-    unsigned int& n_valid_tracks);
+    unsigned int* n_valid_tracks);
```
krasznaa, line 42: Why? We should only pass variables by pointer that are allowed to be `nullptr`. When something can only be a valid reference, we might as well make it clear in the API. 🤔

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -91,7 +91,7 @@ __global__ void find_tracks(
     const unsigned int step, const unsigned int n_max_candidates,
     bound_track_parameters_collection_types::view out_params_view,
     vecmem::data::vector_view<candidate_link> links_view,
-    unsigned int& n_candidates) {
+    unsigned int* n_candidates) {
```
krasznaa, line 94: To be clear, I'm absolutely happy about making this into a pointer. On the `__global__` function we should use pointers indeed, as it makes the intentions clearer. But on `__device__` functions I much prefer references.

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -204,80 +204,46 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
-                m_mr.main);
-
-    // Global counter object in Host memory
-    device::finding_global_counter global_counter_host;
-
     /*****************************************************************
      * Measurement Operations
      *****************************************************************/
 
+    unsigned int n_modules;
```
krasznaa, line 211: I agree that this is correct. But I'm a bit surprised that we didn't get any compiler warnings about the variable being uninitialized. 🤔 I would consider:

```suggestion
    unsigned int n_modules{};
```

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -204,80 +204,46 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
-                            uniques.begin(), measurement_equal_comp());
-    unsigned int n_modules = end - uniques.begin();
+    {
+        assert(is_contiguous_on<measurement_collection_types::const_device>(
+            measurement_module_projection(), m_mr.main, m_copy, m_stream,
+            measurements));
+
+        measurement_collection_types::device uniques(uniques_buffer);
+
+        measurement* end = thrust::unique_copy(
```
krasznaa, line 226: I completely agree with the intention here. But I would go even one step further, as I have in #722, and call it `uniques_end` instead of `end`.

Since the scope is limited now, `end` is technically good enough. Still, as long as we're going for readability... 🤔

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -286,76 +252,111 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
-        // Reset the number of tracks per seed
-        m_copy.memset(n_tracks_per_seed_buffer, 0)->ignore();
+    {
+        bound_track_parameters_collection_types::device in_params(
+            in_params_buffer);
+        bound_track_parameters_collection_types::const_device seeds(
+            seeds_buffer);
+        thrust::copy(thrust::cuda::par.on(stream), seeds.begin(), seeds.end(),
+                     in_params.begin());
+    }
```
krasznaa, line 279: Why are you using `thrust::copy` here instead of vecmem's copy object? 😕

krasznaa (changes requested):
After every new vecmem buffer, you should always call `vecmem::copy::setup(...)->ignore()`! For non-resizable 1D buffers this is not technically not necessary, but it would help my OCD if we finally started doing this consistently. 🤔

Generally on board with the PR of course.

device/cuda/src/finding/finding_algorithm.cu
stephenswat: This is how the code already was! I didn't change anything here.

device/cuda/src/finding/finding_algorithm.cu
stephenswat: There's no code path in which this is used before it is assigned, so it's not a compiler warning (nor is it a problem).

device/common/include/traccc/finding/device/build_tracks.hpp
stephenswat: Fair, updated. :+1:

device/cuda/src/finding/finding_algorithm.cu
beomki-yeo: We can change it to vecmem::copy for sure

stephenswat: Trying to figure out the right syntax for this.

stephenswat:
Updated. :+1:

beomki-yeo:
I think @krasznaa would be OK now, so let me merge it

PR #729, opened 2024-10-07
Use surface link for module projection

beomki-yeo:
I think module links do not have to be used after clusterization. This also helps the currect refactoring of #692

### DIFF
--- a/core/include/traccc/clusterization/impl/measurement_creation.ipp
+++ b/core/include/traccc/clusterization/impl/measurement_creation.ipp
@@ -129,7 +129,6 @@ TRACCC_HOST_DEVICE inline void fill_measurement(
     const auto module_dd = det_descr.at(module_idx);
 
     // Fill the measurement object.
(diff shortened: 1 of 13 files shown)


beomki-yeo:
Actually `module_link` can be dropped from the measurement as it is not used anywhere but some tests - this will be followed up later

beomki-yeo:
@stephenswat Any thought?

krasznaa:
I believe this is fine. But this PR by itself is really meaningless in my mind. For this to be meaningful, you should:
  - Remove the now unnecessary variable;
  - Remove the code that fills that unnecessary variable;
  - Do this change, which I believe is the last remaining client for this variable.

Could you add the first 2 points to this PR as well?

beomki-yeo:
OK I removed the module link of measurements

beomki-yeo:
@krasznaa Could you review the PR?

krasznaa (approved):
👍

PR #730, opened 2024-10-07
Fix register overload in CUDA spacepoint formation

stephenswat:
PR #719 refactored the spacepoint formation to use the detray detector. While this was and remains a good idea, the introduction of the detray detector massively increases the register use of the kernel, to the point that I was seeing "too many resources requested for launch" errors. This commit modifies the kernel to specify appropriate launch bounds, allowing the compiler to better tune the register usage and resolve the bug.

### DIFF
--- a/device/cuda/src/seeding/spacepoint_formation_algorithm.cu
+++ b/device/cuda/src/seeding/spacepoint_formation_algorithm.cu
@@ -18,11 +18,11 @@ namespace traccc::cuda {
 namespace kernels {
 
 template <typename detector_t>
-__global__ void form_spacepoints(
-    typename detector_t::view_type det_view,
-    measurement_collection_types::const_view measurements_view,
-    const unsigned int measurement_count,
-    spacepoint_collection_types::view spacepoints_view) {
+__global__ void __launch_bounds__(1024, 1)
+    form_spacepoints(typename detector_t::view_type det_view,
+                     measurement_collection_types::const_view measurements_view,
+                     const unsigned int measurement_count,
+                     spacepoint_collection_types::view spacepoints_view) {
 
     device::form_spacepoints<detector_t>(threadIdx.x + blockIdx.x * blockDim.x,
                                          det_view, measurements_view,


device/cuda/src/seeding/spacepoint_formation_algorithm.cu
```diff
@@ -18,11 +18,11 @@ namespace traccc::cuda {
 namespace kernels {
 
 template <typename detector_t>
-__global__ void form_spacepoints(
-    typename detector_t::view_type det_view,
-    measurement_collection_types::const_view measurements_view,
-    const unsigned int measurement_count,
-    spacepoint_collection_types::view spacepoints_view) {
+__global__ void __launch_bounds__(1024, 1)
```
krasznaa, line 21: I don't understand this. 😕 This says that the kernel must be launched with `<=` 1024 threads per block, and at least one block, no? But that's not a restriction at all... 😕

Or did I completely misunderstand what these numbers are?

https://docs.nvidia.com/cuda/cuda-c-programming-guide/#launch-bounds

Don't get me wrong, I'm very happy to put in guidelines to the compiler like this one. I just don't understand how these exact numbers are helping things.

stephenswat: Without this specifier, the compiler is free to optimize the kernel for running with, e.g. one block of 512 threads. So out of your 65536 threads per SM, it may choose to use $\frac{65536}{512} = 128$ per thread. If we then launch the kernel with 1024 threads, we're using $128 \times 1024 = 131072$ threads, exceeding the limit.

This statement tells the compiler that at least one block of 1024 threads must be able to run per SM, so it can use a maximum of 64 registers per thread.

Unfortunately the register usage of detray is a bit ludicrous, so these kinds of measures are necessary. :cry:

krasznaa: Okay, I trust you on this. 👍

krasznaa (approved):

PR #731, opened 2024-10-07
Change the particle hypo of KF in case of charge is changed

beomki-yeo:
@stephenswat Can you test this if it does work? Otherwise please let me know the command line so I can reproduce


stephenswat:
I tested this, but now the CKF seems to run forever... Very odd!

beomki-yeo:
You mean KF runs forever? Or is this something independent

beomki-yeo:
With this PR, I can run `mt_cuda` with ODD mu200 data (RelWithDebInfo build). Probably you need to wait a couple of minute because it took about 2 mins for warming up

stephenswat (approved):
Okay, then I'm happy. Thanks!

PR #732, opened 2024-10-07
Make CKF propagation a 1:1 kernel

stephenswat:
Right now, we have two big kernels in the CKF. The first is `find_tracks`, which takes $N$ tracks and turns it into $M$ tracks, where $M$ can be either smaller than, equal to, or greater than $M$. The second kernel is the propagation kernel, which takes $N$ tracks and turns those into $M$ tracks, where $M$ is equal to or smaller than $N$, but never greater.

The problem with kernels where the number of outputs differs from the number of inputs is that you need to atomically insert the elements into the output, and then you lose the ordering of your inputs.

In this PR, I make propagation a 1:1 kernel, e.g. the size of the output is always equal to the size of the input. I do this by keeping track of which kernels are alive.

This has the following benefits:

1. It makes the code easier to understand, as there is a bit less indirection.
2. It makes the code faster, although only by a small ~5% margin.
3. It promotes future optimization by preserving the order of tracks, i.e. we can perhaps try to do sorting only at the start.
4. This model will be useful for the proposed non-combinatorial KF.


device/common/include/traccc/finding/device/impl/apply_interaction.ipp
```diff
@@ -31,24 +33,29 @@ TRACCC_DEVICE inline void apply_interaction(
...
-    const typename detector_t::geometry_context ctx{};
-
-    // Apply interactor
-    typename interactor_type::state interactor_state;
-    interactor_type{}.update(
-        ctx,
-        detail::correct_particle_hypothesis(cfg.ptc_hypothesis, bound_param),
-        bound_param, interactor_state,
-        static_cast<int>(detray::navigation::direction::e_forward), sf);
+    if (params_liveness.at(globalIndex) != 0u) {
```
beomki-yeo, line 45: At the first step of CKF, I think all values of `params_liveness` is 0u, which is a default value, so this will not be passed?

beomki-yeo (commented):
Looks good to me. But I do not get how this works in the apply interaction kernel

device/common/include/traccc/finding/device/impl/apply_interaction.ipp
stephenswat: Right; to make sure that everything works as normal, this sets the values to 1 for all parameters: https://github.com/acts-project/traccc/pull/732/files#diff-90d760f48590c5507580da9993439d4bbaa1758d54e6c05d332631c9e256fbe9R280

beomki-yeo: Ah OK I missed `memset`

beomki-yeo (approved):
I think this should go in for marginal performance gain

krasznaa (approved):
As much as I can tell... 🤔

PR #733, opened 2024-10-08
Fix the bug that the particle hypothesis is not changed as intended

beomki-yeo:
Ah. I found that #731 does nothing


beomki-yeo:
This is a simple bug fix so any approval will be appreciated 🫠

stephenswat (approved):

PR #734, opened 2024-10-08
Enhance the track matching condition

beomki-yeo:
Currently, a track is considered to be unmatched if its measurements are from multiple particles.
However, the ACTS defines that the track is matched if the number of measurements of major particle is larger than the half of the number of measurements. This increases the actual tracking efficieny of `seq_example` slightly

Related code in ACTS:
https://github.com/acts-project/acts/blob/69e0b6b181a14daab9b99e1ae6b576914a060afb/Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackTruthMatcher.cpp#L97

### DIFF
--- a/performance/src/efficiency/finding_performance_writer.cpp
+++ b/performance/src/efficiency/finding_performance_writer.cpp
@@ -173,7 +173,8 @@ void finding_performance_writer::write_common(
         // Consider it being matched if hit counts is larger than the half
         // of the number of measurements
         assert(measurements.size() > 0u);
-        if (particle_hit_counts.at(0).hit_counts > measurements.size() / 2) {
+        if (particle_hit_counts.at(0).hit_counts / measurements.size() >
+            matching_ratio) {
             const auto pid = particle_hit_counts.at(0).ptc.particle_id;
             match_counter[pid]++;
         } else {
--- a/performance/src/efficiency/finding_performance_writer.cpp
+++ b/performance/src/efficiency/finding_performance_writer.cpp
@@ -170,14 +170,15 @@ void finding_performance_writer::write_common(
         std::vector<particle_hit_count> particle_hit_counts =
             identify_contributing_particles(measurements, evt_map.meas_ptc_map);
 
-        if (particle_hit_counts.size() == 1) {
-            auto pid = particle_hit_counts.at(0).ptc.particle_id;
+        // Consider it being matched if hit counts is larger than the half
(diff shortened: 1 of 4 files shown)


performance/src/efficiency/finding_performance_writer.cpp
```diff
@@ -170,14 +170,15 @@ void finding_performance_writer::write_common(
         std::vector<particle_hit_count> particle_hit_counts =
             identify_contributing_particles(measurements, evt_map.meas_ptc_map);
 
-        if (particle_hit_counts.size() == 1) {
-            auto pid = particle_hit_counts.at(0).ptc.particle_id;
+        // Consider it being matched if hit counts is larger than the half
+        // of the number of measurements
+        assert(measurements.size() > 0u);
+        if (particle_hit_counts.at(0).hit_counts > measurements.size() / 2) {
```
asalzburger, line 176: Can we make this configurable?

Usually we define:
- loose (0.5)
- thight (0.7)
- full (1.0)
matching criteria, it would be good to be able to reproduce those.

performance/src/efficiency/seeding_performance_writer.cpp
```diff
@@ -70,12 +70,13 @@ void seeding_performance_writer::write(
...
+        const auto measurements = sd.get_measurements(spacepoints_view);
+
         // Check which particle matches this seed.
         std::vector<particle_hit_count> particle_hit_counts =
-            identify_contributing_particles(
-                sd.get_measurements(spacepoints_view), evt_map.meas_ptc_map);
+            identify_contributing_particles(measurements, evt_map.meas_ptc_map);
 
-        if (particle_hit_counts.size() == 1) {
+        if (particle_hit_counts.at(0).hit_counts > measurements.size() / 2) {
```
asalzburger, line 79: Same here.

asalzburger (changes requested):

performance/src/efficiency/finding_performance_writer.cpp
beomki-yeo: Yup. Added `matching_ratio` now

performance/src/efficiency/seeding_performance_writer.cpp
beomki-yeo: Same

asalzburger (approved):
Happy now!

PR #735, opened 2024-10-08
Improve the seeding and finding performance

beomki-yeo:
The efficiency drop is from bad seeding config (which was optimized for tml detector) and loose particle selection in performance writers. I also synchronized the truth cut for seeding and finding performance writer

After fixing them, good efficiency is obtained for single muon ODD events

```
./bin/traccc_seq_example --digitization-file=geometries/odd/odd-digi-geometric-config.json --detector-file=geometries/odd_full/odd-detray_geometry_detray.json --grid-file=geometries/odd_full/odd-detray_surface_grids_detray.json --material-file=geometries/odd_full/odd-detray_material_detray.json --use-detray-detector --use-acts-geom-source --input-directory=odd/geant4_10muon_100GeV/ --input-events=100 --check-performance
```

image

### DIFF
--- a/core/include/traccc/seeding/detail/seeding_config.hpp
+++ b/core/include/traccc/seeding/detail/seeding_config.hpp
@@ -19,12 +19,8 @@ struct seedfinder_config {
     seedfinder_config() { setup(); }
 
     // limiting location of measurements
-    // Beomki's note: this value introduces redundant bins
-    // without any spacepoints
-    // m_config.zMin = -2800.;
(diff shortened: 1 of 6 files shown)


beomki-yeo:
I set the r and z range of particle selection to cover the pixel barrel of ODD by default:

image

asalzburger (approved):
This makes a lot of sense, the adapted range sound reasonable, and also the increase on the scattering parameter should happen ODD has way more material than TrackML.

PR #736, opened 2024-10-13
Wrap the phi of track parameters in the range of [-pi,pi]

beomki-yeo:
This is for constraining the values of phi from Kalman updater and smoother to be within [-pi,pi].
If we don't do this, very occasionally, the while loop of[ `wrap_to_pi` of `is_same_angle`](https://github.com/acts-project/traccc/blob/bc55abf73a82625497a63e3a50d8ddb179a227b4/performance/src/performance/details/is_same_angle.cpp#L16) can take very long time.

### DIFF
--- a/core/include/traccc/edm/track_parameters.hpp
+++ b/core/include/traccc/edm/track_parameters.hpp
@@ -9,6 +9,7 @@
 
 // traccc include
 #include "traccc/definitions/common.hpp"
+#include "traccc/definitions/math.hpp"
 #include "traccc/definitions/primitives.hpp"
 #include "traccc/definitions/qualifiers.hpp"
 #include "traccc/definitions/track_parametrization.hpp"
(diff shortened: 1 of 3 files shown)


krasznaa (approved):
It would've been nicer to make this into member code inside of `detray::bound_track_parameters`, but I guess this is the next best thing. 🤔

stephenswat:
Wouldn't it be better to use fmod (
https://en.cppreference.com/w/c/numeric/math/fmod) in the method that has
to wrap the phi values?

beomki-yeo:
I think this PR is using `fmod`

core/include/traccc/edm/track_parameters.hpp
```diff
@@ -31,4 +32,20 @@ using bound_covariance = bound_track_parameters::covariance_type;
 using bound_track_parameters_collection_types =
     collection_types<bound_track_parameters>;
 
+// Wrap the phi of track parameters to [-pi,pi]
+TRACCC_HOST_DEVICE
+inline void wrap_phi(bound_track_parameters& param) {
+
+    traccc::scalar phi = param.phi();
+    static constexpr traccc::scalar TWOPI =
+        2. * traccc::constant<traccc::scalar>::pi;
```
niermann999, line 41:
```suggestion
        2.f * traccc::constant<traccc::scalar>::pi;
```

PR #737, opened 2024-10-14
ref: Update to detray version 0.79.0 and algebra-plugins version 0.25.0

niermann999:
Mostly addressing maintainability issues in detray:
 
- Rename vc to vc_aos
- Remove enable_if 
- Use ranges algorithms
- Refactor type traits into concepts
- Split dependency between benchmarks and svgtools and make validation utils lib

### DIFF
--- a/extern/algebra-plugins/CMakeLists.txt
+++ b/extern/algebra-plugins/CMakeLists.txt
@@ -18,7 +18,7 @@ message( STATUS "Building Algebra Plugins as part of the TRACCC project" )
 
 # Declare where to get Algebra Plugins from.
 set( TRACCC_ALGEBRA_PLUGINS_SOURCE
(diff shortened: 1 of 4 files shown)


stephenswat:
Gentle nudge :angel: 

https://github.com/acts-project/detray/blob/main/CMakeLists.txt#L9

niermann999:
> Gentle nudge 👼
> 
> https://github.com/acts-project/detray/blob/main/CMakeLists.txt#L9

Thanks! Absolutely forgot about that again. Is there maybe a way to automatize this?

krasznaa:
> > Gentle nudge 👼
> > https://github.com/acts-project/detray/blob/main/CMakeLists.txt#L9
> 
> Thanks! Absolutely forgot about that again. Is there maybe a way to automatize this?

No real "automation" that I would be aware of. 😦 The only thing that can be automated is to create TGZ source files with the correct version number automatically. But then the version number in git would never be reliable. 🤔  So remembering to update the version after every tag is pretty much the simplest thing that we can do in my mind.

stephenswat (approved):

PR #738, opened 2024-10-15
Skip the time component calculation in track param estimation

beomki-yeo:
The motivation is the time component becomes `inf` very occasionally in the case of `invTanTheta=0` or `vz=0`. And This fails CKF making all other parameters `nan` values.

I could come up with some ideas to avoid this inifinite values but I decided to remove all the relevant codes because this does not seem to be very useful. Also ACTS takes an actual measured time instead of analytically calculated one. 

https://github.com/acts-project/acts/blob/fb3ddf48fdcf2bc08052e0a925a0c0531fa056f8/Core/include/Acts/Seeding/EstimateTrackParamsFromSeed.hpp#L272


krasznaa (approved):
This looks sane... 🤔

I'll refer to your judgement on this.

PR #740, opened 2024-10-15
Put CKF tips into a single vector

stephenswat:
Currently, we unnecessarily put tips into a jagged vector-like structure; this commit turns this into a single, resizable array to save on copies and to make the code a bit simpler.

### DIFF
--- a/device/cuda/src/finding/finding_algorithm.cu
+++ b/device/cuda/src/finding/finding_algorithm.cu
@@ -288,10 +288,11 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
     std::map<unsigned int, vecmem::data::vector_buffer<candidate_link>>
         link_map;
 
-    // Create a map for tip links
-    std::map<unsigned int, vecmem::data::vector_buffer<
-                               typename candidate_link::link_index_type>>
-        tips_map;
+    // Create a buffer of tip links
+    vecmem::data::vector_buffer<typename candidate_link::link_index_type>
+        tips_buffer{m_cfg.max_num_branches_per_seed * n_seeds, m_mr.main,
+                    vecmem::data::buffer_type::resizable};
+    m_copy.setup(tips_buffer)->wait();
 
     // Link size
     std::vector<std::size_t> n_candidates_per_step;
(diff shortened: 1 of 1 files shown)


device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -466,41 +462,13 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
-            prefix_sum += n_tips;
-        }
-    }
-
     /*****************************************************************
      * Kernel6: Build tracks
      *****************************************************************/
 
+    // Get the number of tips
+    auto n_tips_total = m_copy.get_size(vecmem::get_data(tips_buffer));
```
krasznaa, line 470: 🤔 I would probably skip out on calling `vecmem::get_data` explicitly. It's not technically needed here, and I don't think it improves readability.

```suggestion
    auto n_tips_total = m_copy.get_size(tips_buffer);
```

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -288,10 +288,11 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
     std::map<unsigned int, vecmem::data::vector_buffer<candidate_link>>
         link_map;
 
-    // Create a map for tip links
-    std::map<unsigned int, vecmem::data::vector_buffer<
-                               typename candidate_link::link_index_type>>
-        tips_map;
+    // Create a buffer of tip links
+    vecmem::data::vector_buffer<typename candidate_link::link_index_type>
+        tips_buffer{m_cfg.max_num_branches_per_seed * n_seeds, m_mr.main,
```
krasznaa, line 293: 🤔 How expensive is this over-allocation? Since with this we allocate a little more than with the current code, right?

Though I guess `candidate_link::link_index_type` is a 32-bit integer anyway, right? So this should not be too worrisome either way I guess. 🤔

krasznaa (commented):
Your proposal looks sound to me. But Beomki will have to give the final say on this one.

device/cuda/src/finding/finding_algorithm.cu
stephenswat: `max_num_branches_per_seed` is (currently) 10, `n_seeds` is probably somewhere around 50,000 for ⟨μ⟩, so we're talking about around 2 MB. Pretty negligible, I'd say.

beomki-yeo (approved):
Looks good to me. I think I could not come up with this because `max_num_branches_per_seed` is introduced by Shima later.

PR #741, opened 2024-10-16
Check if the result of Kalman update is OK

beomki-yeo:
In this PR, Kalman update fail when the $\theta$ is out of bound $[0, \pi]$ and $\phi$ is not a finite value. This only happens for FP32


krasznaa (approved):
As much as I can tell, this looks reasonable. 🤔

beomki-yeo:
Hmm, I think I will add the same check in the smoother for potential risk

PR #743, opened 2024-10-17
Change the default value of chi2 max to 10

beomki-yeo:
This seems to be the good number for ODD. The current value of 30 finds 3 times more tracks from CKF with almost same tracking effieicny


krasznaa (approved):
Looks reasonable to me.

PR #745, opened 2024-10-17
Double matching condition for track finding effieicny

beomki-yeo:
Based on #692 

Double matching condition:

1. The number of measurements from the major particle is larger than a half of the number of the measurements of the found track
2. The number of measurements from the major particle is larger than a half of the number of measurements produced by the particle

We have been only using the first condition. This PR adds the second one. 

Relevant ACTS code can be found here: https://github.com/acts-project/acts/blob/f935e2512cc72577eb9aa9934a6f7b52f4c0dd53/Examples/Algorithms/TruthTracking/ActsExamples/TruthTracking/TrackTruthMatcher.cpp#L105

As expected, the efficiency drops but not so much

Before

After


beomki-yeo:
This is the final PR before creating the tag...

performance/src/efficiency/finding_performance_writer.cpp
```diff
@@ -171,18 +171,31 @@ void finding_performance_writer::write_common(
...
-        assert(measurements.size() > 0u);
-        if (particle_hit_counts.at(0).hit_counts / measurements.size() >
-            m_cfg.matching_ratio) {
-            const auto pid = particle_hit_counts.at(0).ptc.particle_id;
+        assert(found_measurements.size() > 0u);
+        assert(truth_measurements.size() > 0u);
+        const bool reco_matched =
+            n_major_hits / found_measurements.size() > m_cfg.matching_ratio;
+        const bool truth_matched =
+            n_major_hits / truth_measurements.size() > m_cfg.matching_ratio;
```
krasznaa, line 194: I am getting **very** worried about this code. 😦 Is `n_major_hits` a floating point number? I don't think it is...

But this division should of course be done at floating point precision. We cannot divide two integers. And now I'm getting worried about this type of issue possibly being present in other places as well. Even though the efficiency plots have been okay-looking. 🤔

performance/src/efficiency/finding_performance_writer.cpp
```diff
@@ -171,18 +171,31 @@ void finding_performance_writer::write_common(
...
-                measurements, evt_data.m_found_meas_to_ptc_map);
+                found_measurements, evt_data.m_found_meas_to_ptc_map);
         } else {
             particle_hit_counts = identify_contributing_particles(
-                measurements, evt_data.m_meas_to_ptc_map);
+                found_measurements, evt_data.m_meas_to_ptc_map);
         }
 
+        const auto major_ptc = particle_hit_counts.at(0).ptc;
+        const auto n_major_hits = particle_hit_counts.at(0).hit_counts;
```
krasznaa, line 181:
```suggestion
        const double n_major_hits = static_cast<double>(particle_hit_counts.at(0).hit_counts);
```

krasznaa (changes requested):
Way to scare be in the home stretch... 😟

performance/src/efficiency/finding_performance_writer.cpp
stephenswat: I agree this is worrying, but shouldn't `-Wconversion` pick up on this?

stephenswat: EDIT: Nope, we only use `-Wfloat-conversion`. I would have sworn someone added `-Wconversion` at some point...

beomki-yeo: Damn - there was bug in track classification since long. You are right. FIxed now

beomki-yeo:
The plot gets more sense now...

Before
image

After
image

beomki-yeo:
The recovery at eta~0 makes sense:
As the track is orthogonal to the module, the cluster size (# number of cells) are usually small so the hit count sum frequently gets smaller than the measurements size which makes the value of division zero

krasznaa (approved):
Fine. Let's make the tag finally.

PR #746, opened 2024-10-18
Enable `-Wconversion`

stephenswat:
This commit replaces the existing `Wfloat-conversion` flag with the more strict `-Wconverion` and fixes related errors. These were mostly harmless integer conversions, but certainly also some nasty improper floating point-to-integer conversions which may constitute bugs.

### DIFF
--- a/tests/common/tests/kalman_fitting_test.cpp
+++ b/tests/common/tests/kalman_fitting_test.cpp
@@ -85,7 +85,7 @@ void KalmanFittingTests::ndf_tests(
     const fitting_result<traccc::default_algebra>& fit_res,
     const track_state_collection_types::host& track_states_per_track) {
 
-    scalar dim_sum = 0;
+    unsigned int dim_sum = 0;
     std::size_t n_effective_states = 0;
 
     for (const auto& state : track_states_per_track) {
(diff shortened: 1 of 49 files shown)


stephenswat:
False positives from Sonarcloud.

stephenswat:
Finally passed the tests. @krasznaa

beomki-yeo (approved):
Let's get this first

tests/common/tests/kalman_fitting_test.cpp
```diff
@@ -85,7 +85,7 @@ void KalmanFittingTests::ndf_tests(
     const fitting_result<traccc::default_algebra>& fit_res,
     const track_state_collection_types::host& track_states_per_track) {
 
-    scalar dim_sum = 0;
+    unsigned int dim_sum = 0;
```
beomki-yeo, line 88: Ah btw, I meant this to be scalar object rather than scalar. In some extensions of fitter, it is the summation of (dimension * weight) where the weight is scalar

stephenswat: Ah okay, that makes sense. I'll change it back

core/src/finding/find_tracks.hpp
```diff
@@ -76,6 +76,8 @@ track_candidate_container_types::host find_tracks(
     using propagator_type =
         detray::propagator<stepper_t, navigator_t, actor_type>;
 
+    assert(config.min_track_candidates_per_track > 1);
```
beomki-yeo, line 79: Why not `>=`? I think 1 should be allowed

stephenswat: Good catch, changed to `>= 1`

PR #747, opened 2024-10-18
Add an option for align vector for telescope plane placement

beomki-yeo:
This is to enable to create telescope detector in different axes, e.g. x-axis. Creating only in z-axis may not be desirable as it is numerically instable


niermann999 (approved):
Looks good!

PR #749, opened 2024-10-23
De-Template Spacepoint Formation, main branch (2024.10.23.)

krasznaa:
Following the example of #722, here I remove the "class level templating" from the host implementation of the spacepoint formation algorithm.

Beside all of the code re-organization, I also renamed the algorithm to a much more descriptive name (`traccc::host::silicon_pixel_spacepoint_formation_algorithm`). To make it very clear what it's meant to be able to do. Later when/if we introduce spacepoint formation from strip measurements, we can choose to either add a separate algorithm for that, or to make this one smarter, while also renaming it. 🤔


stephenswat (approved):
Sure.

beomki-yeo:
Let's do things one-by-one - Please don't rename the `spacepoint_formation` to `silicon_pixel_spacepoint_formation` in this PR.

beomki-yeo:
Wth

beomki-yeo:
Yeah please do so. I think there is no advantage in differentiating this algorithm to different data types.
 
In case we want to this, we will need to put a set of pixel measurements only to the `silicon_pixel_spacepoint_formation`. Current approach seems to be error-prone

PR #750, opened 2024-10-24
Update detray to v0.81.0

stephenswat:
Brings changes to the stepper states and the wire chambers.

### DIFF
--- a/core/include/traccc/finding/actors/ckf_aborter.hpp
+++ b/core/include/traccc/finding/actors/ckf_aborter.hpp
@@ -41,7 +41,7 @@ struct ckf_aborter : detray::actor {
 
         // Abort at the next sensitive surface
         if (navigation.is_on_sensitive() &&
-            stepping._s > abrt_state.min_step_length) {
+            stepping.path_from_surface() > abrt_state.min_step_length) {
             prop_state._heartbeat &= navigation.abort();
             abrt_state.success = true;
         }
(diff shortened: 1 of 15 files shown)


core/include/traccc/finding/actors/ckf_aborter.hpp
```diff
@@ -41,7 +41,7 @@ struct ckf_aborter : detray::actor {
 
         // Abort at the next sensitive surface
         if (navigation.is_on_sensitive() &&
-            stepping._s > abrt_state.min_step_length) {
+            stepping.step_size() > abrt_state.min_step_length) {
```
niermann999, line 44:
```suggestion
            stepping.path_from_surface() > abrt_state.min_step_length) {
```

niermann999 (changes requested):

core/include/traccc/finding/actors/ckf_aborter.hpp
stephenswat: Fixed, thanks!

niermann999 (approved):

krasznaa (approved):
Sure, all looks reasonable. 🤔

stephenswat:
I'm getting a bit tired of the OneAPI compiler running into some internal error every other week... :frowning_face:

stephenswat:
Turns out that the current version of detray, 0.81.0, manages to resolve the compiler bug.

niermann999 (approved):

PR #751, opened 2024-10-24
x86-64-v2, main branch (2024.10.24.)

krasznaa:
Use `-march=x86-64-v2` for the build when appropriate. While still allowing builds to use something more aggressive, like:

```
CXXFLAGS="-march=native" cmake ...
```

This was triggered by recent discussions about vectorization. So I thought we might as well use the same code for this that the ATLAS offline build uses. :thinking:

https://gitlab.cern.ch/atlas/atlasexternals/-/blob/main/Build/AtlasCMake/modules/AtlasCompilerSettings.cmake?ref_type=heads#L103


stephenswat (commented):
I mean, sure; but let's then also turn this on for OneAPI's x86 targets so we don't end up compiling our host code and our x86-target SYCL code for different architectures.

stephenswat:
Also, x86-64-**v3** has been around for well over a decade now, so to me it makes sense to target that. :stuck_out_tongue:

krasznaa:
For v3: We could, but at the moment we only use v2 on the grid with the ATLAS code. That's why I went with v2 out of the box. Again, one can set something more aggressive in local tests. But for the ATLAS offline code we've seen practically no improvement with more aggressive settings.

For SYCL: Do you have a suggestion? 😕 Right now, building SYCL code is still sort of an expert task. Asking people to set up, among other things, the `$SYCLFLAGS` environment variable in a non-trivial way to make the build do exactly what they want. That by itself is a complicated enough procedure that it would be hard to tweak the settings safely from the project's CMake code.

stephenswat:
> For SYCL: Do you have a suggestion? 😕 Right now, building SYCL code is still sort of an expert task. Asking people to set up, among other things, the `$SYCLFLAGS` environment variable in a non-trivial way to make the build do exactly what they want. That by itself is a complicated enough procedure that it would be hard to tweak the settings safely from the project's CMake code.

I don't think it needs to be so complicated; if we slap a `-match=whatever` flag on the SYCL flags we'll get architectural support for at least the host code, and hopefully also for whatever device code is compiled with `x86_64` SYCL targets.

stephenswat (approved):

PR #752, opened 2024-10-25
Add theta range input

beomki-yeo:
This PR allows to set a theta range input instead of eta when `use-theta-range-input` is ON. 

Currently this is the best I could come up with but I wonder if we can block the set of `theta-range` when `use-theta-range-input` is OFF.

### DIFF
--- a/examples/options/include/traccc/options/generation.hpp
+++ b/examples/options/include/traccc/options/generation.hpp
@@ -11,6 +11,7 @@
 #include "traccc/definitions/primitives.hpp"
 #include "traccc/options/details/interface.hpp"
 #include "traccc/options/details/value_array.hpp"
+#include "traccc/utils/ranges.hpp"
 
 // detray include(s).
 #include "detray/definitions/pdg_particle.hpp"
@@ -42,6 +43,9 @@ class generation : public interface {
     opts::value_array<float, 2> phi_range{-180.f, 180.f};
     /// Range of eta
     opts::value_array<float, 2> eta_range{-2.f, 2.f};
+    /// Range of theta [rad] (corresponding to eta range of [-2,2])
+    opts::value_array<float, 2> theta_range{eta_to_theta_range(eta_range)};
+
     /// PDG number for particle type (Default: muon)
     int pdg_number = 13;
 
@@ -50,8 +54,6 @@ class generation : public interface {
     /// @name Derived options
     /// @{
 
-    /// Range of theta [rad]
-    opts::value_array<float, 2> theta_range{0.f, 0.f};
     /// Particle type
     detray::pdg_particle<traccc::scalar> ptc_type =
         detray::muon<traccc::scalar>();
--- a/examples/options/src/generation.cpp
+++ b/examples/options/src/generation.cpp
@@ -49,18 +49,35 @@ generation::generation() : interface("Particle Generation Options") {
         "gen-eta",
         po::value(&eta_range)->value_name("MIN:MAX")->default_value(eta_range),
         "Range of eta");
+    m_desc.add_options()("gen-theta",
+                         po::value(&theta_range)
+                             ->value_name("MIN:MAX")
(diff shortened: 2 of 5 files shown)


krasznaa:
🤔 How about going something like this?

https://www.boost.org/doc/libs/1_86_0/libs/program_options/example/real.cpp

I.e. define both `--gen-eta` and `--gen-theta`, and throw an error if the user specified both. (With a help message explaining this behaviour.)

As far as I could find with a quick search, there's no built-in functionality in `boost::program_options` for expressing this relationship between two options. But we can put in an `if(...)` statement ourselves for this. 🤔

niermann999:
yes, we do smth like that also in detray : https://github.com/acts-project/detray/blob/a05a161a063b2ae4d71e408424c052f0d679a42d/tests/tools/include/detray/options/track_generator_options.hpp#L137
Btw, can't we do smth like in https://github.com/acts-project/traccc/blob/main/examples/options/src/track_propagation.cpp and fill the detray track generator config directly (https://github.com/acts-project/detray/blob/main/tests/include/detray/test/utils/simulation/event_generator/random_track_generator_config.hpp)? Then you could maybe pass the theta or eta ranges to the detray config in the ```read``` method, which converts them on the fly anyway

beomki-yeo:
I followed the example code provided by @krasznaa . @niermann999 's sugesstion sounds reasonable but let's do this in the next PRs

examples/options/src/generation.cpp
```diff
@@ -49,18 +49,38 @@ generation::generation() : interface("Particle Generation Options") {
...
+            std::string("Conflicting options 'gen-eta' and 'gen-theta'"));
+    } else if (vm.count("gen-eta") && !vm["gen-eta"].defaulted()) {
+        theta_range = eta_to_theta_range(eta_range);
+    } else if (vm.count("gen-theta") && !vm["gen-theta"].defaulted()) {
+        theta_range *= detray::unit<float>::degree;
+        eta_range = theta_to_eta_range(theta_range);
+    } else {
+        // When none of eta and theta is specified use the default eta range
+        theta_range = eta_to_theta_range(eta_range);
+    }
```
krasznaa, line 82: Do we need this very last part? I thought you chose the default values of the `eta_range` and `theta_range` variables such that they would be equivalent. 🤔

examples/options/include/traccc/options/generation.hpp
```diff
@@ -42,6 +42,11 @@ class generation : public interface {
     opts::value_array<float, 2> phi_range{-180.f, 180.f};
     /// Range of eta
     opts::value_array<float, 2> eta_range{-2.f, 2.f};
+    /// Range of theta [rad] (corresponding to eta range of [-2,2]
+    opts::value_array<float, 2> theta_range{
+        15.415f * detray::unit<float>::degree,
+        164.585f * detray::unit<float>::degree};
```
krasznaa, line 48: To be absolutely sure that we do the right thing, how about:

```suggestion
    opts::value_array<float, 2> theta_range{
        eta_to_theta_range(eta_range)};
```

?

krasznaa (commented):
I'm very on board with this, just let's see about these technicalities first. 😉

examples/options/src/generation.cpp
beomki-yeo: Removed

examples/options/include/traccc/options/generation.hpp
beomki-yeo: OK. I also changed the CMake `traccc_performance` to the public link to make this work

krasznaa (approved):
I was a bit hesitant about the public dependency at first, but in the end this is part of the "example code". Having `traccc::options` publicly depend on `traccc::performance` is not actually a big deal. And keeping the private dependency would make the code a whole lot uglier. 🤔

So yeah, let's just go with this! 👍

PR #753, opened 2024-10-25
Update Presets, main branch (2024.10.25.)

krasznaa:
Re-designed the project's build presets a little.
  - Made explicit FP32 and FP64 presets, to make it a bit easier for users to switch between the two;
  - Ditched the `array` and `eigen` presets, as even I never used those so far;
  - Introduced a pair of `full-fp32` / `full-fp64` presets, which attempt to build everything that the repository has to offer, at the same time.
    * Which I can't actually use in any of my own environments at the moment. 😢 The latest versions of CUDA and oneAPI just do not like to work at the same time. 😦

Technically a lot more combinations could still be added. But I thought this may be the golden middle ground between not having enough variety, and having too much. 🤔

I'm also pondering about reforming our CI setup a little. 🤔 More may very well come to this PR soon...

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -2,57 +2,113 @@
    "version" : 3,
    "configurePresets": [
       {
-         "name" : "base",
-         "displayName" : "Base Configuration",
+         "name" : "base-fp32",
+         "displayName" : "Base FP32 Configuration",
          "warnings": {
             "deprecated": true
          },
          "cacheVariables": {
-            "CMAKE_BUILD_TYPE"        : "RelWithDebInfo",
-            "TRACCC_BUILD_TESTING"    : "TRUE",
-            "TRACCC_BUILD_EXAMPLES"   : "TRUE",
-            "TRACCC_FAIL_ON_WARNINGS" : "TRUE",
-            "TRACCC_USE_ROOT"         : "FALSE"
+            "CMAKE_BUILD_TYPE"         : "RelWithDebInfo",
+            "TRACCC_BUILD_TESTING"     : "TRUE",
+            "TRACCC_BUILD_EXAMPLES"    : "TRUE",
+            "TRACCC_BUILD_BENCHMARKS"  : "TRUE",
+            "TRACCC_FAIL_ON_WARNINGS"  : "TRUE",
+            "TRACCC_USE_ROOT"          : "FALSE",
+            "TRACCC_ALGEBRA_PLUGINS"   : "ARRAY",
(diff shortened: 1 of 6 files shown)


krasznaa:
Turns out, we were not building the benchmarks in the CI so far. 😱 While at it, I also added a 64-bit build of the host code. For the device libraries I didn't add such things, but we could consider doing so at one point... 🤔

Generally the PR is now ready to be looked at. 😉

CMakePresets.json
```diff
@@ -2,83 +2,140 @@
...
       {
-         "name" : "sycl",
-         "displayName" : "SYCL Code Development",
-         "inherits": ["base"],
+         "name" : "cuda-fp64",
+         "displayName" : "CUDA FP64 Code Development",
+         "inherits": ["base-fp64"],
+         "cacheVariables": {
+            "TRACCC_BUILD_CUDA" : "TRUE",
+            "VECMEM_BUILD_CUDA_LIBRARY" : "TRUE"
```
beomki-yeo, line 55: Is `VECMEM_BUILD_CUDA_LIBRARY=TRUE` (or `VECMEM_BUILD_SYCL_LIBRARY=TRUE`) necessary?

beomki-yeo: I thought TRACCC_BUILD_CUDA=true handles other flags well

krasznaa: Ahh, that's what you meant... 🤦

Technically, vecmem does auto-detect what it is likely to be able to build. When building the project without presets, one can expect that as long as CUDA is in one's environment, vecmem will automatically build `vecmem::cuda`. But that's not because `TRACCC_BUILD_CUDA` would turn on `VECMEM_BUILD_CUDA_LIBRARY`. But instead because vecmem by itself turns on that flag as long as CUDA is available.

In this way, if CUDA is available in the build environment, even with `--preset base-fp32` the `vecmem::cuda` library will be built.

The reason I ask vecmem explicitly to turn on the build of its CUDA library is to get a slightly clearer error message when things go wrong. If vecmem is told explicitly to build its CUDA code, but for some reason it can't, it will give a relatively clear error message about it. On the other hand if vecmem unexpectedly is unable to build CUDA code, we might end up with an error message in the traccc code that says that `vecmem::cuda` was never defined. At which point users might not understand right away that vecmem's language detection was not recognizing CUDA in the environment.

Long story short: Yes, we should set these things explicitly in the presets. 😉

beomki-yeo: OK

beomki-yeo (approved):
Looks good to me

PR #754, opened 2024-10-26
Fix fitting performance writer by changing the measurement comparator

beomki-yeo:
The measurement comparison operator with `float_epsilon` is causing a bug in `event_data`.  When the difference between the new measurement and one in the map is bigger than `float_epsilon`, the measurement is not added to the map.
(You can find the simple code that reproduces this behavior: https://godbolt.org/z/3hP9P5Eo9)

I fixed the issue by removing the `float_epsilon`. Also put `at` instead of `[]` wherever possible to make sure that there is the measurement in the map

### DIFF
--- a/core/include/traccc/edm/measurement.hpp
+++ b/core/include/traccc/edm/measurement.hpp
@@ -62,11 +62,16 @@ TRACCC_HOST_DEVICE inline bool operator<(const measurement& lhs,
 
     if (lhs.surface_link != rhs.surface_link) {
         return lhs.surface_link < rhs.surface_link;
-    } else if (math::fabs(lhs.local[0] - rhs.local[0]) > float_epsilon) {
-        return (lhs.local[0] < rhs.local[0]);
-    } else {
-        return (lhs.local[1] < rhs.local[1]);
+    } else if (lhs.local[0] != rhs.local[0]) {
+        return lhs.local[0] < rhs.local[0];
+    } else if (lhs.local[1] != rhs.local[1]) {
+        return lhs.local[1] < rhs.local[1];
+    } else if (lhs.variance[0] != rhs.variance[0]) {
+        return lhs.variance[0] < rhs.variance[0];
+    } else if (lhs.variance[1] != rhs.variance[1]) {
+        return lhs.variance[1] < rhs.variance[1];
     }
+    return false;
 }
 
 /// Equality operator for measurements
--- a/performance/src/utils/event_data.cpp
+++ b/performance/src/utils/event_data.cpp
@@ -190,17 +190,17 @@ void event_data::setup_csv(bool use_acts_geom_source, const detector_type* det,
             auto hid = csv_meas_hit_ids[meas_id].hit_id;
             const auto& iohit = csv_hits[hid];
 
-            const auto meas = m_measurement_map[meas_id];
+            const auto meas = m_measurement_map.at(meas_id);
             meas_to_cluster_map[meas].push_back(iocell);
 
-            const auto& ptc = m_particle_map[iohit.particle_id];
+            const auto& ptc = m_particle_map.at(iohit.particle_id);
             m_cell_to_particle_map[iocell] = ptc;
         }
 
         // Fill the meas_to_particle_map
         for (auto const& [ms, cluster] : meas_to_cluster_map) {
             for (const auto& cell : cluster) {
-                const auto& ptc = m_cell_to_particle_map[cell];
+                const auto& ptc = m_cell_to_particle_map.at(cell);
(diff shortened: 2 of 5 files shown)


core/include/traccc/edm/measurement.hpp
```diff
@@ -62,22 +62,24 @@ TRACCC_HOST_DEVICE inline bool operator<(const measurement& lhs,
...
 /// Equality operator for measurements
 TRACCC_HOST_DEVICE
 inline bool operator==(const measurement& lhs, const measurement& rhs) {
 
     return ((lhs.surface_link == rhs.surface_link) &&
-            (math::fabs(lhs.local[0] - rhs.local[0]) < float_epsilon) &&
-            (math::fabs(lhs.local[1] - rhs.local[1]) < float_epsilon) &&
-            (math::fabs(lhs.variance[0] - rhs.variance[0]) < float_epsilon) &&
-            (math::fabs(lhs.variance[1] - rhs.variance[1]) < float_epsilon));
+            (lhs.local == rhs.local) && (lhs.variance == rhs.variance));
```
krasznaa, line 82: Do we have reasonable definitions for these operators? You're not requiring bit-wise equivalence for these floating point arrays, right? 😕

performance/src/utils/event_data.cpp
```diff
@@ -229,8 +230,10 @@ void event_data::setup_csv(bool use_acts_geom_source, const detector_type* det,
         m_ptc_to_meas_map[ptc].push_back(meas);
 
         if (!include_silicon_cells) {
-            // Fill measurement to particle map
-            m_meas_to_ptc_map[meas][ptc]++;
+            // Each measurement is created by a single particle unless we use
+            // the clusterization results
+            assert(m_meas_to_ptc_map.find(meas) == m_meas_to_ptc_map.end());
+            m_meas_to_ptc_map[meas][ptc] = 1u;
```
krasznaa, line 236: Should we only look out for this during debugging? 🤔

To make sure that we insert a new element into the map here, we can use [std::map::insert](https://en.cppreference.com/w/cpp/container/map/insert). It is indeed quite a bit more awkward, but you could do let's say:

```c++
if (m_meas_to_ptc_map.insert({meas, {}}).second == false) {
   throw std::runtime_error(...);
}
if (m_meas_to_ptc_map.at(meas).insert({ptc, 1u}).second == false) {
   throw std::runtime_error(...);
}
```

Not impossible that I would've made a mistake in the example, but something like this could technically work I believe. 🤔

And yeah, the second insert could also be kept as an `operator[]` expression. Since by definition you're filling an empty map there. 🤔

krasznaa (commented):
No big objections on my part, just the two technical questions/suggestions...

core/include/traccc/edm/measurement.hpp
beomki-yeo: I am meaning to have bit-level equivalence here. This will make CPU and GPU measurements different but I don't think that's a problem. For tolerances with epsilon, `is_same_object` can do exactly the same thing

beomki-yeo: I decided to leave `operator==` as it is. I found that it is going to drop the CKF track_candidate_vector matching rate to 0% as the measurement from CPU and GPU will be different..

The comparison of track candidate element (which is a measurement) just uses the `operator ==` instead of `is_same_object<measurement>`:
https://github.com/acts-project/traccc/blob/8dd31a49ee5bd9fbf2bfc1de24b700a4ae4f99f3/performance/src/performance/details/is_same_object.cpp#L127

I think this line should be using `is_same_object<measurement>` but let me do this in the later PR..

krasznaa (commented):
The rest I'm happy with.

And if you push back firmly enough about the whole `std::map::insert` point, I can drop that as well. Since your code technically already works correctly like this as well. 😉

performance/src/utils/event_data.cpp
```diff
@@ -229,8 +230,14 @@ void event_data::setup_csv(bool use_acts_geom_source, const detector_type* det,
...
-            // Fill measurement to particle map
-            m_meas_to_ptc_map[meas][ptc]++;
+            if (m_meas_to_ptc_map.find(meas) != m_meas_to_ptc_map.end()) {
+                throw std::runtime_error(
+                    "The new measurement should not exist in the "
+                    "measurement-to-particle map");
+            }
+            // Each measurement is created by a single particle unless we use
+            // the clusterization results
+            m_meas_to_ptc_map[meas][ptc] = 1u;
```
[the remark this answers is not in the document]
beomki-yeo: OK. I changed as suggested.

PR #756, opened 2024-10-28
De-Template (Host) Track Fitting, main branch (2024.10.28.)

krasznaa:
This is the next step in my crusade to make the reconstruction algorithms use non-templated interfaces. 😄

In this I renamed `traccc::fitting_algorithm` to `traccc::host::kf_algorithm`, along exactly the same lines as what I did for `traccc::host::ckf_algorithm` in #722.

As before, one important aspect of this is to get the memory needs of the build under control. In the current `main` branch we use the following amounts of memory for compiling a few representative source files:
  - `seq_example.cpp`: 1.4 GB
  - `cpu/throughput_st.cpp`: 700 MB

This PR introduces 3 new source files, and reduces the memory needs of the previous examples like:
  - `kf_algorithm.cpp`: 320 MB
  - `kf_algorithm_teldet_cfield.cpp`: 475 MB
  - `kf_algorithm_defdet_cfield.cpp`: 650 MB
  - `seq_example.cpp`: 1.0 GB
  - `cpu/throughput_st.cpp`: 575 MB

Clearly this is not the end of the road yet, but I'm fairly happy with what the PR does for the build's memory usage.


core/include/traccc/fitting/fitting_algorithm.hpp
```diff
@@ -1,88 +0,0 @@
...
-#include "traccc/edm/track_state.hpp"
-#include "traccc/fitting/fitting_config.hpp"
-#include "traccc/fitting/kalman_filter/kalman_fitter.hpp"
-#include "traccc/utils/algorithm.hpp"
-
-namespace traccc {
-
-/// Fitting algorithm for a set of tracks
-template <typename fitter_t>
-class fitting_algorithm
```
beomki-yeo, line 22: Hmm, why are you changing the name? The motivation of this name and template parameter is using various fitters with the same interface

core/include/traccc/fitting/kf_algorithm.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include "traccc/geometry/detector.hpp"
+#include "traccc/utils/algorithm.hpp"
+
+// Detray include(s).
+#include <detray/detectors/bfield.hpp>
+
+namespace traccc::host {
+
+/// Kalman filter based track fitting algorithm
+class kf_algorithm : public algorithm<track_state_container_types::host(
```
beomki-yeo, line 23: This should be just called `kalman_fitter` instead of `kf_algorithm`

core/src/fitting/kf_algorithm_defdet_cfield.cpp
```diff
@@ -0,0 +1,38 @@
+/** TRACCC library, part of the ACTS project (R&D line)
```
beomki-yeo, line 1: I don't think defdet or teldet is very understandable to general users... how about 
`kalman_fitter_constant_field`
`kalman_fitter_constant_field_telescope` ?

beomki-yeo (commented):
There are some naming issues that concern me..

core/include/traccc/fitting/fitting_algorithm.hpp
krasznaa: My overall points with these PRs is that we can't have such an interface for our GPU algorithms. Users cannot freely choose between different template specialisations. We'll provide them with a few specific ones, and they'll have to use one of those.

So the algorithm that I'm introducing, `traccc::host::kf_algorithm`, is specifically an algorithm applying a Kálmán fit to the tracks. With one specific setup.

Once/if we have a different valid setup for fitting as well, we'll create another algorithm class. One that will reflect in its name that it is performing a different kind of fit.

In the `src/` directory we can still have a single `traccc::host::details::fit_tracks` function. Which would get a different fitter template argument in all of these cases. But it would not be up to the user to construct the fitter themselves. 🤔

This latter point in my mind will actually make it a lot easier to use these algorithms. Since right now you really need expert level knowledge to just know how to even instantiate one of these algorithms such that they would make any sense.

core/include/traccc/fitting/kf_algorithm.hpp
krasznaa: The `kalman_fitter` fitter class still exists.

🤔 We could call it `traccc::host::kalman_fitter_algorithm` instead, but since `traccc::host::ckf_algorithm` already went through, I thought I'd go with this short name here as well.

core/src/fitting/kf_algorithm_defdet_cfield.cpp
krasznaa: I don't have strong feelings about these file names. But keep in mind that users will not be exposed to any of these names. These are just implementation details behind the scenes. That's why I went with relatively cryptic names. Because people who are not developers of this code, would not be exposed to these files anyway.

beomki-yeo: I believe having more verbosity is going to be helpful for devs as well.

beomki-yeo: We will keep expanding the list of this specialization for different detector and inhomogeneous field
Being cryptic makes the maintenance more difficult I guess

krasznaa: As I said, I'm very open to naming these files differently. I'll make this change shortly.

core/include/traccc/fitting/fitting_algorithm.hpp
beomki-yeo: If so, we might need to put those files in `fitting/kalman_filter`. In case we add the triplet fitter, global chi square fitter, DAF, etc, we may not want to dump them and their specialization into the same `fitting/` directory

core/include/traccc/fitting/kf_algorithm.hpp
beomki-yeo: This is what is exposed to the user directly and naming it into `kf_algorithm` feels less obvious to me. `kalman_fitting_algorithm` might work better
`ckf_algorithm` probably need to be changed to `ckf_finding_algorithm`  or sth else

core/include/traccc/fitting/fitting_algorithm.hpp
beomki-yeo: Splitting them would make more sense as far as I can tell.. OK

krasznaa:
Please have a look @paulgessinger. I now still have a specific algorithm interface for performing the fitting "in specific ways", but also expose the template code that would allow clients to do different things as well if they so wished.

I think this design could be extended to the device code as well. 🤔

stephenswat (approved):

PR #757, opened 2024-10-29
VecMem Update, main branch (2024.10.29.)

krasznaa:
Updated the project to using [vecmem-1.11.0](https://github.com/acts-project/vecmem/releases/tag/v1.11.0).

This update, most importantly, includes: https://github.com/acts-project/vecmem/pull/299 Which was made by @stephenswat to get the memory usage of our code under control. It also includes https://github.com/acts-project/vecmem/pull/286, which is the reason for all the actual code changes in the PR...

Currently `traccc_throughput_st_cuda` uses the following amount of memory to reconstruct 100 mu=140 events:

current_mu140

So, about 3.85 GB at its peak.

With this update included, this looks like:

newvecmem_mu140

About 2.77 GB at its peak.

To complete the picture, this is what I get when using [vecmem::pool_memory_resource](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/memory/pool_memory_resource.hpp) instead of [vecmem::binary_page_memory_resource](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/memory/binary_page_memory_resource.hpp):

pool_mu140

About 3.3 GB at its peak.

And finally, if I turn off memory caching completely, I get:

nocache_mu140

I.e. 200-700 MB of memory usage.

In all these cases, on the RTX 2060 GPU of my desktop machine, the throughput of reconstructing mu=140 events in a single thread/stream remained in +-10%. Without caching being the slowest of course, and `vecmem::pool_memory_resource` (which is based on [CUB](https://docs.nvidia.com/cuda/cub/index.html) and [std::pmr::unsynchronized_pool_resource](https://en.cppreference.com/w/cpp/memory/unsynchronized_pool_resource), being the fastest. But only by little. Apparently with the full chain running, the caching is not dominating the performance at the moment. 🤔

### DIFF
--- a/benchmarks/cuda/toy_detector_cuda.cpp
+++ b/benchmarks/cuda/toy_detector_cuda.cpp
@@ -111,15 +111,19 @@ BENCHMARK_F(ToyDetectorBenchmark, CUDA)(benchmark::State& state) {
             traccc::spacepoint_collection_types::buffer spacepoints_cuda_buffer(
                 static_cast<unsigned int>(spacepoints_per_event.size()),
                 mr.main);
+            async_copy.setup(spacepoints_cuda_buffer)->ignore();
             async_copy(vecmem::get_data(spacepoints_per_event),
-                       spacepoints_cuda_buffer);
+                       spacepoints_cuda_buffer)
+                ->ignore();
 
             traccc::measurement_collection_types::buffer
                 measurements_cuda_buffer(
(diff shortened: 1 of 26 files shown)


beomki-yeo:
Great! I will let @stephenswat take a look into this for detailed review

stephenswat:
Closes #645.

stephenswat (approved):
Happy to see the memory usage drop, and happy to approve this PR.

PR #759, opened 2024-10-30
TBB Re-Configuration Fix, main branch (2024.10.30.)

krasznaa:
As discussed with @stephenswat earlier today, @paulgessinger's update in #682 revealed a deeper issue with our build.

What happened was that [oneDPL](https://github.com/oneapi-src/oneDPL) would call `find_package(TBB)` to access TBB, when using the `"dpcpp"` backend. (The one we actually use.)

https://github.com/oneapi-src/oneDPL/blob/main/CMakeLists.txt#L163

This would:
  - Look for an external TBB, and also finding it as part of either oneAPI or the general setup that we would have for ROOT;
  - Set up the `TBB_DIR` cache variable, pointing at the external TBB version that was found.

This would work ~fine on the first CMake configuration. Even though at this point oneDPL would link against a different version of TBB than the one building together with the project. 😦 But on a re-configuration, we would now bump into:

https://github.com/oneapi-src/oneTBB/blob/v2021.7.0/CMakeLists.txt#L210

I.e. on this re-configuration we would completely abandon our own TBB version, and rather pick up the external version that was previously found by oneDPL.

To fix this, I had to do two things:
  - Upgrade to a newer version of TBB. One in which defining `TBB_DIR` by itself would not make it abandon setting up its own version of TBB. (In https://github.com/oneapi-src/oneTBB/blob/v2021.13.0/CMakeLists.txt#L236 the previous `OR` was changed to an `AND`.)
  - Make use of a new feature in [FetchContent_Declare](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_declare) that makes any subsequent calls to `find_package(TBB)` be practically a no-op. This way ensuring that oneDPL would actually use the TBB version that we are building ourselves.

Yepp, some non-trivial CMake-ing going on here... 😛

P.S. I moved the setup of TBB before (Roc)Thrust, because those can also conditionally use TBB themselves. Even if in our current build they don't.

### DIFF
--- a/extern/tbb/CMakeLists.txt
+++ b/extern/tbb/CMakeLists.txt
@@ -1,11 +1,11 @@
 # TRACCC library, part of the ACTS project (R&D line)
 #
-# (c) 2022 CERN for the benefit of the ACTS project
+# (c) 2022-2024 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
 # CMake include(s).
-cmake_minimum_required( VERSION 3.14 )
+cmake_minimum_required( VERSION 3.24 )
 include( FetchContent )
 
 # Silence FetchContent warnings with CMake >=3.24.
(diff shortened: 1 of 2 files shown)


krasznaa:
Let me also tag you Beomki, as fixing this would help my other developments quite a bit. 😉

extern/tbb/CMakeLists.txt
```diff
@@ -1,11 +1,11 @@
 # TRACCC library, part of the ACTS project (R&D line)
 #
-# (c) 2022 CERN for the benefit of the ACTS project
+# (c) 2022-2024 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
 # CMake include(s).
-cmake_minimum_required( VERSION 3.14 )
+cmake_minimum_required( VERSION 3.24 )
```
stephenswat, line 8: Isn't this quite recent? I'm a bit surprised all our CI machines have this!

stephenswat (approved):
Happy to approve this.

extern/tbb/CMakeLists.txt
krasznaa: I was pleasantly surprised as well. This is what I need for `OVERRIDE_FIND_PACKAGE`, so was very happy that the CI accepted it without further effort. 😄

PR #760, opened 2024-11-01
De-Template the SYCL Library, main branch (2024.11.01.)

krasznaa:
Following in the footsteps of #756 and #749, here I want to make the two SYCL algorithms that currently have a template parameter, not have one. :thinking:

The logic is exactly the same as in the previous PRs. I make the algorithms deal only with specific detector and magnetic field types, and use a specific setup for track fitting in the case of `traccc::sycl::kalman_fitting_algorithm`.

I wanted to do this before finally introducing a CKF algorithm for SYCL. Since technically we should be able to make that work since a while...


krasznaa:
@stephenswat, @beomki-yeo, @paulgessinger, any comments?

This is not the be-all / end-all yet. As I told @stephenswat yesterday, I have some ideas of how to expose the underlying kernels in the API of the device libraries, but I'd like to separate that from this PR. As I'll first want to add SYCL track finding, once this one goes in. 🤔

stephenswat (approved):

PR #762, opened 2024-11-04
ref: Make the interaction register an observing actor

niermann999:
Make the interaction_register actor an observing actor to the pointwise_material_interactor in order to break the dependency on the interactor state during state construction. Also uses a utility function from the detray actor chain to contruct the sate tuple and the state reference tuple if all states are default constructible.


niermann999:
Needs some modification in detray before it will compile

niermann999:
The interaction register has been removed

PR #763, opened 2024-11-04
ref: Do not copy the bound track parameters for the gain matrix updater

niermann999:
Try to avoid the copying of the bound track parameters when doing the CKF Kalman update

### DIFF
--- a/core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
+++ b/core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
@@ -78,7 +78,7 @@ struct kalman_actor : detray::actor {
     TRACCC_HOST_DEVICE void operator()(state& actor_state,
                                        propagator_state_t& propagation) const {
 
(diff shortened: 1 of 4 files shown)


stephenswat (commented):
Makes sense to me but I will refer to @beomki-yeo.

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
```diff
@@ -114,6 +114,9 @@ struct kalman_actor : detray::actor {
                 return;
             }
 
+            // Update the propagation flow
+            propagation._stepping.bound_params() = trk_state.filtered();
```
beomki-yeo, line 118: how about `stepping.bound_params() = trk_state.filtered()` after removing the `const` qualifier above?

niermann999: Sure, done

beomki-yeo (approved):

PR #764, opened 2024-11-07
consistently use io::get_absolute_path when dealing with user-input f…

flg:
minor change to handle absolute paths in example user input, as it was for digitization file


krasznaa (approved):
😄 I've been thinking for a while about making the same changes. As it has been bugging me quite a bit as well not to be able to pick up geometry files from random places.

Well done! 👍

PR #765, opened 2024-11-08
Adapt seeding cpu test to acts v37.3.0

CarloVarni:
This adapts the seeding cpu test and the atlas cuts to Acts v37.3.0. since the ACTS seeding has changed quite a lot, I've also added some comments to explain what we do.

From local test, the seed matching between Acts and traccc is ~0.9995

@krasznaa

### DIFF
--- a/tests/cpu/compare_with_acts_seeding.cpp
+++ b/tests/cpu/compare_with_acts_seeding.cpp
@@ -1,4 +1,4 @@
-/** TRACCC library, part of the ACTS project (R&D line)
+/** TRACCC library, part of the ACTS projAect (R&D line)
  *
  * (c) 2021-2024 CERN for the benefit of the ACTS project
  *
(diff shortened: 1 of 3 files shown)


krasznaa:
image

No way! That was time very well spent with you this afternoon... 😄

tests/cpu/compare_with_acts_seeding.cpp
```diff
@@ -42,6 +41,57 @@
...
+    std::size_t size_impl() const { return storage().size(); }
+    float x_impl(std::size_t idx) const { return storage()[idx]->x(); }
+    float y_impl(std::size_t idx) const { return storage()[idx]->y(); }
+    float z_impl(std::size_t idx) const { return storage()[idx]->z(); }
+    float varianceR_impl(std::size_t idx) const {
+        return storage()[idx]->varianceR;
+    }
+    float varianceZ_impl(std::size_t idx) const {
+        return storage()[idx]->varianceZ;
+    }
```
krasznaa, line 70: Interesting... 🤔 I **think** we should also be able to skip defining a `SpacePoint` class entirely, and just use `traccc::spacepoint` directly, no?

In any case, that's something that I'll be pretty interested to try for myself. As the spacepoints will soon need to go SoA as well anyway... 🤔

krasznaa (approved):
I'm happy with it myself. 😄 But let's wait for @beomki-yeo to have a look as well.

tests/cpu/compare_with_acts_seeding.cpp
CarloVarni: indeed you can do it

beomki-yeo (approved):
What an amazing work.. Thanks!

PR #767, opened 2024-11-09
Parallelize Cell Reading For Throughput Measurements, main branch (2024.11.09.)

krasznaa:
This is something that I had in the back of my mind ever since running the throughput tests for @beomki-yeo's CHEP 2024 results. But until now I thought that I should do this in `traccc::io`, by reviving the same style of function that we had back in the day with OpenMP.

But I had to realize that the throughput application itself, where this is actually relevant, already has everything available to parallelize its own file reading. :thinking: So I just used the simplest API of [TBB](https://github.com/oneapi-src/oneTBB), [tbb::parallel_for](https://oneapi-src.github.io/oneTBB/main/tbb_userguide/parallel_for_os.html). :wink:

On my home PC it for instance brings down the reading of 500 mu=140 ODD events from almost 4 minutes to ~10 seconds. (Yepp, [Threadripper](https://www.amd.com/en/products/processors/workstations/ryzen-threadripper.html) still FTW. :stuck_out_tongue:)

### DIFF
--- a/examples/run/common/throughput_mt.ipp
+++ b/examples/run/common/throughput_mt.ipp
@@ -46,6 +46,7 @@
 #else
 #include <tbb/global_control.h>
 #endif
+#include <tbb/parallel_for.h>
 #include <tbb/task_arena.h>
 #include <tbb/task_group.h>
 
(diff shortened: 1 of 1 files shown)


beomki-yeo (approved):
Good. I also suspected that the number of read events is the main factor for computing time instead of the number of processed events

PR #768, opened 2024-11-11
Update Host Track Finding, main branch (2024.11.11.)

krasznaa:
This is to update the host track finding algorithm to be more in line with #756.
  - I renamed `traccc::host::ckf_algorithm` into `traccc::host::combinatorial_kalman_filter_algorithm`, and made the "specialization source file" names a little longer / more elaborate as well.
  - I moved `traccc::host::details::find_tracks` to be publicly visible in `traccc::core`. So that clients could use it directly, in case they want to use a detector / stepper / navigator that we don't support at the moment.


krasznaa:
SonarCloud is correct that range based algorithms would be nice to use in the host code. But for that we'll first need to make vecmem device containers work correctly with those algorithms... 🤔 Because it seems that they don't. (But I also didn't investigate too deeply at this point.)

beomki-yeo (approved):
`combinatorial_kalman_filter_algorithm` seems quite long but it is probably better than `ckf_algorithm` :thinking:

PR #769, opened 2024-11-12
traccc::sycl::full_chain_algorithm Fixes, main branch (2024.11.12.)

krasznaa:
As reported by @flg, the SYCL throughput applications have been crashing since some time. 😦

With SYCL it is paramount that the underlying [::sycl::queue](https://github.khronos.org/SYCL_Reference/iface/queue.html) object would be deleted after all objects have stopped using it. (Most importantly, after all allocated memory has been freed.) So now I made this **super explicit**.

This reminded me that the clusterization algorithm also manages some memory itself. So that algorithm needs to be very manually deleted as well. 🤔

At the same time did some general maintenance as well, to make the code just a little bit better.

### DIFF
--- a/examples/run/sycl/full_chain_algorithm.hpp
+++ b/examples/run/sycl/full_chain_algorithm.hpp
@@ -119,13 +119,13 @@ class full_chain_algorithm
 
     private:
     /// Private data object
-    details::full_chain_algorithm_data* m_data;
+    std::unique_ptr<details::full_chain_algorithm_data> m_data;
     /// Host memory resource
-    vecmem::memory_resource& m_host_mr;
+    std::reference_wrapper<vecmem::memory_resource> m_host_mr;
     /// Device memory resource
-    std::unique_ptr<vecmem::sycl::device_memory_resource> m_device_mr;
+    vecmem::sycl::device_memory_resource m_device_mr;
     /// Device caching memory resource
-    std::unique_ptr<vecmem::binary_page_memory_resource> m_cached_device_mr;
+    mutable vecmem::binary_page_memory_resource m_cached_device_mr;
     /// Memory copy object
     mutable vecmem::sycl::async_copy m_copy;
 
--- a/examples/run/sycl/full_chain_algorithm.sycl
+++ b/examples/run/sycl/full_chain_algorithm.sycl
@@ -9,7 +9,7 @@
 #include "full_chain_algorithm.hpp"
 
 // SYCL include(s).
-#include <CL/sycl.hpp>
+#include <sycl/sycl.hpp>
 
 // System include(s).
 #include <exception>
@@ -35,7 +35,14 @@ namespace traccc::sycl {
 namespace details {
 
 struct full_chain_algorithm_data {
+
+    /// Constructor
+    explicit full_chain_algorithm_data(const ::sycl::async_handler& handler)
+        : m_queue{handler} {}
+
+    /// The native SYCL queue object
     ::sycl::queue m_queue;
+    /// Wrapper around the SYCL queue object
     queue_wrapper m_queue_wrapper{&m_queue};
 };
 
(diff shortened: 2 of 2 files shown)


examples/run/sycl/full_chain_algorithm.sycl
```diff
@@ -35,8 +35,32 @@ namespace traccc::sycl {
...
+        : m_queue{handler} {}
+
+    /// Destructor
+    ~full_chain_algorithm_data() {
+        // Delete the objects on the heap explicitly, so that they would
+        // definitely be deleted before the SYCL queue object. Even take care
+        // of deleting them in the correct order.
+        m_cached_device_mr.reset();
+        m_device_mr.reset();
+    }
```
stephenswat, line 50: This destructor is strictly unnecessary, as members are automatically destroyed in reverse declaration order.

examples/run/sycl/full_chain_algorithm.sycl
```diff
@@ -85,69 +108,78 @@ full_chain_algorithm::full_chain_algorithm(
...
 }
 
 full_chain_algorithm::~full_chain_algorithm() {
     // Need to ensure that objects would be deleted in the correct order.
-    delete m_data;
+    m_clusterization.reset();
+    m_device_det_descr.reset();
+    m_device_detector.reset();
+    m_data.reset();
```
stephenswat, line 167: This destructor should be removed.

examples/run/sycl/full_chain_algorithm.hpp
```diff
@@ -119,32 +125,26 @@ class full_chain_algorithm
...
-    host_detector_type::buffer_type m_device_detector;
+    std::unique_ptr<host_detector_type::buffer_type> m_device_detector;
     /// View of the detector's payload on the device
     host_detector_type::view_type m_device_detector_view;
 
     /// @name Sub-algorithms used by this full-chain algorithm
     /// @{
     /// Clusterization algorithm
-    clusterization_algorithm m_clusterization;
+    std::unique_ptr<clusterization_algorithm> m_clusterization;
```
stephenswat, line 147: Why do these need to be unique pointers? :confused:

flg:
Hi Attila,

Since you've also done some general maintenance I have a small question/suggestion.

If I understand correctly a local sycl::queue is created in full_chain_algorithm constructor that is never used:
https://github.com/acts-project/traccc/blob/29ca228cc30a213f9f283f8bfa211e41c297da96/examples/run/sycl/full_chain_algorithm.sycl#L89

If I'm not wrong it should be removed and more importantly the following should wait on the m_data->m_queue not this local one since m_copy will use m_data->m_queue:
https://github.com/acts-project/traccc/blob/29ca228cc30a213f9f283f8bfa211e41c297da96/examples/run/sycl/full_chain_algorithm.sycl#L97

Or is this local queue really used for something?

stephenswat (changes requested):
No, this is not the way to do this... We walked into this in #664; the solution here is to _remove_ the destructors, not make them more complicated. The C++ standard guarantees that members are destroyed in reverse declaration order anyway, so I don't see what these destructors bring us.

examples/run/sycl/full_chain_algorithm.sycl
krasznaa: The destructor was made `default`. It can't be removed, as it needs to delete `traccc::sycl::details::full_chain_algorithm_data`. Which can only be done in that source file.

But yeah, the custom destructor could be removed.

krasznaa:
The code could indeed be simplified. 🤔

In a separate PR I'll add unit tests that would run some of our example applications. As we should really test it in our CI that the applications work to some basic level at least... 🤔

krasznaa:
> Hi Attila,
> 
> Since you've also done some general maintenance I have a small question/suggestion.
> 
> If I understand correctly a local sycl::queue is created in full_chain_algorithm constructor that is never used:
> 
> https://github.com/acts-project/traccc/blob/29ca228cc30a213f9f283f8bfa211e41c297da96/examples/run/sycl/full_chain_algorithm.sycl#L89
> 
> If I'm not wrong it should be removed and more importantly the following should wait on the m_data->m_queue not this local one since m_copy will use m_data->m_queue:
> 
> https://github.com/acts-project/traccc/blob/29ca228cc30a213f9f283f8bfa211e41c297da96/examples/run/sycl/full_chain_algorithm.sycl#L97
> 
> Or is this local queue really used for something?

Yepp, I saw that too. Didn't search the history for how that code happened, but indeed it was doing some silly things. The PR now removes the erroneous lines.

PR #770, opened 2024-11-12
Run Kalman Smoother in CKF

beomki-yeo:
This is to use the greedy ambiguity solver before the KF, which requires `chi2`


beomki-yeo:
The further development on this PR might be pending because storing all track states even for poor-quality tracks does not look very convincing

beomki-yeo:
Let's not run Kalman Smoother in the CKF, which does not make sense to me a lot for the following reasons
1. RTS smoother seems not a right model for complex geometries
2. Two-filters method is too expensive as it is going to x2 the computing time

PR #771, opened 2024-11-13
Acts Seed-Finding Comparison Update, main branch (2024.11.13.)

krasznaa:
Triggered by #765, this PR tries to simplify the comparison code a little bit.

With the latest Acts version it's not necessary to create a dummy `SpacePoint` class anymore. We can put a wrapper around `traccc::spacepoint` directly. So this is what I did. At this point we can just use the existing `==` operator for `traccc::spacepoint` as well.

At the same time I simplified / cleaned the code a little, so that it would be more to my liking...


tests/cpu/compare_with_acts_seeding.cpp
```diff
@@ -405,20 +358,24 @@ TEST_P(CompareWithActsSeedingTests, Run) {
...
+                    spacepoints_per_event.at(traccc_seed.spM_link);
+                const traccc::spacepoint& traccc_spT =
+                    spacepoints_per_event.at(traccc_seed.spT_link);
+
+                const traccc::spacepoint& acts_spB =
+                    acts_seed.sp()[0]->externalSpacePoint();
+                const traccc::spacepoint& acts_spM =
+                    acts_seed.sp()[1]->externalSpacePoint();
+                const traccc::spacepoint& acts_spT =
+                    acts_seed.sp()[2]->externalSpacePoint();
```
CarloVarni, line 378: just a though: at this point ACTS has created Seeds of proxies to traccc space points, and proxies have an index (corresponding to the position of the original space point in the original collection). Meaning you can simply access the indexes and use them for the comparison.

So here you can actually compare the `traccc::seed::link_type` against the acts space point proxy `index()` instead of comparing the coordinates (which I believe is what the `==` does, right?)

CarloVarni (commented):
I'm ok with the changes. Just added a though about the traccc-acts comparison. Up to you on what to do with it

CarloVarni (approved):

PR #772, opened 2024-11-14
Digitization Config JSON I/O Update, main branch (2024.11.14.)

krasznaa:
Added the ability to write `traccc::digitization_config` to JSON files. This is in a hope that @nribaric will be able to use this code to write a (possibly very large) ITk digitization file using Athena. 🤔 So that we could attempt ITk studies outside Athena as well. 😉

At the same time moved the JSON reading code to be more in line with how we have the CSV and OBJ file reading/writing code in the library.


beomki-yeo (approved):
Makes sense to me

PR #774, opened 2024-11-15
Refactor the Parameter interface for Kalman fitting test

beomki-yeo:
In the current `main` branch, all tests derived from `KalmanFittingTests` share the same input parameters. 
This made it impossible to give the input parameters specific to detector types (e.g. Telescope, toy geometry and wire chamber).

In this PR, `KalmanFittingTests` does not take input parameters anymore. Instead, `KalmanFittingTelescopeTests`, `KalmanFittingToyDetectorTests`, and `KalmanFittingWireChamberTests` have their own input parameters.

`KalmanFittingTelescopeTests` also sets the planes (20 mm, 40 mm, ..., 180 mm) instead of (0 mm, 20 mm, ..., 160 mm)


niermann999:
> `KalmanFittingTelescopeTests` also sets the planes (20 mm, 40 mm, ..., 180 mm) instead of (0 mm, 20 mm, ..., 160 mm)

Be careful to either move the track generator origin into the detector space or set ```cfg.envelope(21mm)``` such that the portals encompass the position (0,0,0). Otherwise the navigation might start outside of the detector...

tests/common/tests/kalman_fitting_telescope_test.hpp
```diff
@@ -62,19 +77,31 @@ class KalmanFittingTelescopeTests : public KalmanFittingTests {
...
+                                      static_cast<scalar>(i) * spacing *
+                                          unit<scalar>::mm);
+        }
+
         detray::tel_det_config<> tel_cfg{rectangle};
         tel_cfg.positions(plane_positions);
         tel_cfg.module_material(mat);
         tel_cfg.mat_thickness(thickness);
         tel_cfg.pilot_track(traj);
+        tel_cfg.envelope(offset * 2.f);
```
beomki-yeo, line 104: I think this is good enough

beomki-yeo:
Review anyone?

krasznaa (commented):
Okay... I don't have particularly strong opinions about this testing code. :thinking:

The one thing I noticed: Should the SYCL test not be updated as well?

https://github.com/acts-project/traccc/blob/main/tests/sycl/test_kalman_fitter_telescope.sycl

beomki-yeo:
It's been a while since SYCL KF test was disabled. (See https://github.com/acts-project/traccc/issues/655)

Maybe there are more things to fix other than the change in this PR. So I suggest to do it with a dedicated PR

krasznaa (approved):
Fine, let's do SYCL separately. I believe we'll need oneAPI 2025.0.0 Docker images as a first step... (And then we'll need to update a bunch of code to be compatible with the new version...)

PR #778, opened 2024-11-22
Count the number of holes in the KF

beomki-yeo:
Based on #774 

The number of holes will be required in the ambiguity solver.
To count the number of holes even after all track candidates are visited, the abort condition in the KF is removed. This will increase the CPU computing time but impact on the GPU KF will be much smaller.

Also added a test to make sure the new feature works correctly

### DIFF
--- a/core/include/traccc/edm/track_state.hpp
+++ b/core/include/traccc/edm/track_state.hpp
@@ -32,6 +32,10 @@ struct fitting_result {
 
     /// Chi square of fitted track
     scalar_type chi2{0};
+
+    // The number of holes (The number of sensitive surfaces which do not have a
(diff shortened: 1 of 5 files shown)


stephenswat:
Don't we already have the hole count form the CKF? Why do we need to recalculate it in the KF?

stephenswat (approved):
Makes sense to me, thank you for the explanation.

beomki-yeo:
OK I didn't change the propagator algorithm and  KF and CKF might yield the same result for the holes. But it is not bad to have the hole counting in KF in case we combine with other track finding methods that don't do any hole counting (e.g. GNN...)

PR #779, opened 2024-11-22
ITk Processing Fixes, main branch (2024.11.22.)

krasznaa:
This is a collection of changes I made to be able to use the files produced by @paradajzblond in the example applications.
  - Had to teach all (relevant) executables to respect the `--input-skip` and `--use-acts-geom-source` command line flags;
  - Turned off memory caching for the host-based throughput applications. I just noticed it while debugging an issue (see next bullet point) that we've been unnecessarily using caching with those applications so far. :thinking:
  - I had to find that with #756 I introduced a bug into `traccc::host::full_chain_algorithm`. :frowning: (Which would allow the code to access invalid memory.) This PR now fixes that.
    * While debugging that, I also found a minor issue in `traccc::host::details::fit_tracks(...)`, where it would use the default memory resource instead of the explicitly provided one for one of the temporary containers,

With all this in place, I'm able to exercise all the executables on Neza's files. :smile:

### DIFF
--- a/core/include/traccc/fitting/details/fit_tracks.hpp
+++ b/core/include/traccc/fitting/details/fit_tracks.hpp
@@ -51,7 +51,7 @@ track_state_container_types::host fit_tracks(
 
         // Make a vector of track states for this track.
         vecmem::vector<track_state<typename fitter_t::algebra_type> >
-            input_states;
+            input_states{&mr};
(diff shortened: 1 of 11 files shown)


beomki-yeo (approved):
Looks reasonable

PR #781, opened 2024-11-23
Fail the KF if the smoother result is invalid

beomki-yeo:


beomki-yeo:
I don't see the same error anymore so let me close the PR at the moment

PR #785, opened 2024-11-26
Make alpaka work with SYCL

StewMH:
I added a queue argument for vecmem::sycl::copy. Not in a very clean way, but this all needs cleaned up (as discussed with @CrossR) and hopefully coming soon in an MR from him or me.

With this I can get the alpaka seeding example to work on the CERN GPU Flex 170:
```
==> Statistics ...
- read    77602 spacepoints
- created  (cpu)  0 seeds
- created (alpaka)  66540 seeds
==>Elapsed times...
            Hit reading  (cpu)  1012 ms
              Seeding (alpaka)  6472 ms
         Track params (alpaka)  1847 ms
                     Wall time  9334 ms
```
It's very slow compared to the CUDA implementation, but I think this is largely because I had to enable FP64 emulation:
>* Some Intel GPUs do not support the `double` type for device code. alpaka will not check this.
  >You can enable software emulation for `double` precision types with
  ```bash
  export IGC_EnableDPEmulation=1
  export OverrideDefaultFP64Settings=1
  ```


examples/run/alpaka/seeding_example_alpaka.cpp
```diff
@@ -43,6 +43,10 @@
 #include "detray/propagator/propagator.hpp"
 #include "detray/propagator/rk_stepper.hpp"
 #include "traccc/alpaka/utils/vecmem_types.hpp"
+#ifdef ALPAKA_ACC_SYCL_ENABLED
+#include <CL/sycl.hpp>
```
krasznaa, line 47: Please use

```suggestion
#include <sycl/sycl.hpp>
```

instead. I'll soon update the project to use this (SYCL2020) convention. So that we could use oneAPI 2025.0.0.

examples/run/alpaka/seeding_example_alpaka.cpp
```diff
@@ -63,8 +67,15 @@ int seq_run(const traccc::opts::track_seeding& seeding_opts,
     using Idx = uint32_t;
 
     using Acc = ::alpaka::ExampleDefaultAcc<Dim, Idx>;
+#ifdef ALPAKA_ACC_SYCL_ENABLED
+    ::sycl::queue q;
+    vecmem::sycl::queue_wrapper qw{&q};
+    traccc::alpaka::vecmem::host_device_types<
+        ::alpaka::trait::AccToTag<Acc>::type>::device_copy copy(qw);
```
krasznaa, line 74: Technically you don't **have to** create a queue_wrapper object like this. You could also do:

```suggestion
    traccc::alpaka::vecmem::host_device_types<
        ::alpaka::trait::AccToTag<Acc>::type>::device_copy copy(&q);
```

And yeah... I'll be interested in your ideas on how to abstract this in a better way... 🤔 (I only have some vague ideas at the moment.)

examples/run/alpaka/seeding_example_alpaka.cpp
```diff
@@ -63,8 +67,15 @@ int seq_run(const traccc::opts::track_seeding& seeding_opts,
...
+#else
     traccc::alpaka::vecmem::host_device_types<
         ::alpaka::trait::AccToTag<Acc>::type>::device_copy copy;
+#endif
     traccc::alpaka::vecmem::host_device_types<
         ::alpaka::trait::AccToTag<Acc>::type>::host_memory_resource host_mr;
```
krasznaa, line 80: Note that it's not just the copy object, but all memory resources also need to be constructed explicitly on top of the same `::sycl::queue` object. Otherwise all sorts of weird behaviours can happen. 🤔 (Though not necessarily using all backends. And with some backends, only if you have multiple possible devices available.)

krasznaa (changes requested):
Overall, very happy about the developments. But we'll indeed need to find some good way of passing in the common `::sycl::queue` from the outside to all the SYCL objects.

I also wonder where we could pass a queue to Alpaka. Or how we could get the queue that Alpaka uses. Since all of these will have to be synced up. 🤔

examples/run/alpaka/CMakeLists.txt
```diff
@@ -13,15 +13,16 @@ set(TRACCC_ALPAKA_EXAMPLE_SOURCES
 
 if(alpaka_ACC_GPU_CUDA_ENABLE)
   enable_language(CUDA)
-  include( traccc-compiler-options-cuda )
   set_source_files_properties(${TRACCC_ALPAKA_EXAMPLE_SOURCES} PROPERTIES LANGUAGE CUDA)
 
-  list (APPEND EXTRA_LIBS vecmem::cuda traccc::cuda)
+  list (APPEND EXTRA_LIBS vecmem::cuda)
 elseif(alpaka_ACC_GPU_HIP_ENABLE)
-  enable_language(HIP)
```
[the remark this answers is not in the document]
StewMH: So, my original plan was to move all of the `enable_language` functions to a `enable_language_alpaka` function. However, the documentation was quite surprising:

>It must be called in the highest directory common to all targets using the named language directly for compiling >sources or indirectly through link dependencies. It is simplest to enable all needed languages in the top-level >directory of a project.

So my intention (I made some mistakes in this PR) is to move this to `traccc/CMakeLists.txt` in the TRACCC_SETUP_ALPAKA section so we're only calling it once and we always have it.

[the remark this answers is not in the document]
krasznaa: 🤔 Indeed, in the most ideal world `traccc::alpaka` should only be using Alpaka privately. While providing a "pure C\+\+" interface. (There is more to this, but let's say this on first order.)

At that point the example applications should indeed not know anything about Alpaka at all.

examples/run/alpaka/seeding_example_alpaka.cpp
StewMH: OK, this makes a lot of sense. I guess they should also be taking the queue as argument?

StewMH: I added queue_wrapper to all the memory_resources.

CMakeLists.txt
```diff
@@ -241,6 +241,16 @@ if( TRACCC_SETUP_ALPAKA )
...
+   #enable_language cannot be called by a function: put it in the top-level directory so it's available for all client code
+   if(alpaka_ACC_GPU_CUDA_ENABLE)
+     enable_language(CUDA)
+     include( traccc-compiler-options-cuda )
+   elseif(alpaka_ACC_GPU_HIP_ENABLE)
+     enable_language(HIP)
+   elseif(alpaka_ACC_SYCL_ENABLE)
+     enable_language(SYCL)
+     include( traccc-compiler-options-sycl )
+   endif()
```
krasznaa, line 253: I don't like this. :thinking: This way we set our own specific flags for all the externals as well. The reason that we don't include `traccc-compile-options-cpp` in the main file is not to pollute the settings of the externals.

You already have these blocks in the appropriate places. If you don't want to have them in 3 different places, then introduce a macro for doing this.

https://cmake.org/cmake/help/latest/command/macro.html

As the name suggests, macros run in the scope in which they are launched. So any variable set inside of a macro, "leaks" into the caller's scope. (Which in this case is what we'd want.)

krasznaa (commented):
Not for this PR, but we should try to find a way of using "factory functions" for instantiating the appropriate type of memory resource and copy objects. To eliminate all appearances of preprocessor statements from the "user code".

But for this PR, as long as you clean up the CMake code a little using a new macro, I'll be happy to let it through. (The change in the main `CMakeLists.txt` file affects others as well, that's why I'd like to get that fixed.)

krasznaa (approved):
Okay, this setup I can live with. :smile:

krasznaa:
Sorry, didn't realize that you would not be allowed to merge once I approved it. :stuck_out_tongue:

PR #786, opened 2024-11-27
Collection Comparison Progress Bar, main branch (2024.11.27.)

krasznaa:
Added a progress bar for the collection comparisons.

For high object counts these comparisons can take a while. To give some feedback to the user about the progress of a comparison, added a progress bar that would provide this information. Now that we have access to such a thing. (#780)

And for why this is really necessary:

image

Hopefully this will help (at least in morals) with debugging the current ITk weirdnesses.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,8 +352,13 @@ if( TRACCC_SETUP_BENCHMARKS )
 endif()
 
 # Set up indicators.
+set( _indicatorsDefault FALSE )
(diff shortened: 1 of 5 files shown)


krasznaa:
Unfortunately SonarCloud is completely missing the point of that code...

image

If I wanted to construct the data object in the class body, I wouldn't need to use [PImpl](https://en.cppreference.com/w/cpp/language/pimpl) in the first place...

stephenswat (approved):
Agreed the SonarCloud warning is nonsense.

PR #788, opened 2024-11-30
Backward filter for Two-filter Smoothing

beomki-yeo:
This PR implements the two-filters smoother where a smoothed state is calculated by taking a weighted average between filtered state from the forward filter and predicted state from the backward filter (See Eq (3.38) of [Pattern Recognition, Tracking and Vertex Reconstruction in Particle Detectors](https://link.springer.com/book/10.1007/978-3-030-65771-0))

Following changes are made to implement the two-filters method:
1. Backward propagator, whose detray actor sequence is reversed, is implemented to run the backward filter
2. Initial covariance is always inflated for both forward and backward filter
3. Test for the uniform distribution of p-values is also added

However, the default smoothing algorithm will still be the RTS smoother (i.e. smoothing algorithm which is already implemented) until we improve the two-filters method with the DirectNavigator of ACTS. The DirectNavigator is required to make sure that forward and backward filter propagate through the same set of surfaces.

### DIFF
--- a/core/include/traccc/edm/track_parameters.hpp
+++ b/core/include/traccc/edm/track_parameters.hpp
@@ -48,4 +48,20 @@ inline void wrap_phi(bound_track_parameters& param) {
     param.set_phi(phi);
 }
 
+/// Covariance inflation used for track fitting
+TRACCC_HOST_DEVICE
+inline void inflate_covariance(bound_track_parameters& param,
+                               const traccc::scalar inf_fac) {
+    auto& cov = param.covariance();
+    for (unsigned int i = 0; i < e_bound_size; i++) {
+        for (unsigned int j = 0; j < e_bound_size; j++) {
+            if (i == j) {
+                getter::element(cov, i, i) *= inf_fac;
+            } else {
+                getter::element(cov, i, j) = 0.f;
+            }
+        }
+    }
+}
+
 }  // namespace traccc
--- a/core/include/traccc/edm/track_state.hpp
+++ b/core/include/traccc/edm/track_state.hpp
@@ -37,6 +37,14 @@ struct fitting_result {
     // The number of holes (The number of sensitive surfaces which do not have a
     // measurement for the track pattern)
     unsigned int n_holes{0u};
+
+    /// Reset the statistics
+    TRACCC_HOST_DEVICE
+    void reset_statistics() {
+        ndf = 0.f;
+        chi2 = 0.f;
+        n_holes = 0u;
+    }
 };
 
 /// Fitting result per measurement
@@ -160,6 +168,14 @@ struct track_state {
     TRACCC_HOST_DEVICE
     inline const scalar_type& filtered_chi2() const { return m_filtered_chi2; }
 
+    /// @return the non-const chi square of backward filter
+    TRACCC_HOST_DEVICE
+    inline scalar_type& backward_chi2() { return m_backward_chi2; }
+
+    /// @return the const chi square of backward filter
+    TRACCC_HOST_DEVICE
+    inline scalar_type backward_chi2() const { return m_backward_chi2; }
+
     /// @return the non-const filtered parameter
     TRACCC_HOST_DEVICE
     inline bound_track_parameters_type& filtered() { return m_filtered; }
@@ -200,6 +216,7 @@ struct track_state {
     bound_track_parameters_type m_filtered;
     scalar_type m_smoothed_chi2 = 0.f;
     bound_track_parameters_type m_smoothed;
+    scalar_type m_backward_chi2 = 0.f;
 };
 
 /// Declare all track_state collection types
--- a/core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
+++ b/core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
@@ -11,6 +11,7 @@
 #include "traccc/definitions/qualifiers.hpp"
 #include "traccc/edm/track_state.hpp"
 #include "traccc/fitting/kalman_filter/gain_matrix_updater.hpp"
+#include "traccc/fitting/kalman_filter/two_filters_smoother.hpp"
 #include "traccc/utils/particle.hpp"
 
 // detray include(s).
@@ -33,6 +34,7 @@ struct kalman_actor : detray::actor {
         state(vector_t<track_state_type>&& track_states)
             : m_track_states(std::move(track_states)) {
             m_it = m_track_states.begin();
+            m_it_rev = m_track_states.rbegin();
         }
 
         /// Constructor with the vector of track states
@@ -40,24 +42,42 @@ struct kalman_actor : detray::actor {
         state(const vector_t<track_state_type>& track_states)
             : m_track_states(track_states) {
             m_it = m_track_states.begin();
+            m_it_rev = m_track_states.rbegin();
         }
 
         /// @return the reference of track state pointed by the iterator
         TRACCC_HOST_DEVICE
-        track_state_type& operator()() { return *m_it; }
+        track_state_type& operator()() {
+            if (!backward_mode) {
+                return *m_it;
+            } else {
+                return *m_it_rev;
+            }
+        }
 
         /// Reset the iterator
         TRACCC_HOST_DEVICE
-        void reset() { m_it = m_track_states.begin(); }
+        void reset() {
+            m_it = m_track_states.begin();
+            m_it_rev = m_track_states.rbegin();
+        }
 
         /// Advance the iterator
         TRACCC_HOST_DEVICE
-        void next() { m_it++; }
+        void next() {
+            if (!backward_mode) {
+                m_it++;
+            } else {
+                m_it_rev++;
+            }
+        }
 
         /// @return true if the iterator reaches the end of vector
         TRACCC_HOST_DEVICE
-        bool is_complete() const {
-            if (m_it == m_track_states.end()) {
+        bool is_complete() {
+            if (!backward_mode && m_it == m_track_states.end()) {
+                return true;
+            } else if (backward_mode && m_it_rev == m_track_states.rend()) {
                 return true;
             }
             return false;
@@ -69,9 +89,15 @@ struct kalman_actor : detray::actor {
         // iterator for forward filtering
         typename vector_t<track_state_type>::iterator m_it;
 
+        // iterator for backward filtering
+        typename vector_t<track_state_type>::reverse_iterator m_it_rev;
+
         // The number of holes (The number of sensitive surfaces which do not
         // have a measurement for the track pattern)
         unsigned int n_holes{0u};
+
+        // Run back filtering for smoothing, if true
+        bool backward_mode = false;
     };
 
     /// Actor operation to perform the Kalman filtering
@@ -99,31 +125,43 @@ struct kalman_actor : detray::actor {
             // Increase the hole counts if the propagator fails to find the next
             // measurement
             if (navigation.barcode() != trk_state.surface_link()) {
-                actor_state.n_holes++;
+                if (!actor_state.backward_mode) {
+                    actor_state.n_holes++;
+                }
                 return;
             }
 
             // This track state is not a hole
-            trk_state.is_hole = false;
+            if (!actor_state.backward_mode) {
+                trk_state.is_hole = false;
+            }
 
             // Run Kalman Gain Updater
             const auto sf = navigation.get_surface();
 
-            const bool res =
-                sf.template visit_mask<gain_matrix_updater<algebra_t>>(
+            bool res = false;
+
+            if (!actor_state.backward_mode) {
+                // Forward filter
+                res = sf.template visit_mask<gain_matrix_updater<algebra_t>>(
                     trk_state, propagation._stepping.bound_params());
+
+                // Update the propagation flow
+                stepping.bound_params() = trk_state.filtered();
+
+                // Set full jacobian
+                trk_state.jacobian() = stepping.full_jacobian();
+            } else {
+                // Backward filter for smoothing
+                res = sf.template visit_mask<two_filters_smoother<algebra_t>>(
+                    trk_state, propagation._stepping.bound_params());
+            }
 
             // Abort if the Kalman update fails
             if (!res) {
                 propagation._heartbeat &= navigation.abort();
                 return;
             }
-
-            // Update the propagation flow
-            stepping.bound_params() = trk_state.filtered();
-
-            // Set full jacobian
-            trk_state.jacobian() = stepping.full_jacobian();
 
             // Change the charge of hypothesized particles when the sign of qop
             // is changed (This rarely happens when qop is set with a poor seed
--- a/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
+++ b/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
@@ -17,6 +17,7 @@
 #include "traccc/fitting/kalman_filter/kalman_actor.hpp"
 #include "traccc/fitting/kalman_filter/kalman_step_aborter.hpp"
 #include "traccc/fitting/kalman_filter/statistics_updater.hpp"
+#include "traccc/fitting/kalman_filter/two_filters_smoother.hpp"
 #include "traccc/utils/particle.hpp"
 
 // detray include(s).
@@ -67,9 +68,16 @@ class kalman_fitter {
         detray::actor_chain<detray::dtuple, aborter, transporter, interactor,
                             fit_actor, resetter, kalman_step_aborter>;
 
+    using backward_actor_chain_type =
+        detray::actor_chain<detray::dtuple, aborter, transporter, fit_actor,
+                            interactor, resetter, kalman_step_aborter>;
+
     // Propagator type
     using propagator_type =
         detray::propagator<stepper_t, navigator_t, actor_chain_type>;
+
+    using backward_propagator_type =
+        detray::propagator<stepper_t, navigator_t, backward_actor_chain_type>;
 
     /// Constructor with a detector
     ///
@@ -104,6 +112,14 @@ class kalman_fitter {
                                m_resetter_state, m_step_aborter_state);
         }
 
+        /// @return the actor chain state
+        TRACCC_HOST_DEVICE
+        typename backward_actor_chain_type::state backward_actor_state() {
+            return detray::tie(m_aborter_state, m_transporter_state,
+                               m_fit_actor_state, m_interactor_state,
+                               m_resetter_state, m_step_aborter_state);
+        }
+
         /// Individual actor states
         typename aborter::state m_aborter_state{};
         typename transporter::state m_transporter_state{};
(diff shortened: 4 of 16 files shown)


core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
```diff
@@ -33,31 +34,50 @@ struct kalman_actor : detray::actor {
...
+            }
+        }
 
         /// @return true if the iterator reaches the end of vector
         TRACCC_HOST_DEVICE
         bool is_complete() const {
-            if (m_it == m_track_states.end()) {
+            if (!backward_mode && m_it == m_track_states.end()) {
+                return true;
+            } else if (backward_mode && m_it_rev == m_track_states.rend()) {
```
beomki-yeo, line 80: @krasznaa 

Can you figure out why the following error appears (only for CUDA):
```
/mnt/nvme0n1/byeo/projects/traccc/traccc/core/include/traccc/fitting/kalman_filter/kalman_actor.hpp(80): error: no operator "==" matches these operands
            operand types are: const vecmem::details::reverse_iterator<traccc::track_state<traccc::default_algebra> *> == vecmem::details::reverse_iterator<const traccc::track_state<traccc::default_algebra> *>
              } else if (backward_mode && m_it_rev == m_track_states.rend()) {
                                                   ^
/mnt/nvme0n1/byeo/projects/traccc/traccc/core/include/traccc/edm/measurement.hpp(79): note #3326-D: function "traccc::operator==" does not match because argument #1 does not match parameter
  inline bool operator==(const measurement& lhs, const measurement& rhs) {
              ^
/mnt/nvme0n1/byeo/projects/traccc/traccc_build/_deps/vecmem-src/core/include/vecmem/containers/impl/reverse_iterator.ipp(161): note #3327-D: candidate function template "vecmem::details::operator==" failed deduction
  __attribute__((host)) __attribute__((device)) bool operator==(const reverse_iterator<T>& itr1,
                                                     ^
/mnt/nvme0n1/byeo/projects/traccc/traccc/core/include/traccc/fitting/kalman_filter/kalman_actor.hpp(80): note #3328-D: built-in operator==(<promoted arithmetic>, <promoted arithmetic>) does not match because argument #1 does not match parameter
              } else if (backward_mode && m_it_rev == m_track_states.rend()) {
                                                   ^
/mnt/nvme0n1/byeo/projects/traccc/traccc/core/include/traccc/fitting/kalman_filter/kalman_actor.hpp(80): note #3328-D: built-in operator==(<nullptr>, <nullptr>) does not match because argument #1 does not match parameter
              } else if (backward_mode && m_it_rev == m_track_states.rend()) {
                                                   ^
          detected during:
```

stephenswat: It's because the right hand side is const and the left hand side is not. I think this will require a fix in vecmem.

beomki-yeo: OK I don't get how `rend` has const type

stephenswat: `m_track_states` is const, and calling `rend` on a const value returns a const iterator. `rend` on a non-const value gives a non-const iterator, and `crend` always gives a const iterator.

stephenswat: By the way, you can also set `m_it_rev` to be a `const_reverse_iterator`.

beomki-yeo: It compiles after removing the constness of the `is_complete` function

I think the const function tries to call the `const_reverse_iterator` which causes the problem
```c++
    /// Return a reverse iterator pointing at the beginning of the vector
    VECMEM_HOST_AND_DEVICE
    reverse_iterator rend();
    /// Return a constant reverse iterator pointing at the beginning of the
    /// vector
    VECMEM_HOST_AND_DEVICE
    const_reverse_iterator rend() const;
```

beomki-yeo: Ah our comments crossed - I will use the const iterator

stephenswat: That's intentional and matches the way [`std::vector::end`](https://en.cppreference.com/w/cpp/container/vector/end) works.

Indeed you can either:

1. Remove the const qualifier on `is_complete` which makes `m_track_states` non-const and thus makes the iterator non-const; or
2. Turn `m_it_rev` from a `reverse_iterator` to a `const_reverse_iterator`; or
3. Modify vecmem to allow you to compare const and non-const iterators.

stephenswat: > EDIT: Maybe not - the const iterator conflict with the function of `operator()`..

What do you mean? That sounds like a problem. :thinking:

stephenswat: Fix in https://github.com/acts-project/vecmem/pull/307.

tests/cpu/test_ckf_sparse_tracks_telescope.cpp
```diff
@@ -123,9 +126,10 @@ TEST_P(CkfSparseTrackTelescopeTests, Run) {
     // Finding algorithm configuration
     typename traccc::finding_config cfg;
     cfg.ptc_hypothesis = ptc;
-    cfg.chi2_max = 30.f;
+    cfg.chi2_max = 200.f;
```
beomki-yeo, line 129: After I set the B field to z-axis and Telescope planes set along the x-axis, chi2 cut value of 30 for the CKF selection did not work for a few measurements :/ 

I believe the cut value of 30 should be OK in most cases from the distribution of chi2 of the Kalman Fitter (See the plot below). The change to 200 is just to cover the outliers which happen very rarely

andiwand:
If I read the ACTS implementation correctly it just rewinds the direction and uses the usual KF formalism but then sets `smoothed` instead of `filtered` https://github.com/acts-project/acts/blob/2bb167c0ba1be5c66b55ca76308f8b84e609f489/Core/include/Acts/TrackFitting/KalmanFitter.hpp#L696

That does not seem correct to me 🤔 since we would use measurements twice without accounting this in the covariance.

Have you looked at this @beomki-yeo? I suppose you use the formalism provided in the reference?

cc @asalzburger @pbutti

beomki-yeo:
> it just rewinds the direction and uses the usual KF formalism but then sets smoothed instead of filtered

If I read the code correctly, `apply_backwards` function sets the backward-filtered state to the smoothed state? In this PR, we use the forward-fileted state and backward-predicted state to calculate the smoothed state as stated in the reference. 

And It seems the ACTS backward filter does not inflate covariance. I inflated the covariance of both forward and backward propagator. In the various references, I found that there are two boundary conditions:
1. smoothed-state of the last measurement = forward-filtered state of the last measurement 
2. inverse of initial covariance of backward propagator = zero matrix, which means a very big initial covariance matrix

The second condition sounds a bit weird because there is no inverse matrix for a zero matrix but I interpreted this as a big diagonal initial covariance matrix.

beomki-yeo:
That does not seem correct to me 🤔 since we would use measurements twice without accounting this in the covariance.

> Yeah it seems some measurements are used twice;

core/include/traccc/edm/track_state.hpp
```diff
@@ -37,6 +37,14 @@ struct fitting_result {
     // The number of holes (The number of sensitive surfaces which do not have a
     // measurement for the track pattern)
     unsigned int n_holes{0u};
+
+    // Reset the statistics
```
niermann999, line 41:
```suggestion
    /// Reset the statistics
```

core/include/traccc/edm/track_state.hpp
```diff
@@ -160,6 +168,14 @@ struct track_state {
     TRACCC_HOST_DEVICE
     inline const scalar_type& filtered_chi2() const { return m_filtered_chi2; }
 
+    /// @return the non-const chi square of backward filter
+    TRACCC_HOST_DEVICE
+    inline scalar_type& backward_chi2() { return m_backward_chi2; }
+
+    /// @return the const chi square of backward filter
+    TRACCC_HOST_DEVICE
+    inline const scalar_type& backward_chi2() const { return m_backward_chi2; }
```
niermann999, line 177:
```suggestion
    inline scalar_type backward_chi2() const { return m_backward_chi2; }
```

core/include/traccc/edm/track_parameters.hpp
```diff
@@ -48,4 +48,20 @@ inline void wrap_phi(bound_track_parameters& param) {
     param.set_phi(phi);
 }
 
+// Covariance inflation used for track fitting
```
niermann999, line 51:
```suggestion
/// Covariance inflation used for track fitting
```

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
```diff
@@ -33,31 +34,50 @@ struct kalman_actor : detray::actor {
         state(vector_t<track_state_type>&& track_states)
             : m_track_states(std::move(track_states)) {
             m_it = m_track_states.begin();
+            m_it_rev = m_track_states.rbegin();
```
niermann999, line 37: Does it have to be a reverse iterator? Is it not possible to set a bidirectionl iterator to the last element (```m_itr = detray::detail::prev(m_track_states.end())```) and go back? Then it would not need a second iterator and the direction can be reversed at any time. This could be put at a single point in :
```
void reset() {
     m_it = backward_mode ? detray::detail::prev(m_track_states.end()) : m_track_states.begin();
 }
```

Also, if there were two different methods ```next()``` and ```previous()```, they could be called in the if-else block in line 141 and there wound not need to be an if-else branch in ```next()```

niermann999: Otherwise, I am not so sure why these two workflows are contained in a single class? They seem to be pretty separate from each other, except that they act on the same data. We continue to switch between them with a boolean flag at every step of the algorithm? There are two different actor chains for fitting and smoothing anyway...

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
```diff
@@ -109,22 +135,31 @@ struct kalman_actor : detray::actor {
             // Run Kalman Gain Updater
             const auto sf = navigation.get_surface();
 
-            const bool res =
-                sf.template visit_mask<gain_matrix_updater<algebra_t>>(
+            bool res = false;
+
+            // Forward filter
+            if (!actor_state.backward_mode) {
```
niermann999, line 141: Is the whole counting needed again during smoothing or does it need to be moved into this if-block?

core/include/traccc/fitting/kalman_filter/two_filters_smoother.hpp
```diff
@@ -0,0 +1,173 @@
...
+            bound_params.covariance();
+
+        const matrix_type<e_bound_size, e_bound_size> predicted_cov_inv =
+            matrix::inverse(predicted_cov);
+        const matrix_type<e_bound_size, e_bound_size> filtered_cov_inv =
+            matrix::inverse(trk_state.filtered().covariance());
+
+        // Eq (3.38) of "Pattern Recognition, Tracking and Vertex
+        // Reconstruction in Particle Detectors"
+        const matrix_type<e_bound_size, e_bound_size> smoothed_cov_inv =
```
niermann999, line 91: Is this not a 
```suggestion
        const bound_matrix_type smoothed_cov_inv =
```
?
Same in other places and also for the ```bound_vector_type```

tests/common/tests/kalman_fitting_test.cpp
```diff
@@ -81,6 +83,58 @@ void KalmanFittingTests::pull_value_tests(
 #endif  // TRACCC_HAVE_ROOT
 }
 
+void KalmanFittingTests::p_value_tests(std::string_view file_name) const {
+
+    // Avoid unused variable warnings when building the code without ROOT.
+    (void)file_name;
```
niermann999, line 89:
```suggestion
void KalmanFittingTests::p_value_tests([[maybe_unused]] std::string_view file_name) const {
```

tests/cpu/test_kalman_fitter_wire_chamber.cpp
```diff
@@ -124,6 +124,7 @@ TEST_P(KalmanFittingWireChamberTests, Run) {
         static_cast<float>(mask_tolerance);
     fit_cfg.propagation.navigation.search_window = search_window;
     fit_cfg.ptc_hypothesis = ptc;
+    fit_cfg.covariance_inflation_factor = 1.f;
```
niermann999, line 127: Is this something specific for the wire chamber?

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
beomki-yeo: Sorry, what is the counting?

niermann999: A sorry, typo! I meant hole counting (l.127)

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
```diff
@@ -104,6 +112,14 @@ class kalman_fitter {
                                m_resetter_state, m_step_aborter_state);
         }
 
+        /// @return the actor chain state
+        TRACCC_HOST_DEVICE
```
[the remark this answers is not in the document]
beomki-yeo: I implemented this because compiler spits `no matching function for call` error with `actor_state()`

tests/cpu/test_kalman_fitter_wire_chamber.cpp
beomki-yeo: It is for not changing the algorithm used in wire chamber test ( RTS smoother and no covariance inflation like in the past )

And yes, it is not working for the wire chamber after inflating covariance which will introduce an error in navigation result. Eventually, the surface in the track candidate won't be reached because cell of wire chamber is so small (~ 1 cm scale)

The navigator used for **forward** fitter needs a fix to make it find the sensitive surface correctly. (To be clear, I am not talking about the direct navigator). We need to make the mask tolerance for sensitive surfaces in the `track_candidate` be large enough.

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
niermann999: > I think it is convenient to have them in the same actor because smoother need the filtered state information from the forward filter.

Does the smoothing step need to own the track states? It also seems a bit inconvenient to always have to add an ```if(backward_mode) {...} else {...}``` everywhere. But I leave that up to you

> Using `detail::prev(m_track_states.end())` sounds interesting but I am not very familiar with this grammar. I don't think it is going to make big difference either

It is the same as ```std::prev```, just working on device. It moves an iterator one position backwards, since ```end()``` is one position behind the last element of the range

beomki-yeo: > Does the smoothing step need to own the track states? 

As long I don't want to copy the forward filtered state to somewhere else, yes it need to own

beomki-yeo: > It is the same as std::prev, just working on device.

Anyway you will still need to have if-block for `next` or `is_complete`. (Not even sure how I can write `is_complete` function with `std::prev`)

core/include/traccc/fitting/kalman_filter/two_filters_smoother.hpp
beomki-yeo: To be honest, I prefer using `matrix_type<int, int>`, which is more informative to readers...

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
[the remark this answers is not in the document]
niermann999: Oh, you are right. I think we removed some templating on the propagate function call and now it requires a particular actor state tuple type...

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
niermann999: > > It is the same as std::prev, just working on device.
> 
> Anyway you will still need to have if-block for `next` or `is_complete`. (Not even sure how I can write `is_complete` function with `std::prev`)

My initial point was that if you call ```next()``` in the if block in l.141 and a new method ```previous()``` in the else block below, the if-else block in ```next()``` would not actually be needed... The ```operator()``` would also not need an if-else block anymore.

Yes, ```is_complete()``` would need to be checked against ```begin()```, that is true

beomki-yeo: > I am not following, it just dereferences an iterator to get a reference to the track state or not? It can do that without owning the track_state container that the iterator points to or what am I missing?

You are right. So is your suggestion to separate the `track_states` from the actor and have it accessed by two different actors made for forward filter and backward filter?

beomki-yeo: Someone might attempt that in the next PRs but let's keep the changes minimal in this PR.

core/include/traccc/edm/track_state.hpp
beomki-yeo: Fixed

tests/cpu/test_kalman_fitter_wire_chamber.cpp
niermann999: OK, interesting, thanks for explaining! I always found making the propagation integration test work for the wirechamber quite fiddly. Maybe we can start with the propagation conf used there. I am also worried by various overlaps that are still present in that geometry

core/include/traccc/edm/track_state.hpp
beomki-yeo: Fixed

core/include/traccc/edm/track_parameters.hpp
beomki-yeo: Fixed

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
beomki-yeo: Fixed

tests/common/tests/kalman_fitting_test.cpp
beomki-yeo: Fixed

core/include/traccc/fitting/kalman_filter/two_filters_smoother.hpp
niermann999: Sure enough! Having just gone through the algebra plugins templating cleanup, I slightly prefer centrally defined type aliases that I don't have to chase through the code, but that is a matter of taste I guess :)

tests/cpu/test_kalman_fitter_wire_chamber.cpp
beomki-yeo: For the moment, making the mask tolerance bigger only for the track candidate surface seems to be enough to me. Making it bigger for all surfaces will be a disaster. I don't know if such selective tolerance setup is possible in detray

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
niermann999: > > I am not following, it just dereferences an iterator to get a reference to the track state or not? It can do that without owning the track_state container that the iterator points to or what am I missing?
> 
> You are right. So is your suggestion to separate the `track_states` from the actor and have it accessed by two different actors made for forward filter and backward filter?

Yes, instead of moving the ```input_states``` into the fitter state, the fitter/smoother states could simply be views onto the track states (views in the sense of std::ranges not vecmem)

tests/cpu/test_kalman_fitter_wire_chamber.cpp
niermann999: From what I understand atm, we need a new mechanism in the navigator anyway that forces it to stop at some externally provided surfaces. Among them also the perigee perhaps

niermann999 (approved):
OK, lets get this in an fix stuff later

PR #773, opened 2024-11-15
SYCL Track Finding, main branch (2024.11.14.)

krasznaa:
This is finally a CKF algorithm for SYCL. 🥳

Still some technical stuff to set up, I just wanted to open it already, so people would be aware. 😄 (I'll open at least one PR off of this first, which should go in before this one.)

PR #790, opened 2024-12-02
Thrust Usage Update, main branch (2024.12.02.)

krasznaa:
Changed how [Thrust](https://nvidia.github.io/cccl/thrust/) would be used in the build.

Instead of setting up a single `traccc::Thrust` target that would take care of everything related to Thrust, the code rather sets up `traccc::Thrust` in a way that allows very little access in `traccc::core` to Thrust.

Instead, `traccc::cuda` and `traccc::sycl` (and also `traccc_test_cuda`) privately link against Thrust in specific ways. So that the appropriate "device headers" would get included everywhere.

Note that this shall not only help the SYCL developments (the commit comes directly from #773...), but the HIP ones as well. (Pinging @flg and @StewMH.)

### DIFF
--- a/device/cuda/CMakeLists.txt
+++ b/device/cuda/CMakeLists.txt
@@ -106,7 +106,14 @@ target_compile_options( traccc_cuda
   PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:--expt-relaxed-constexpr> )
 target_link_libraries( traccc_cuda
   PUBLIC traccc::core detray::core vecmem::core covfie::core
-  PRIVATE CUDA::cudart traccc::Thrust traccc::device_common vecmem::cuda )
+  PRIVATE CUDA::cudart traccc::device_common vecmem::cuda )
+
+# Set up Thrust specifically for the traccc::cuda library.
+thrust_create_target( traccc::cuda_thrust
+  HOST CPP
+  DEVICE CUDA )
+target_link_libraries( traccc_cuda
+  PRIVATE traccc::cuda_thrust )
 
 # For CUDA 11 turn on separable compilation. This is necessary for using
 # Thrust 2.1.0.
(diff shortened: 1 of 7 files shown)


device/cuda/CMakeLists.txt
```diff
@@ -106,7 +106,14 @@ target_compile_options( traccc_cuda
...
   PUBLIC traccc::core detray::core vecmem::core covfie::core
-  PRIVATE CUDA::cudart traccc::Thrust traccc::device_common vecmem::cuda )
+  PRIVATE CUDA::cudart traccc::device_common vecmem::cuda )
+
+# Set up Thrust specifically for the traccc::cuda library.
+thrust_create_target( traccc::cuda_thrust
+  HOST CPP
+  DEVICE CUDA )
+target_link_libraries( traccc_cuda
+  PRIVATE traccc::cuda_thrust )
```
stephenswat, line 116: Out of curiosity, if you link a target against multiple of these Thrust libraries, what happens? Does it give an error at configuration time or does it cause include ambiguities? :thinking:

stephenswat (approved):
Looks good. Happy to approve if the CI accepts this.

device/cuda/CMakeLists.txt
krasznaa: I'm desperately trying to find it in the code, but don't remember all the details anymore. :frowning:

I **think** the issue comes from the specific Thrust targets setting some custom target properties on themselves. See:

https://github.com/NVIDIA/cccl/blob/v2.6.1/thrust/thrust/cmake/thrust-config.cmake#L447-L448

If I remember correctly what I saw was that in case a target ends up being linked against let's say both `Thrust::CPP::Device` and `Thrust::CUDA::Device`, then CMake freaks out during the build system generation stage. Stating that the value of `THRUST_INTERFACE_DEVICE` couldn't be determined for the user's target. Or something like that.

But definitely, already during the CMake configuration one gets an error if trying to link against multiple "host" or "device" libraries at the same time. This is why `traccc::Thrust` is set up in this hacky way now. That I really can't use `thrust_create_target(...)` to make a target that would be appropriate for `traccc::core`. :frowning:

PR #791, opened 2024-12-02
Atomic Reference Fixes, main branch (2024.12.02.)

krasznaa:
Make atomic references use correct address spaces in the track finding device code.

With CUDA there is no difference in the API for using atomic operations on global or local/shared memory. But with SYCL there is. So this is a necessary update for #773. (But also, if we ever start using [cuda::atomic_ref](https://nvidia.github.io/cccl/libcudacxx/extended_api/memory_model.html) behind `vecmem::device_atomic_ref`, this may become relevant for CUDA as well. :thinking: Maybe...


stephenswat (approved):
Yup, makes sense. :+1:

PR #792, opened 2024-12-03
Externals Update, main branch (2024.12.03.)

krasznaa:
Continuing with the updates necessary for #773, this PR updates the externals to a state that oneAPI 2025.0.0 would be on board with.

Increased the required CMake version to 3.25 in the externals, to be able to finally simplify the CMake code setting them up, at least a little. Mainly to be able to use the `SYSTEM` flag everywhere.

At the same time simplified how cache variables (like `BUILD_TESTING`) would be used. After some careful testing, it turns out that one can override cache variables in subdirectories much more easily than I thought in the past.

At the same time, updated the versions of:
  - Covfie;
  - Detray;
  - Algebra Plugins (This one uses a custom patch from my own fork. Once @niermann999's updates settle down around Algebra Plugins and Detray, of course we shall switch back to an "official tag".);
  - Vecmem;
  - TBB.

All so that oneAPI 2025.0.0 could be used for the (SYCL) build.


stephenswat (approved):
:+1:

PR #795, opened 2024-12-03
Add an argument for B field in the Telescope KF test

beomki-yeo:
Perhaps I will also need this for #788


krasznaa (approved):
I'm fine with this, but I wonder... Why not make the test completely generic at this point? Instead of passing in a single vector for generating a constant field with, why not instead pass in a Covfie object that may have any kind of field?

I'm not sure what the best formalism for that would be, but that's where @stephenswat would come in. :wink:

But as I started, this is fine as-is for me.

beomki-yeo:
Any kind of field including inhomogeneous ones? If I am correct, homogeneous and inhomogeneous ones are defined with different classes so maybe not an easy thing to implement :thinking:

PR #796, opened 2024-12-04
Use Switch Instead of If, main branch (2024.12.04.)

krasznaa:
As a bit of a palette cleanser between larger / messier PRs, I present this simple one. :smile:

I came across these `if(...)` blocks while working on the SoA-ification of `traccc::measurement`. :thinking: I think these changes should allow the compiler to generate more efficient code. Though probably not by a whole lot.

### DIFF
--- a/core/include/traccc/edm/track_state.hpp
+++ b/core/include/traccc/edm/track_state.hpp
@@ -79,14 +79,11 @@ struct track_state {
     TRACCC_HOST_DEVICE inline matrix_type<D, 1> measurement_local() const {
         static_assert(((D == 1u) || (D == 2u)),
                       "The measurement dimension should be 1 or 2");
+        assert((m_measurement.subs.get_indices()[0] == e_bound_loc0) ||
+               (m_measurement.subs.get_indices()[0] == e_bound_loc1));
 
         matrix_type<D, 1> ret;
         switch (m_measurement.subs.get_indices()[0]) {
-            default:
-                assert(
-                    "The measurement index out of e_bound_loc0 and "
-                    "e_bound_loc1 should not happen.");
-                [[fallthrough]];
             case e_bound_loc0:
                 matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
                 if constexpr (D == 2u) {
@@ -101,6 +98,8 @@ struct track_state {
                         m_measurement.local[0];
                 }
                 break;
+            default:
+                __builtin_unreachable();
         }
         return ret;
     }
@@ -110,14 +109,11 @@ struct track_state {
     TRACCC_HOST_DEVICE inline matrix_type<D, D> measurement_covariance() const {
         static_assert(((D == 1u) || (D == 2u)),
                       "The measurement dimension should be 1 or 2");
+        assert((m_measurement.subs.get_indices()[0] == e_bound_loc0) ||
+               (m_measurement.subs.get_indices()[0] == e_bound_loc1));
 
         matrix_type<D, D> ret;
         switch (m_measurement.subs.get_indices()[0]) {
-            default:
-                assert(
-                    "The measurement index out of e_bound_loc0 and "
-                    "e_bound_loc1 should not happen.");
-                [[fallthrough]];
             case e_bound_loc0:
                 matrix_operator().element(ret, 0, 0) =
                     m_measurement.variance[0];
@@ -138,6 +134,8 @@ struct track_state {
                         m_measurement.variance[0];
                 }
                 break;
+            default:
+                __builtin_unreachable();
         }
         return ret;
     }
--- a/core/include/traccc/edm/track_state.hpp
+++ b/core/include/traccc/edm/track_state.hpp
@@ -81,26 +81,27 @@ struct track_state {
                       "The measurement dimension should be 1 or 2");
 
         matrix_type<D, 1> ret;
-        if (m_measurement.subs.get_indices()[0] == e_bound_loc0) {
-            matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
-            if constexpr (D == 2u) {
-                matrix_operator().element(ret, 1, 0) = m_measurement.local[1];
-            }
-        } else if (m_measurement.subs.get_indices()[0] == e_bound_loc1) {
-            matrix_operator().element(ret, 0, 0) = m_measurement.local[1];
-            if constexpr (D == 2u) {
-                matrix_operator().element(ret, 1, 0) = m_measurement.local[0];
-            }
-        } else {
-            assert(
-                "The measurement index out of e_bound_loc0 and e_bound_loc1 "
-                "should not happen.");
-            matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
-            if constexpr (D == 2u) {
-                matrix_operator().element(ret, 1, 0) = m_measurement.local[1];
-            }
+        switch (m_measurement.subs.get_indices()[0]) {
+            default:
+                assert(
+                    "The measurement index out of e_bound_loc0 and "
+                    "e_bound_loc1 should not happen.");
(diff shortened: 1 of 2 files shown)


core/include/traccc/edm/track_state.hpp
```diff
@@ -81,26 +81,27 @@ struct track_state {
...
-        } else {
-            assert(
-                "The measurement index out of e_bound_loc0 and e_bound_loc1 "
-                "should not happen.");
-            matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
-            if constexpr (D == 2u) {
-                matrix_operator().element(ret, 1, 0) = m_measurement.local[1];
-            }
+        switch (m_measurement.subs.get_indices()[0]) {
+            default:
```
stephenswat, line 85: Why is this at the top and not at the bottom as is more common in switch statements?

core/include/traccc/edm/track_state.hpp
```diff
@@ -111,36 +112,32 @@ struct track_state {
...
-            if constexpr (D == 2u) {
-                matrix_operator().element(ret, 0, 1) = 0.f;
-                matrix_operator().element(ret, 1, 0) = 0.f;
-                matrix_operator().element(ret, 1, 1) =
+        switch (m_measurement.subs.get_indices()[0]) {
+            default:
+                assert(
+                    "The measurement index out of e_bound_loc0 and "
+                    "e_bound_loc1 should not happen.");
+                [[fallthrough]];
```
stephenswat, line 120: Same comments here.

core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp
```diff
@@ -45,13 +45,14 @@ struct gain_matrix_updater {
...
-            return update<1u, shape_type>(trk_state, bound_params);
-        } else if (D == 2u) {
-            return update<2u, shape_type>(trk_state, bound_params);
+        switch (D) {
+            case 1u:
+                return update<1u, shape_type>(trk_state, bound_params);
+            case 2u:
+                return update<2u, shape_type>(trk_state, bound_params);
+            default:
+                return false;
```
stephenswat, line 54: I'd rather finish this with a `__builtin_unreachable`.

stephenswat (changes requested):
No, I don't like this. Not only are switch statements arguably less clear than if-else constructs, but putting the default case at the top, falling through after assertions, etc. are all a no-go as far as I am concerned.

I appreciate that this mimics the original code but let's redesign this properly rather than patching it up with `switch` statements. :confused:

core/include/traccc/edm/track_state.hpp
```diff
@@ -81,26 +81,27 @@ struct track_state {
...
-            matrix_operator().element(ret, 0, 0) = m_measurement.local[0];
-            if constexpr (D == 2u) {
-                matrix_operator().element(ret, 1, 0) = m_measurement.local[1];
-            }
+        switch (m_measurement.subs.get_indices()[0]) {
+            default:
+                assert(
+                    "The measurement index out of e_bound_loc0 and "
+                    "e_bound_loc1 should not happen.");
+                [[fallthrough]];
```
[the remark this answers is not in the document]
krasznaa: Unfortunately that's not how `assert(...)` works. :frowning: I think people have used stuff like

```c++
assert(false && "Some explanation");
```

in their codes. (https://stackoverflow.com/questions/3692954/add-custom-messages-in-assert) But this is not super portable either.

You're absolutely right though, that this is an incorrect use of `assert(...)` in general.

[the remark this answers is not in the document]
krasznaa: I just took the current behaviour, as you've seen as well. Not sure what we should do here instead in an optimised build. :thinking: Generally, I'd like to print an error message. And then either make the kernel stop with an error, or let the kernel continue.

This ties into discussions that we had not long ago about some organised logging infrastructure for our GPU code. :thinking:

core/include/traccc/edm/track_state.hpp
krasznaa: To be able to fall through it. I thought that was clear.

Indeed, this is not a standard setup. But it seemed like a neat way of achieving the current behaviour.

Whether we want to change the current behaviour is a discussion to be had though.

core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp
krasznaa: It's not quite unreachable though, is it? :thinking: In a release build the assertion would let this code be reached. At which point we can once again go back to my previous thoughtline.
  - Print an error message;
  - Either stop the execution, or let it continue as it currently does.
    * Letting it continue, with an error communicated back to the caller about failing the update, does not seem that crazy to me. :thinking:

core/include/traccc/edm/track_state.hpp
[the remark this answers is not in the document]
stephenswat: Right yes, sorry; should be `false && ...`; but just putting in a string will ensure that the assertion always succeeds, defeating the point.

core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp
stephenswat: Of course it is not _technically_ unreachable but it is semantically unreachable, which is exactly what is modelled by the `assert`. I think it's fair to say that the code reaching this point without `D` being either 1 or 2 can be declared undefined behaviour, which is exactly what marking it as unreachable does.

core/include/traccc/edm/track_state.hpp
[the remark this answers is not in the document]
stephenswat: We should absolutely change the behaviour! Just as below I believe the correct way to do this is to define a case for `loc0` and for `loc1` and to mark the default case as unreachable.

krasznaa:
Okay, I added `__builtin_unreachable()` to these blocks. Have a look!

Of course [std::unreachable()](https://en.cppreference.com/w/cpp/utility/unreachable) would be even nicer, but that will be some while away still...

core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp
```diff
@@ -45,13 +45,18 @@ struct gain_matrix_updater {
...
-        } else if (D == 2u) {
-            return update<2u, shape_type>(trk_state, bound_params);
+        bool result = false;
+        switch (D) {
+            case 1u:
+                result = update<1u, shape_type>(trk_state, bound_params);
+                break;
+            case 2u:
+                result = update<2u, shape_type>(trk_state, bound_params);
+                break;
```
stephenswat, line 55:
```suggestion
            case 1u:
                return update<1u, shape_type>(trk_state, bound_params);
            case 2u:
                return update<2u, shape_type>(trk_state, bound_params);
```

stephenswat (commented):
Thanks! Looks a lot better already. :smile:

core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp
krasznaa: The reason I went with that setup was two-fold:
  - I'm not 100% sure what happens with `__builtin_unreachable()` on all the platforms that this code will run on. 

I was considering putting

```c++
default:
    __builtin_unreachable();
    return false;
```

there, "just to be safe". But this formalism seemed better.

  - GPU functions can sometimes have hangups about returning at different points. Many coding rules even forbid multiple return points from functions. So I thought this may actually help.

But admittedly, I'm not absolutely sure about either of these points. :thinking:

stephenswat (approved):
Happy with this! :+1:

PR #798, opened 2024-12-05
ref: Update to detray version 0.85.0 and adapt to new algebra-plugins API

niermann999:
Test if the new algebra plugin integration works.

Edit: It turns out that algebra plugins has included the arithmetic operators for ```std::array``` vector types in the global namespace from where they were picked up in the ACTS VectorHelper by Eigen types. The operators are now moved into the algebra namespace and from there imported into the detray and traccc namspaces.

### DIFF
--- a/core/include/traccc/definitions/primitives.hpp
+++ b/core/include/traccc/definitions/primitives.hpp
@@ -14,7 +14,7 @@
 #elif ALGEBRA_PLUGINS_INCLUDE_SMATRIX
 #include "traccc/plugins/algebra/smatrix_definitions.hpp"
 #elif ALGEBRA_PLUGINS_INCLUDE_VC
-#include "traccc/plugins/algebra/vc_definitions.hpp"
+#include "traccc/plugins/algebra/vc_aos_definitions.hpp"
 #elif ALGEBRA_PLUGINS_INCLUDE_VECMEM
 #include "traccc/plugins/algebra/vecmem_definitions.hpp"
 #endif
@@ -40,7 +40,9 @@ using point3 = detray::dpoint3D<default_algebra>;
 using vector3 = detray::dvector3D<default_algebra>;
 using variance3 = point3;
 using transform3 = detray::dtransform3D<default_algebra>;
-template <std::size_t ROW, std::size_t COL>
-using matrix = detray::dmatrix<default_algebra, ROW, COL>;
+
+namespace getter = detray::getter;
+namespace vector = detray::vector;
+namespace matrix = detray::matrix;
 
 }  // namespace traccc
--- a/io/src/csv/read_measurements.cpp
+++ b/io/src/csv/read_measurements.cpp
@@ -44,7 +44,7 @@ void read_measurements(measurement_collection_types::host& measurements,
 
         // Construct the measurement object.
         traccc::measurement meas;
-        std::array<typename transform3::size_type, 2u> indices{0u, 0u};
+        std::array<detray::dsize_type<default_algebra>, 2u> indices{0u, 0u};
         meas.meas_dim = 0u;
 
         // Local key is a 8 bit char and first and last bit are dummy value. 2 -
--- a/core/include/traccc/edm/track_state.hpp
+++ b/core/include/traccc/edm/track_state.hpp
@@ -8,6 +8,7 @@
 #pragma once
 
 // Project include(s).
+#include "traccc/definitions/primitives.hpp"
 #include "traccc/definitions/qualifiers.hpp"
 #include "traccc/edm/container.hpp"
 #include "traccc/edm/measurement.hpp"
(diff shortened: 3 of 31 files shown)


krasznaa:
Just to make it obvious that this is still a draft, as it uses test/development versions of algebra-plugins and detray.

niermann999:
> Just to make it obvious that this is still a draft, as it uses test/development versions of algebra-plugins and detray.

  The corresponding algebra-plugins tag is ready, as is the PR in detray (883)

stephenswat (approved):
Happy with this!

stephenswat:
But please rename the PR to something more meaningful before we merge

core/include/traccc/definitions/primitives.hpp
```diff
@@ -32,6 +32,7 @@ using channel_id = unsigned int;
 // Default algebra type
 using default_algebra = ALGEBRA_PLUGIN<traccc::scalar>;
 
+using size_type = detray::dsize_type<default_algebra>;
```
krasznaa, line 35: I know why algebras need a size type. But I'm not super happy about using a global `traccc::size_type`. 🤔 We will still have different size types here and there. The `traccc::size_type` name, I fear, will be too misleading.

core/include/traccc/geometry/detector.hpp
```diff
@@ -7,6 +7,9 @@
 
 #pragma once
 
+// Project include(s).
+#include "traccc/definitions/primitives.hpp"
```
krasznaa, line 11: Was this needed here? I'm looking at the file, but can't see it. 😕

krasznaa (changes requested):
As far as I can see, `traccc::size_type` is only used in the I/O code. 🤔 I would much rather use `std::size_t` in those 2 places instead.

core/include/traccc/geometry/detector.hpp
niermann999: I did not understand that either, but without, the toy detector benchmarks were not finding the arithmetic operators for vectors

core/include/traccc/definitions/primitives.hpp
niermann999: Ok, I just tried replicate what I found in the IO code

io/src/csv/read_measurements.cpp
```diff
@@ -44,7 +44,7 @@ void read_measurements(measurement_collection_types::host& measurements,
 
         // Construct the measurement object.
         traccc::measurement meas;
-        std::array<typename transform3::size_type, 2u> indices{0u, 0u};
+        std::array<traccc::size_type, 2u> indices{0u, 0u};
```
krasznaa, line 47: I feel quite strongly about this... Let's not introduce a new global type, just to read and write measurements. We can just use a local type inside this code.

After all, we are trying to get rid of global floating point types to begin with. :thinking:

core/include/traccc/geometry/detector.hpp
krasznaa: Doesn't that mean then that the benchmark needs this include? :confused: The code in this file really doesn't seem to require those definitions. :thinking:

niermann999: That's what I thought as well, but the operators were not found in detray code that the benchmarks depended on. I tried a few things, including defining the operators again in the benchmark files, but in the end it depended on the include ordering of this file into the benchmark header (before or after the other traccc includes), unless I included ```primitives``` here. Having this depend on the include order seemed more worrying to me than adding this include.
Let me try again, after the dust has settled now in detray. I agree that something is/was definitely fishy with how we have been using these operators so far.

If we go to the templated detray code, should this file contain fully defined types or templated alias types? I.e.
```
/// Default detector (also used for ODD)
template<typename algebra_t>
using default_detector = detector<detray::default_metadata<algebra_t>>;
```
because for fully defined types, we will need this include anyway, right?

io/src/csv/read_measurements.cpp
niermann999: I don't have a strong opinion on this, I can get the size type locally from the global default algebra. That is essentially also what has been done in the code so far. The size type is still defined globally for the algebra plugin, just not as visible.

core/include/traccc/geometry/detector.hpp
niermann999: Seems to work now without! All the better

krasznaa (approved):
:+1:

beomki-yeo:
This PR seems to slow down the benchmark?

stephenswat:
That's curious. :thinking:

niermann999:
> This PR seems to slow down the benchmark?

This is strange indeed, since I was just now working on the detray benchmarks and I am not seeing the same problem...

PR #799, opened 2024-12-07
SYCL Tweaks, main branch (2024.12.07.)

krasznaa:
Continuing with trying  to make #773 work...

That PR by now compiles ~successfully for all platforms, but produces weird runtime issues on anything beside an NVIDIA backend. Since I wasn't sure how to best tackle that issue, decided to rather do some cleanup that's been long overdue. But as it happens, I may have found at least part of the problem.

One of the commits of the PR is just taken from #773 as is, simplifying the sanity checking code a little.

For the rest, I set out to move the non-public parts of the SYCL code under `device/sycl/src`, so that unsuspecting users would not be exposed to `<sycl/sycl.hpp>`. (At the same time I also made `traccc::sycl` publicly depend on `traccc::device_common`. Since it does.)

While doing that, at first I just wanted to tweak `traccc::sycl::thread_id1` and `traccc::sycl::barrier` a little, so that I'd like them more. But I actually found some implementation issues with both of them, at which point I decided to fully re-design them. :thinking:
  - They now both have a template parameter for the dimensionality of the grid/kernel that they should operate on. Set to `1` by default.
  - Moved both of them into the `traccc::sycl::details::` namespace, just because I decided I liked them better there. :stuck_out_tongue:
  - Modified the member functions of `traccc::sycl::details::thread_id` (the artist formerly known as `traccc::sycl::thread_id1`) to (hopefully) return correct values finally. As some of the member functions didn't so far. :frowning:
    * While doing this, I noticed that the `traccc::device::concepts::thread_id1` concept didn't require the member functions of these types to be `const`. Even though our code itself assumed it. So I modified the concept to require constness on the functions.
  - The issue with `traccc::sycl:details::barrier` (formerly `traccc::sycl::barrier`) was even weirder. It currently uses a function called `::sycl::nd_item<...>::barrier()`. But that function is not actually part of the SYCL2020 standard! (https://github.khronos.org/SYCL_Reference/iface/nd_item.html)
    * It's a function that is there in oneAPI, looking like: https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/nd_item.hpp#L201-L206
    * Notice that it has a default argument, very much suggesting that a global synchronization would be attempted by default.
    * To be SYCL2020 compatible, I rather switched to using [sycl::group_barrier](https://github.khronos.org/SYCL_Reference/iface/group-functions.html).
    * At the same time, I had to realize that the member functions of this class didn't need to be non-const. So I made them const, and updated the concept to require these functions to be const.

I had to tweak some common and test code to account for these changes, but I think this just made those a bit easier to use.

Unfortunately I'm not sure whether these changes will fix my observed issues in #773 or not. (Since these changes ended up diverging a bit too much from that PR.) But I do believe that they are good fixes in either case. And fingers crossed that they'll help with the track finding code as well...


device/cuda/src/utils/barrier.hpp
```diff
@@ -1,29 +1,28 @@
...
-    bool blockOr(bool predicate) { return __syncthreads_or(predicate); }
+    __device__ inline bool blockOr(bool predicate) const {
+        return __syncthreads_or(predicate);
+    }
 
-    TRACCC_DEVICE
-    int blockCount(bool predicate) { return __syncthreads_count(predicate); }
+    __device__ inline int blockCount(bool predicate) const {
+        return __syncthreads_count(predicate);
+    }
```
stephenswat, line 25: Hmm, why? The previous version felt more consistent with the rest of the code.

device/sycl/src/clusterization/clusterization_algorithm.sycl
```diff
@@ -153,15 +154,15 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
                         *mutex_ptr);
 
                     // Barrier used in the algorithm.
-                    traccc::sycl::barrier barry_r(item);
-                    const sycl::thread_id1 thread_id(item);
+                    const details::barrier barrier{item};
```
stephenswat, line 157: RIP Barry.

device/cuda/src/utils/barrier.hpp
krasznaa: Why should we use a macro in a file that can only be used in CUDA device code? Much like how I removed `TRACCC_DEVICE` from the SYCL code (as it served really no purpose there), I thought this would also be a good simplification.

The macros are really only needed in the shared code. CUDA specific code should be free to use `__device__`, `__global__`, etc.

device/common/include/traccc/device/concepts/barrier.hpp
```diff
@@ -18,7 +18,7 @@ namespace traccc::device::concepts {
  * @tparam T The barrier-like type.
  */
 template <typename T>
-concept barrier = requires(T& b) {
+concept barrier = requires(const T& b) {
```
stephenswat, line 21: I don't love this; requiring the concept to work for a non-const `T` is a stronger requirement than having it work with a const `T`. Let's not make our concepts weaker for no reason.

device/common/include/traccc/device/concepts/thread_id.hpp
```diff
@@ -19,7 +19,7 @@ namespace traccc::device::concepts {
  * @tparam T The thread identifier-like type.
  */
 template <typename T>
-concept thread_id1 = requires(T& i) {
+concept thread_id1 = requires(const T& i) {
```
stephenswat, line 22: Same as above.

stephenswat, line 22: If we're making the `thread_id` objects templated on the dimensionality, why not the `thread_id` concept too? :wink:

device/sycl/src/utils/barrier.hpp
```diff
@@ -8,40 +8,59 @@
...
+/// A barrier type for SYCL kernels.
+template <std::size_t DIMENSIONS = 1>
 struct barrier {
-    barrier(::sycl::nd_item<1> item) : m_item(item) {}
 
-    TRACCC_DEVICE
-    void blockBarrier() { m_item.barrier(); }
+    /// Verify that the barrier is used on a supported dimensionality.
+    static_assert(DIMENSIONS > 0 && DIMENSIONS < 4,
+                  "Only 1D, 2D, and 3D barriers are supported.");
```
stephenswat, line 24: `static_assert` is _so_ C++17; let's turn this into a nice `requires` clause.

device/sycl/src/utils/thread_id.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+namespace traccc::sycl::details {
+
+/// A SYCL thread identifier type
+template <std::size_t DIMENSIONS = 1>
+struct thread_id {
+
+    /// Verify that the thread identifier is used on a supported dimensionality.
+    static_assert(DIMENSIONS > 0 && DIMENSIONS < 4,
+                  "Only 1D, 2D, and 3D barriers are supported.");
```
stephenswat, line 25: Could be a `requires` clause!

device/sycl/src/utils/barrier.hpp
```diff
@@ -8,40 +8,59 @@
...
-    bool blockAnd(bool predicate) {
-        m_item.barrier();
+    /// Dimensions of the barrier
+    static constexpr std::size_t dimensions = DIMENSIONS;
+
+    /// Constructor
+    ///
+    /// @param item The @c ::sycl::nd_item<1> to use for the barrier.
+    ///
+    explicit barrier(const ::sycl::nd_item<dimensions>& item) : m_item(item) {}
```
stephenswat, line 33: Should we add some template deduction guides? That would be a lot more elegant than having the default template argument.

device/sycl/src/utils/thread_id.hpp
```diff
@@ -0,0 +1,67 @@
...
+// Project include(s).
+#include "traccc/device/concepts/thread_id.hpp"
+
+// SYCL include(s).
+#include <sycl/sycl.hpp>
+
+namespace traccc::sycl::details {
+
+/// A SYCL thread identifier type
+template <std::size_t DIMENSIONS = 1>
```
stephenswat, line 20: What's the proposed strategy for extending this to 2 and 3 dimensions? Wouldn't it be nicer to declare this as an empty type and then specialize it for $D \in \{1, 2, 3\}$?

device/sycl/src/utils/barrier.hpp
```diff
@@ -8,40 +8,59 @@
...
+
+    inline void blockBarrier() const {
+        ::sycl::group_barrier(m_item.get_group());
+    }
```
stephenswat, line 37: Wondering if we should accept `m_item.get_group()` as the constructor argument, then we don't need to template the barrier on the number of dimensions. :thinking:

device/sycl/src/clusterization/clusterization_algorithm.sycl
krasznaa: I did enjoy that name. But indeed, decided that we had enough fun with it. :thinking:

stephenswat:
> Modified the member functions of traccc::sycl::details::thread_id (the artist formerly known as traccc::sycl::thread_id1) to (hopefully) return correct values finally. As some of the member functions didn't so far. 😦 

Where? I cant say I see what you mean.

device/common/include/traccc/device/concepts/barrier.hpp
krasznaa: Okay, I don't feel strongly about this.

device/common/include/traccc/device/concepts/thread_id.hpp
krasznaa: I thought about that. But decided that I like the non-templated concept better. :thinking:

We use very specific function names in the concept. In a higher dimensionality concept we would probably replace `X` in a couple of the function names with `Y`.

I think I'd rather just add `thread_id2` as a concept once we need it. (I imagine it would require the `thread_id1` concept, and add some more required functions.) That would still allow us to be pretty free with how we provide the implementation classes.

device/sycl/src/utils/barrier.hpp
krasznaa: I'll have to look up how to do that. :stuck_out_tongue: But sure, that sounds like an idea.

device/sycl/src/utils/barrier.hpp
krasznaa: I thought about this. :thinking: But having never written code like that before, wasn't sure how to do it exactly.

You're right though. It would be very cool to construct the correct type using:

```c++
const ::sycl::nd_item<2> item = ...;
traccc::sycl::details::barrier b{item};
```

Any guidelines on how to do this? (To save me some google-ing.)

stephenswat: Should be as simple as:

```cpp
template<std::size_t N>
barrier(::sycl::nd_item<N>) -> barrier<N>;
```

And can be placed just behind the class.

device/sycl/src/utils/thread_id.hpp
krasznaa: Not sure. :thinking: But I wouldn't jump to template specialization with this. That just has so much code repetition built in...

device/sycl/src/utils/barrier.hpp
krasznaa: I thought about that for a while. The reason I didn't go that way is that all these functions receive templated group objects as arguments, by value. We could mimic that in this class, making it into a generic template having a `GROUP` member. :thinking:

Holding on to the "more concrete" item object just seemed a bit more robust against SYCL implementation changes. :thinking:

device/sycl/src/utils/thread_id.hpp
stephenswat: Thinking about it, we can simply declare them as:

```
getLocalThreadIdY() requires(DIMENSIONS >= 2) {
...
}
```

So I think that's probably a fine solution. :+1:

krasznaa:
> > Modified the member functions of traccc::sycl::details::thread_id (the artist formerly known as traccc::sycl::thread_id1) to (hopefully) return correct values finally. As some of the member functions didn't so far. 😦
> 
> Where? I cant say I see what you mean.

Let me give one example, you can then look at the rest. See:

https://github.com/acts-project/traccc/blob/main/device/sycl/include/traccc/sycl/utils/thread_id.hpp#L22

It doesn't return the "X component" of the global thread ID, but rather the "linearized global thread ID".

Since we don't use multi-dimensional kernels actively, I don't think it would've hurt us yet. But these should still do what they advertise.

stephenswat:
> > > Modified the member functions of traccc::sycl::details::thread_id (the artist formerly known as traccc::sycl::thread_id1) to (hopefully) return correct values finally. As some of the member functions didn't so far. 😦
> > 
> > 
> > Where? I cant say I see what you mean.
> 
> Let me give one example, you can then look at the rest. See:
> 
> https://github.com/acts-project/traccc/blob/main/device/sycl/include/traccc/sycl/utils/thread_id.hpp#L22
> 
> It doesn't return the "X component" of the global thread ID, but rather the "linearized global thread ID".
> 
> Since we don't use multi-dimensional kernels actively, I don't think it would've hurt us yet. But these should still do what they advertise.

I see your point but the implementation is fully correct in the context of a 1D thread identifier lookup. Just trying to understand if this is related to your SYCL woes, but I am guessing not then.

krasznaa:
> I see your point but the implementation is fully correct in the context of a 1D thread identifier lookup. Just trying to understand if this is related to your SYCL woes, but I am guessing not then.

I suspect / hope that the changes in the barrier class will have an effect on track finding. The ID class indeed probably won't change the behaviour.

device/sycl/src/utils/barrier.hpp
stephenswat: Yeah alright, makes sense; it is rather unfortunate to have to template this class but sure. :+1:

stephenswat:
Let's hope :crossed_fingers:

device/sycl/src/utils/barrier.hpp
```diff
@@ -8,40 +8,64 @@
...
     private:
-    ::sycl::nd_item<1> m_item;
+    /// Item object coming from the SYCL kernel
+    const ::sycl::nd_item<dimensions>& m_item;
 };
 
-}  // namespace traccc::sycl
+/// Template deduction guide for the barrier type.
+template <int N>
+barrier(::sycl::nd_item<N>) -> barrier<N>;
```
krasznaa, line 63: Template deduction guides are much more fragile than I thought. :confused: When using `std::size_t` here, Clang told me that the expression "it's trying to figure out" uses `int`, so it can't use the guide with `std::size_t`.

I then tried to use the `std::integer` concept as part of the deduction guide, but that also failed. Since in that case it's the integer type that becomes the template parameter, not the integer value.

So for now I just used `int`. We may need to add other "overloads" later on as well, let's see.

stephenswat: Actually this makes total sense: if you look at the template definition of `nd_item` it is indeed templated on `https://github.khronos.org/SYCL_Reference/iface/nd_item.html`.

stephenswat (approved):

krasznaa:
Unfortunately I already tried this morning, that these changes by themselves are not doing much to what I'm observing in #773 with the Intel and AMD backends. 😦

Still, these were good cleanups to have.

PR #800, opened 2024-12-09
chore: Remove obsolete ACTS include for JSON

paulgessinger:
This will be dropped in ACTS v39


stephenswat:
Is this still relevant? Seems like this compiles with ACTS 46.0.0.

PR #802, opened 2024-12-11
Spacepoint formation algorithm change

arkslobodkins:
Changed block size to 256, launching with 1024 threads per block would cause 'too many resources requested for launch' on a desktop GPU.


krasznaa (approved):
👍

stephenswat:
Hmm what GPU is this? The kernel is set to be launchable for 1024 threads;
perhaps the CUDA architecture you build with (`CMAKE_CUDA_ARCHITECTURES`)
does not match the device you are running on?

stephenswat:
No, please do not merge this yet

arkslobodkins:
I believe it's Nvidia RTX 500 Ada Generation Laptop GPU. CMAKE_CUDA_ARCHITECTURES was set to native in my CMakeLists so that flag should have propagated to traccc. Setting threads to 256 solves the issue...

PR #805, opened 2024-12-19
ref: Update to detray version 0.87.0

niermann999:
Update to detray version 0.87.0 and algebra plugins version 0.26.2, which mainly contain the full algebra templating of detray

### DIFF
--- a/core/include/traccc/finding/combinatorial_kalman_filter_algorithm.hpp
+++ b/core/include/traccc/finding/combinatorial_kalman_filter_algorithm.hpp
@@ -28,12 +28,14 @@ namespace traccc::host {
 class combinatorial_kalman_filter_algorithm
     : public algorithm<track_candidate_container_types::host(
           const default_detector::host&,
-          const detray::bfield::const_field_t::view_t&,
+          const detray::bfield::const_field_t<
+              default_detector::host::scalar_type>::view_t&,
           const measurement_collection_types::const_view&,
           const bound_track_parameters_collection_types::const_view&)>,
       public algorithm<track_candidate_container_types::host(
           const telescope_detector::host&,
-          const detray::bfield::const_field_t::view_t&,
+          const detray::bfield::const_field_t<
+              telescope_detector::host::scalar_type>::view_t&,
           const measurement_collection_types::const_view&,
           const bound_track_parameters_collection_types::const_view&)> {
 
@@ -58,7 +60,8 @@ class combinatorial_kalman_filter_algorithm
     ///
     output_type operator()(
         const default_detector::host& det,
-        const detray::bfield::const_field_t::view_t& field,
+        const detray::bfield::const_field_t<
+            default_detector::host::scalar_type>::view_t& field,
         const measurement_collection_types::const_view& measurements,
         const bound_track_parameters_collection_types::const_view& seeds)
         const override;
@@ -75,7 +78,8 @@ class combinatorial_kalman_filter_algorithm
     ///
     output_type operator()(
         const telescope_detector::host& det,
-        const detray::bfield::const_field_t::view_t& field,
+        const detray::bfield::const_field_t<
+            telescope_detector::host::scalar_type>::view_t& field,
         const measurement_collection_types::const_view& measurements,
         const bound_track_parameters_collection_types::const_view& seeds)
         const override;
--- a/core/include/traccc/fitting/kalman_fitting_algorithm.hpp
+++ b/core/include/traccc/fitting/kalman_fitting_algorithm.hpp
@@ -29,11 +29,13 @@ namespace traccc::host {
 class kalman_fitting_algorithm
     : public algorithm<track_state_container_types::host(
           const default_detector::host&,
-          const detray::bfield::const_field_t::view_t&,
+          const detray::bfield::const_field_t<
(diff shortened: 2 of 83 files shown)


core/include/traccc/finding/combinatorial_kalman_filter_algorithm.hpp
```diff
@@ -28,12 +28,14 @@ namespace traccc::host {
 class combinatorial_kalman_filter_algorithm
     : public algorithm<track_candidate_container_types::host(
           const default_detector::host&,
-          const detray::bfield::const_field_t::view_t&,
+          const detray::bfield::const_field_t<
+              telescope_detector::host::scalar_type>::view_t&,
```
krasznaa, line 32:
```suggestion
              default_detector::host::scalar_type>::view_t&,
```

Since it's "just" the scalar type that's being given as a template, I wonder if there may be a better way of providing it to the b-field type. 🤔

core/include/traccc/finding/combinatorial_kalman_filter_algorithm.hpp
```diff
@@ -58,7 +60,8 @@ class combinatorial_kalman_filter_algorithm
     ///
     output_type operator()(
         const default_detector::host& det,
-        const detray::bfield::const_field_t::view_t& field,
+        const detray::bfield::const_field_t<
+            telescope_detector::host::scalar_type>::view_t& field,
```
krasznaa, line 64:
```suggestion
            default_detector::host::scalar_type>::view_t& field,
```

I won't mark the rest. 😉

core/include/traccc/fitting/kalman_fitting_algorithm.hpp
```diff
@@ -29,11 +29,13 @@ namespace traccc::host {
...
           const default_detector::host&,
-          const detray::bfield::const_field_t::view_t&,
+          const detray::bfield::const_field_t<
+              default_detector::host::scalar_type>::view_t&,
           const track_candidate_container_types::const_view&)>,
       public algorithm<track_state_container_types::host(
           const telescope_detector::host&,
-          const detray::bfield::const_field_t::view_t&,
+          const detray::bfield::const_field_t<
+              default_detector::host::scalar_type>::view_t&,
```
krasznaa, line 38:
```suggestion
              telescope_detector::host::scalar_type>::view_t&,
```

I marked this one after all. 😛 Some of the mistakes were made in this direction as it seems.

plugins/algebra/array/include/traccc/plugins/algebra/array_definitions.hpp
```diff
@@ -23,6 +23,8 @@
 #include <map>
 #include <tuple>
 
+#define ALGEBRA_PLUGIN detray::array
```
krasznaa, line 26: Curious. Where is `ALGEBRA_PLUGIN` used? You added these macros, but I don't see any new code that would use them.

I anyway wonder, even if we need these macros, whether it would not be better to define them in the build system. 🤔 But first let's make sure that we actually need them.

krasznaa (commented):
I'm happy with the overall design of the changes. I think the code looks good like this.

Just some technical comments. 😉

plugins/algebra/array/include/traccc/plugins/algebra/array_definitions.hpp
niermann999: In detray I removed it, but traccc still uses a ```default_algebra``` in many cases, which is defined here
https://github.com/acts-project/traccc/blob/862076927a59bdfc969cea0ebd10cfea00ee7391/core/include/traccc/definitions/primitives.hpp#L33

core/include/traccc/finding/combinatorial_kalman_filter_algorithm.hpp
niermann999: I am not sure. What is the plan for the algebra plugins integration in traccc long term? Will we stay with the centrally defined algebra types? Then we could just do the same with the B-field types.  We will eventually need more types than just the constant field here, right?

krasznaa: Since Detray already started to define some b-field typedefs, we could just leave them there. But yes, we will definitely need at least a "parametrized" one (which loads a map from a file and interpolates between the values) and I guess some "functional" one, which would allow using some user function for calculating the magnetic field at position X-Y-Z.

@stephenswat will have to advise about the exact type names on those. (And whether the latter is even possible/practical.)

plugins/algebra/array/include/traccc/plugins/algebra/array_definitions.hpp
krasznaa: Hmm... Let's keep the macro in this PR then. But I'll ponder a bit about what best to do in a follow-up. As even with global type definitions, we should still not use macros like this here. 🤔

core/include/traccc/finding/combinatorial_kalman_filter_algorithm.hpp
niermann999: Hope, I got them all now

stephenswat (approved):
:+1:

niermann999:
I also removed some `detray::get_data` calls now that are not needed

PR #810, opened 2025-01-09
Thread ID Cleanup, main branch (2025.01.09.)

krasznaa:
Following #808, I thought it would be time to clean up the rest of the device functions as well with how they receive thread identifiers.
  - Made all functions that need a "global index" use `traccc::device::global_index_t`;
    * This allowed removing a number of `static_cast`-s from these functions;
  - All functions that receive a `thread_id` and/or `barrier` object, now do so using constant references;
    * All such types provide only `const` functions. So I didn't see much reason for using non-const references here. 🤔

After this, I went and harmonized the CUDA, SYCL and ALPAKA codes a little as well.
  - Moved all `thread_id` types to be private headers in their libraries;
  - Added automatic, compile-time checks that they would all fulfill the `traccc::device::concepts::thread_id1` concept;
  - Introduced `traccc::cuda::details::global_index1()` and `traccc::sycl::details::global_index(...)` as helper functions for generating `traccc::device::global_index_t` values;

While doing all of this, I tried to fix up the includes in all the touched files a bit. Since many of them were doing very questionable things. (Including way more files than necessary, hiding missing includes in some of the common device headers.)

I'm pretty happy with these updates myself, but am interested in your opinions.


device/common/include/traccc/fitting/device/impl/fit.ipp
```diff
@@ -34,8 +34,7 @@ TRACCC_HOST_DEVICE inline void fit(
         return;
     }
 
-    const unsigned int param_id =
-        param_ids.at(static_cast<unsigned int>(globalIndex));
+    const unsigned int param_id = param_ids.at(globalIndex);
```
stephenswat, line 37: This should be `global_index_t`.

device/cuda/src/sanity/contiguous_on.cuh
```diff
@@ -60,8 +61,8 @@ template <std::equality_comparable T>
 __global__ void is_contiguous_on_all_unique(
     vecmem::data::vector_view<T> in_view, bool* out) {
 
-    int tid_x = threadIdx.x + blockIdx.x * blockDim.x;
-    int tid_y = threadIdx.y + blockIdx.y * blockDim.y;
+    const unsigned int tid_x = threadIdx.x + blockIdx.x * blockDim.x;
+    const unsigned int tid_y = threadIdx.y + blockIdx.y * blockDim.y;
```
stephenswat, line 65: These should be `global_index_t`.

device/cuda/src/utils/global_index.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+#pragma once
+
+// Project include(s).
+#include "traccc/device/global_index.hpp"
+
+namespace traccc::cuda::details {
+
+/// Function creating a global index in a 1D CUDA kernel
+__device__ inline device::global_index_t global_index1() {
```
stephenswat, line 16: No 2 and 3 dimensional versions like for SYCL?

device/cuda/src/utils/global_index.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+    return blockIdx.x * blockDim.x + threadIdx.x;
```
stephenswat, line 18: Also, I don't think we need this; the thread identifier classes already serve this purpose. :confused:

device/cuda/src/utils/thread_id.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+    inline unsigned int __device__ getGlobalThreadIdX() const {
+        return threadIdx.x + blockIdx.x * blockDim.x;
+    }
+
+    inline unsigned int __device__ getBlockIdX() const { return blockIdx.x; }
+
+    inline unsigned int __device__ getBlockDimX() const { return blockDim.x; }
+
+    inline unsigned int __device__ getGridDimX() const { return gridDim.x; }
```
stephenswat, line 40: Recall we discussed before the break that making these unsigned integers puts us at risk of accidental integer truncation.

IMO these should all return `std::size_t` and then we can cast them to unsigned ints at the use sites.

device/sycl/src/utils/global_index.hpp
```diff
@@ -0,0 +1,36 @@
...
+inline device::global_index_t global_index(const ::sycl::nd_item<2>& item) {
+
+    return static_cast<device::global_index_t>(item.get_global_linear_id());
+}
+
+/// Function creating a global index in a 3D SYCL kernel
+inline device::global_index_t global_index(const ::sycl::nd_item<3>& item) {
+
+    return static_cast<device::global_index_t>(item.get_global_linear_id());
+}
```
stephenswat, line 34: This seems very silly to me; we're replacing a SYCL native call to `get_global_linear_id` by a custom function which simply wraps it for no good reason _and_ also unsafely narrows the value to a 32-bit integer... This is just unnecessary complexity plus unnecessary code fragility; suggest we keep it as it was.

stephenswat: Or, indeed, we use the thread identifier types, which are intended for exactly this purpose.

device/sycl/src/utils/thread_id.hpp
```diff
@@ -33,23 +33,33 @@ requires(DIMENSIONS >= 1 && DIMENSIONS <= 3) struct thread_id {
     /// @name Function(s) implementing @c traccc::device::concepts::thread_id1
     /// @{
 
-    inline auto getLocalThreadId() const {
-        return m_item.get_local_linear_id();
+    inline unsigned int getLocalThreadId() const {
```
stephenswat, line 36: FYI you forgot this change in the Alpaka code.

stephenswat (changes requested):
I generally dislike this idea that because vecmem only supports 32-bit integers we should be downcasting all out accesses in a 64-bit memory space to 32-bit at the generation site of those indices, rather than generating them at the appropriate size and only downcasting them necessary.

device/common/include/traccc/fitting/device/impl/fit.ipp
krasznaa: Should it? We take an element out of the `param_ids` vector. It has `unsigned int` values, true, but these are still different things from `global_index_t`. 🤔

device/cuda/src/sanity/contiguous_on.cuh
krasznaa: I was thinking about it... 🤔 I'm happy to turn these into `global_index_t` if you think that's the correct way about this, but in the end these are not really "global" indices here.

Still, I have no strong opinion about this. 🤔

device/common/include/traccc/fitting/device/impl/fit.ipp
stephenswat: I guess it depends a bit on the semantics of the `global_index_t` type; if the semantics are exactly that they represent thread identifiers and nothing else then yes, although then I feel it might also be useful to have a separate index type for array indices. :thinking:

device/cuda/src/sanity/contiguous_on.cuh
stephenswat: In what way are they not global indices? Or do you mean that the global index is strictly a _linear_ index in the global thread space? And because these are 2D they do not meet that description?

device/cuda/src/utils/global_index.hpp
krasznaa: 🤔 It's just that those need some actual new code for them. As we don't have examples of doing this in any other place.

So it was mostly laziness on my part, not wanting to write new code plus the accompanying unit tests. 😛

stephenswat: Sure, that's a valid reason in my mind. :smile:

device/cuda/src/utils/thread_id.hpp
krasznaa: Also recall what I wrote to that. 🤔

We'd have to `static_cast` every single `threadIdx.x`, etc. element to `std::size_t` first so that the multiplication / addition would be done as `std::size_t` as well. Without that, all the calculations happen as `unsigned int` anyway. So casting to `std::size_t` at the very last step doesn't save us from any overflows. But it does introduce a tremendous number of `static_cast`-s in the client code.

Note that right now we use `auto` here. Which evaluates to `unsigned int`, since all the elements in the calculations are `unsigned int`-s. The only reason I replaced `auto` with `unsigned int` is that `nvcc` couldn't evaluate the `static_assert` with that setup. Otherwise I would've loved to keep the `auto`-s to skip this discussion. 🤔

stephenswat: nvcc doesn't let you evaluate the concept on the type because of the auto type? That's interesting (and smells a bit like a bug in nvcc). :thinking:

device/sycl/src/utils/thread_id.hpp
krasznaa: I didn't. 🤔 I was pondering about removing `auto` from there as well.

But I think you're right, we should do an explicit cast in that class as well. 🤔

device/cuda/src/utils/thread_id.hpp
stephenswat: But yes in general, why don't we make these explicitly return `std::size_t`? I feel like that would be a nicer solution (the non-composite indices which are guaranteed to be less than UINT_MAX can stay unsigned ints).

krasznaa: Because then we'd need to `static_cast` them back to `unsigned int` in O(50) places. As we use these as vector indices all around the place. (As we should.)

So I would like to make the casts in the `thread_id` and `global_index()` functions "once", instead of N times all over the place.

device/common/include/traccc/fitting/device/impl/fit.ipp
krasznaa: We could start using `global_index_t` in a number of other places as well. But as long as the `param_ids` variable uses `unsigned int` directly, we should use `unsigned int` here as well. (We'd need to switch to using `global_index_t` for the `param_ids` variables everywhere before I'd want to change the type on this line.)

device/cuda/src/utils/thread_id.hpp
krasznaa: For info, this is what I get when replacing all the `unsigned int`-s with `auto` in this code:

```
[ 64%] Building CUDA object device/cuda/CMakeFiles/traccc_cuda.dir/src/clusterization/clusterization_algorithm.cu.o
/data/ssd-1tb/projects/traccc/traccc/device/cuda/src/clusterization/../utils/thread_id.hpp:46:33: error: static assertion failed
   46 | static_assert(traccc::device::concepts::thread_id1<thread_id1>);
      |               ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
/data/ssd-1tb/projects/traccc/traccc/device/cuda/src/clusterization/../utils/thread_id.hpp:46:33: note: constraints not satisfied
/data/ssd-1tb/projects/traccc/traccc/device/common/include/traccc/device/concepts/thread_id.hpp:21:28:   required by the constraints of ‘template<class T> concept traccc::device::concepts::thread_id1’
/data/ssd-1tb/projects/traccc/traccc/device/common/include/traccc/device/concepts/thread_id.hpp:21:41:   in requirements with ‘T& i’ [with T = traccc::cuda::details::thread_id1]
/data/ssd-1tb/projects/traccc/traccc/device/common/include/traccc/device/concepts/thread_id.hpp:22:26: note: ‘i.getLocalThreadId()’ does not satisfy return-type-requirement
   22 | concept thread_id1 = requires(T& i) {
      |        ~~~~~~~~~~~~~~~~~~^~
/data/ssd-1tb/projects/traccc/traccc/device/common/include/traccc/device/concepts/thread_id.hpp:23:27: note: ‘i.getLocalThreadIdX()’ does not satisfy return-type-requirement
   23 |     /*
      |                           ^ 
/data/ssd-1tb/projects/traccc/traccc/device/common/include/traccc/device/concepts/thread_id.hpp:24:27: note: ‘i.getGlobalThreadId()’ does not satisfy return-type-requirement
   24 |      * This function should return the local thread identifier in a *flat* way,
      |        ~~~~~~~~~~~~~~~~~~~^~
/data/ssd-1tb/projects/traccc/traccc/device/common/include/traccc/device/concepts/thread_id.hpp:25:28: note: ‘i.getGlobalThreadIdX()’ does not satisfy return-type-requirement
   25 |      * e.g. compressing two or three dimensional blocks into one dimension.
      |        ~~~~~~~~~~~~~~~~~~~~^~
/data/ssd-1tb/projects/traccc/traccc/device/common/include/traccc/device/concepts/thread_id.hpp:26:21: note: ‘i.getBlockIdX()’ does not satisfy return-type-requirement
   26 |      */
      |                     ^ 
/data/ssd-1tb/projects/traccc/traccc/device/common/include/traccc/device/concepts/thread_id.hpp:27:21: note: ‘i.getBlockIdX()’ does not satisfy return-type-requirement
   27 |     { i.getLocalThreadId() }
      |        ~~~~~~~~~~~~~^~
/data/ssd-1tb/projects/traccc/traccc/device/common/include/traccc/device/concepts/thread_id.hpp:28:21: note: ‘i.getBlockIdX()’ does not satisfy return-type-requirement
   28 |     ->std::integral;
      |        ~~~~~~~~~~~~~^ 
cc1plus: note: set ‘-fconcepts-diagnostics-depth=’ to at least 2 for more detail
gmake[2]: *** [device/cuda/CMakeFiles/traccc_cuda.dir/build.make:222: device/cuda/CMakeFiles/traccc_cuda.dir/src/clusterization/clusterization_algorithm.cu.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:3788: device/cuda/CMakeFiles/traccc_cuda.dir/all] Error 2
gmake: *** [Makefile:166: all] Error 2
```

This is with CUDA 12.6.3. I also think that it's a shortcoming in CUDA...

device/sycl/src/utils/global_index.hpp
krasznaa: The main reason I introduced these helper functions for SYCL is to do the static casts in as few places as possible.

Let's stop worrying about not fitting into 32 bits with our thread count! I've never seen any NVIDIA engineer advise against using `unsigned int` with this. (I believe the hardware has much lower limits than 2^32 threads...)

SYCL rightfully used `std::size_t` in its functions. That's what I would've done as well. But as long as we need to use `unsigned int` for indices "for atomic reasons", we're stuck with that type... 😦

stephenswat: FYI the maximum grid size in CUDA is $2^{57}$ threads, so there certainly is reason to be worried.

Also, thinking of it, the maximum number of threads in the $x$-axis is $2^{41}$ so that also presents some issues with the unsigned integer type. It's $2^{26}$ threads in the $y$ and $z$ axes, so no problems there.

But okay, fine, we can do it this way.

stephenswat (approved):

PR #811, opened 2025-01-09
SYCL Track Finding, main branch (2025.01.09.)

krasznaa:
With everything finally in place, this is a (hopefully) reviewable replacement of #773.

Unfortunately the code continues to not work on Intel GPUs, but for now I "solved" this by turning off the unit tests for the Intel platform. :frowning:

At least on an NVIDIA platform, the tests work as expected. :partying_face:

```
[bash][Legolas]:out > ./build/sycl-fp32/bin/traccc_test_sycl --gtest_filter="*Ckf*"
Running main() from /data/ssd-1tb/projects/traccc/traccc/out/build/sycl-fp32/_deps/googletest-src/googletest/src/gtest_main.cc
Note: Google Test filter = *Ckf*
[==========] Running 5 tests from 2 test suites.
[----------] 2 tests from SYCLCkfCombinatoricsTelescopeValidation/CkfCombinatoricsTelescopeTests
[ RUN      ] SYCLCkfCombinatoricsTelescopeValidation/CkfCombinatoricsTelescopeTests.Run/0
Created test queue on device: NVIDIA GeForce RTX 3080

Detector check: OK
[       OK ] SYCLCkfCombinatoricsTelescopeValidation/CkfCombinatoricsTelescopeTests.Run/0 (696 ms)
[ RUN      ] SYCLCkfCombinatoricsTelescopeValidation/CkfCombinatoricsTelescopeTests.Run/1
Created test queue on device: NVIDIA GeForce RTX 3080

Detector check: OK
[       OK ] SYCLCkfCombinatoricsTelescopeValidation/CkfCombinatoricsTelescopeTests.Run/1 (587 ms)
[----------] 2 tests from SYCLCkfCombinatoricsTelescopeValidation/CkfCombinatoricsTelescopeTests (1284 ms total)

[----------] 3 tests from SYCLCkfToyDetectorValidation/CkfToyDetectorTests
[ RUN      ] SYCLCkfToyDetectorValidation/CkfToyDetectorTests.Run/0
Created test queue on device: NVIDIA GeForce RTX 3080

Detector check: OK
[       OK ] SYCLCkfToyDetectorValidation/CkfToyDetectorTests.Run/0 (278 ms)
[ RUN      ] SYCLCkfToyDetectorValidation/CkfToyDetectorTests.Run/1
Created test queue on device: NVIDIA GeForce RTX 3080

Detector check: OK
[       OK ] SYCLCkfToyDetectorValidation/CkfToyDetectorTests.Run/1 (5096 ms)
[ RUN      ] SYCLCkfToyDetectorValidation/CkfToyDetectorTests.Run/2
Created test queue on device: NVIDIA GeForce RTX 3080

Detector check: OK
[       OK ] SYCLCkfToyDetectorValidation/CkfToyDetectorTests.Run/2 (5084 ms)
[----------] 3 tests from SYCLCkfToyDetectorValidation/CkfToyDetectorTests (10459 ms total)

[==========] 5 tests from 2 test suites ran. (11743 ms total)
[  PASSED  ] 5 tests.
[bash][Legolas]:out >
```


device/sycl/src/finding/find_tracks.hpp
```diff
@@ -0,0 +1,502 @@
...
+            n_in_params * config.max_num_branches_per_surface;
+
+        bound_track_parameters_collection_types::buffer updated_params_buffer(
+            n_in_params * config.max_num_branches_per_surface, mr.main);
+        copy.setup(updated_params_buffer)->wait();
+
+        vecmem::data::vector_buffer<unsigned int> updated_liveness_buffer(
+            n_in_params * config.max_num_branches_per_surface, mr.main);
+        copy.setup(updated_liveness_buffer)->wait();
+        copy.memset(updated_liveness_buffer, 0)->wait();
```
stephenswat, line 224: Why is this necessary?

device/sycl/src/finding/combinatorial_kalman_filter_algorithm_constant_field_default_detector.sycl
```diff
@@ -0,0 +1,60 @@
...
+
+struct kernels {
+    using make_barcode_sequence_kernel_type = make_barcode_sequence;
+    using apply_interaction_kernel_type = apply_interaction;
+    using find_tracks_kernel_type = find_tracks;
+    using fill_sort_keys_kernel_type = fill_sort_keys;
+    using propagate_to_next_surface_kernel_type = propagate_to_next_surface;
+    using build_tracks_kernel_type = build_tracks;
+    using prune_tracks_kernel_type = prune_tracks;
+};  // namespace kernels
```
stephenswat, line 38: This is not the right way to approach this problem. Instead, we should have a set of structures templated - defined not for any specific detector - on the detector type, the navigator type, the stepper type, etc. and specialize those at the kernel call sites.

device/sycl/src/finding/combinatorial_kalman_filter_algorithm_constant_field_telescope_detector.sycl
```diff
@@ -0,0 +1,60 @@
...
+struct prune_tracks;
+
+struct kernels {
+    using make_barcode_sequence_kernel_type = make_barcode_sequence;
+    using apply_interaction_kernel_type = apply_interaction;
+    using find_tracks_kernel_type = find_tracks;
+    using fill_sort_keys_kernel_type = fill_sort_keys;
+    using propagate_to_next_surface_kernel_type = propagate_to_next_surface;
+    using build_tracks_kernel_type = build_tracks;
+    using prune_tracks_kernel_type = prune_tracks;
```
stephenswat, line 37: Same as above; this approach does not scale well.

device/sycl/src/finding/find_tracks.hpp
```diff
@@ -0,0 +1,502 @@
...
+
+    // Prepare input parameters with seeds
+    bound_track_parameters_collection_types::buffer in_params_buffer(n_seeds,
+                                                                     mr.main);
+    copy.setup(in_params_buffer)->wait();
+    copy(seeds, in_params_buffer, vecmem::copy::type::device_to_device)->wait();
+    vecmem::data::vector_buffer<unsigned int> param_liveness_buffer(n_seeds,
+                                                                    mr.main);
+    copy.setup(param_liveness_buffer)->wait();
+    copy.memset(param_liveness_buffer, 1)->wait();
```
stephenswat, line 158: Make of these operations can be performed concurrently is we launch them asynchronously; I'd suggest we start using the event API that SYCL provides (as in e.g. device/sycl/src/sanity/contiguous_on.hpp) as the easy of its task graph building APIs is one of the biggest strengths of SYCL.

stephenswat (changes requested):
Looks good for the most part, although we need to rework the way we handle kernel identifiers.

device/sycl/src/finding/find_tracks.hpp
krasznaa: As step 1 I was just not caring about efficiency. Only about correctness.

Indeed, there are 2 "more correct" ways of doing this:
  - If we require nothing about the queue that is given to the algorithm, then we need to start holding on to the "event objects" returned by these calls, and only wait for them in the appropriate places.
    * Though this will not give too much of an improvement unfortunately.
  - We can require `::sycl::queue` objects to always be "in order". (https://github.khronos.org/SYCL_Reference/iface/queue.html#is-in-order) Then we can implement this function practically in the same way between CUDA and SYCL.

It's the latter that I'll want to go for. But decided not to make this "first" PR even more complicated with that. :thinking:

stephenswat: I guess this is a reply to https://github.com/acts-project/traccc/pull/811#discussion_r1908679856, right? :thinking:

device/sycl/src/finding/combinatorial_kalman_filter_algorithm_constant_field_default_detector.sycl
krasznaa: Not sure that you understand what these are for. :thinking: These types are here to allow SYCL to assign user-readable identifiers/names to the kernels that we launch.

Now... The types ("names") given to `::sycl::handler::parallel_for(...)` have to do a couple of things:
  - They can't be templated;
  - They need to be PoDs. At least they must be trivially constructible if I remember correctly.

So... None of the existing types that we use, qualify. :frowning: That's the first thing that I tried. If we want to see understandable names for the kernels in our profiles, this was the only good way I found of doing it. By defining a bunch of simple structs, all with unique names.

krasznaa: :confused: I know that I ran into some sort of issue with the types that we template track finding on. It might be the trivial constructability then. But I was not able to hide this from the API of the `find_tracks` function. :frowning:

It's also an option that we say that we're okay with not-super-friendly kernel names in VTune / NSys / etc. I was seriously considering that as well. Because with that none of this would be needed.

device/sycl/src/finding/find_tracks.hpp
krasznaa: As I wrote, I'm very open to exploring all of these. Not in this PR though. I decidedly made the choice of just waiting for everything in this first version.

device/sycl/src/finding/combinatorial_kalman_filter_algorithm_constant_field_default_detector.sycl
stephenswat: Yeah I think that's also an option. It's just a lot of boilerplate to have to define aliases for all of these types for each of the different detector types that we want to support, I think it would be better to avoid that. Admittedly I don't know if you are allowed to use the same type for different kernels, I assume that would cause some kind of name collision.

We can alternatively declare the structs for the `find_tracks` and other kernels independently of the detector type, and then specialize them with a label struct signifying the detector type (`defdet`, etc.).

device/sycl/src/finding/find_tracks.hpp
stephenswat: No but wait I am talking about the memset operation itself. Why is that necessary? I don't remember that being in the CUDA code.

krasznaa: My understanding is that allocated memory is not guaranteed to be null-initialized with SYCL. I'm happy to be proven wrong though.

device/sycl/src/finding/combinatorial_kalman_filter_algorithm_constant_field_default_detector.sycl
krasznaa: :thinking: Let me give that a try. Indeed, as long as the sort of templating that's happening in `contiguous_on.hpp` is accepted, there would be a better way of generating unique names here...

device/sycl/src/finding/find_tracks.hpp
stephenswat: It's not a point about automatic zeroing. Rather, if I recall correctly (which I may not), the `find_tracks` kernel guarantees that the updated liveness is written to a meaningful value before the array is read for all indices below `n_candidates`: https://github.com/acts-project/traccc/blob/main/device/common/include/traccc/finding/device/impl/find_tracks.ipp; if that's true then the memset is unnecessary.

Have you found that the code doesn't run without this memset?

krasznaa: I just now looked at it. This buffer is being set in the `find_tracks` kernel, with the common function only ever setting elements of the buffer to `1`.
  - https://github.com/acts-project/traccc/blob/main/device/common/include/traccc/finding/device/impl/find_tracks.ipp#L230
  - https://github.com/acts-project/traccc/blob/main/device/common/include/traccc/finding/device/impl/find_tracks.ipp#L286

So we'll either have to teach that kernel to set `0` values as well, or teach the CUDA code to do this memset.

krasznaa: :confused: What's the point of the buffer then? If the first N elements of the track container are guaranteed to be "live", then what do we use these liveness buffers for at all? I assumed that tracks would be allowed to be "sparsely alive". :confused:

But sure, if we can kill these buffers, I'm all for it. :thinking:

device/sycl/src/finding/combinatorial_kalman_filter_algorithm_constant_field_default_detector.sycl
krasznaa: It was indeed possible to make this look a whole lot nicer.

stephenswat: Fantastic! Thanks! :smile:

device/sycl/src/finding/find_tracks.hpp
krasznaa: The unit tests succeed without it as well, so let's remove it for now.

But we should have a discussion about the logic-flow here soon. Apparently it could be simplified a bit.

beomki-yeo:
The PR looks OK from my side... :+1:

stephenswat (approved):
LGTM. :+1:

PR #821, opened 2025-01-23
Fix several SYCL, OneDPL, and Kokkos build issues

stephenswat:
It turns out that the SYCL build system is in a significantly worse state than I though it was, and needs some urgent fixing. This commit fixes four primary issues:

1. OneDPL was previously being required as "DPL" but the actual package name is "oneDPL" so it was impossible to find.
2. OneDPL and Kokkos both set a bunch of unwanted interface compiler flags which we need to manually remove.
3. OneDPL cannot be configured with the traccc standard of having a separate C++ compiler and a SYCL compiler, so it needs to be tricked into thinking the SYCL compiler works.
4. OneDPL contains a boatload of compiler warnings which I had to silence with diagnostics pragmas.


beomki-yeo (approved):
Looks good to me

PR #822, opened 2025-01-23
Add structured configuration printing

stephenswat:
The current configuration printing code is a hodgepodge of unstructured printing which facilitates reading by neither humans nor machines. This commit replaces the existing output system by a much more structured, JSON-like format which pretty prints the output and will, in the future, allow us to very easily output configurations in machine-readable formats.


beomki-yeo (approved):
Looks good to me

PR #824, opened 2025-01-24
Add logging infrastructure to all algorithms

stephenswat:
This commit adds ACTS-style loggers to all existing traccc algorithms, and requires that all future algorithms have the same. It also carefully threads logger objects through the computation chain. Currently, little is added in the way of useful logging outputs, as that can happen in a future commit: this is just infrastructure.


stephenswat:
Apologies for the high code volume, the vast majority is boilerplate.

core/CMakeLists.txt
```diff
@@ -116,7 +116,7 @@ traccc_add_library( traccc_core core TYPE SHARED
   "src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cpp" )
 target_link_libraries( traccc_core
   PUBLIC Eigen3::Eigen vecmem::core detray::core detray::detectors
-         traccc::algebra )
+         traccc::algebra ActsCore )
```
beomki-yeo, line 119: Can't we please not include the ActsCore in the `traccc::core`?
It will require Boost in compiling traccc core which should be pre-installed.

beomki-yeo (changes requested):
Having logger is a good thing but let's find more silent way than this PR. At least I would not  include it as an argument of the constructors as it will break all projects that use traccc. Also users are not supposed to struggle with loggers

core/CMakeLists.txt
stephenswat: Is it a big issue for `traccc::core` to depend on Boost? If so I am sure we can work around it but Boost is installed basically everywhere already. :thinking:

stephenswat:
> Having logger is a good thing but let's find more silent way than this PR. At least I would not include it as an argument of the constructors as it will break all projects that use traccc. Also users are not supposed to struggle with loggers

Would it be an acceptable solution for you if the logger had a default parameter so the external behaviour of the constructors would remain identical to the current behaviour?

core/CMakeLists.txt
beomki-yeo: I believe not being depdendent on Boost greatly reduces stress. Upgrading the Boost version has been always painful to me :/ ..

beomki-yeo:
That would be acceptable

core/CMakeLists.txt
stephenswat: Okay, fair enough! I'll work out a way to do this without depending directly on `Acts::Core`. :slightly_smiling_face:

beomki-yeo: Thanks!

paulgessinger: Boost is one of the most universally used C++ libs, is readily available anywhere (interesting that you can't get it on Perlmutter), and building + upgrading it is ~ trivial, since they essentially never break API, and a CMake-based build possible now.

All in all: having this as a dependency should not be a concern, really. Specifically with the Athena integration, having a single logging system and not multiple ones is beneficial.

We will eventually have to make Acts Core a dependency of traccc anyway when we continue unifying the code bases, so why not do it now? I don't think we should be duplicating work "just because".

gagnonlg: @beomki-yeo I think the sort-of-official stance on such things at NERSC is to just use a container. I can help you get setup with that if you want, just lmk

beomki-yeo: Sweet. Thanks but I have already asked you to demonstrate that when we prepared the tutorial. I don't think I have heard anything on it from you since then

gagnonlg: Well you made it work before I got to it ^^ Sorry about that, let's talk next week

beomki-yeo: > We will eventually have to make Acts Core a dependency of traccc anyway when we continue unifying the code bases,

It is opposite If I am correct. What is being developed in ACTS is making traccc a dependency of ACTS.

beomki-yeo: > I  think the sort-of-official stance on such things at NERSC is to just use a container. 

Also please make sure that if you are talking about NERSC standard or ATLAS standard. As far as I can see [here](https://docs.nersc.gov/development/compilers/wrappers/), NERSC is serious in supporting various compilers to general users without relying on containers

paulgessinger: > It is opposite If I am correct. What is being developed in ACTS is making traccc a dependency of ACTS.

With this change, it depends on the core library, which is what we want so we can eventually share definitions and conventions from ACTS core headers. The traccc plugin effectively makes it a component of the main project.

stephenswat:
There we go, all the algorithms now have default loggers so the external behaviour is unchanged, and the code now uses the ACTS core only conditionally, i.e. if it is already available; so the code continues to work without ACTS core.

core/CMakeLists.txt
beomki-yeo: @stephenswat  thanksfully handled the issue so I will resolve the conversation. @paulgessinger We can start discussing whether to include ActsCore as a part of traccc only if you have a concrete idea on it. I am pretty sure that most of Acts EDM or headers will not work in traccc out of the box because they are not GPU friendly. We cannot make `traccc::core` to be dependent on ActsCore just to use a couple of utility headers from it.

beomki-yeo: > a CMake-based build possible now.

Is it possible from https://github.com/boostorg/cmake ? It might be worth trying it out

paulgessinger (changes requested):
This is pretty messy. 

I suggest we just add ActsCore as a dependency here and start making our code more unified and less complex, rather than increasing complexity. 

I think littering our code with preprocessor instructions is a pretty bad idea and we should refrain from doing this to avoid tech debt.

krasznaa (changes requested):
Without having gone through the entire thing completely...

I agree with @beomki-yeo that I don't want the current `traccc::core` algorithm to depend on `ActsCore`. At the same time, clearly future `ActsSYCL`, `ActsCUDA`, etc. libraries will publicly depend on `ActsCore`. So it's a good idea to use a carbon copy of Acts's logger code.

But I would just always use the carbon copy. As long as we have to have a copy in our code, what do we win by sometimes not using it?

I would remove the conditional compilation flag, and just use the "imported" logger as-is.

Now... did you think at all about logging (error reporting...) in the GPU code? Since to me that is the actually interesting part here. For the host-side logging it was clear that we would follow what Acts has been doing since forever. (Which itself is a variation on what Gaudi does. Not sure if you've seen [Heretic](https://www.imdb.com/title/tt28015403/)... Its main premise applies here as well. 😆)

core/include/traccc/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.hpp
```diff
@@ -156,6 +159,10 @@ class greedy_ambiguity_resolution_algorithm
         state_t& final_state) const;
 
     config_t _config;
+
+    std::unique_ptr<const Logger> _logger;
+
+    const Logger& logger() const override { return *_logger; }
```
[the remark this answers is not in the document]
stephenswat: Unfortunately this went out of the window when we started having some algorithms depend on the `traccc::algorithm` class multiple times, which would then require multiple loggers and be a general nuisance. The only reasonable solution would be a loggable algorithm mixin but I think the current solution is probably the most elegant.

stephenswat:
> But I would just always use the carbon copy. As long as we have to have a copy in our code, what do we win by sometimes not using it?

The problem is that the ACTS and Athena code will be passing in the _actual_ ACTS logger, not out carbon copy. And you and I know that those types are functionally the same, but the compiler obviously doesn't - so it will reject the code with a compilation error.

stephenswat:
> Now... did you think at all about logging (error reporting...) in the GPU code? Since to me that is the actually interesting part here.

AFAIK nobody has come up with a better solution than what I proposed in #640, so as far as I am concerned that is going to be the way to go unless someone comes up with something really fancy!

CMakeLists.txt
```diff
@@ -370,6 +370,20 @@ option( TRACCC_ENABLE_NVTX_PROFILING
 # option for algebra plugins (ARRAY EIGEN SMATRIX VC VECMEM)
 set(TRACCC_ALGEBRA_PLUGINS ARRAY CACHE STRING "Algebra plugin to use in the build")
 
+# Set up either the ACTS or traccc logger.
+if(TRACCC_SETUP_ACTS OR TARGET ActsCore)
+   message(STATUS "Enabling the ACTS logger, because ACTS is set to be enabled or already exists.")
+   set(TRACCC_USE_ACTS_LOGGER ON)
+else()
+   set(TRACCC_USE_ACTS_LOGGER OFF)
+endif()
```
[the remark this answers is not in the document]
stephenswat: This is a nice idea. :+1:

core/include/traccc/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.hpp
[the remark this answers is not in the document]
krasznaa: Okay... 🤔 Then how about adding a class called `traccc::log_writer` / `traccc::messaging` / etc., which all the algorithms would inherit from? It would be the one providing a uniform interface for the macros.

In the end this is what we do in Athena as well:

https://gitlab.cern.ch/atlas/athena/-/blob/main/Control/AthenaBaseComps/AthenaBaseComps/AthMessaging.h?ref_type=heads

core/include/traccc/utils/logging.hpp
```diff
@@ -0,0 +1,322 @@
...
+
+#define TRACCC_ERROR(x) TRACCC_LOG(::traccc::logging::ERROR, x)
+
+#define TRACCC_FATAL(x) TRACCC_LOG(::traccc::logging::FATAL, x)
+
+namespace traccc {
+namespace logging {
+enum Level { VERBOSE = 0, DEBUG, INFO, WARNING, ERROR, FATAL, MAX };
+
+inline std::string_view levelName(Level level) {
```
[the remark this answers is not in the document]
stephenswat: Sure thing, I don't mind putting it there. :smile:

core/include/traccc/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.hpp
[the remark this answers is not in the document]
stephenswat: Yup that's what I meant by mixin. If you prefer I can do it that way, I think that's acceptable too. I'd usually aim to reduce the amount of OOP magic to a minimum (OOP considered harmful etc., etc.) but I am flexible.

krasznaa:
> > Now... did you think at all about logging (error reporting...) in the GPU code? Since to me that is the actually interesting part here.
> 
> AFAIK nobody has come up with a better solution than what I proposed in #640, so as far as I am concerned that is going to be the way to go unless someone comes up with something really fancy!

Not quite what I meant. 🤔 Though it's good that you reminded me that we'll need some generic functionality for "monitoring" as well.

I was rather thinking of something along the lines of the vecmem logging code.

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/utils/debug.hpp

But I don't think that quite scales to these algorithms. That's sort of good enough for debugging low-level functionality in special builds of the code, but in this project I would rather go with a combination of
  - warning / error messages, that are always printed when something bad is detected in the GPU code;
  - "monitoring information" being collected through a runtime flag.

We don't have to tackle those here. But we should start more seriously thinking about them. 🤔

krasznaa:
> > But I would just always use the carbon copy. As long as we have to have a copy in our code, what do we win by sometimes not using it?
> 
> The problem is that the ACTS and Athena code will be passing in the _actual_ ACTS logger, not our carbon copy. And you and I know that those types are functionally the same, but the compiler obviously doesn't - so it will reject the code with a compilation error.

Wait, why?

Once the code is actually merged into [acts-project/acts](https://github.com/acts-project/acts), then of course we will just have a single class. The current (or maybe slightly improved) `Acts::Logger`.

But as long as Athena and Acts use this project as an external, why couldn't they construct its own loggers? As long as this carbon copy can forward its messages to an outside logger like `Acts::Logger` does with `MsgStream`, I really don't see a problem here.

paulgessinger:
@krasznaa why the duplication and tech debt now though? Just adopting the core logger type at this point already we're improving things already, rather than increasing divergence. 

Unless your argument is NIH, this has basically no downsides at this point.

CMakeLists.txt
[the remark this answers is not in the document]
stephenswat: Done.

core/include/traccc/utils/logging.hpp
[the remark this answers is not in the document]
stephenswat: Done.

examples/run/alpaka/seq_example_alpaka.cpp
```diff
@@ -52,7 +52,9 @@ int seq_run(const traccc::opts::detector& detector_opts,
             const traccc::opts::clusterization& clusterization_opts,
             const traccc::opts::track_seeding& seeding_opts,
             const traccc::opts::performance& performance_opts,
-            const traccc::opts::accelerator& accelerator_opts) {
+            const traccc::opts::accelerator& accelerator_opts,
+            std::unique_ptr<const traccc::Logger> _logger) {
```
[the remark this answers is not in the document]
stephenswat: Renamed to remove the underscore.

stephenswat:
Okay, I've addressed all the topics of conversation. Clearly we have reached some impasse when it comes to the reliance on the ACTS Core library, so I think the solution of selecting either the ACTS logger or our own carbon copy of it is the best solution until the integration continues. I'd like to move towards merging this now.

core/include/traccc/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.hpp
```diff
@@ -156,6 +159,10 @@ class greedy_ambiguity_resolution_algorithm
         state_t& final_state) const;
 
     config_t _config;
+
+    std::unique_ptr<const Logger> m_logger;
+
+    const Logger& logger() const override { return *m_logger; }
```
krasznaa, line 165: Let's talk in the afternoon in person. Since I still very much dislike the amount of boilerplate code added in this PR. :frowning: Since this is something that we'll keep as a design definitely for a while, I do want to come to something that we can all live with...

paulgessinger: This is the ACTS design, if we want to use ACTS_XXX macros, this is what it'll be.

krasznaa: There **has to** be a way of factorizing this with a mix-in class. I have a very hard time accepting that every single client has to have their own member functions defined separately. We stopped doing that in Athena a decade ago...

core/include/traccc/utils/logging.hpp
```diff
@@ -0,0 +1,40 @@
...
+}  // namespace traccc
+
+#define TRACCC_LOCAL_LOGGER(x) ACTS_LOCAL_LOGGER(x)
+#define TRACCC_LOG(level, x) ACTS_LOG(level, log)
+#define TRACCC_VERBOSE(x) ACTS_VERBOSE(x)
+#define TRACCC_DEBUG(x) ACTS_DEBUG(x)
+#define TRACCC_INFO(x) ACTS_INFO(x)
+#define TRACCC_WARNING(x) ACTS_WARNING(x)
+#define TRACCC_ERROR(x) ACTS_ERROR(x)
+#define TRACCC_FATAL(x) ACTS_FATAL(x)
```
paulgessinger, line 40: I'm guessing if we want to move away from these macro aliases, they should be easy to replace.

stephenswat:
Ping.

core/include/traccc/seeding/doublet_finding.hpp
```diff
@@ -14,6 +14,7 @@
 #include "traccc/seeding/detail/spacepoint_type.hpp"
 #include "traccc/seeding/doublet_finding_helper.hpp"
 #include "traccc/utils/algorithm.hpp"
+#include "traccc/utils/logging_mixin.hpp"
```
krasznaa, line 17: In some files you include both `logging.hpp` and `logging_mixin.hpp`. But somewhere you only include `logging_mixin.hpp`.

What should be the convention? As long as the "logger base class" is expected to pull in the `Logger` type, I think it's fine to only include `logging_mixin.hpp`. We have plenty of examples of that in our other projects. No Athena algorithms ever explicitly include headers related to logging, status codes, and such. As they should by definition be available.

This is not a super important point, but could still be worth thinking about.

core/include/traccc/utils/logging.hpp
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * (c) 2021-2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+/*
+ * This file was copied from the ACTS repository, which is also available
+ * under the Mozilla Public License Version 2.0
+ */
```
krasznaa, line 11: I guess this is a leftover? Just our normal header should be appropriate for the current state of the file I think.

core/include/traccc/utils/logging_mixin.hpp
```diff
@@ -0,0 +1,30 @@
...
+    logging_mixin() = delete;
+
+    const Logger& logger() const {
+        assert(m_logger.get() != nullptr);
+        return *m_logger;
+    }
+
+    private:
+    std::unique_ptr<const Logger> m_logger;
+};
```
krasznaa, line 29: I'm not in love with the class name. :thinking:

This is not a deciding factor for sure, but in the Gaudi/Athena ecosystem we landed on the "messaging" name for these sort of classes.
  - https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/GaudiKernel/include/GaudiKernel/CommonMessaging.h?ref_type=heads
  - https://gitlab.cern.ch/atlas/athena/-/blob/main/Control/AthenaBaseComps/AthenaBaseComps/AthMessaging.h?ref_type=heads
  - https://gitlab.cern.ch/atlas/athena/-/blob/main/Control/AthToolSupport/AsgMessaging/AsgMessaging/AsgMessaging.h?ref_type=heads

In keeping with some of our other conventions, how about calling it `traccc::host_messaging` or `traccc::host::messaging`? :thinking:

Not necessarily for this PR, but I think the class should also get:
  - An explicit (defaulted) move constructor and move assignment operator. Which I would tuck away into a `.cpp` file. I agree that with the `logger()` function we probably want to be `inlined`, but the constructor code does not need to be super efficient. :thinking:
  - A function like `set_logger(...)`, so that the logger could be modified on an existing algorithm / component. I think this could come in handy in a number of situations in the future.

But it's really the name of the class that I'd like to open up for debate in this PR... :thinking:

examples/options/include/traccc/options/program_options.hpp
```diff
@@ -28,7 +29,7 @@ class program_options {
     program_options(
         std::string_view description,
         const std::vector<std::reference_wrapper<interface> >& options,
-        int argc, char* argv[]);
+        int argc, char* argv[], std::unique_ptr<const traccc::Logger> ilogger);
```
krasznaa, line 32: Why no default argument here? I think a default name could be conveniently given for messages coming from this code.

io/src/csv/read_cells.cpp
```diff
@@ -60,8 +63,7 @@ read_deduplicated_cells(std::string_view filename) {
         }
     }
     if (nduplicates > 0) {
-        std::cout << "WARNING: @traccc::io::csv::read_cells: " << nduplicates
-                  << " duplicate cells found in " << filename << std::endl;
+        TRACCC_WARNING(nduplicates << " duplicate cells found in " << filename);
```
krasznaa, line 66: It is not for this PR. But this probably shouldn't be a warning in the end, but rather just a debug message. Since it's something that we expect to happen with G4 simulations coming from Acts.

But since these warnings have been good progress indicators for our file reading, let's think about normalizing the messages around here, in a follow-up PR.

io/src/read_cells.cpp
```diff
@@ -19,6 +19,7 @@ namespace traccc::io {
 
 void read_cells(edm::silicon_cell_collection::host& cells, std::size_t event,
                 std::string_view directory,
+                std::unique_ptr<const Logger> ilogger,
```
krasznaa, line 22: For standalone functions, would it not be acceptable to just use a hardcoded logger name? :thinking: For classes of course we want to have the ability to provide the logging name from the outside. But for stateless functions, is it necessary? Would it not be okay to just use `"traccc::io::read_cells"`, `"traccc::io::csv::read_cells"`, etc. for the source of the messages? :thinking:

Then we wouldn't need to bother with changing the API of so many of these functions.

krasznaa (commented):
I'm generally on board with the way the code is organized now. It's mainly the name of the mix-in class that I'd like to change. Everything else we can tweak later on. (For instance I'd be in favour of having meaningful default names for the algorithms. So that clients not overriding the names, would still get something meaningful.) But once we commit to a given name for that class, that is probably not changing anymore.

core/include/traccc/utils/logging_mixin.hpp
stephenswat: `messaging` is okay. :+1:

io/src/csv/read_cells.cpp
stephenswat: Agreed!

io/src/read_cells.cpp
stephenswat: This is not quite how loggers work here. The loggers carry much more than just a name, they also carry formatting and configuration options. I agree we can accept a default argument, but the only valid default argument is the dummy logger; we cannot have a default name for loggers as you suggest.

examples/options/include/traccc/options/program_options.hpp
stephenswat: See message below, but I will add a default argument.

io/src/read_cells.cpp
krasznaa: That's unfortunate. Is it really not possible to have a default logger with some specific name, that prints messages to `std::cout` with level `INFO`? :confused: I thought the idea was that if one was okay with messages going "with the Acts formatting" to `std::cout`, they wouldn't need to bother too much with setting up their own loggers. Only when they want to pipe messages "into their own system", would they need to become explicit.

I accept that even the I/O functions need to be able to receive a logger from the outside. (Though I'll try to think of a better API for that in a follow-up.) But why can't the default still be something like

```
= getDefaultLogger("meaningfulDefaultName", Logging::INFO)
```

? Why must it be a dummy logger by default?

stephenswat: It's possible, but we should not risk writing anything to `std::cout` without explicit user instruction to do so. That's why the only reasonable default logger is the dummy one.

krasznaa: Why? :confused: As long as there is a clear and well documented way for how users can override it, what's the harm in having default messages be meaningful/pretty? At that point, why use `getDummyLogger()` at all? Why don't we just force users to **always** supply a logger that they constructed?

stephenswat: > Why don't we just force users to **always** supply a logger that they constructed?

That's exactly the current state of the code. :D

core/include/traccc/clusterization/clusterization_algorithm.hpp
```diff
@@ -41,7 +43,9 @@ class clusterization_algorithm
     ///
     /// @param mr The memory resource to use for the result objects
     ///
-    clusterization_algorithm(vecmem::memory_resource& mr);
+    clusterization_algorithm(
+        vecmem::memory_resource& mr,
+        std::unique_ptr<const Logger> logger = getDummyLogger().clone());
```
krasznaa, line 48: :confused: Technically clients don't need to change their code when instantiating these classes. They'll "just" get absolutely meaningless printouts if they don't.

This is what I don't understand. As long as we use default arguments here, why not make them such that in "standalone mode" we wouldn't even need to override them? Only once we integrate the tools into a larger framework. (...cough... Athena ...cough...)

core/include/traccc/seeding/doublet_finding.hpp
stephenswat: Fixed; cases where only the mixin can be included have been made to do so.

core/include/traccc/utils/logging.hpp
stephenswat: Well spotted! Removed.

examples/options/include/traccc/options/program_options.hpp
stephenswat: Default added.

core/include/traccc/clusterization/clusterization_algorithm.hpp
stephenswat: I'm not sure I really understand your argument, but I think your point is that you want users to be able to get meaningful logging even with just default parameters? Afaik that is only possible by having some kind of global state, and I absolutely want to avoid that.

io/src/read_cells.cpp
stephenswat: Default added here.

core/include/traccc/utils/logging_mixin.hpp
stephenswat: `s/logging_mixin/messaging/g`

examples/run/sycl/seq_example_sycl.sycl
```diff
@@ -434,29 +450,27 @@ int seq_run(const traccc::opts::detector& detector_opts,
...
+    TRACCC_INFO("==>Elapsed times...\n" << elapsedTimes);
 
     return 0;
 }
 //
 // The main routine
 //
 int main(int argc, char* argv[]) {
+    std::unique_ptr<const traccc::Logger> logger = traccc::getDefaultLogger(
+        "TracccExampleSeqSycl", traccc::Logging::Level::INFO);
```
krasznaa, line 473: I might very well not understand the Acts logging code...

But providing default logger objects to the algorithms with this formalism would not be possible? What sort of a global state would that mean, if by default they all get their own loggers printing to `std::cout` at an `INFO` level?

The user would still be able to override everything at their pleasure. But it would seem possible to provide meaningful default objects to the algorithms. :thinking: What am I missing here?

stephenswat: Right, but we cannot print to users' `stdout` without their permission; that's just bad software design. If we want to provide a default logger (which I agree we should) then the only proper way to do that is to provide a default logger that doesn't print anything.

The only remotely viable alternative is to to have a traccc-global logger which the user can set which is then used for all instances where a traccc function needs a default. But as I said I think we should aim to avoid this kind of setup.

PR #826, opened 2025-01-29
Remove obsolete Thrust functionality check

stephenswat:
This check was possibly useful in the past, but now produces spurious warnings. As the code runs fine even without this check on the Thrust installation, the easiest solution is simply to remove it.


krasznaa (approved):
👍

PR #827, opened 2025-01-29
Version Update, main branch (2025.01.29.)

krasznaa:
Update the project to version `0.19.0`. So that the upcoming tag of @paradajzblond would report the correct version itself.

Plus switched to require CMake 3.25 on the project level already. Since subdirectories require this ever since #792 anyway. 🤔

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,12 +1,12 @@
 # TRACCC library, part of the ACTS project (R&D line)
 #
-# (c) 2021-2024 CERN for the benefit of the ACTS project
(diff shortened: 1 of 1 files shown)


paradajzblond:
Can we pull in commit #825 before this tag change please :) I need to check that it fixes all the issues.

stephenswat:
> Can we pull in commit #825 before this tag change please :) I need to check that it fixes all the issues.

Don't worry, merging this doesn't actually tag a new version yet; that's a separate process. This simply prepares the repository for a new tag!

stephenswat (approved):

PR #828, opened 2025-01-29
Fix broken assertion in prune tracks kernel

stephenswat:
Mea culpa, I made a mistake in a recent assertion added to the `prune_tracks` kernel which was too assertive, requiring an equality relation rather than a greater-or-equal relation between the number of tips and the number of tracks.


krasznaa (approved):
Sure. This time around I guess you runtime tested it. 😛

stephenswat:
> This time around I guess you runtime tested it. 😛

Of course not. :D

PR #829, opened 2025-01-29
Make Thrust a private dependency of the CUDA library

stephenswat:
This was previously a public dependency due to a lingering include of the Thrust pair code in a header, but this dependency is no longer necessary. Thus, we can remove the include from the header file and thereby turn Thrust into a private dependency of the CUDA library. This should also fix the install interface, which is currently broken due to the lack of Thrust setup code.


krasznaa (approved):
👍

PR #831, opened 2025-01-30
Add code for per-commit benchmarking

stephenswat:
This PR adds a bunch of code to the extras directory with the code that I used to plot the performance of individual kernels over the course of the repository's history, e.g.:

test


krasznaa:
We should definitely add your code to the repository for future, similar debugging sessions.

But something that would be even better, would be to add measurements like this to the CI. 🤔 Similar to what we have already on: https://github.com/acts-project/traccc/blob/main/.github/workflows/benchmarks.yml Which is obviously not as good as it should be. 😭

If we could make "your kind of" plots automatically, that would be the dream... 🤔

stephenswat:
Hmm yes, agree; I'll have to see if I can implement this in that Gitlab repository!

beomki-yeo:
Yeah it would be greater if we can see this continuously

krasznaa (approved):
With all that being said, adding a new subdirectory with scripts under `extras/` is not a controversial thing. Let's save these scripts of yours!

PR #832, opened 2025-01-30
Make traccc::alpaka executables work in pure C++

StewMH:
Move device code to utils. Built on work by @CrossR (currently in a private branch with other changes for track fitting).

### DIFF
--- a/device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
+++ b/device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
@@ -8,12 +8,12 @@
 #pragma once
 
 // Project include(s)
+#include "traccc/alpaka/utils/vecmem_types.hpp"
 #include "traccc/edm/seed.hpp"
 #include "traccc/edm/spacepoint.hpp"
 #include "traccc/edm/track_parameters.hpp"
 #include "traccc/utils/algorithm.hpp"
 #include "traccc/utils/memory_resource.hpp"
-#include "traccc/alpaka/utils/vecmem_types.hpp"
 
 namespace traccc::alpaka {
 
--- a/device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
+++ b/device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
@@ -13,6 +13,7 @@
 #include "traccc/edm/track_parameters.hpp"
 #include "traccc/utils/algorithm.hpp"
 #include "traccc/utils/memory_resource.hpp"
+#include "traccc/alpaka/utils/vecmem_types.hpp"
 
 namespace traccc::alpaka {
 
@@ -30,7 +31,7 @@ struct track_params_estimation
     /// @param copy The copy object to use for copying data between device
     ///             and host memory blocks
     track_params_estimation(const traccc::memory_resource& mr,
-                            vecmem::copy& copy);
+                            traccc::alpaka::vecmem::device_copy& copy);
 
     /// Callable operator for track_params_estimation
     ///
@@ -58,7 +59,7 @@ struct track_params_estimation
     /// Memory resource used by the algorithm
     traccc::memory_resource m_mr;
     /// Copy object used by the algorithm
-    vecmem::copy& m_copy;
+    vecmem::device_copy& m_copy;
 };
 
 }  // namespace traccc::alpaka
--- a/device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
+++ b/device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
@@ -8,7 +8,8 @@
 #pragma once
 
 // Project include(s)
-#include "traccc/alpaka/utils/vecmem_types.hpp"
+#include <vecmem/utils/copy.hpp>
+
 #include "traccc/edm/seed.hpp"
 #include "traccc/edm/spacepoint.hpp"
 #include "traccc/edm/track_parameters.hpp"
@@ -31,7 +32,7 @@ struct track_params_estimation
     /// @param copy The copy object to use for copying data between device
     ///             and host memory blocks
     track_params_estimation(const traccc::memory_resource& mr,
-                            traccc::alpaka::vecmem::device_copy& copy);
+                            vecmem::copy& copy);
 
     /// Callable operator for track_params_estimation
     ///
--- a/device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
+++ b/device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
@@ -59,7 +59,7 @@ struct track_params_estimation
     /// Memory resource used by the algorithm
     traccc::memory_resource m_mr;
     /// Copy object used by the algorithm
-    vecmem::device_copy& m_copy;
+    ::vecmem::copy& m_copy;
 };
 
 }  // namespace traccc::alpaka
--- a/device/alpaka/include/traccc/alpaka/utils/get_device_info.hpp
+++ b/device/alpaka/include/traccc/alpaka/utils/get_device_info.hpp
@@ -7,30 +7,13 @@
 
 #pragma once
 
-#include <alpaka/acc/Tag.hpp>
+#include <string>
 
 namespace traccc::alpaka {
-
-// Get alpaka tag for current device
-#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
-using AccTag = ::alpaka::TagGpuCudaRt;
-#elif defined(ALPAKA_ACC_GPU_HIP_ENABLED)
-using AccTag = ::alpaka::TagGpuHipRt;
-#elif defined(ALPAKA_ACC_SYCL_ENABLED)
-#if defined(ALPAKA_SYCL_ONEAPI_CPU)
-using AccTag = ::alpaka::TagCpuSycl;
-#elif defined(ALPAKA_SYCL_ONEAPI_FPGA)
-using AccTag = ::alpaka::TagFpgaSyclIntel;
-#elif defined(ALPAKA_SYCL_ONEAPI_GPU)
-using AccTag = ::alpaka::TagGpuSyclIntel;
-#endif
-#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLED)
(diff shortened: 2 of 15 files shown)


stephenswat:
Non-technical point, but GitHub is failing to pick up both @CrossR and your own contribution to this because I guess your email being unknown to GitHub, and Ryan not being mentioned as a co-author.

If Ryan is reponsible for part of this code I'd suggest adding a co-authorship trailer to the end of the commit:

`Co-authored-by: Ryan Cross <ryan'semail@example.ac.uk>`

StewMH:
OK, do I need a new commit now? or can that be done on the merge commit?

stephenswat:
I can try doing it via the merge commit. :+1:

device/alpaka/src/utils/utils.hpp
```diff
@@ -26,16 +25,34 @@ using Dim = ::alpaka::DimInt<1>;
...
-    4;
-#endif
+template <typename TAcc>
+consteval std::size_t getWarpSize() {
+    if constexpr (::alpaka::accMatchesTags<TAcc, ::alpaka::TagGpuCudaRt,
+                                           ::alpaka::TagGpuHipRt,
+                                           ::alpaka::TagGpuSyclIntel>) {
+        return 32;
+    } else {
+        return 4;
```
stephenswat, line 53: Where does the value 4 come from here?

tests/alpaka/CMakeLists.txt
```diff
@@ -26,6 +25,7 @@ endif()
 
 traccc_add_test( alpaka
    ${TRACCC_ALPAKA_TEST_SOURCES}
+   test_cca.cpp
```
stephenswat, line 28: Is there a reason for this change? And indeed for keeping `TRACCC_ALPAKA_TEST_SOURCES` around at all?

device/alpaka/src/utils/utils.hpp
```diff
@@ -26,16 +25,34 @@ using Dim = ::alpaka::DimInt<1>;
...
-    32;
-#else
-    4;
-#endif
+template <typename TAcc>
+consteval std::size_t getWarpSize() {
+    if constexpr (::alpaka::accMatchesTags<TAcc, ::alpaka::TagGpuCudaRt,
+                                           ::alpaka::TagGpuHipRt,
+                                           ::alpaka::TagGpuSyclIntel>) {
+        return 32;
```
stephenswat, line 51: Also, note that AMD RDNA devices have a configurable warp size of either 32 or 64, and CDNA have a fixed size of 64. So I'm not sure this choice makes sense for AMD.

tests/alpaka/CMakeLists.txt
StewMH: `test_cca.cpp` is now a C++ rather than CUDA/HIP/SYCL executable. But indeed, no need to keep the variable any more.

device/alpaka/src/utils/utils.hpp
StewMH: Thanks, indeed Ryan made this change in his branch as well

device/alpaka/src/utils/utils.hpp
StewMH: 4 threads for threaded CPU.

stephenswat (approved):

device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
```diff
@@ -58,7 +59,7 @@ struct track_params_estimation
     /// Memory resource used by the algorithm
     traccc::memory_resource m_mr;
     /// Copy object used by the algorithm
-    vecmem::copy& m_copy;
+    vecmem::device_copy& m_copy;
```
krasznaa, line 62: Why would you need to use a specific type here all of a sudden? 😕 We use the generic `vecmem::copy` type in the CUDA and SYCL algorithms as well.

The important part is to **instantiate** the correct type. Not to have reference to the correct type.

device/alpaka/include/traccc/alpaka/utils/get_device_info.hpp
```diff
@@ -0,0 +1,36 @@
...
+#if defined(ALPAKA_SYCL_ONEAPI_CPU)
+using AccTag = ::alpaka::TagCpuSycl;
+#elif defined(ALPAKA_SYCL_ONEAPI_FPGA)
+using AccTag = ::alpaka::TagFpgaSyclIntel;
+#elif defined(ALPAKA_SYCL_ONEAPI_GPU)
+using AccTag = ::alpaka::TagGpuSyclIntel;
+#endif
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLED)
+using AccTag = ::alpaka::TagCpuThreads;
+#endif
```
krasznaa, line 29: I don't like these sitting here. They are not used in the interface of `get_device_info()`. So if the client needs them, they should be put into a header that has a more descriptive name. 🤔

device/alpaka/include/traccc/alpaka/utils/get_device_info.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+/// Function that prints the current device information to the console.
+/// Included as part of the traccc::alpaka namespace, to avoid having to include
+/// alpaka headers in any users of the library.
+void get_device_info();
```
krasznaa, line 34: The correct API would be:

```suggestion
std::string get_device_info();
```

We should never print stuff to `std::cout` with no override possibilities. (I'll be changing some of @stephenswat's recent updates as well...)

device/alpaka/src/utils/utils.hpp
```diff
@@ -8,34 +8,47 @@
...
+#elif defined(ALPAKA_SYCL_ONEAPI_FPGA)
+using Acc = ::alpaka::AccFpgaSyclIntel<Dim, Idx>;
+#elif defined(ALPAKA_SYCL_ONEAPI_GPU)
+using Acc = ::alpaka::AccGpuSyclIntel<Dim, Idx>;
+#endif
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLED)
+using Acc = ::alpaka::AccCpuThreads<Dim, Idx>;
+#else
+#error "No supported backend selected." //we definitely want to fail the build if no matching accelerator is found
+#endif
```
krasznaa, line 35: Could this not appear just once? In this PR this is the second time I see the exact same thing. Or is it not exactly the same?

device/alpaka/include/traccc/alpaka/utils/vecmem_types.hpp
```diff
@@ -102,4 +104,13 @@ struct host_device_types<::alpaka::TagGpuSyclIntel> {
...
     using device_copy = ::vecmem::sycl::copy;
 };
+
+using device_memory_resource =
+    typename host_device_types<AccTag>::device_memory_resource;
+using host_memory_resource =
+    typename host_device_types<AccTag>::host_memory_resource;
+using managed_memory_resource =
+    typename host_device_types<AccTag>::managed_memory_resource;
+using device_copy = typename host_device_types<AccTag>::device_copy;
```
krasznaa, line 114: 🤔 Not necessarily for this PR, but I'd really like if the device libraries would not publicly depend on the device/language specific VecMem libraries. Much like how they shouldn't publicly depend on CUDA/SYCL/etc.

For this, I'd imagine that Alpaka could provide a helper class, that could provide (non)-const references to memory resources and copy objects of the appropriate types. Through their abstract vecmem interfaces. So that the "decision" about what object to instantiate, could sit all in the `traccc_alpaka` private source code. 🤔

krasznaa (changes requested):
Let's talk about this for just a little more...

device/alpaka/src/utils/utils.hpp
stephenswat: I think one of them defines Accelerator _tagS_ and this one is just Accelerators; I noticed the same but I think they are different.

StewMH: Right, the tags are a way to refer to different accelerators without having to include `alpaka/alpaka.hpp` at which point you have to use the device compiler. It's annoying to have basically the same logic twice but it lets us define`Acc` in the source and then `Tag` in the header.

device/alpaka/include/traccc/alpaka/utils/get_device_info.hpp
StewMH: `device_properties.hpp`?

krasznaa: Sure, that would sound okay.

device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
```diff
@@ -30,7 +31,7 @@ struct track_params_estimation
     /// @param copy The copy object to use for copying data between device
     ///             and host memory blocks
     track_params_estimation(const traccc::memory_resource& mr,
-                            vecmem::copy& copy);
+                            traccc::alpaka::vecmem::device_copy& copy);
```
krasznaa, line 34: The member may have changed, but the interface still expects a concrete type. This should also stay as-is. (Adding a `::` prefix to the `vecmem` namespace may be needed here as well though. Or you could avoid introducing a `traccc::alpaka::vecmem` namespace. 🤔)

device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
```diff
@@ -8,6 +8,7 @@
 #pragma once
 
 // Project include(s)
+#include "traccc/alpaka/utils/vecmem_types.hpp"
```
krasznaa, line 11: To be clear, this include should also not be here.

device/alpaka/src/seeding/track_params_estimation.cpp
```diff
@@ -10,6 +10,7 @@
 
 // Project include(s).
 #include "traccc/alpaka/seeding/track_params_estimation.hpp"
+#include "traccc/alpaka/utils/vecmem_types.hpp"
```
krasznaa, line 13: Not needed here either.

krasznaa (changes requested):

krasznaa (approved):
Okay, we could go ahead like this. The further design improvements/changes will come later. :wink:

PR #833, opened 2025-01-31
Use configuration types as SSOT for defaults

stephenswat:
As it stands, some of our default values for configuration parameters, e.g. those of the track finding, have defaults defined both in the configuration type as well as in the corresponding options type in the examples. This is confusing as @paradajzblond discovered today. In this commit, I remove the defaults defined in the examples code and treat the configuration type defaults as the single source of truth for default values.

### DIFF
--- a/examples/options/include/traccc/options/clusterization.hpp
+++ b/examples/options/include/traccc/options/clusterization.hpp
@@ -32,14 +32,8 @@ class clusterization
     std::unique_ptr<configuration_printable> as_printable() const override;
 
     private:
-    /// @name Options
-    /// @{
-    /// The number of cells to merge in a partition
-    unsigned int threads_per_partition;
-    unsigned int max_cells_per_thread;
-    unsigned int target_cells_per_thread;
-    unsigned int backup_size_multiplier;
-    /// @}
+    /// Internal configuration object
+    clustering_config m_config;
 };  // class clusterization
 
 }  // namespace traccc::opts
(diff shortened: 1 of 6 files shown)


examples/options/src/track_propagation.cpp
```diff
@@ -23,6 +23,8 @@ namespace po = boost::program_options;
 
 track_propagation::track_propagation()
     : interface("Track Propagation Options") {
+    m_search_window[0] = config.navigation.search_window[0];
+    m_search_window[1] = config.navigation.search_window[1];
 
     m_desc.add_options()("constraint-step-size-mm",
                          po::value(&(config.stepping.step_constraint))
```
niermann999, line 30: Can you please also add the config default values as the io default values for the rest of this file?

niermann999 (changes requested):

krasznaa (commented):
I like it. Just the `m_` prefixes from my side... 🤔

examples/options/src/track_propagation.cpp
stephenswat: Updated to draw all the configuration parameters from the detray default.

examples/options/include/traccc/options/clusterization.hpp
```diff
@@ -32,14 +32,8 @@ class clusterization
...
-    /// @name Options
-    /// @{
-    /// The number of cells to merge in a partition
-    unsigned int threads_per_partition;
-    unsigned int max_cells_per_thread;
-    unsigned int target_cells_per_thread;
-    unsigned int backup_size_multiplier;
-    /// @}
+    /// Internal configuration object
+    clustering_config config;
```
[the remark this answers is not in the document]
stephenswat: All the members in the touched files now have the `m_` prefix!

niermann999 (approved):

krasznaa (approved):
:+1:

PR #835, opened 2025-02-03
Use caching allocators in Thrust

stephenswat:
I was very pleasantly surprised to learn that using our vecmem memory resources in Thrust is trivial! You simply create a polymorphic allocator from them and Thrust will accept them as-is. This should reduce the amount of memory allocations significantly throughput traccc.

With this change, I observe a 10-15% improvement of the full-application throughput of traccc. :partying_face:


niermann999 (approved):

device/cuda/src/clusterization/measurement_sorting_algorithm.cu
```diff
@@ -32,9 +32,11 @@ measurement_sorting_algorithm::operator()(
         m_copy.get().get_size(measurements_view);
 
     // Sort the measurements in place
-    thrust::sort(thrust::cuda::par.on(stream), measurements_view.ptr(),
-                 measurements_view.ptr() + n_measurements,
-                 measurement_sort_comp());
+    thrust::sort(
+        thrust::cuda::par(std::pmr::polymorphic_allocator(&(m_mr.main)))
```
krasznaa, line 36: Strange that you needed to be so very explicit. :thinking: `std::pmr::polymoprhic_allocator` doesn't put `explicit` on its constructor for sure. That's how we can create `vecmem::vector` objects without always writing out the full allocator name in their constructors.

I guess I'll try `thrust::cuda::par(&(m_mr.main)).on(stream)` tomorrow, to see if that would also work...

krasznaa (commented):
I'm also pretty sure that we could avoid creating these policy objects on every single call separately. :thinking: Again, something to try.

I'm **very** happy though about the performance improvement that you found!

beomki-yeo:
Yeah - What a free lunch..

PR #836, opened 2025-02-03
Increase collection comparison performance

stephenswat:
Although the collection comparison code is not performance-critical, it is still very, very slow. In part, this is because it is written very naively. This commit improves the performance of this code by several orders of magnitude by sorting the collections, and by avoiding checks for elements which have already been matched at a lower uncertainty level. After these optimisations, it turned out that the Iterators-based progress bar was the performance-limiting factor, so I had to remove that as well.

We can now do GPU-vs-CPU physics comparisons for pile-up 200 events in a fraction of a second. :smile:


krasznaa (commented):
I'm on board with this, since the code is indeed painfully slow at the moment.

Just, you know... make it pass the CI first. :wink:

krasznaa (approved):
Sure!

PR #838, opened 2025-02-04
Add Spack environment specification

stephenswat:
This commit adds a YAML-based Spack environment specification which allows users to more easily install traccc's dependencies.

### DIFF
--- a/spack.yaml
+++ b/spack.yaml
@@ -1,0 +1,33 @@
+# traccc library, part of the ACTS project (R&D line)
+#
+# (c) 2025 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+spack:
+  specs:
+    # Build tools
+    - "cmake@3.26:"
+    - "intel-oneapi-compilers@2024.2.0"
+    - "cuda@12.5"
+    - "gcc@13"
+    # HEP dependencies
+    - "acts@38.2.0 +json"
+    - "detray@0.87"
+    - "covfie@0.10.0 +cuda"
+    - "vecmem@1.13.0%oneapi@2024.2.0 +cuda+sycl"
+    - "acts-algebra-plugins@0.26.2"
+    - "root cxxstd=20"
+    # General dependencies
+    - "intel-oneapi-tbb@2021.12.0"
+    - "kokkos"
+    - "alpaka"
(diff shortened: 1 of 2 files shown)


paulgessinger:
Should we add this to the build cache populater as well at some point?

spack.yaml
```diff
@@ -0,0 +1,33 @@
...
+# (c) 2025 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+spack:
+  specs:
+    # Build tools
+    - "cmake@3.26:"
+    - "intel-oneapi-compilers@2024.2.0"
+    - "cuda@12.5"
+    - "gcc@13"
```
paulgessinger, line 12: Is spack able to use this compiler yet? Last I checked this was still pending and waiting for a change to include compilers in the dependency graph.

stephenswat: Indeed that is true, the purpose of this is just to make gcc 13 available to the user, rather than to require it be used for compiling e.g. ACTS.

stephenswat:
> Should we add this to the build cache populater as well at some point?

I think that would be a good idea!

beomki-yeo (commented):
Using  `-DTRACCC_USE_SYSTEM_LIBS=ON` flag is mandatory?

Can we let spack install compilers, CMake, Boost and ROOT and let traccc CMake install other dependencies

stephenswat:
> Using `-DTRACCC_USE_SYSTEM_LIBS=ON` flag is mandatory?
> 
> Can we let spack install compilers, CMake, Boost and ROOT and let traccc CMake install other dependencies?

It's certainly not mandatory, but I don't see a good reason _not_ to use it. If you have a package manager set up already there is not really any use in having CMake download the dependencies for you, right? :thinking: 

But yes, it is of course possible to install only the compilers, CMake, Boost, and ROOT with Spack and to use CMake for the rest if you prefer!

beomki-yeo:
I thought it would be bothersome to keep the versions of all dependencies updated. 🤔
I don't have a strong opinion on this so probably you can go with the setup you prefer.

paulgessinger (approved):

stephenswat:
I changed this slightly, taking out the less stable dependencies for now (detray, vecmem, etc.).

PR #839, opened 2025-02-04
Fix missing typename in sanity checks

stephenswat:
https://github.com/acts-project/acts/pull/4068 exposes a mistake in our sanity checks, revealing that there are some missing typename keywords. This commit fixes those errors by adding the necessary keywords.

### DIFF
--- a/core/include/traccc/sanity/contiguous_on.hpp
+++ b/core/include/traccc/sanity/contiguous_on.hpp
@@ -54,11 +54,11 @@ is_contiguous_on(P&& projection, const CONTAINER& in) {
     // Compress adjacent elements
     for (std::size_t i = 0; i < n; ++i) {
         if (i == 0) {
-            iout[iout_size++] =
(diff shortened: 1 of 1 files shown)


niermann999 (approved):

krasznaa (approved):
Curious that none of the CI compilers complained about this. :confused: Do we possibly need some additional Debug builds? :thinking: Though we do have Debug builds with both GCC and Clang (oneAPI).

Should we add "vanilla" Clang Release+Debug builds as well to the CI? :thinking:

asalzburger (approved):
Thanks! I might need a new release then for this.

niermann999:
> > Can we build a patch release 19.0.1 with the fix?
> 
> That could absolutely be done. We may as well build `20.0.0`, since updates continue coming in. But as I saw it, Detray will also need some fixes. We should test that things work correctly with the `main` branch in Acts before making any new tags...

I am just guessing, but hopefully https://github.com/acts-project/detray/pull/926 will do the trick

PR #840, opened 2025-02-04
Implement less aggressive and configurable caching

stephenswat:
This commit relaxes our caching a little bit by implementing a custom caching algorithm which caches only small allocations. This prevents us from allocating massive amounts of memory only to never free them. Also generalizes the way we currently handle caching memory resource handling in order to make it more flexible. The caching threshold is taken as a command line parameter. This should help with the memory issues when processing ITk data seen by @paradajzblond.

### DIFF
--- a/examples/utils/include/traccc/examples/utils/caching_memory_resource.hpp
+++ b/examples/utils/include/traccc/examples/utils/caching_memory_resource.hpp
@@ -1,0 +1,31 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2023-2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <vecmem/memory/binary_page_memory_resource.hpp>
+#include <vecmem/memory/memory_resource.hpp>
+
+namespace traccc {
+class caching_memory_resource final : public vecmem::memory_resource {
+    public:
+    caching_memory_resource(vecmem::memory_resource& base_mr,
+                            std::size_t threshold);
+
+    void* do_allocate(std::size_t bytes, std::size_t alignment) final;
+    void do_deallocate(void* p, std::size_t bytes, std::size_t alignment) final;
+    bool do_is_equal(const vecmem::memory_resource& other) const noexcept final;
+
+    private:
+    vecmem::memory_resource& m_base_mr;
+    std::unique_ptr<vecmem::binary_page_memory_resource> m_caching_mr;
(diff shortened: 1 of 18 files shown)


examples/utils/include/traccc/examples/utils/caching_memory_resource.hpp
```diff
@@ -0,0 +1,31 @@
...
+    bool do_is_equal(const vecmem::memory_resource& other) const noexcept final;
+
+    private:
+    vecmem::memory_resource& m_base_mr;
+    std::unique_ptr<vecmem::binary_page_memory_resource> m_caching_mr;
+    std::size_t m_threshold;
+};
+
+std::unique_ptr<vecmem::memory_resource> make_caching_memory_resource(
+    vecmem::memory_resource& base_mr, std::size_t threshold);
```
krasznaa, line 30: What exactly does this additional function bring to the table here? :confused: We don't use this pattern anywhere else.

examples/utils/include/traccc/examples/utils/caching_memory_resource.hpp
```diff
@@ -0,0 +1,31 @@
...
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <vecmem/memory/binary_page_memory_resource.hpp>
+#include <vecmem/memory/memory_resource.hpp>
+
+namespace traccc {
+class caching_memory_resource final : public vecmem::memory_resource {
```
krasznaa, line 14: I like the idea, but I would not like to have this here. Under the examples in this repository.

I would propose to teach `vecmem::binary_page_memory_resource` about this threshold itself. :thinking: Similar to:

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/memory/pool_memory_resource.hpp#L75

Or... we may not need to do anything actually... :thinking: Instead of creating a new memory resource class from scratch, you should just use:

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/memory/conditional_memory_resource.hpp

That's the thing that **you** wrote it for years ago already after all.

krasznaa (changes requested):
The initiative I'm very much on board with. But we don't need to re-invent the wheel here.

examples/utils/include/traccc/examples/utils/caching_memory_resource.hpp
krasznaa: Actually, the class to be used should be:

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/memory/choice_memory_resource.hpp

PR #841, opened 2025-02-04
ref: update detray to v0.88.1

niermann999:
Update to the new detray version. I also made sure that the traccc custom scalar type is handed through to detray, so that it will be built in double precision when required. Apart from that, I removed a detray thrust related cmake variable, which doesn't exist anymore.
The Kalman Fit actor state was switched to use the non-owning `vecmem::device_vector` type always, since it just needs a view on the track states returned from the CKF and the detray detector no longer provides a host/device vector type publicly.

### DIFF
--- a/core/include/traccc/utils/seed_generator.hpp
+++ b/core/include/traccc/utils/seed_generator.hpp
@@ -13,12 +13,7 @@
 // detray include(s).
 #include <detray/geometry/barcode.hpp>
 #include <detray/geometry/tracking_surface.hpp>
-#include <detray/propagator/actor_chain.hpp>
-#include <detray/propagator/actors/aborters.hpp>
-#include <detray/propagator/actors/parameter_resetter.hpp>
-#include <detray/propagator/actors/parameter_transporter.hpp>
-#include <detray/propagator/actors/pointwise_material_interactor.hpp>
-#include <detray/propagator/base_actor.hpp>
+#include <detray/propagator/actors.hpp>
 #include <detray/propagator/propagator.hpp>
 
 // System include(s).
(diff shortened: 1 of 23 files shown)


extern/detray/CMakeLists.txt
```diff
@@ -13,13 +13,13 @@ message( STATUS "Building Detray as part of the TRACCC project" )
 
 # Declare where to get Detray from.
 set( TRACCC_DETRAY_SOURCE
-"URL;https://github.com/acts-project/detray/archive/refs/tags/v0.87.0.tar.gz;URL_MD5;0e30107edebfd41fbb3fc1974f665ea9"
+"GIT_REPOSITORY;https://github.com/niermann999/detray;GIT_TAG;ref-algebra-detail"
```
beomki-yeo, line 16: Typo?

niermann999: Not really. I would like to include that branch in the new detray version

krasznaa: So it's a draft PR, right...? Because of course we won't merge in this exact code. :thinking: Even a specific hash of your fork would be very iffy. But a branch name is of course absolutely a no-go.

niermann999: Sorry, I unmarked the draft, because it would otherwise not trigger the CUDA CI, which is what I am trying to fix on this detray branch

core/include/traccc/utils/seed_generator.hpp
```diff
@@ -76,8 +72,10 @@ struct seed_generator {
 
         for (std::size_t i = 0; i < e_bound_size; i++) {
 
-            bound_param[i] = std::normal_distribution<scalar>(
-                bound_param[i], m_stddevs[i])(m_generator);
+            if (m_stddevs[i] != scalar{0}) {
+                bound_param[i] = std::normal_distribution<scalar>(
+                    bound_param[i], m_stddevs[i])(m_generator);
+            }
```
stephenswat, line 78: Shouldn't this still set the parameter to the mean if the deviation is zero?

niermann999: Yes, but I think the mean is `bound_param[i]`, so in this case nothing needs to be done. Or am I reading this wrong?

stephenswat (approved):

PR #842, opened 2025-02-05
Change the default smoother to two-filters smoother

beomki-yeo:
This will fix the issue on bad matching rate between CPU and GPU, reported by @stephenswat and @krasznaa

### DIFF
--- a/core/include/traccc/fitting/fitting_config.hpp
+++ b/core/include/traccc/fitting/fitting_config.hpp
@@ -27,7 +27,7 @@ struct fitting_config {
         detray::muon<traccc::scalar>();
 
     /// Smoothing with backward filter
-    bool use_backward_filter = false;
+    bool use_backward_filter = true;
     traccc::scalar covariance_inflation_factor = 1e3f;
 };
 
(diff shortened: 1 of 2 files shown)


beomki-yeo:
Yeah it's not working for the drift chamber where the navigation can get unstable.
Fitting with wire chamber won't be OK until we implement (1) direct navigator and (2) deterministic annealing filter.

krasznaa (approved):
Okay... Let's treat this as a quick fix to the observed issue.

What I'll (or somebody else will) do is to introduce `traccc::opts::track_finding`, to make this (and the other track finding options) configurable from the command line. As it really should be...

beomki-yeo:
This kills the performance a bit but it is expected as propagation runs twice..

krasznaa:
> This kills the performance a bit but it is expected as propagation runs twice..

"A bit" may be a bit of an understatement unfortunately. :sob: Lately our code has slowed down a lot... Something to be discussed in tomorrow's meeting.

beomki-yeo:
It's a drop from correcting the mathematics so... take it as a necessary evil..

beomki-yeo:
And KF will be trivial once we implement the ambinsolver after CKF.

If i am correct, KF is used in refitting but it is not a part of ITk chain

PR #844, opened 2025-02-06
Record track quality to the CKF tracks

beomki-yeo:
track quality is the struct for the NDF, chi2 and the number of holes.
This has been recorded in the KF but we can also calculate it for the CKF.

This will allow us to apply the ambiguity resolution to the CKF output directly, and reduce the KF time significantly.


krasznaa (commented):
I'm generally on board. Code-wise I like how you did things.

But do you have some measurements to what it does to our runtime performance? I understand that the KF should become faster (not necessary?) after this. But do you have any quantitative numbers on that? 🤔

beomki-yeo:
>  I understand that the KF should become faster (not necessary?) after this.

No this PR is not for performance improvement. 
Instead, this PR will allow us moving the ambi-solver, which needs chi2 information, before the KF. Then, KF will fit less tracks.

niermann999:
> > I understand that the KF should become faster (not necessary?) after this.
> 
> No this PR is not for performance improvement. Instead, this PR will allow us moving the ambi-solver, which needs chi2 information, before the KF. Then, KF will fit less tracks.

Sorry if this is a stupid question, but do we have/plan to add a GPU ambi solver or is this a CPU-only development?

stephenswat:
> Sorry if this is a stupid question, but do we have/plan to add a GPU ambi solver or is this a CPU-only development?

If we end up running the ambiguity resolution between the CKF and the KF we will want to build a GPU ambiguity resolver.

beomki-yeo:
I will move the current cpu amisolver to between ckf and kf, and implement the gpu ambl-solver in a similar fasion

beomki-yeo:
Can I go with the PR? There are still some PRs left before I can move the ambiguity resolver between the CKF and KF.

stephenswat (approved):

PR #846, opened 2025-02-08
Inflate covariance ckf

beomki-yeo:
While working on #844, I found that I didn't add covariance inflation in the CKF. This PR adds it for CPU CUDA and SYCL implementations.

It seems thrust does not support the lambda expression by default so I defined a struct for covariance inflation to make a functor.
However, SYCL does not seem to work with a functor. For example, when I use the following  code:

```c++
    covariance_inflator cov_inf(config.covariance_inflation_factor);
    oneapi::dpl::for_each(policy, in_params_device.begin(),
                          in_params_device.end(), cov_inf);
```

I got the following error:

```
/opt/intel/oneapi/dpl/2022.7/include/oneapi/dpl/pstl/hetero/../hetero/dpcpp/parallel_backend_sycl.h:242:17: error: no matching function for call to object of type 'const oneapi::dpl::unseq_backend::walk_n<oneapi::dpl::execution::device_policy<> &, traccc::covariance_inflator>'
```

If I use the lambda expression for SYCL like in this PR, the compilation takes forever... any ideas?


examples/options/src/track_finding.cpp
```diff
@@ -66,9 +66,11 @@ track_finding::track_finding() : interface("Track Finding Options") {
         po::value(&m_config.max_num_skipping_per_cand)
             ->default_value(m_config.max_num_skipping_per_cand),
         "Maximum allowed number of skipped steps per candidate");
-    m_desc.add_options()("particle-hypothesis",
-                         po::value(&m_pdg_number)->default_value(m_pdg_number),
-                         "PDG number for the particle hypothesis");
```
stephenswat, line 71: Is there a a reason to remove this argument?

device/sycl/src/finding/find_tracks.hpp
```diff
@@ -172,6 +172,17 @@ track_candidate_container_types::buffer find_tracks(
...
+
+    // Inflate the covariance
+    bound_track_parameters_collection_types::device in_params_device(
+        in_params_buffer);
+    oneapi::dpl::for_each(
+        policy, in_params_device.begin(), in_params_device.end(),
+        [config](bound_track_parameters& params) {
+            covariance_inflator cov_inf(config.covariance_inflation_factor);
+            cov_inf(params);
+        });
```
stephenswat, line 184: Have you tried implementing this not using DPL but just as a SYCL kernel? Should be trivial.

examples/options/src/track_finding.cpp
beomki-yeo: Mistake - will fix it

beomki-yeo:
The increased covariance makes CKF pick up the measurements very far from the track. ($$\chi^2  \propto  distance/covariance$$). Eventually CKF will produce lots of fake tracks and increase the wall time significantly :thinking: 
I will leave this PR draft until I find a solution

stephenswat:
Would it help if we found measurements in a sorted way, i.e. we select the $N$ best measurement (ranked by $\chi^2$) instead of what we do now which is selecting $M$ random measurements within a $\chi^2$ cut; with $N < M$?

stephenswat:
Note this is something I have discussed as being useful with @paradajzblond as well.

beomki-yeo:
We already have `max_num_branches_per_surface` defined in finding_config.hpp, so it won't be very difficult. Maybe we should have implemented that already? 

However, I am afraid this approach still does not solve the issue from this PR. CKF will still make branches as many as $N_{max}$.

device/sycl/src/finding/find_tracks.hpp
beomki-yeo: Nah, just found that it is equally slow in the main branch.

stephenswat:
Closing this as superseded by #1167.

PR #848, opened 2025-02-10
Seed Parameter Estimation Fix, main branch (2025.02.10.)

krasznaa:
This is a follow-up about the issue discussed (amongst other places) in #842. That in `RelWithDebInfo` mode we would see the following output from `traccc_seq_example_cuda`:

```
[bash][Celeborn]:out > ./build/cuda-fp64/bin/traccc_seq_example_cuda --compare-with-cpu

Running Full Tracking Chain Using CUDA

Detector Options:
├ Detector file:                   geometries/odd/odd-detray_geometry_detray.json
├ Material file:                   geometries/odd/odd-detray_material_detray.json
├ Surface grid file:               geometries/odd/odd-detray_surface_grids_detray.json
├ Use detray detector:             true
└ Digitization file:               geometries/odd/odd-digi-geometric-config.json
...
WARNING: @traccc::io::csv::read_cells: 28 duplicate cells found in /home/krasznaa/ATLAS/projects/traccc/traccc/data/odd/geant4_10muon_10GeV/event000000000-cells.csv
===>>> Event 0 <<<===
Number of measurements: 631 (host), 631 (device)
  Matching rate(s):
    - 100% at 0.01% uncertainty
    - 100% at 0.1% uncertainty
    - 100% at 1% uncertainty
    - 100% at 5% uncertainty
Number of spacepoints: 451 (host), 451 (device)
  Matching rate(s):
    - 100% at 0.01% uncertainty
    - 100% at 0.1% uncertainty
    - 100% at 1% uncertainty
    - 100% at 5% uncertainty
Number of seeds: 181 (host), 181 (device)
  Matching rate(s):
    - 100% at 0.01% uncertainty
    - 100% at 0.1% uncertainty
    - 100% at 1% uncertainty
    - 100% at 5% uncertainty
Number of track parameters: 181 (host), 181 (device)
  Matching rate(s):
    - 0% at 0.01% uncertainty
    - 0% at 0.1% uncertainty
    - 0% at 1% uncertainty
    - 0% at 5% uncertainty
Number of track candidates (header): 183 (host), 183 (device)
  Matching rate(s):
    - 0% at 0.01% uncertainty
    - 0% at 0.1% uncertainty
    - 0% at 1% uncertainty
    - 0% at 5% uncertainty
  Track candidates (item) matching rate: 100%
Number of track states: 183 (host), 183 (device)
  Matching rate(s):
    - 98.9071% at 0.01% uncertainty
    - 98.9071% at 0.1% uncertainty
[...]

PR #849, opened 2025-02-10
Track Parameter CSV Writing, main branch (2025.02.10.)

krasznaa:
While debugging the issue fixed in #848, I wrote this code for dumping the properties of reconstructed bound track parameters.

While `detray::bound_track_parameters` has a print operator itself (which I'm very much on board with, and could've technically used for this debugging as well...), I thought this code may still be good to preserve. It could come in handy in some future debugging sessions as well. 🤔

### DIFF
--- a/io/CMakeLists.txt
+++ b/io/CMakeLists.txt
@@ -1,6 +1,6 @@
 # TRACCC library, part of the ACTS project (R&D line)
 #
-# (c) 2021-2024 CERN for the benefit of the ACTS project
+# (c) 2021-2025 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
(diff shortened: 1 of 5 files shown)


niermann999 (commented):
I must say, that I am not completely happy of doing free track param csv IO in detray (https://github.com/acts-project/detray/blob/main/io/include/detray/io/csv/track_parameters.hpp) and bound track param IO in traccc. I think, it would be better to keep the IO implementation of detray types bundled in one place. Sooner or later, somebody decides this is needed for some detray debugging as well, and nobody remembers that traccc has this implemented and then we have two versions...
Besides, detray io is already a traccc dependency, right? So, it should not be too difficult to extend/adapt the detray implementation of the csv IO and then adding an interface for the traccc container types?

krasznaa:
I was not aware of the CSV capabilities in Detray...

Sure, I think it's fair that such "pure Detray collections" would have I/O code in Detray. I'll see how to port this code into that project...

PR #857, opened 2025-02-11
Move alpaka vecmem logic to functions

StewMH:
Templated functions returning shared_ptrs that create `vecmem::hip::device_memory_resource` etc. Note that because of the casting to a derived class this has to be done via pointers.

@krasznaa think this might be along the lines of what you were thinking of?

Waiting for a substantial PR from @CrossR first


CrossR:
Does

https://github.com/CrossR/traccc/blob/CrossR/Unify/device/alpaka/include/traccc/alpaka/utils/get_vecmem_resource.hpp

and 

https://github.com/CrossR/traccc/blob/CrossR/Unify/device/alpaka/src/utils/get_vecmem_resource.cpp

Look closer to what you imagined @krasznaa?
There is a touch more logic required than your basic example, as we need to deal with SYCL requiring a queue wrapper input for most things (vs the others that don't), but I think it broadly matches what you were thinking.

Still need to move to more appropriate files / names, and see if I can think up a tidier way of dealing with the 3 cases (CUDA/HIP vs SYCL vs Threaded CPU, as CPU doesn't have an equivalent `deviceNativeQueue`, so fails to compile at the mo), but wanted to check this is more inline with what we were thinking, before I spend more time.

PR #859, opened 2025-02-12
Minor fixes to the track building kernel

stephenswat:
This commit makes some minor fixes and improvements to the track building kernel. In particular, it:

1. Fixes an error where we were comparing measurements indices to see if they were _larger_ than the total number of measurements, but they should be check also for _equality_.
2. Fail more gracefully if we cannot find any true measurements, i.e. register the failed track as a candidate for pruning.
3. Simplify the control flow of the kernel by removing an unnecessary break statement.

### DIFF
--- a/core/include/traccc/finding/details/find_tracks.hpp
+++ b/core/include/traccc/finding/details/find_tracks.hpp
@@ -362,7 +362,7 @@ track_candidate_container_types::host find_tracks(
         unsigned int n_skipped{0u};
         while (true) {
 
-            if (L.meas_idx > n_meas) {
+            if (L.meas_idx >= n_meas) {
                 n_skipped++;
             }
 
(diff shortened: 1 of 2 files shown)


beomki-yeo:
Nice catch. Could you fix the same bugs in CPU code: https://github.com/acts-project/traccc/blob/main/core/include/traccc/finding/details/find_tracks.hpp ?

stephenswat:
> Nice catch. Could you fix the same bugs in CPU code: https://github.com/acts-project/traccc/blob/main/core/include/traccc/finding/details/find_tracks.hpp ?

Ah yes, of course!

stephenswat:
There we go! :smile:

krasznaa (approved):
👍

stephenswat:
Clearly this warrants some more investigation... :thinking:

beomki-yeo:
I will update the branch by myself and merge it once the CI passes

PR #861, opened 2025-02-12
Return error state if PDG number is unknown

stephenswat:
Currently, we assume that any unknown PDG number is a muon, but this is obviously false. In order to remove this silent failure state, we should instead return some kind of invalid PDG particle. For this, we can use the same standard that ACTS uses, e.g. to use the number 0.

### DIFF
--- a/core/include/traccc/utils/particle.hpp
+++ b/core/include/traccc/utils/particle.hpp
@@ -41,7 +41,8 @@ particle_from_pdg_number(const int pdg_num) {
             return detray::pion_minus<scalar_t>();
     }
 
-    return detray::muon<scalar_t>();
+    // TODO: Replace with `detray::invalid` in the future
+    return detray::pdg_particle<scalar_t>(0, 0.f, 0.f);
 }
 
 // Apply the charge operator to return the antimatter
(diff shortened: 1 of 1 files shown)


core/include/traccc/utils/particle.hpp
```diff
@@ -41,7 +41,8 @@ particle_from_pdg_number(const int pdg_num) {
             return detray::pion_minus<scalar_t>();
     }
 
-    return detray::muon<scalar_t>();
+    // TODO: Replace with `detray::invalid` in the future
+    return detray::pdg_particle<scalar_t>(0, 0.f, 0.f);
```
krasznaa, line 45: Do we expect to reach this "legitimately"? I wonder if we should "error out" more aggressively.

Since this code may be used in both host and device code (Though I doubt that device code actually calls this. Does it?), throwing an exception is not on the table. But maybe something else?

krasznaa (commented):
I'm a little afraid of moving the point where the application encounters an error, very far from where we detect this error. 🤔 How do we expect applications to behave with an unknown particle?

beomki-yeo:
This makes sense to me if there is no particle with pdg number of 0 :thinking:

stephenswat:
> This makes sense to me if there is no particle with pdg number of 0 🤔

As far as I can tell there isn't: https://pdg.lbl.gov/2007/reviews/montecarlorpp.pdf

ACTS also uses 0 to represent invalid states:

https://github.com/acts-project/acts/blob/36cc0cb978e6e797f840ba2560bd18cd2daa92fc/Core/include/Acts/Definitions/PdgParticle.hpp#L17

beomki-yeo (approved):

Issue #858, opened 2025-02-12
RTS smoother silently fails on low-determinant matrices


stephenswat:
I've recently started running into a Detray assertion that fires when we try to set the $\phi$ value of  our track parameters at:

https://github.com/acts-project/traccc/blob/a54d64a38c6b22ae32a81308eed0b83c3c25c8ca/core/include/traccc/edm/track_parameters.hpp#L48

After some digging, I found out that the cause of this is that the $\phi$ value turns out to be NaN, and that NaN is produced because we are, in the gain matrix smoother, trying to invert a matrix:

https://github.com/acts-project/traccc/blob/a54d64a38c6b22ae32a81308eed0b83c3c25c8ca/core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp#L91-L94

It turns out that the pathological matrix is the following (found as the first state on the first fitting step):

```
regularized_predicted_cov:
[  0.00019195,   0.00005403,  -0.00014609,  -0.00000383,   0.00095228,   0.00000000]
[ -0.00082664,  -0.00024949,  -0.00001079,   0.00002236,   0.00012880,   0.00000000]
[ -0.00014095,  -0.00003770,   0.00121523,   0.00001438,  -0.00829925,   0.00000000]
[  0.00000147,   0.00000009,   0.00001426,   0.00000042,   0.00000299,   0.00000000]
[  0.00095218,   0.00012900,  -0.00829925,   0.00000299,   0.09977844,   0.00000000]
[  0.00000000,   0.00000000,   0.00000000,   0.00000000,   0.00000000, 89875512.000]
```

There are some worrying things about this matrix, namely:

1. ~~The very small covariance values~~
2. The very large time variance value
3. The fact that the determinant of this matrix is $-8.3 \times 10^{-14}$
4. The non-symmetric covariance values

I am fairly convinced that we are seeing these NaN values because the determinant is so close to zero that our single-precision calculation is simply producing NaN values here.

We will need to either:

1. Figure out the root cause for this odd covariance matrix and fix it; or
2. Make sure that we gracefully exit the fitter if we fail to invert.

beomki-yeo:
Small covariance value is not a worry. Non-symmetric offdiagonal elements are worrysome though.

[...]

PR #862, opened 2025-02-12
Add failure propagation to Kálmán filter code

stephenswat:
This commit enables our Kálmán filter code to fail, e.g. it allows the smoothers and updaters to return a success value, and that success value is returned to the outside algorithm. This allows us to elegantly handle edge cases like those described in #858, as well as cases where tracks propagate exactly along the z-axis. Some of this code taken from #781, thanks @beomki-yeo.

### DIFF
--- a/core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
+++ b/core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
@@ -11,6 +11,7 @@
 #include "traccc/definitions/qualifiers.hpp"
 #include "traccc/edm/track_parameters.hpp"
 #include "traccc/edm/track_state.hpp"
+#include "traccc/fitting/status_codes.hpp"
 
 // Detray inlcude(s)
 #include <detray/geometry/shapes/line.hpp>
@@ -43,7 +44,7 @@ struct gain_matrix_smoother {
     ///
     /// @return true if the update succeeds
     template <typename mask_group_t, typename index_t>
-    TRACCC_HOST_DEVICE inline void operator()(
+    TRACCC_HOST_DEVICE [[nodiscard]] inline kalman_fitter_status operator()(
         const mask_group_t& /*mask_group*/, const index_t& /*index*/,
         track_state<algebra_t>& cur_state,
         const track_state<algebra_t>& next_state) {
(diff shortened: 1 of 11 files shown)


core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
```diff
@@ -102,6 +104,14 @@ struct gain_matrix_smoother {
         cur_state.smoothed().set_vector(smt_vec);
         cur_state.smoothed().set_covariance(smt_cov);
 
+        // Return false if track is parallel to z-axis or phi is not finite
+        const scalar theta = cur_state.smoothed().theta();
+        if (theta <= 0.f || theta >= constant<traccc::scalar>::pi ||
+            !std::isfinite(cur_state.smoothed().phi()) ||
+            std::abs(cur_state.smoothed().qop()) == 0.f) {
```
beomki-yeo, line 111: When do you hit the condition of `qop = 0`? I observed this from the buggy behavior of smoother. Once I fix the bug, this may not be observed

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
```diff
@@ -244,6 +252,14 @@ class kalman_fitter {
                 detray::navigation::direction::e_backward);
             fitter_state.m_fit_actor_state.backward_mode = true;
 
+            if (propagation._stepping().dir()[0] == 0.f ||
```
beomki-yeo, line 255: && instead of ||? But when do you observe this exactly?

beomki-yeo (commented):
Looks good but I would like to understand the situation as I am also debugging the smoother

core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
stephenswat: Ah, it is very possible that your fix will solve this problem, too!

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
stephenswat: Sorry this should read:

```cpp
if (vector::theta(propagation._stepping().dir()) == 0.f) {
```

Reason for this is that tracks with theta 0 will produce division by zero errors in detray:

https://github.com/acts-project/detray/blob/2d6be9366462218e52723a2ac90dcd39a0e9812c/core/include/detray/propagator/detail/jacobian_engine.hpp#L60

beomki-yeo: I think this should include $\pi$:
```
if (vector::theta(propagation._stepping().dir()) == 0.f || vector::theta(propagation._stepping().dir()) == pi) { 
```
But isn't the following better because trigonometric used in `vector::theta` might be expensive?:
```
if (propagation._stepping().dir()[0] ==0 && propagation._stepping().dir()[1] ==0){
}
```

core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
beomki-yeo: Yeah but if you are having an issue from `qop=0` right know, I am onboard with it

core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
stephenswat: Updated.

beomki-yeo:
It seems the smoother used in the wire chamber fitting is failing - I think It is OK to change the requirement to
`ASSERT_NEAR(n_tracks, n_truth_tracks, n_truth_tracks * 0.02)` to pass the test

krasznaa (commented):
You guys will have to sort out the technical details, but just to say, I'm very on board with these code changes. 😉

PR #863, opened 2025-02-12
Fix the bug in getting the first smoother state

beomki-yeo:
A little bit critial bug.. when the measurement of the last track state is not found during the forward filtering

After merging #862, we will need to `return false` in `smoothe` if all track states are holes.


stephenswat (approved):
Looks good!

PR #867, opened 2025-02-13
ref: Harmonize track parameters types

niermann999:
Use the traccc track parameter types everywhere and allow templating of the algebra type


beomki-yeo (commented):
Do we still want to allow other algebras? @stephenswat @krasznaa 
TBH, I would like to focus on the development of default algebra

krasznaa:
> Do we still want to allow other algebras? @stephenswat @krasznaa TBH, I would like to focus on the development of default algebra

It's not a black or white thing in my eyes. But in end effect, I believe we should be going this way, yes. I'm very much on board with making it possible to use different algebras in the code.

With that being said, the final code, in 1-2 year's time, may not use quote so much templating. But for now, I believe this is the way to go. 🤔

stephenswat (approved):

PR #868, opened 2025-02-13
Divide default propagation options by proper units

beomki-yeo:
Currently overstep tolerenace is set to  -300 um (detray default) * um = -0.3 um, which makes propagation unstable


beomki-yeo:
@stephenswat It seems that we were shadow-boxing recently with too small overstep tolerance :/ ...

niermann999 (approved):

stephenswat:
Thanks for this PR, but please do note that this change breaks the idea that the defaults for the CLI options are taken from the defaults for the configuration objects. Now we are back to a scenario where the CLI defaults are different; I'll aim to fix this in a new PR today.

stephenswat:
Never mind the aforementioned is incorrect and all is well. :+1:

PR #869, opened 2025-02-14
Determine succesfully fitted tracks

beomki-yeo:
Now depends on #881

-------------------------------------------------
This PR adds a boolean variable in `fitting_result` to determine whether the fitting was successful. This should be harmonized with #862 as well

Success conditions:
1. NDF >0
2. All non-hole track states should be smoothed (i.e. In case of two-filters smoother, forward and backward filter need to pass the same modules.)

The second condition will be much easier to be met once we implement the direct navigator.

Some plots about fitting efficiency will follow soon...

----------------------------------------------

Here is the plot of fitting efficiency (Number of successfully fitted track / Total number of tracks) as a function of mask tolerance. 

Fitting efficiency drops a lot for momentum < 10 GeV/c, which is expected due to the larger material effect + unstable navigation.

- Firx overstep tolerance = default value (-300 um)

- Fixed mask tolerance

Following is the script used:

<details>

<summary>Script</summary>

```sh
#!/bin/bash
n_ptc=2000
mom_array=(1 2 5 10 100)
tol_array=(0 1 2 3 4)

for mom in "${mom_array[@]}"
do

    command_simulate="
    ./bin/traccc_simulate \
    --output-directory=odd/simplified_${n_ptc}muon_${mom}GeV/ \
    --gen-events=1 \
    --gen-nparticles=${n_ptc} \
    --gen-mom-gev=${mom}:${mom} \
    --gen-eta=-1:1 \
    --rk-tolerance=1e-8
    "

    ${command_simulate}

    for tol in "${tol_array[@]}"
    do

        echo "Mom setup: ${mom}, Tol setup: ${tol}"

        command_truth_fit="
        ./bin/traccc_truth_fitting_example \
        --use-detray-detector=true \
        --use-acts-geom-source=false \
        --input-directory=odd/simplified_${n_ptc}muon_${mom}GeV/ \
        --input-events=1 \
        --min-mask-tolerance=${tol} \
        --max-mask-tolerance=${tol} \
        --check-performance 
        "

        ${command_truth_fit}

        mv performance_track_fitting.root performance_track_fitting_${mom}_${tol}.root
    done
done
```

</details>

@niermann999 advised that mask tolerances of the first figure should not be touched. I also think the mask tolerance won't be even an issue once we implement the direct navigator. However, the current default overstep tolerance of (-300 um) may be too small for small momentum particle. It would be nice to make it larger or adaptive for different values of momentum.

### DIFF
--- a/performance/include/traccc/resolution/fitting_performance_writer.hpp
+++ b/performance/include/traccc/resolution/fitting_performance_writer.hpp
@@ -61,6 +61,17 @@ class fitting_performance_writer {
         static_assert(std::same_as<typename detector_t::algebra_type,
                                    traccc::default_algebra>);
 
+        if (fit_res.fit_outcome != fitter_outcome::SUCCESS) {
+            return;
+        }
+
+        // Get the first smoothed track state
+        const auto& trk_state = *std::find_if(
+            track_states_per_track.begin(), track_states_per_track.end(),
+            [](const auto& st) { return st.is_smoothed; });
+        assert(!trk_state.is_hole);
+        assert(trk_state.is_smoothed);
+
         std::map<measurement, std::map<particle, std::size_t>> meas_to_ptc_map;
         std::map<measurement, std::pair<point3, point3>> meas_to_param_map;
 
@@ -72,8 +83,6 @@ class fitting_performance_writer {
             meas_to_param_map = evt_data.m_meas_to_param_map;
         }
 
-        // Get the track state at the first surface
-        const auto& trk_state = track_states_per_track[0];
         const measurement meas = trk_state.get_measurement();
 
         // Find the contributing particle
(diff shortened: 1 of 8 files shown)


beomki-yeo:
@niermann999 I don't know how to set the min/max mask tolerance so I just set them to the same value. Let me know if it has to be changed.

niermann999:
The maximum mask tolerance should not be made smaller than it is currently, I believe (making it larger should be fine, as it determines the mask tolerance for surfaces that are far away from the track). The minimum mask tolerance is effectively what we require when the track is already on the surface. Not sure that should be much looser than it is (which corresponds to the propagation position error that was found in the ATLAS paper, if I remember correctly). The two tolerances should however not be set to the same value imo. 
The scale factor, that sets the mask tolerances in between the min and max values depending on the distance to the track is probably what we should be looking at.

beomki-yeo:
@stephenswat I made a new enum class of `fitting_outcome` which is recorded in the `fitting_result`. I wanted to rather record the unsuccessfully fitted tracks unless it is not an error categorized in `fitting_status`.

performance/include/traccc/resolution/fitting_performance_writer.hpp
```diff
@@ -103,7 +112,9 @@ class fitting_performance_writer {
         if (fit_res.ndf > 0 && !trk_state.is_hole) {
             write_res(truth_param, trk_state.smoothed(), ptc);
         }
-        write_stat(fit_res, track_states_per_track);
+        if (fit_res.fit_outcome == fitter_outcome::SUCCESS) {
```
stephenswat, line 115: I guess this check is technically not necessary because you return if the outcome was not a success on lines 61-63.

stephenswat (approved):
LGTM.

PR #871, opened 2025-02-17
Add pre-commit job to check quote includes

stephenswat:
PR #865 moved our detray includes from quotes to angle brackets, which I like a lot. This commit adds a new pre-commit job to check that this doesn't regress, and makes the same change for our other dependencies.

### DIFF
--- a/.github/check_quote_includes.sh
+++ b/.github/check_quote_includes.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+if ! [ -x "$(command -v rg)" ]; then
+   GREP="grep -R"
+else
+   GREP="rg -j 1 --no-heading -N"
+fi
+
+${GREP} "#include \"vecmem" "$@" ; test $? -eq 1 || exit 1
+${GREP} "#include \"detray" "$@" ; test $? -eq 1 || exit 1
+${GREP} "#include \"Acts" "$@" ; test $? -eq 1 || exit 1
+${GREP} "#include \"covfie" "$@" ; test $? -eq 1 || exit 1
+${GREP} "#include \"algebra" "$@" ; test $? -eq 1 || exit 1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,3 +5,10 @@ repos:
       - id: clang-format
         types_or: [file]
         files: \.(cpp|hpp|ipp|cu|cuh|sycl|hip)$
+
+  - repo: local
+    hooks:
+      - id: check_quote_includes
+        name: Check includes with quotes
+        language: system
+        entry: .github/check_quote_includes.sh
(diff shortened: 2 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -5,3 +5,10 @@ repos:
       - id: clang-format
         types_or: [file]
         files: \.(cpp|hpp|ipp|cu|cuh|sycl|hip)$
+
+  - repo: local
+    hooks:
+    - id: check_quote_includes
+      name: Check includes with quotes
+      language: system
+      entry: .github/check_quote_includes.sh
```
krasznaa, line 14: It might just be my browser, but is the indentation on this block different from lines 5-7?

.github/check_quote_includes.sh
```diff
@@ -0,0 +1,13 @@
...
+   GREP="grep -R"
+else
+   GREP="rg -j 1 --no-heading -N"
+fi
+
+${GREP} "#include \"vecmem" "$@" ; test $? -eq 1 || exit 1
+${GREP} "#include \"detray" "$@" ; test $? -eq 1 || exit 1
+${GREP} "#include \"Acts" "$@" ; test $? -eq 1 || exit 1
+${GREP} "#include \"covfie" "$@" ; test $? -eq 1 || exit 1
+${GREP} "#include \"algebra" "$@" ; test $? -eq 1 || exit 1
```
krasznaa, line 13: 🤔 I think you're right, this is probably the most reasonable way to implement it. Technically it would be better to do this in a way that checks whether a file is part of this repository (to allow local subdirectories called `"vecmem"` for instance), but it's not worth putting that much effort into this. 🤔

krasznaa (commented):
Apart from that stylistic issue with indentation, I'm on board. This is my coding preference as well.

I wonder if there may be a possibility of checking include orders as well (I prefer going from the "more local" to the "more generic" ones), but that's not for this PR.

.github/check_quote_includes.sh
stephenswat: Mind you pre-commit does this for us already, so only files which we want to check end up in `$@`.

stephenswat:
> I wonder if there may be a possibility of checking include orders as well (I prefer going from the "more local" to the "more generic" ones), but that's not for this PR.

Include order is checked via clang-format!

.pre-commit-config.yaml
stephenswat: Fixed.

.github/check_quote_includes.sh
stephenswat: (See https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includecategories)

krasznaa: As discussed, I meant a slightly different thing here. But the functionality in clang-format I didn't know about, that was a good pointer.

krasznaa (approved):
👍

PR #872, opened 2025-02-17
ref: Measure wall time in multithreaded benchmark and reduce number of events

niermann999:
To get an accurate result in a multithreaded benchmark, measure the wall clock time.
Until we find out, why the benchmark produces a segmentation fault, I have reduced the number of events, so that it at least runs. At the same time, I also increased the number of tracks per event to simulate higher pileup

### DIFF
--- a/benchmarks/common/benchmarks/toy_detector_benchmark.hpp
+++ b/benchmarks/common/benchmarks/toy_detector_benchmark.hpp
@@ -19,11 +19,7 @@
 
 // Detray include(s).
 #include <detray/detectors/bfield.hpp>
-#include <detray/io/frontend/detector_reader.hpp>
 #include <detray/io/frontend/detector_writer.hpp>
-#include <detray/navigation/navigator.hpp>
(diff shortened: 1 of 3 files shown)


beomki-yeo:
Let's see if #862 will solve the current issue with benchmark. If it does, we don't have to change the setup

beomki-yeo (approved):

beomki-yeo:
The benchmark still fails.. have you tested it?
https://gitlab.cern.ch/acts/traccc-benchmark/-/jobs/51326838

niermann999:
> The benchmark still fails.. have you tested it? https://gitlab.cern.ch/acts/traccc-benchmark/-/jobs/51326838

Hm, yes, I tried the CPU benchmark again just now and on my local machine its still fine...

beomki-yeo:
It's also seg fault in my setup

niermann999:
Sounds like undefined behavior. I am trying a debug run now, but that will take some time

PR #874, opened 2025-02-18
Manage include order via clang-format

stephenswat:
Currently we manage the grouping of include files via a home-grown system of comments interspersed between includes, but this looks messy and is subject to user error. clang-format contains functionality for grouping and sorting includes automatically, which this commit moves towards.


krasznaa (commented):
As we discussed the other day, I generally like the idea of automating this.

But let's put this on ice for now. This update literally touches every single file in the repository. Messing with everyone's ongoing developments.

On the timescale on which we should be bothering with this level of organization, we'll be moving the code into Acts anyway.

So let's just not do this. We have much bigger issues to sort out at the moment.

stephenswat:
Closing, this will happen during the integration into ACTS.

PR #876, opened 2025-02-19
ref: Remove detray unit types

niermann999:
Remove some more detray units and set the rk-tolerance to mm, which internally it already seems to be


examples/options/src/track_propagation.cpp
```diff
@@ -28,34 +28,34 @@ track_propagation::track_propagation()
...
         "search-window",
         po::value(&m_search_window)->default_value(m_search_window),
         "Size of the grid surface search window");
-    // TODO: Use the unit of [mm] for rk tolerance
-    m_desc.add_options()("rk-tolerance",
+    m_desc.add_options()("rk-tolerance [mm]",
                          po::value(&(m_config.stepping.rk_error_tol))
-                             ->default_value(m_config.stepping.rk_error_tol),
+                             ->default_value(m_config.stepping.rk_error_tol /
+                                             traccc::unit<float>::mm),
```
beomki-yeo, line 58: You also need to multiply `mm` in the `read` function below

niermann999: Absolutely! Done

krasznaa (approved):
Looks good to me... 🤔

stephenswat:
A pre-commit check to ensure we don't re-introduce these detray namespaces would be nice... :smile:

PR #877, opened 2025-02-19
Add B-field conversion and generation tools

stephenswat:
This commit adds two new tools which both create magnetic field files for us. The `convert_csv_bfield` tool takes a space-separated magnetic field like the one we have of the ATLAS field and converts it to a binary covfie file. The `generate_constant_bfield` executable generates 2T magnetic fields at different sizes, either with 201x201x301 field points like the ATLAS field, or with 2x2x2 field points.


examples/tools/bfield_type.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <covfie/core/algebra/affine.hpp>
+#include <covfie/core/backend/primitive/array.hpp>
+#include <covfie/core/backend/transformer/affine.hpp>
+#include <covfie/core/backend/transformer/nearest_neighbour.hpp>
+#include <covfie/core/backend/transformer/strided.hpp>
+#include <covfie/core/field.hpp>
+#include <covfie/core/parameter_pack.hpp>
+
+// TODO: Put these types in traccc core.
+using backend_t = covfie::backend::affine<covfie::backend::nearest_neighbour<
```
beomki-yeo, line 17: Can we use the linear approximation instead of NN? We can include NN after we confirm that physics performance stays good

examples/tools/convert_csv_bfield.cpp
```diff
@@ -0,0 +1,284 @@
...
+            Bx = static_cast<float>(std::atof(word.c_str()));
+            std::getline(ss, word, ' ');
+            CHECK_IOSTATE(ss);
+            By = static_cast<float>(std::atof(word.c_str()));
+            std::getline(ss, word);
+            Bz = static_cast<float>(std::atof(word.c_str()));
+
+            field_t::view_t::output_t& p = fv.at(xp, yp, zp);
+
+            p[0] = Bx * 0.000299792458f;
```
beomki-yeo, line 238: You can use `traccc::unit<scalar>::T` instead of hardcoded value

examples/tools/bfield_type.hpp
stephenswat: We can use the nearest-neighbour to make building the b-field easier, but we can then convert it to linear interpolation afterwards. But you are right the type that lives in the core should not be this NN type.

examples/tools/convert_csv_bfield.cpp
stephenswat: Fixed, thanks!

examples/tools/bfield_type.hpp
stephenswat: I removed the comment saying this should go into the core lib; we can use a linearly interpolated type in core.

beomki-yeo (approved):

krasznaa (commented):
These tools would be expected to live for a while, no? Should we maybe not put them under `examples/`, but into the main directory instead? 🤔 Though I'm not sure myself what would be better.

Other than that, sure, let's add the helper executables!

PR #878, opened 2025-02-19
SoA Spacepoints, main branch (2025.02.19.)

krasznaa:
With all its problems, let me open this monster PR, so that all of you could have a first look.

As I've been advertising for a while now, I was working on making `traccc::spacepoint` (and `traccc::seed`) have an SoA memory layout. Using the same `vecmem::edm` code that I pushed in for the cells as well.

To jump right to the conclusion, the code is currently a little slower with the SoA layout. 😭 Even though I would've dreamt of some cheap performance gains with all this effort... But let's come back to this further down.

Some of the main points in this PR:
  - Introduced `traccc::edm::spacepoint<t>`, `traccc::edm::spacepoint_collection`, `traccc::edm::seed<T>` and `traccc::edm::seed_collection`;
    * Spacepoints no longer have a `traccc::measurement` member, but rather just an integer with the index of the measurement that the spacepoint was made out of.
    * This could later on be extended so that the spacepoint would point at two strip measurements. In case we end up writing such a thing.
  - Removed `traccc::internal_spacepoint` completely, and modified the spacepoint grid type to:
    * Follow the same naming scheme that we use for all our containers. I.e. `traccc::details::spacepoint_grid_types::host`, `traccc::details::spacepoint_grid_types::buffer`, etc.
    * I used the `traccc::details` namespace for it, since regular users should not really need to interact with such objects directly.
    * The grid now uses `unsigned int` elements, with the indices of the spacepoints that belong to a given bin in the grid.
  - I demoted all of the `traccc::<foo>::spacepoint_binning` and `traccc::<foo>::seed_finding` classes from "being algorithms", to just being some helper tools that the main `traccc::<foo>::seeding_algorithm` classes would use.
    * I also moved all of them into `traccc::<foo>::details::` namespaces, to show that these would not be the main interfaces to use in client code. (I considered completely hiding these classes into the `src/` directories, but in the end decided against that.)
  - Updated all of the I/O code to the new classes;
  - Updated `traccc::performance` to be able to compare SoA containers to each other.
    * This is done through a new `traccc::soa_comparator` class, which uses the same machinery as the existing `traccc::collection_comparator`. (I tried to extend the existing class for a bit to make it compatible with both AoS and SoA types, but it was just not practical to do.)
  - Updated all of the tests and examples to use the new EDM types.

Coming back to the performance... On our trusty old A5000, at $\mu$=140, we go from:

```
 Time (%)  Total Time (ns)  Instances    Avg (ns)       Med (ns)      Min (ns)     Max (ns)    StdDev (ns)                                                   Name
 --------  ---------------  ---------  -------------  -------------  -----------  -----------  ------------  ----------------------------------------------------------------------------------------------------
...
      1.5      470,490,231        110    4,277,183.9    3,780,795.5    1,370,243   14,717,057   2,201,624.2  traccc::cuda::kernels::count_triplets(traccc::seedfinder_config, detray::const_grid2_view<detray::g…
...
      0.7      230,992,739        110    2,099,934.0    1,630,692.5      527,521   10,839,928   1,571,123.8  traccc::cuda::kernels::find_doublets(traccc::seedfinder_config, detray::const_grid2_view<detray::gr…
      0.5      171,477,070        110    1,558,882.5    1,158,307.5      345,345   19,859,050   2,248,613.7  traccc::cuda::kernels::find_triplets(traccc::seedfinder_config, traccc::seedfilter_config, detray::…
...
      0.3       89,031,135        110      809,374.0      602,642.0      214,465    4,017,866     655,702.0  traccc::cuda::kernels::count_doublets(traccc::seedfinder_config, detray::const_grid2_view<detray::g…
...
      0.1       34,684,545        110      315,314.0      224,912.5       93,952    1,779,783     312,483.6  traccc::cuda::kernels::select_seeds(traccc::seedfilter_config, vecmem::data::vector_view<const trac…
...
      0.1       16,780,777        110      152,552.5      115,008.5       21,888      829,538     145,592.3  traccc::cuda::kernels::estimate_track_params(vecmem::data::vector_view<const traccc::spacepoint>, v…
...
      0.0       12,371,711        110      112,470.1       59,296.5       19,552      528,097     109,092.5  traccc::cuda::kernels::update_triplet_weights(traccc::seedfilter_config, detray::const_grid2_view<d…
      0.0        8,864,118        110       80,582.9       45,296.0       22,720      368,385      69,339.7  void traccc::cuda::kernels::form_spacepoints<detray::detector<detray::default_metadata<algebra::plu…
 ...
      0.0        4,239,799        110       38,543.6       29,376.0       19,456      213,505      26,191.3  traccc::cuda::kernels::populate_grid(traccc::seedfinder_config, vecmem::data::vector_view<const tra…
      0.0        3,109,098        110       28,264.5       23,200.0       14,975      106,528      14,777.9  traccc::cuda::kernels::count_grid_capacities(traccc::seedfinder_config, detray::axis2::circular<std…
...
      0.0        2,480,457        110       22,549.6       11,600.0        7,360      113,793      18,672.3  traccc::cuda::kernels::reduce_triplet_counts(vecmem::data::vector_view<const traccc::device::double…
...
```

To:

```
 Time (%)  Total Time (ns)  Instances    Avg (ns)       Med (ns)      Min (ns)     Max (ns)    StdDev (ns)                                                   Name
 --------  ---------------  ---------  -------------  -------------  -----------  -----------  ------------  ----------------------------------------------------------------------------------------------------
...
      1.8      601,845,107        110    5,471,319.2    4,951,645.5    1,737,715   17,843,903   2,667,566.6  traccc::cuda::kernels::count_triplets(traccc::seedfinder_config, vecmem::edm::view<vecmem::edm::sch…
      1.1      374,323,769        110    3,402,943.4    2,881,532.5    1,434,646   19,223,364   2,234,771.4  traccc::cuda::kernels::find_doublets(traccc::seedfinder_config, vecmem::edm::view<vecmem::edm::sche…
      0.8      272,225,133        110    2,474,773.9    1,712,515.5      589,308   23,448,033   3,162,311.5  traccc::cuda::kernels::find_triplets(traccc::seedfinder_config, traccc::seedfilter_config, vecmem::…
...
      0.6      204,124,161        110    1,855,674.2    1,549,077.0      489,917    5,415,800   1,074,744.7  traccc::cuda::kernels::count_doublets(traccc::seedfinder_config, vecmem::edm::view<vecmem::edm::sch…
...
      0.1       45,660,333        110      415,093.9      262,094.0      128,127    2,542,445     458,869.7  traccc::cuda::kernels::select_seeds(traccc::seedfilter_config, vecmem::edm::view<vecmem::edm::schem…
...
      0.1       18,027,751        110      163,888.6      156,639.0       22,016      730,811     125,213.4  traccc::cuda::kernels::estimate_track_params(vecmem::data::vector_view<const traccc::measurement>, …
...
      0.1       17,326,347        110      157,512.2       60,160.0       26,496    3,417,801     354,702.6  traccc::cuda::kernels::update_triplet_weights(traccc::seedfilter_config, vecmem::edm::view<vecmem::…
...
      0.0        5,860,878        110       53,280.7       26,735.5       10,848      216,286      51,295.1  void traccc::cuda::kernels::form_spacepoints<detray::detector<detray::default_metadata<algebra::plu…
...
      0.0        2,855,926        110       25,963.0       20,448.0       13,920      135,007      16,604.9  traccc::cuda::kernels::populate_grid(traccc::seedfinder_config, vecmem::edm::view<vecmem::edm::sche…
      0.0        2,389,644        110       21,724.0       19,103.5       13,600       62,591       9,177.2  traccc::cuda::kernels::count_grid_capacities(traccc::seedfinder_config, detray::axis2::circular<std…
...
```

The spacepoint grid creation speeds up a little, but everything else slows down. 😦 We discussed a little about this with @stephenswat already, I'll try some tricks still. But it seems to make it pretty clear that just blindly switching to an SoA layout is not a silver bullet. As the memory throughput of the affected kernels clearly went down with the current state of this PR. 🤔


tests/cpu/test_seeding.cpp
```diff
@@ -41,27 +57,30 @@ TEST(seeding, case1) {
...
+    measurement_collection_types::host measurements(&host_mr);
+    edm::spacepoint_collection::host spacepoints{host_mr};
 
     // Spacepoints from 16.62 GeV muon
-    spacepoints.push_back({{36.6706f, 10.6472f, 104.131f}, {}});
-    spacepoints.push_back({{94.2191f, 29.6699f, 113.628f}, {}});
-    spacepoints.push_back({{149.805f, 47.9518f, 122.979f}, {}});
-    spacepoints.push_back({{218.514f, 70.3049f, 134.029f}, {}});
-    spacepoints.push_back({{275.359f, 88.668f, 143.378f}, {}});
+    add_spacepoint(measurements, spacepoints, {36.6706f, 10.6472f, 104.131f});
```
beomki-yeo, line 67: If you want to push the idea of SoA for all EDMs, could you come up with a user-friendly interface in adding elements?  

For example, can we just keep the command the same as before?: 
`spacepoints.push_back({{36.6706f, 10.6472f, 104.131f}, {}});`

What I heard in the past is that something called a proxy enables this kind of interface.

core/include/traccc/edm/spacepoint.hpp
```diff
@@ -1,73 +0,0 @@
-/** TRACCC library, part of the ACTS project (R&D line)
```
beomki-yeo, line 1: You can use AoS EDM type to keep the original command in adding elements. There is no need to kill them.

tests/cpu/test_seeding.cpp
krasznaa: Indeed, this has been on my radar since a while. 🤔 Not sure yet exactly how to do it, but template magic should be able to reach that far... 🤔

Note though that in traccc this is mostly an inconvenience in the testing code. That's why it wasn't the highest priority for me so far. Because the core libraries don't need this functionality all that much.

But yes, I do want to make it nicer in the future. 👍

core/include/traccc/edm/spacepoint.hpp
krasznaa: 🤔 No, I think keeping AoS and SoA classes side by side would be very confusing. It would obfuscate even more what is used where.

There should be a way of doing this sort of a thing in [vecmem::edm::host](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/edm/host.hpp) and [vecmem::edm::device](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/edm/device.hpp). 🤔

tests/cpu/test_seeding.cpp
beomki-yeo: The reason that this is mostly an inconvenience in the testing code is that our example codes are mostly running on the full chain. However, users might need to use only the part of algorithms depending on the experiments or their situations. IO or examples directory is not very recommended to be used by arbitrary experiments as well.

You can find tutorial codes to see what I expect with `traccc::core` application: https://github.com/beomki-yeo/traccc-tutorial/blob/main/tutorials/seeding.cpp

krasznaa:
Detailed performance tests to come on Monday. But with the modified `traccc::edm::spacepoint<T>` class I now get the following.
  - With the current `main` branch I get the following on my home PC:

```
./build/bin/traccc_throughput_mt_cuda --input-directory /data/hdd-4tb/acts_data/odd-20240509/geant4_ttbar_mu140/ --input-events 100 --cpu-threads 2 --deterministic

Running Multi-threaded CUDA GPU throughput tests
...
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Reconstructed track parameters: 4766973
Time totals:
                  File reading  3325 ms
            Warm-up processing  2421 ms
              Event processing  21890 ms
Throughput:
            Warm-up processing  242.162 ms/event, 4.12946 events/s
              Event processing  218.906 ms/event, 4.56816 events/s
```

  - And with this PR's code I see:

```
./build/cuda-native-fp32/bin/traccc_throughput_mt_cuda --input-directory /data/hdd-4tb/acts_data/odd-20240509/geant4_ttbar_mu140/ --input-events 100 --cpu-threads 2 --deterministic

Running Multi-threaded CUDA GPU throughput tests
...
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Reconstructed track parameters: 4767233
Time totals:
                  File reading  3280 ms
            Warm-up processing  2426 ms
              Event processing  21936 ms
Throughput:
            Warm-up processing  242.633 ms/event, 4.12145 events/s
              Event processing  219.37 ms/event, 4.55851 events/s
```

But don't read too much into it, I've seen fairly varied results. (The GPU is also doing 2x 4K graphics during all of this... :stuck_out_tongue:) At least now they're in the same ballpark... :thinking:

stephenswat:
> But with the modified `traccc::edm::spacepoint<T>` class I now get the following.

I assume this means the global coordinates are stored as an array of structs rather than a struct of arrays?

krasznaa:
> I assume this means the global coordinates are stored as an array of structs rather than a struct of arrays?

? Have a look! I pushed the changes before posting the comment.

Yes, I switched back to using `traccc::point3` as-is. I also believe I understand where the remaining performance difference comes from. We can discuss about it in a little bit...

stephenswat:
> ? Have a look! I pushed the changes before posting the comment.

image

:stuck_out_tongue:

krasznaa:
Worse than that, the changes don't even show up as I imagined. I might have messed up with the push after all... 🤔

In the end the new `spacepoint_collection.hpp` file should be the important part. Everything else is just there to support that new thing.

stephenswat:
> I also believe I understand where the remaining performance difference comes from.

I think the remainder is about 0.01 Hz, right? I think that's an acceptable change, doesn't warrant too much investigation as far as I'm concerned.

krasznaa:
I might as well spoil the surprise. 😛

I kept the variance variables from `internal_spacepoint`.

https://github.com/acts-project/traccc/blob/main/core/include/traccc/edm/internal_spacepoint.hpp#L80-L84

The ones that have **always** been hardcoded to zero... Now they are stored as actual variables in the SoA. With their values always set to zero. 🤔 This of course would introduce some additional I/O.

So we'll have to see how much we actually want these variances in the seed finding. At least for the moment. 🤔

krasznaa:
So... let's ignore the build failure for a moment...

The current code of the PR, on the same A5000 card as before, behaves like this:

```
nsys profile --stats=true -o seeding_new2.profile ./build-new/bin/traccc_throughput_mt_cuda --input-directory /data/Acts/odd-simulations-20240509/geant4_ttbar_mu140/ --input-events 100 --cpu-threads 2 --deterministic 2>&1 | tee seeding_new2.profile.log
...
 Time (%)  Total Time (ns)  Instances    Avg (ns)       Med (ns)      Min (ns)    Max (ns)    StdDev (ns)                                                   Name                                                
 --------  ---------------  ---------  -------------  -------------  ----------  -----------  ------------  ----------------------------------------------------------------------------------------------------
...
      1.8      541,780,618        110    4,925,278.3    4,801,199.5   1,934,854   11,273,987   1,975,679.1  traccc::cuda::kernels::count_triplets(traccc::seedfinder_config, vecmem::edm::view<vecmem::edm::sch…
...
      0.8      227,799,499        110    2,070,904.5    1,894,950.5     912,321    5,418,155     847,702.1  traccc::cuda::kernels::find_doublets(traccc::seedfinder_config, vecmem::edm::view<vecmem::edm::sche…
      0.6      168,275,466        110    1,529,777.0    1,386,307.0     538,785   12,902,792   1,197,117.2  traccc::cuda::kernels::find_triplets(traccc::seedfinder_config, traccc::seedfilter_config, vecmem::…
      0.5      162,648,091        110    1,478,619.0    1,493,715.5     316,609    3,315,369     543,757.1  traccc::cuda::kernels::count_doublets(traccc::seedfinder_config, vecmem::edm::view<vecmem::edm::sch…
...
      0.1       22,085,158        110      200,774.2      195,952.0     114,208      559,553      56,230.2  traccc::cuda::kernels::select_seeds(traccc::seedfilter_config, vecmem::edm::view<vecmem::edm::schem…
...
      0.0        9,887,572        110       89,887.0       43,184.0      24,640      310,785      67,368.5  traccc::cuda::kernels::estimate_track_params(vecmem::data::vector_view<const traccc::measurement>, …
...
      0.0        4,828,110        110       43,891.9       11,055.5       8,384      147,872      47,505.2  traccc::cuda::kernels::reduce_triplet_counts(vecmem::data::vector_view<const traccc::device::double…
      0.0        4,640,174        110       42,183.4       39,520.0      24,032      155,073      15,700.1  traccc::cuda::kernels::update_triplet_weights(traccc::seedfilter_config, vecmem::edm::view<vecmem::…
...
      0.0        3,019,749        110       27,452.3       15,776.0       9,216       93,760      23,093.6  void traccc::cuda::kernels::form_spacepoints<detray::detector<detray::default_metadata<algebra::plu…
...
      0.0        2,190,561        110       19,914.2       18,799.5      13,120       49,440       5,830.7  traccc::cuda::kernels::populate_grid(traccc::seedfinder_config, vecmem::edm::view<vecmem::edm::sche…
      0.0        2,043,941        110       18,581.3       17,472.0      12,577       57,984       6,019.7  traccc::cuda::kernels::count_grid_capacities(traccc::seedfinder_config, detray::axis2::circular<std…
...
```

Note though that there is a fair amount of variance on these numbers. :thinking: Re-running the same test on the current state of the main branch I would get ~5% variances even on the runtime of `traccc::cuda::kernels::ccl_kernel`. Which is absolutely not affected by this PR. :confused: One of the slower runs with the current main branch look like this for instance:

```
nsys profile --stats=true ./build-current/bin/traccc_throughput_mt_cuda --input-directory /data/Acts/odd-simulations-20240509/geant4_ttbar_mu140/ --input-events 100 --cpu-threads 2 --deterministic
...
 Time (%)  Total Time (ns)  Instances    Avg (ns)       Med (ns)      Min (ns)     Max (ns)    StdDev (ns)                                                   Name                                                
 --------  ---------------  ---------  -------------  -------------  -----------  -----------  ------------  ----------------------------------------------------------------------------------------------------
...
      1.8      516,405,282        110    4,694,593.5    4,480,770.0    1,578,508   13,249,062   2,161,830.1  traccc::cuda::kernels::count_triplets(traccc::seedfinder_config, detray::const_grid2_view<detray::g…
...
      0.6      174,097,132        110    1,582,701.2    1,160,745.5      580,709    5,464,202     992,292.8  traccc::cuda::kernels::find_doublets(traccc::seedfinder_config, detray::const_grid2_view<detray::gr…
      0.4      125,428,299        110    1,140,257.3    1,098,185.0      370,114    2,470,451     387,936.4  traccc::cuda::kernels::find_triplets(traccc::seedfinder_config, traccc::seedfilter_config, detray::…
...
      0.3       86,367,906        110      785,162.8      701,397.5      198,882    2,067,408     376,686.2  traccc::cuda::kernels::count_doublets(traccc::seedfinder_config, detray::const_grid2_view<detray::g…
...
      0.1       19,817,147        110      180,155.9      177,713.0      104,000      369,699      45,532.0  traccc::cuda::kernels::select_seeds(traccc::seedfilter_config, vecmem::data::vector_view<const trac…
...      
      0.0        9,832,528        110       89,386.6       42,881.0       20,928      585,636     125,096.4  traccc::cuda::kernels::update_triplet_weights(traccc::seedfilter_config, detray::const_grid2_view<d…
      0.0        9,513,894        110       86,489.9       36,496.0       21,568      306,818      73,636.3  traccc::cuda::kernels::estimate_track_params(vecmem::data::vector_view<const traccc::spacepoint>, v…
...
      0.0        5,305,766        110       48,234.2       37,824.5       21,281      288,706      34,678.8  void traccc::cuda::kernels::form_spacepoints<detray::detector<detray::default_metadata<algebra::plu…
...
      0.0        3,374,773        110       30,679.8       26,944.0       17,024      263,362      23,959.4  traccc::cuda::kernels::populate_grid(traccc::seedfinder_config, vecmem::data::vector_view<const tra…
...
      0.0        2,786,839        110       25,334.9       10,480.0        8,160      141,857      29,897.2  traccc::cuda::kernels::reduce_triplet_counts(vecmem::data::vector_view<const traccc::device::double…
      0.0        2,451,603        110       22,287.3       21,520.5       16,032       49,761       5,758.1  traccc::cuda::kernels::count_grid_capacities(traccc::seedfinder_config, detray::axis2::circular<std…
...
```

As discussed at our meeting this morning, the fact that the code now properly keeps track of the (currently always zero) variances of the spacepoints, slows things down. But we absolutely have to do this. The code was artificially faster previously. :thinking: 

I'll still do some compute profiles, and fix the compilation issue(s). But generally I think we should go forward like this. :thinking:

krasznaa:
The NSight Compute profile didn't reveal much. :thinking: I believe we still have performance to be gained overall, but that should not be done in this PR. It's already doing more than it really should...

krasznaa:
Since I couldn't resist, I tried what would happen if I tweaked `doublet_finding_helper::transform_coordinates` a bit. With just some trivial modifications, I now get:

```
./traccc/out/build/cuda-native-fp32/bin/traccc_throughput_mt_cuda --input-directory /data/hdd-4tb/acts_data/odd-20240509/geant4_ttbar_mu140/ --input-events 100 --cpu-threads 2 --deterministic
...
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Reconstructed track parameters: 4767399
Time totals:
                  File reading  28576 ms
            Warm-up processing  2388 ms
              Event processing  21230 ms
Throughput:
            Warm-up processing  238.884 ms/event, 4.18614 events/s
              Event processing  212.308 ms/event, 4.71013 events/s
```

This can be directly compared with the numbers in https://github.com/acts-project/traccc/pull/878#issuecomment-2675546640. So it will be possible to improve things. :smile:

But that will be better to put into a separate, much smaller PR. Once this beast gets in. :wink:

core/include/traccc/seeding/triplet_finding_helper.hpp
```diff
@@ -31,25 +31,27 @@ struct triplet_finding_helper {
...
         scalar& curvature, scalar& impact_parameter);
 };
 
+template <typename T>
 bool TRACCC_HOST_DEVICE triplet_finding_helper::isCompatible(
-    const internal_spacepoint<spacepoint>& spM, const lin_circle& lb,
-    const lin_circle& lt, const seedfinder_config& config,
-    const scalar& iSinTheta2, const scalar& scatteringInRegion2,
-    scalar& curvature, scalar& impact_parameter) {
+    const edm::spacepoint<T>& spM, const lin_circle& lb, const lin_circle& lt,
```
beomki-yeo, line 44: Is `edm::spacepoint<T>& spM` accessing the spacepoint collection directly so we can access memory in coalesced manner? If it is an AoS elements created from the SoA container, it is not surprising that this PR is not gaining any speedup

device/common/include/traccc/seeding/device/impl/count_triplets.ipp
```diff
@@ -47,19 +50,20 @@ inline void count_triplets(
...
 
     const unsigned int counter_link = mid_bot.counter_link;
     const doublet_counter doublet_counts = dc_device.at(counter_link);
 
     // middle spacepoint
     const sp_location spM_loc = doublet_counts.m_spM;
-    const traccc::internal_spacepoint<traccc::spacepoint> spM =
-        internal_sp_device.bin(spM_loc.bin_idx)[spM_loc.sp_idx];
+    const auto spM =
+        spacepoints.at(sp_device.bin(spM_loc.bin_idx)[spM_loc.sp_idx]);
```
beomki-yeo, line 62: Could you teach me how this `at` function works? If this creates an AoS-like object from the SoA container, it won't be easy to get the speedup.

core/include/traccc/seeding/triplet_finding_helper.hpp
beomki-yeo: Hmm ho, @stephenswat also made the same comment..:
https://github.com/acts-project/traccc/pull/878#issuecomment-2677688853

device/common/include/traccc/seeding/device/impl/count_triplets.ipp
krasznaa: The code is a bit complicated by now, but the intention is meant to be relatively simple.

The `at(...)` functions and `operator[]` operators of [vecmem::edm::host](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/edm/host.hpp) and [vecmem::edm::device](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/edm/device.hpp) create "proxy objects". These are implemented using the highly templated [vecmem::edm::proxy](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/edm/proxy.hpp) type.

Now... It is indeed fairly well hidden in the code by now, but "proxy objects" by original design don't hold any payload. They just point to elements of a container. Through a bunch of (non-)const references. Which references are calculated from the container's pointers when creating the proxy.

Long story short, calling `.at(...)` or `[...]` on an SoA container doesn't do any I/O by itself. Only once you access one of the functions of the proxy, is the memory load/store initiated. Or, to be even more correct, once the code really accesses the variable returned by one of those functions, is a load initiated.

It's not too easy to demonstrate it exactly on these lines, since accessing the `spM_loc` member variables on this particular line does trigger a memory load by itself, but let's look at this MR's `doublet_finding_helper::transform_coordinates` function:

ncu

Notice how the first memory load only happens on line 91. Since `radius()` needs to use `x()` and `y()` to calculate a value on the fly. (This is also how I was able to speed things up by re-arranging things a bit in that function. But I don't want to spoil the surprise that I have teed up with that...)

Now... how `vecmem::edm::device::push_back(...)` and `vecmem::edm::host::push_back(...)` work is somewhat of a similar, but still different story. That's the thing that I needed to invent using `vecmem::edm::proxy` last week... 🤔

beomki-yeo: > Notice how the first memory load only happens on line 91.

Mind my ignorance - how do we know if the memory access is coalesced here?

core/include/traccc/seeding/triplet_finding_helper.hpp
krasznaa: What Stephen meant there (correctly) is that separating the `x`, `y` and `z` values into 3 separate `scalar` arrays was hurting performance. You see, seed finding very rarely accesses spacepoints that would be side-by-side. 😦 So accesses are rarely coalesced. 😢 Nothing that an SoA layout can help with unfortunately.

So yes, `x`, `y` and `z` are now stored as a single `point3` array in `traccc::edm::spacepoint_collection`. So that the 4-word load operations would at least be mostly efficient on those 3 floats.

But for perfect coalescence we need to make sure that neighboring threads would access neighboring spacepoints. Not really something that we can do...

device/common/include/traccc/seeding/device/impl/count_triplets.ipp
krasznaa: See comment above.

core/include/traccc/seeding/triplet_finding_helper.hpp
beomki-yeo: > You see, seed finding very rarely accesses spacepoints that would be side-by-side. 😦 So accesses are rarely coalesced. 

This makes sense OK..

> So yes, x, y and z are now stored as a single point3 array in traccc::edm::spacepoint_collection. So that the 4-word load operations would at least be mostly efficient on those 3 floats.

If I am correct, even for `point3`, `x`,`y`,`z` will be picked up separately so it won't make any benefit 
https://stackoverflow.com/questions/55688311/can-float3-enjoy-cuda-memory-coalescing

beomki-yeo: If the point3 is not working as we expect (in terms of efficient memory access), we can also consider using more native type such as float4 which would contain x,y,z,r. Though I am not sure if my argument is valid

krasznaa: I meant the [traccc::point3](https://github.com/acts-project/traccc/blob/main/core/include/traccc/definitions/primitives.hpp#L42) type. Which, in our default algebra, is `std::array<float, 3>`. That type works great for memory management. The 3 floats in such an array are by definition sitting side-by-side memory. So when you load the first element (`x`), the other two elements are also loaded on the same cache line.

Since `x`, `y` and `z` are practically always needed together, and because the seed-finging kernels practically never access side-by-side spacepoints in side-by-side threads, grouping those variables like this makes sense.

Notice that with the SoA infrastructure we can easily control how we group certain variables together. With our current AoS layout this type of tweaking is not available. I'm encouraged by my experience (https://github.com/acts-project/traccc/pull/878#issuecomment-2678166293) that we'll be able to do some clever things with re-arranging our code's data access for some performance gains. Since NSight Systems tells us that most of our seed-finding kernels are using single digit percentages of the GPU both in terms of compute and memory bandwidth at the moment. (Ironically the slowest kernel, `count_triplets`, is the one with the highest compute and memory bandwidth utilization already...)

I strongly believe that making the EDM use an SoA layout where we can group certain variables together for better I/O performance, is the way to go here. 🤔

beomki-yeo: If `std::array<float,3>` is automatically aligned for 16 byte cache size, then it seems fine

krasznaa: Interesting idea. Though not really for this PR to sort out.

I'd really like to get feedback about the technicalities of how the algorithmic code changed in the PR. Because I think you guys didn't even notice yet how I re-arranged the host code. (Though to be fair, the host code should not be our main focus...) Or the I/O and performance code.

krasznaa: > If `std::array<float,3>` is automatically aligned for 16 byte cache size, then it seems fine

The [vecmem::edm::buffer](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/edm/buffer.hpp) objects try to be very particular about alignment. 👍 Making use of the generic code that Stephen wrote for VecMem a long time ago.

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/edm/impl/buffer.ipp#L98-L105

beomki-yeo: OK then let's resolve this thread for now. Coalescing and memory layout in the GPU programming is a confusing part to me tbh..

krasznaa:
Choosing a local rebase on this branch was not the right call... :frowning: Git clearly got very confused about how to transplant my 13 commits on top of the current `main` branch. First time that I'm running into such an issue. :thinking:

It's not he end of the world since the commits would've been squashed in the end anyway, but still...

tests/cpu/test_track_params_estimation.cpp
```diff
@@ -46,18 +47,23 @@ TEST(track_params_estimation, helix_negative_charge) {
...
     // Make three spacepoints with the helix
-    spacepoint_collection_types::host spacepoints;
-    spacepoints.push_back({hlx(50 * unit<scalar>::mm), {}});
-    spacepoints.push_back({hlx(100 * unit<scalar>::mm), {}});
-    spacepoints.push_back({hlx(150 * unit<scalar>::mm), {}});
+    measurement_collection_types::host measurements(&host_mr);
+    edm::spacepoint_collection::host spacepoints{host_mr};
+    measurements.resize(3);
+    spacepoints.reserve(3);
+    spacepoints.push_back({0, hlx(50 * unit<scalar>::mm), 0.f, 0.f});
```
beomki-yeo, line 54: What is the index of 0,1,2?

beomki-yeo (approved):
It looks good to me but let's wait for @stephenswat's approval and further comments as the PR is huge

tests/cpu/test_track_params_estimation.cpp
krasznaa: Probably I could simplify the test here a bit. 🤔

I wanted the test spacepoints to point at "valid" measurements. That's why I resized the measurement container to `3`.

But no algorithmic code tries to access measurements through these spacepoints in this test. So even creating the measurement collection is really unnecessary in the end.

So `0`, `1` and `2` are meant to be the indices of the elements in `measurements`. Though in this test it really doesn't matter...

krasznaa: Never mind, let's not "simplify" this test.

The track parameter estimation does access one of the measurements. It's fine to have dummy values in the measurements, but it's still best to have existing objects for all three spacepoints in memory.

core/include/traccc/seeding/detail/seed_finding.hpp
```diff
@@ -0,0 +1,65 @@
...
+    ///
+    edm::seed_collection::host operator()(
+        const edm::spacepoint_collection::const_view& spacepoints,
+        const traccc::details::spacepoint_grid_types::host& sp_grid) const;
+
+    private:
+    /// Internal implementation struct
+    struct impl;
+    /// Pointer to the internal implementation
+    std::unique_ptr<impl> m_impl;
```
stephenswat, line 61: Is there a reason this needs to be PIMPL? :thinking:

stephenswat (commented):
Looks okay in general, but there are _so_ many unrelated changes in this PR... I would suggest we work a bit on pull request hygiene because this single 5000-line delta PR is practically unreviewable because it contains what should have been five or six independent PRs... :confused:

performance/include/traccc/performance/details/projector.hpp
```diff
@@ -38,24 +38,6 @@ struct projector<traccc::measurement> {
...
-
-template <>
-struct projector<traccc::seed> {
-    static constexpr bool exists = true;
-
-    float operator()(const traccc::seed& i) {
-        return static_cast<float>(i.z_vertex);
-    }
-};
-
```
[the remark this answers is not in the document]
krasznaa: This concept is not really needed for SoAs. :thinking: If you want to extract one of the columns from the SoA, you can access that column directly, without any transformation.

You caught it correctly that I simplified the comparison code with this new EDM. Since SoA containers can't easily be sorted, and I didn't want to invent a whole thing with an extra container holding the sorted indices.

I actually first translated these specializations to work on `traccc::edm::spacepoint<T>` and `traccc::edm::seed<T>` types. But then had to realize that we wouldn't be able to use these specializations with the SoA EDM anyway. :thinking:

performance/include/traccc/efficiency/nseed_performance_writer.hpp
```diff
@@ -34,6 +34,7 @@ class nseed_performance_writer {
     void initialize();
     void finalize();
 
+    /*
```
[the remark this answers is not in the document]
krasznaa: I may have forgotten about a "temporarily disabled" piece of code. :frowning: Let me check...

io/src/obj/write_seeds.cpp
```diff
@@ -91,10 +81,12 @@ void write_seeds(std::string_view filename,
 
     // Now build the seeds as lines connecting the spacepoint vertices.
     file << "# Seeds\n";
-    for (const seed& s : seeds) {
-        file << "l " << get_spacepoint_index(s.spB_link) << " "
-             << get_spacepoint_index(s.spM_link) << " "
-             << get_spacepoint_index(s.spT_link) << "\n";
+    for (edm::seed_collection::const_device::size_type i = 0; i < seeds.size();
+         ++i) {
```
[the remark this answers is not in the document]
krasznaa: Because there is no easy way of implementing `begin()` and `end()` for `vecmem::edm::device` or `vecmem::edm::host`.

I thought about this since a while. At one point I do intend to write an iterator type that would allow us to iterate over transient SoA proxies. But one doesn't exist yet. So for the moment we need to iterate over SoA containers in such an old-school way.

device/sycl/include/traccc/sycl/seeding/details/spacepoint_binning.hpp
```diff
@@ -11,27 +11,22 @@
...
-#include <functional>
-#include <memory>
 #include <utility>
 
-namespace traccc::sycl {
+namespace traccc::sycl::details {
 
 /// Spacepoing binning executed on a SYCL device
-class spacepoint_binning : public algorithm<sp_grid_buffer(
-                               const spacepoint_collection_types::const_view&)>,
```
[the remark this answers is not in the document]
krasznaa: Indeed, this is one of the not-strictly-related changes. I made this particular change because it was now that I finally was modifying these classes.

I even wanted to hide these classes completely from the users at first. As I didn't think that a regular client of the code should ever use these classes directly.

But since some test code was using them directly, I decided to just push them into (various) `details::` namespaces, and remove their algorithm base class. The latter was also mainly to signal that these are not algorithms that people should use directly.

io/src/obj/write_seeds.cpp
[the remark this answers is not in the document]
stephenswat: Okay...

device/common/include/traccc/seeding/device/impl/find_doublets.ipp
```diff
@@ -37,15 +38,17 @@ inline void find_doublets(
...
+    const traccc::details::spacepoint_grid_types::const_device sp_grid(sp_view);
     device_doublet_collection_types::device mb_doublets(mb_doublets_view);
     device_doublet_collection_types::device mt_doublets(mt_doublets_view);
 
     // Get the spacepoint that we're evaluating in this thread, and treat that
     // as the "middle" spacepoint.
-    const internal_spacepoint<spacepoint> middle_sp =
-        sp_grid.bin(middle_sp_counter.m_spM.bin_idx)
-            .at(middle_sp_counter.m_spM.sp_idx);
+    const auto middle_sp =
```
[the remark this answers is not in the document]
krasznaa: As I wrote before, that would result in some properly dangerous code.

Since I do know that the compiler would actually allow that to be written. We've had such an occurrence in the ATLAS offline code, where one function all of a sudden started returning its thing by value. And clients continued working for a long time. Until one of the clients tried to pass along such a reference to a different scope. Then all of a sudden all hell broke lose. And it took quite some debugging to figure out what was going wrong. (As the code breaking was quite a bit removed from the code causing the issue.)

I also still debate whether we should use `auto` so liberally in these places. :thinking: There are proper named types that we could use here. They are just so very long... :thinking: So I'm not sure what's the best for readability and maintainability...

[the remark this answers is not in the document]
stephenswat: > Until one of the clients tried to pass along such a reference to a different scope. Then all of a sudden all hell broke lose. And it took quite some debugging to figure out what was going wrong. (As the code breaking was quite a bit removed from the code causing the issue.)

Can you elaborate on how passing on a lifetime-extended temporary is dangerous?

core/include/traccc/seeding/detail/seed_finding.hpp
krasznaa: Because I wanted to move all the separate classes that we use in the implementation, out of the public interface of the library. As they are all implementation details. That are only used by the host code.

core/src/seeding/silicon_pixel_spacepoint_formation.hpp
```diff
@@ -38,14 +38,19 @@ spacepoint_collection_types::host silicon_pixel_spacepoint_formation(
...
+    for (unsigned int meas_index = 0u; const auto& meas : measurements) {
         if (traccc::details::is_valid_measurement(meas)) {
-            result.push_back(traccc::details::create_spacepoint(det, meas));
+            const std::size_t sp_index = result.size();
+            result.resize(sp_index + 1u);
+            auto sp = result.at(sp_index);
+            traccc::details::fill_pixel_spacepoint(sp, det, meas);
+            sp.measurement_index() = meas_index;
         }
+        ++meas_index;
```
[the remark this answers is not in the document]
krasznaa: `std::distance`? You mean to switch from this Frankenstein for-loop to an iterator based one?

Yeah, I could be on board with that. Adding the extra integer seemed like the smaller intervention in the code. But I agree, the for-loop is pretty ugly now.

stephenswat:
By the way, I think we should strongly consider testing out the new `[[lifetimebound]]` attribute on these proxies! Not for this PR though.

device/common/include/traccc/seeding/device/impl/find_doublets.ipp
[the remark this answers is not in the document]
krasznaa: My understanding is that the lifetime of the returned object is as one would expect. It last's "the scope that it was received in".

Let's take the following dummy code for instance:

```c++
std::reference_wrapper<const Foo> ref;
{
   const Foo& f = functionReturningValue();
   ref = f;
}
ref.get().doSomething();
```

The thing we encountered in the offline code was something along these lines.

I feel pretty strongly that using references in these places would very easily lead to coding mistakes later down the road. And clearly... adding `&` will do absolutely nothing to the performance of the generated code. So what good would it do? It would only obfuscate the code more in my mind.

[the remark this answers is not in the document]
stephenswat: > ```c++
> std::reference_wrapper<const Foo> ref;
> {
>    const Foo& f = functionReturningValue();
>    ref = f;
> }
> ref.get().doSomething();
> ```

By this logic your current proxy objects are also broken:

```c++
my_edm_container cont;
std::reference_wrapper<my_edm_container::proxy_type> ref;
{
   my_edm_container::proxy_type p = cont[0];
   ref = p;
}
ref.get().x();
```

What I am trying to say is, if your user wants to shoot themselves in the foot this much then there is no way to prevent them to (without switching to Rust).

At the same time, keeping these variables as references makes the code more flexible! If the design of the container changes, a reference will be able to keep up without invisible performance degradation; not the case if they are values.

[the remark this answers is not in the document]
stephenswat: Plus, accepting them as references provides a semantic hint to the user that these types are basically just fancy reference packs, and that they cannot be changed without side-effects.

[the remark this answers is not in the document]
krasznaa: In your example the code literally makes a reference to a temporary. Of course it will break then. :confused: In my example the writer of the code could reasonably believe that they are holding onto a reference that has a lifetime beyond the current scope.

I see your argument about changing future containers, but then let me go ahead and replace all the `auto`-s with concrete types. As I'm coming to think that that would result in the most readable code going forward...

performance/include/traccc/performance/impl/is_same_spacepoint.ipp
```diff
@@ -1,32 +1,40 @@
...
-                    const bool include_meas = true) const;
+    bool operator()(const edm::spacepoint<T>& obj) const {
+
+        return (is_same_scalar(obj.x(), m_ref.x(), m_unc) &&
+                is_same_scalar(obj.y(), m_ref.y(), m_unc) &&
+                is_same_scalar(obj.z(), m_ref.z(), m_unc));
+    }
 
     private:
     /// The reference object
```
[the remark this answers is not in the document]
krasznaa: I'm only looking at these "outdated" comments now. What do you mean here?

"Reference" here is not in the C++ sense, but rather that "this is the reference that we compare our test object against".

device/common/include/traccc/seeding/device/impl/find_doublets.ipp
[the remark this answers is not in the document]
krasznaa: I went for removing/replacing all usages of `auto` for the new types. This way if/when the definition of `edm::spacepoint_collection::const_device::const_proxy_type` changes, the code here should change correctly as well. :thinking:

performance/include/traccc/efficiency/nseed_performance_writer.hpp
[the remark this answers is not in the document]
krasznaa: This was a good catch!

That code is not used anywhere in practice, but I now still tried to update it to be compatible with the new EDM.

It's very likely to still have issues, but I was not going to write client code for this class just to test it.

core/src/seeding/silicon_pixel_spacepoint_formation.hpp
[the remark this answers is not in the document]
krasznaa: Have a look!

I fear though that I only made the code (even) uglier. :frowning:

core/include/traccc/seeding/detail/triplet_sorter.hpp
```diff
@@ -0,0 +1,77 @@
...
+        else {
+            // Access the spacepoints.
+            const auto spB1 = m_spacepoints->at(
+                m_sp_grid->bin(seed1.sp1.bin_idx)[seed1.sp1.sp_idx]);
+            const auto spT1 = m_spacepoints->at(
+                m_sp_grid->bin(seed1.sp3.bin_idx)[seed1.sp3.sp_idx]);
+            const auto spB2 = m_spacepoints->at(
+                m_sp_grid->bin(seed2.sp1.bin_idx)[seed2.sp1.sp_idx]);
+            const auto spT2 = m_spacepoints->at(
+                m_sp_grid->bin(seed2.sp3.bin_idx)[seed2.sp3.sp_idx]);
```
krasznaa, line 55: Dang... forgot about these ones... :cursing_face:

stephenswat (approved):
We can go ahead and put this in.

krasznaa:
> We can go ahead and put this in.

May God have mercy on our souls... :stuck_out_tongue:

PR #880, opened 2025-02-20
fix: Temporarily switch the smoother type in benchmark

niermann999:
Switch the smoother type to prevent a segfault until the direct navigator can be deployed. Also reanables correct navigation.


niermann999:
I tried this in the ubuntu cuda docker container that is run for the benchmark jobs and it seems to fix the issue. But it would be great, if somebody could check

beomki-yeo:
As we discussed earlier, I will introduce the direct navigator soon so please hold the PR.

beomki-yeo:
#882 will fix the current issue of seg fault.

niermann999:
> As we discussed earlier, I will introduce the direct navigator soon so please hold the PR.

The point is, that in the meantime the simulation step will not produce up to 30% of hits. Then the CKF might also miss a similar amount of sensitive surfaces during navigation. At the same time it needs to get more than three hits to make a track in a detector that has four barrel layers. 
I hear you that the smoothing would produce garbage after this PR, but if we leave the navigation broken, I worried that the throughput that is currently being reported in the benchmark is anywhere near accurate. Would it not be better if the chain works up until the smoother at least?

beomki-yeo:
No worries. I will deliver the solution soon.

PR #882, opened 2025-02-21
Exit with an error if the first smoothed state is invalid

beomki-yeo:
Depends on #881.

This will solve the issue of seg fault issue in the benchmark by exiting with an error if the first smoothed state is invalid

Update: This does not solve the seg fault issue actually


core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
```diff
@@ -226,6 +239,16 @@ class kalman_fitter {
             // return false;
         }
         auto& last = *fitter_state.m_fit_actor_state.m_it_rev;
+
+        const scalar theta = last.smoothed().theta();
+        if (theta <= 0.f || theta >= constant<traccc::scalar>::pi) {
```
beomki-yeo, line 244: To make it clear, this is where the fix was done.

extern/covfie/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ message( STATUS "Fetching covfie as part of the traccc project" )
 
 # Declare where to get covfie from.
 set( TRACCC_COVFIE_SOURCE
-   "URL;https://github.com/acts-project/covfie/archive/refs/tags/v0.11.0.tar.gz;URL_MD5;9f41c9308e7801ac6319c470a1ec4212"
+   "URL;https://github.com/acts-project/covfie/archive/refs/tags/v0.12.1.tar.gz;URL_MD5;1512b389d8d7d08ef3bd30d6b18dcee1"
```
krasznaa, line 16: I thought there was a separate Covfie update recently. Did that not go in yet? It would just seem better to factorize such things from each other.

core/include/traccc/fitting/status_codes.hpp
```diff
@@ -0,0 +1,20 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022-2024 CERN for the benefit of the ACTS project
```
krasznaa, line 3: This could be:

```suggestion
 * (c) 2025 CERN for the benefit of the ACTS project
```

😉

core/include/traccc/fitting/status_codes.hpp
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <cstdint>
+namespace traccc {
+enum class kalman_fitter_status : uint32_t {
```
krasznaa, line 12: I'm not in love with the name, but don't have fundamentally better suggestions for the moment. 🤔

But I wonder... Should we rather use `std::uint32_t`? I **think** the C\+\+ standard only guarantees the existence of the `std::` prefixed version.

https://en.cppreference.com/w/cpp/header/cstdint

But we certainly do this pretty inconsistently in the code elsewhere as well. 🤔

krasznaa (commented):
I absolutely like the direction that all of this is going in.

I was wondering for a moment whether we could make use of [<system_error>](https://en.cppreference.com/w/cpp/header/system_error) in all of this, similar to what we do in Gaudi's [StatusCode](https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/GaudiKernel/include/GaudiKernel/StatusCode.h), but on a quick look [std::error_code](https://en.cppreference.com/w/cpp/error/error_code) looks pretty GPU hostile. So simple enums are probably the best way to go about this. 🤔

I believe eventually we'll have a number of algorithm specific enums. And we'll need to think a bit more about the proper way of reporting back about these as well. But overall, I think this is a good direction to start in like this.

Just the unimportant question/suggestion from me, as shown above...

stephenswat:
Mind you this contains a bunch of already merged changes, the PR needs to be rebased.

krasznaa:
> Mind you this contains a bunch of already merged changes, the PR needs to be rebased.

That of course explains things. I'm just surprised that (as long as it can be cleanly rebased) GitHub doesn't do this automatically in showing the diffs. I **think** GitLab can do that. Though I'm not absolutely sure about that...

stephenswat (approved):

PR #883, opened 2025-02-21
Add pre-commit hook to check for detray units

stephenswat:
This commit adds a little pre-commit hook (inspired by #866 and #876) that ensures that we don't re-introduce uses of the `detray::unit` and `detray::constant` namespaces. Can be extended in the future to hold more patterns if desired.

Depends on #876.


niermann999 (approved):

stephenswat:
I added the track parameters too, so this now depends on #867.

PR #884, opened 2025-02-21
Fix edge case detection in CPU track finding

stephenswat:
This commit fixes a critial bug in the CPU track finding where it is not able to deal with cases where the previous step ID is `U?INT_MAX`. This is used to indicate that we are in the first step. The fix in this commit is two-pronged:

1. This commit ensures that invalid step IDs are skipped through.
2. This commit ensures that the invalid step ID value is properly set; we were previously comparing `UINT_MAX` with `INT_MAX` which are obviously different.

I also sprinkled around some `at` calls to replace `operator[]` for better error handling.


beomki-yeo (approved):
Makes sense

beomki-yeo:
Let me update the branch and merge..

PR #885, opened 2025-02-21
Set particle hypothesis in Kálmán fitter

stephenswat:
Currently, we are not actually setting the particle hypothesis in the Kálmán fitter, which I find leads to some assertions about the sign of the $\frac{q}{p}$ value not being equal to the sign of $q$. This commit enforces the particle hypothesis in the fitter and adds an additional assert to ensure that the operation was successful.


beomki-yeo (approved):
Right Thanks for the fix

PR #890, opened 2025-02-25
Fix SFINAE bug in container base type

stephenswat:
This commit fixes a bug in the use of SFINAE in the container base type. In particuler, this issue has three layers:

1. The existing code uses the `typename = std::enable_if` anti-pattern which is discouraged in the C++ standard.
2. The code used `std::enable_if` without getting `::type` from it, meaning that a specialization failure cannot ever occur, rendering the SFINAE useless.
3. The check used about type equality is incorrect, it should rather check constructability.

This commit fixes the issues by converting the code to a C++20 requires clause and by modifying the conditions.

### DIFF
--- a/core/include/traccc/edm/details/container_base.hpp
+++ b/core/include/traccc/edm/details/container_base.hpp
@@ -137,13 +137,11 @@ class container_base {
     /**
      * @brief Constructor from a pair of "view type" objects
      */
-    template <typename header_vector_tp, typename item_vector_tp,
-              typename = std::enable_if<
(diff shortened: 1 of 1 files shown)


niermann999 (approved):

krasznaa (approved):
I was a little afraid that you would've spent too much time on this code, which is on its way out. But glad to see that the size of the change is appropriate. :smile:

Sure, let's do this!

stephenswat:
> which is on its way out.

TIL :laughing:

PR #845, opened 2025-02-07
Configuration Updates, main branch (2025.02.07.)

krasznaa:
This is a replacement for #837. Abandoning changing the classes in `traccc::utils`, and only focusing on the `traccc::opts::` classes.

I made (pretty much) the same updates in the existing classes as in #837. And then added the following:
  - Updated `traccc::opts::detector` and `traccc::opts::input_data` to use ODD muon simulations by default. To make it a bit easier to test the code. (As we mostly use ODD data for that.)
  - Introduced `traccc::opts::track_fitting`, to make the track fitting options configurable. Notably, with the update, I would see:

```
./out/build/cuda-fp32/bin/traccc_throughput_mt_cuda --input-directory /data/hdd-4tb/acts_data/odd-20240509/geant4_ttbar_mu140/ --input-events 100 --cpu-threads 2 --fit-use-backward-filter false

Running Multi-threaded CUDA GPU throughput tests

Detector Options:
├ Detector file:                   geometries/odd/odd-detray_geometry_detray.json
├ Material file:                   geometries/odd/odd-detray_material_detray.json
├ Surface grid file:               geometries/odd/odd-detray_surface_grids_detray.json
├ Use detray detector:             true
└ Digitization file:               geometries/odd/odd-digi-geometric-config.json
...
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Using CUDA device: NVIDIA GeForce RTX 3080 [id: 0, bus: 1, device: 0]
Warm-up processing [==================================================] 100% [00m:00s]                                                                                                                                                                                                                                                                                                                                               
Event processing   [==================================================] 100% [00m:00s]                                                                                                                                                                                                                                                                                                                                               
Reconstructed track parameters: 5573291
Time totals:
                  File reading  2875 ms
            Warm-up processing  1821 ms
              Event processing  16363 ms
Throughput:
            Warm-up processing  182.15 ms/event, 5.48998 events/s
              Event processing  163.638 ms/event, 6.11107 events/s
```

Versus the default:

```
./out/build/cuda-fp32/bin/traccc_throughput_mt_cuda --input-directory /data/hdd-4tb/acts_data/odd-20240509/geant4_ttbar_mu140/ --input-events 100 --cpu-threads 2

Running Multi-threaded CUDA GPU throughput tests

Detector Options:
[...]

PR #891, opened 2025-02-25
Check for `std::format` support in CMake

stephenswat:
Today I tried building traccc with g++12 which broke as #845 introduced a dependency on `std::format` which is not universally available. In order to provide earlier and more readable errors, this commit adds a configuration-time check to ensure that the C++ compiler supports `std::format`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,25 @@ option( TRACCC_USE_SYSTEM_LIBS "Use system libraries be default" FALSE )
 option( TRACCC_USE_SPACK_LIBS "Use pack libraries by default" FALSE )
 option( TRACCC_USE_ROOT "Use ROOT in the build (if needed)" TRUE )
 
+# Check for some compiler features
+if(TRACCC_BUILD_EXAMPLES)
+   message(STATUS "Checking for `std::format` support")
+   file( WRITE
+   "${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp/compile_test_format.cpp"
+   "#include <format>\n"
+   "int main() {\n"
+   "std::format(\"Hello {}\\n\", \"world\");\n"
+   "return 0; }\n" )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -68,6 +68,25 @@ option( TRACCC_USE_SYSTEM_LIBS "Use system libraries be default" FALSE )
...
+   "std::format(\"Hello {}\\n\", \"world\");\n"
+   "return 0; }\n" )
+   try_compile( TRACCCC_INTERNAL_STD_FORMAT_WORKS SOURCES
+   "${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/CMakeTmp/compile_test_format.cpp")
+
+   if (NOT TRACCCC_INTERNAL_STD_FORMAT_WORKS)
+      message(FATAL_ERROR "Examples are enabled, but `std::format` is not supported by the C++ compiler; upgrade to gcc 13 or clang 14 or higher.")
+   endif()
+
+   unset(TRACCCC_INTERNAL_STD_FORMAT_WORKS)
```
krasznaa, line 87: No need to write such low level code. We have [check_cxx_source_compiles(...)](https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html) for this. Which is used extensively in [vecmem](https://github.com/acts-project/vecmem) for instance. (https://github.com/acts-project/vecmem/blob/main/core/CMakeLists.txt#L208)

I'd propose:

```suggestion
   include(CheckCxxSourceCompiles)
   check_cxx_source_compiles( "
      #include <format>
      int main() {
         std::format(\"Hello {}\\n\", \"world\");
      }
      " TRACCC_STD_FORMAT_WORKS )
   if( NOT TRACCC_STD_FORMAT_WORKS )
      message(FATAL_ERROR "Examples are enabled, but `std::format` is not supported by the C++ compiler; upgrade to gcc 13 or clang 14 or higher.")
   endif()
```

krasznaa (changes requested):
Let's indeed add such code. Just in a slightly higher level way. :wink:

PR #892, opened 2025-02-25
Move [[nodiscard]] and [[maybe_unused]] in front of host/device declarations

StewMH:
Make HIP happy


stephenswat:
I like it, but can we somehow add a CI job so we don't keep reintroducing these mistakes?

core/include/traccc/utils/relations.hpp
```diff
@@ -14,7 +14,7 @@
 
 namespace traccc {
 struct [[maybe_unused]] channel0_major_cell_order_relation{
-    template <typename T1, typename T2> TRACCC_HOST_DEVICE [[maybe_unused]] bool
+    template <typename T1, typename T2>[[maybe_unused]] TRACCC_HOST_DEVICE bool
```
krasznaa, line 17: I'm not in love with the lack of a space in front of `[[maybe_unused]]`. :thinking:

krasznaa (commented):
Indeed, we'd need a recipe soon about how you do these builds. So that they could be more automated.

Though, since I expect that things are not working yet, I don't think it's necessarily for this PR to sort it out.

stephenswat (approved):

PR #893, opened 2025-02-25
Remove dependency on dfelibs

stephenswat:
The ACTS project is deprecating its fork of dfelibs, which we use only for CSV reading. Thus, this commit removes our dependency on dfelibs and simply places the relevant code into the traccc repository.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,20 +296,6 @@ if( TRACCC_SETUP_COVFIE )
    endif()
 endif()
 
-# Set up dfelibs.
-option( TRACCC_SETUP_DFELIBS
-   "Set up the dfelibs target(s) explicitly" TRUE )
-option( TRACCC_USE_SYSTEM_DFELIBS
-   "Pick up an existing installation of dfelibs from the build environment"
-   ${TRACCC_USE_SYSTEM_LIBS} )
-if( TRACCC_SETUP_DFELIBS )
-   if( TRACCC_USE_SYSTEM_DFELIBS )
-      find_package( dfelibs REQUIRED )
-   else()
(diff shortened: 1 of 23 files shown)


krasznaa:
Is that GH repository actually going to be removed? 🤔

If not, I would not bother with this here. I see nothing to be gained from making this change. Not on a repository that has a finite lifetime of its own.

Is Acts still going to continue to use this code for its own CSV I/O? If so, we'll just pick it up from Acts's codebase when we migrate  the components of this repository into the Acts one. But if Acts is switching to a different technology, then we'd of course need to switch to that different technology here a priori as well.

But that's not what's happening from what I understand, right?

krasznaa:
Not to mention that if Acts already has its own copy of that code, then we're at option C. We should be picking up the DFE code from Acts ourselves as well, once we switch to "that" version of Acts. Otherwise we could get into all sorts of issues with diverging codebases.

However I look at it, making a copy of the code into this repository does not seem to be the correct choice in any circumstance. 🤔

stephenswat:
Ah no that lives in ActsExamples

krasznaa:
> Ah no that lives in ActsExamples

In that case I'm back to option A. :thinking: Is there anything to be gained with putting effort into this right now? Our build literally just needs an old TGZ file, which hopefully doesn't go away for a while longer.

paulgessinger:
Can we get this in? It's currently producing warnings about CMake versions, and I'm not keen on changing the dfelibs repo at this point.

Reopened at: https://github.com/acts-project/traccc/pull/1011

stephenswat:
With the looming integration into ACTS this is now more relevant than ever. Shall we move ahead with this?

paulgessinger (approved):

PR #894, opened 2025-02-25
Remove some unused executables

stephenswat:
This commit removes the `ccl_example`, `tbb_task_example`, and the OpenMP examples which, as far as I know, nobody is using.


krasznaa (approved):
:+1:

PR #886, opened 2025-02-21
Add additional error checking in Kálmán filter

stephenswat:
This commit further enhances the error checking mechanism in the Kálmán filter by making the following changes:

1. Adding additional checks and return points for possible errors.
2. Adding checks in debug mode that states which pass the rudimentary finiteness checks pass more rigorous checks on _all_ matrix elements.
3. Mark error paths as unlikely, hopefully aiding the compiler to minimize the performance impact.

PR #895, opened 2025-02-25
Protect against negative chi-square in updater

stephenswat:
I've observed an issue in which the gain matrix updater sometimes produces $\chi^2$ values less than zero, which makes no sense. This commit adds a check for this condition and returns an appropriate error state.

### DIFF
--- a/core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
+++ b/core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
@@ -147,6 +147,10 @@ struct gain_matrix_smoother {
         const matrix_type<1, 1> chi2 =
             matrix::transpose(residual) * matrix::inverse(R) * residual;
 
+        if (getter::element(chi2, 0, 0) < 0.f) {
+            return kalman_fitter_status::ERROR_SMOOTHER_CHI2_NEGATIVE;
+        }
+
         cur_state.smoothed_chi2() = getter::element(chi2, 0, 0);
         cur_state.is_smoothed = true;
 
--- a/core/include/traccc/fitting/kalman_filter/two_filters_smoother.hpp
+++ b/core/include/traccc/fitting/kalman_filter/two_filters_smoother.hpp
@@ -119,6 +119,10 @@ struct two_filters_smoother {
                                            matrix::inverse(R_smt) *
                                            residual_smt;
 
+        if (getter::element(chi2_smt, 0, 0) < 0.f) {
+            return kalman_fitter_status::ERROR_SMOOTHER_CHI2_NEGATIVE;
+        }
+
         trk_state.smoothed_chi2() = getter::element(chi2_smt, 0, 0);
 
         /*************************************
@@ -165,6 +169,10 @@ struct two_filters_smoother {
 
         if (std::abs(bound_params.qop()) == 0.f) {
             return kalman_fitter_status::ERROR_QOP_ZERO;
+        }
+
+        if (getter::element(chi2, 0, 0) < 0.f) {
+            return kalman_fitter_status::ERROR_UPDATER_CHI2_NEGATIVE;
         }
 
         // Set backward chi2
--- a/core/include/traccc/fitting/status_codes.hpp
+++ b/core/include/traccc/fitting/status_codes.hpp
@@ -14,6 +14,8 @@ enum class kalman_fitter_status : uint32_t {
     ERROR_QOP_ZERO,
     ERROR_THETA_ZERO,
     ERROR_INVERSION,
+    ERROR_SMOOTHER_CHI2_NEGATIVE,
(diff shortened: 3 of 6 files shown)


krasznaa (commented):
Up to Beomki to say the final word, but from my side, sure, this looks sane.

I don't think it makes any difference for the GPU compilers, but would it make sense to add `[[unlikely]]` on these checks? :thinking:

https://en.cppreference.com/w/cpp/language/attributes/likely

beomki-yeo:
Rigorously speaking, it is an error. The question is how frequently this happens. If it happens for 99% of tracks.. maybe we should go with `unlikely` :skull:

beomki-yeo:
Meanwhile, the error also needs to be included in the two filters smoother :)

stephenswat:
> I don't think it makes any difference for the GPU compilers, but would it make sense to add `[[unlikely]]` on these checks? 🤔

https://github.com/acts-project/traccc/pull/886/files :wink:

stephenswat:
> Meanwhile, the error also needs to be included in the two filters smoother :)

I've updated the two-filter smoother, now!

tests/cpu/test_kalman_fitter_momentum_resolution.cpp
```diff
@@ -219,7 +221,7 @@ TEST_P(KalmanFittingMomentumResolutionTests, Run) {
     float success_rate = static_cast<float>(n_success) /
                          static_cast<float>(n_truth_tracks * n_events);
 
-    ASSERT_FLOAT_EQ(success_rate, 1.00f);
+    ASSERT_FLOAT_GE(success_rate, 0.98f);
```
beomki-yeo, line 224: It seems `ASSERT_FLOAT_GE` does not exist. Also do you know if this efficiency drops is from the forward or backward filter?

core/include/traccc/fitting/status_codes.hpp
```diff
@@ -14,6 +14,7 @@ enum class kalman_fitter_status : uint32_t {
     ERROR_QOP_ZERO,
     ERROR_THETA_ZERO,
     ERROR_INVERSION,
+    ERROR_CHI2_NEGATIVE,
```
beomki-yeo, line 17: Shall we make it two for better debugging?:
`ERROR_FILTERED_CHI2_NEGATIVE`
`ERROR_SMOOTHED_CHI2_NEGATIVE`

The return values above will need to change accordingly

beomki-yeo (commented):
I like the idea while we can still make it more informative

core/include/traccc/fitting/status_codes.hpp
stephenswat: Hmm, sure I can do that. Should we then also change all the other error codes to have corresponding `FILTER` and `SMOOTH` versions?

tests/cpu/test_kalman_fitter_momentum_resolution.cpp
stephenswat: This I do not know.

core/include/traccc/fitting/status_codes.hpp
stephenswat: I've added smoother and filter versions of these errors now.

beomki-yeo: > Should we then also change all the other error codes to have corresponding FILTER and SMOOTH versions?

Yes I think that would be better but not for this PR to sort out

core/include/traccc/fitting/kalman_filter/gain_matrix_smoother.hpp
```diff
@@ -147,6 +147,10 @@ struct gain_matrix_smoother {
         const matrix_type<1, 1> chi2 =
             matrix::transpose(residual) * matrix::inverse(R) * residual;
 
+        if (getter::element(chi2, 0, 0) < 0.f) {
+            return kalman_fitter_status::ERROR_UPDATER_CHI2_NEGATIVE;
```
beomki-yeo, line 151: This should be ERROR_SMOOTHER_CHI2_NEGATIVE

core/include/traccc/fitting/kalman_filter/two_filters_smoother.hpp
```diff
@@ -167,6 +171,10 @@ struct two_filters_smoother {
             return kalman_fitter_status::ERROR_QOP_ZERO;
         }
 
+        if (getter::element(chi2, 0, 0) < 0.f) {
+            return kalman_fitter_status::ERROR_SMOOTHER_CHI2_NEGATIVE;
```
beomki-yeo, line 175: This should be ERROR_UPDATER_CHI2_NEGATIVE or ERROR_FILTER_CHI2_NEGATIVE whichever used in `gain_matrix_updater`.

stephenswat:
Okay, I've updated the return values.

beomki-yeo (approved):

PR #897, opened 2025-02-27
Add CLI options to control logging verbosity

stephenswat:
This commit adds the `-v` and `-q` flag that control verbosity, respectively making the code more verbose or more quiet. Also adds these options to the throughput examples.

### DIFF
--- a/examples/options/src/logging.cpp
+++ b/examples/options/src/logging.cpp
@@ -1,0 +1,83 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2024-2025 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+// Local include(s).
+#include "traccc/options/logging.hpp"
+
+#include <Acts/Utilities/Logger.hpp>
+
+#include "traccc/examples/utils/printable.hpp"
+
+// System include(s).
+#include <format>
+#include <iostream>
+#include <sstream>
+#include <stdexcept>
+
+namespace {
+class verbosity_counter : public boost::program_options::typed_value<int> {
+    public:
+    verbosity_counter(int* store)
+        : boost::program_options::typed_value<int>(store) {
+        default_value(0);
+        zero_tokens();
+    }
+
+    virtual ~verbosity_counter() = default;
+
+    virtual void xparse(boost::any& store,
+                        const std::vector<std::string>&) const {
+        store = boost::any(++m_value);
+    }
+
+    private:
+    mutable int m_value{0};
+};
+}  // namespace
+
+namespace traccc::opts {
+
+logging::logging() : interface("Logging Options") {
+    m_desc.add_options()(
+        "verbose,v", new verbosity_counter(&m_verbosity_decr),
+        "Increase verbosity (can be specified multiple times)")(
+        "quiet,q", new verbosity_counter(&m_verbosity_incr),
+        "Decrease verbosity (can be specified multiple times)");
+}
+
+logging::operator traccc::Logging::Level() const {
+    int verb = m_verbosity_incr - m_verbosity_decr;
+
+    if (verb <= -2) {
+        return traccc::Logging::Level::VERBOSE;
+    } else if (verb == -1) {
+        return traccc::Logging::Level::DEBUG;
+    } else if (verb == 0) {
+        return traccc::Logging::Level::INFO;
+    } else if (verb == 1) {
+        return traccc::Logging::Level::WARNING;
+    } else if (verb == 2) {
+        return traccc::Logging::Level::ERROR;
+    } else {
+        return traccc::Logging::Level::FATAL;
+    }
+}
+
+std::unique_ptr<configuration_printable> logging::as_printable() const {
+    auto cat = std::make_unique<configuration_category>(m_description);
+
+    traccc::Logging::Level lvl(*this);
+
+    std::string verbosity_string{Acts::Logging::levelName(lvl)};
+
+    cat->add_child(std::make_unique<configuration_kv_pair>("Verbosity level",
+                                                           verbosity_string));
+
+    return cat;
+}
+
+}  // namespace traccc::opts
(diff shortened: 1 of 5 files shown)


examples/options/src/logging.cpp
```diff
@@ -0,0 +1,95 @@
...
+}  // namespace
+
+namespace traccc::opts {
+
+logging::logging() : interface("Logging Options") {
+    m_desc.add_options()(
+        "verbose,v", new verbosity_counter(&m_verbosity_decr),
+        "Increase verbosity (can be specified multiple times)")(
+        "quiet,q", new verbosity_counter(&m_verbosity_incr),
+        "Decrease verbosity (can be specified multiple times)");
```
krasznaa, line 47: Interesting idea. But I think a way over-complicated one.

You can choose to either just make it into an integer option (my preference), or a string option. Which would be trivially translated into `traccc::Logging::Level` values.

I personally would find an `--output-level=1` flag much more user friendly than having to write `--verbose --verbose`. 🤔

examples/options/src/logging.cpp
```diff
@@ -0,0 +1,95 @@
...
+        verbosity_string = "Debug";
+    } else if (lvl == traccc::Logging::Level::INFO) {
+        verbosity_string = "Info";
+    } else if (lvl == traccc::Logging::Level::WARNING) {
+        verbosity_string = "Warning";
+    } else if (lvl == traccc::Logging::Level::ERROR) {
+        verbosity_string = "Error";
+    } else if (lvl == traccc::Logging::Level::FATAL) {
+        verbosity_string = "Fatal";
+    }
```
krasznaa, line 87: - This would seem like a perfect place for `switch` -- `case`;
  - Why change the capitalization on the printout wrt. the enum names? I would prefer to keep the same capitalized names for the printout at well.

examples/run/common/throughput_mt.ipp
```diff
@@ -85,14 +85,18 @@ int throughput_mt(std::string_view description, int argc, char* argv[],
...
          finding_opts, propagation_opts, fitting_opts, throughput_opts,
-         threading_opts},
+         threading_opts, logging_opts},
         argc,
         argv,
-        logger().cloneWithSuffix("Options")};
+        prelogger->cloneWithSuffix("Options")};
+
+    TRACCC_LOCAL_LOGGER(
+        prelogger->clone(std::nullopt, traccc::Logging::Level(logging_opts)));
```
krasznaa, line 99: To me this is a good example of why components like `traccc::opts::program_options` should have functional defaults for their logging needs. If we didn't need to supply that object a logger of its own, we could just use `traccc::getDefaultLogger(...)` here conveniently with the now available log level. 🤔

krasznaa (changes requested):
Of course we should get such options for the executables. But I'd prefer the command line interface to be a bit simpler.

examples/options/src/logging.cpp
stephenswat: The idea is that you write `-vv`, which is standard across a very broad range of CLI applications. Requiring the user to specify an integer level is much less ergonomic.

examples/options/src/logging.cpp
stephenswat: Hmm, I personally prefer to avoid switch statements whenever convenient because they more confusing and error prone than if-else statements, but I don't feel tremendously strongly about it.

Neither do I feel strongly about the capitalization! We can keep it all-caps in line with the enum values.

examples/run/common/throughput_mt.ipp
stephenswat: I still couldn't agree less with this argument. A class like `program_options` should not be making any decisions about formatting at all, and it certainly should not be printing values to any given output stream in any given format without user intervention. To have these small classes be that opinionated is not good design.

But that's outside the scope of this PR.

examples/options/src/logging.cpp
krasznaa: I can tell you that in HEP applications specifying the output level is what we usually do. Be it ROOT, Athena, etc.  I know you're right that some projects in the wild use the `-vv` formalism. But the `-qq` one I've definitely not seen yet. 🤔 So I don't think that's such a strong argument for this formalism.

I was trying to find examples underlining my case in https://github.com/acts-project/acts/tree/main/Examples/Scripts/Python, but **none** of those scripts allow you to set the output level. 😕 At least none that I would've found so far...

stephenswat: Well, I don't recall ever seeing something like `--output-level=X` in the wild, and I think there is a good reason for that:

1. The `-v` flag is so engrained in the minds of people using CLI applications that anything other than it feels foreign.
2. It is not clear what numerical verbosity levels mean; is 0 the INFO level? Is 0 the most verbose level? Is 0 the most quiet level? There is no rhyme or reason to it.
3. Writing the log levels out fully requires the user to know which levels are available, and this differs per application.

krasznaa: :white_flag: Fine, let's use `-vv`. This command line interface is not super important in the long run after all.

But you could still simplify the logic here a bit. Could you not have `verbosity_counter` be able to both increment and decrement the counter, based on a second constructor argument? Then you could have just a single integer member variable, which would have the "correct value" by the time the command line parsing finishes.

stephenswat: I tried really hard to make that work, but I have found that Boost program options starts to really break down if you try to have multiple flags set the same variable. I kept running into some edge cases that would break the situation. I agree it would be nicer to have a single variable for this; I'd be very open to any suggestions to make this work.

examples/options/src/logging.cpp
krasznaa: Also, we could just use the following heere:

https://github.com/acts-project/acts/blob/main/Core/include/Acts/Utilities/Logger.hpp#L168

stephenswat: Oh that's a really nice idea!

krasznaa:
What about this one? Anything preventing us from finishing this one?

PR #899, opened 2025-02-27
Remove placement from detector description which is not necessary

beomki-yeo:
`traccc::io::csv::read_surfaces` also needs to be fixed as it does not have to return `transform3` at all. But that will require removing entire csv support.


krasznaa:
I think by now we're basically at the point that we could drip the old-style TML geometry from the code. 🤔 I was thinking about doing that recently myself as well, when I updated the executables to use the ODD by default.

If the CI succeeds with this, I'm not against this simplification at all.

beomki-yeo:
OK!

krasznaa (approved):
👍

PR #900, opened 2025-02-28
Improve assertions in Kálmán filter and fitter

stephenswat:
This commit makes some improvements to the assertions in the Kálmán filter and fitter. Namely, it asserts that the number of measurements added in the track building kernel is the correct number, and it asserts that none of the measurements are duplicated. In the fitting kernel, the assertion on the fitting status is loosened.


beomki-yeo (approved):
Allocating a unique `measurement_id` has been optional but this PR somehow makes it mandatory, which is not bad

PR #901, opened 2025-03-03
Alpaka: Add Fitting + Finding Code

CrossR:
After far too much time, I've managed to get some time away from other work to get this into a functional and (mostly) tidy state.

This PR has:
 - Track finding + fitting for Alpaka
 - Updates to some of the existing code to bring it closer in parity to the CUDA versions
 - Updates and new example code to use said finding + fitting

This is a fairly large PR, so if required, I can split it up further (I imagine the easiest way to do that would be at the very least split up the library code vs the example code, but other suggestions welcome).

I'll stick a few comments on this draft PR for bits I know need further input, so any help there is much appreciated.

This "functions" now, but I'm still hitting a few issues, though it is also probably / possibly the same issue:
  - The seeding examples don't show exact matches all the time. Running over the 5 events from the `odd/geant4_1muon_10GeV` file, I see 48% / 60% / 78% / 46% and 100% agreement with the CPU based version...so something is going wrong, but not all the time, killing the performance.
  - The single-threaded throughput example seems to die after 1 event with a memory error, which it didn't do previously (since I had performance numbers etc out).
  - The multi-threaded throughput example isn't working, but I've also not had any chance to really do a deeper look into the interactions between TBB + Alpaka, so that one is less known at the moment.

I don't think they should be blockers though, the code is a lot more functional and nearer feature parity vs the existing code, so it would be good to get it in (when suitably reviewed, maintainable etcetc of course). Right now, when I do manage to get half a day or day to work on this, I inevitably end up spending it catching up on recent refactors or improvements, so having the code in-repo should help with that.

This branch has some very artificial commits that I made at the very end to group things into logical chunks (vs the absolute mess of commits that happened really: https://github.com/CrossR/traccc/tree/CrossR/alpaka_fitting). I've no idea how the rest of you feel about that, but figured it was better than a behemoth number of commits coming in (and still slightly better than just a single squash commit). I don't mind re-doing that if useful / better.

### DIFF
--- a/device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
+++ b/device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
@@ -66,7 +66,7 @@ struct track_params_estimation
     /// Memory resource used by the algorithm
     traccc::memory_resource m_mr;
     /// Copy object used by the algorithm
-    ::vecmem::copy& m_copy;
+    vecmem::copy& m_copy;
 };
 
 }  // namespace traccc::alpaka
--- a/device/alpaka/include/traccc/alpaka/utils/get_vecmem_resource.hpp
+++ b/device/alpaka/include/traccc/alpaka/utils/get_vecmem_resource.hpp
@@ -1,7 +1,7 @@
 /**
  * traccc library, part of the ACTS project (R&D line)
  *
- * (c) 2024 CERN for the benefit of the ACTS project
+ * (c) 2024-2025 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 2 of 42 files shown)


device/common/include/traccc/finding/device/impl/apply_interaction.ipp
```diff
@@ -17,7 +17,7 @@
 namespace traccc::device {
 
 template <typename detector_t>
-TRACCC_DEVICE inline void apply_interaction(
```
CrossR, line 20: This (and the other `N` files it is duplicated in), is obviously not really okay.

Perhaps needs a change _somewhere_ to bring this into line with what Alpaka uses, when using an Alpaka back end: https://github.com/alpaka-group/alpaka/blob/1793a0760e57bd3cca2b71170507ae1c26b5c0dd/include/alpaka/core/Common.hpp#L20-L42

This is broadly needed as Alpaka defaults to `__device__ __host__` when you use `ALPAKA_FN_ACC` as it expects...but then you hit warnings / errors of device code being called from the host.

device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
```diff
@@ -66,7 +66,7 @@ struct track_params_estimation
     /// Memory resource used by the algorithm
     traccc::memory_resource m_mr;
     /// Copy object used by the algorithm
-    ::vecmem::copy& m_copy;
+    vecmem::copy& m_copy;
```
krasznaa, line 69: This was done by mistake, right? The current version (with `::vecmem`) should be the more robust setup.

device/alpaka/include/traccc/alpaka/utils/get_vecmem_resource.hpp
```diff
@@ -58,49 +58,49 @@ class copy;
 }  // namespace sycl
 }  // namespace vecmem
 
-namespace traccc::alpaka::vecmem {
+namespace traccc::alpaka::vecmem_resources {
```
krasznaa, line 61: As long as we're changing things anyway: Do we need an extra namespace here at all? I myself would be fairly happy with just the `traccc::alpaka::host_device_types<...>`  type names. 🤔

I could possibly imagine a `traccc::alpaka::details::` namespace, in case clients of the library don't need to interact with these types normally. But other sub-namespaces should probably not be invented. 🤔

device/alpaka/src/finding/finding_algorithm.cpp
```diff
@@ -0,0 +1,483 @@
...
+    Idx threadsPerBlock = getWarpSize<Acc>() * 2;
+
+    // Copy setup
+    m_copy.setup(seeds_buffer)->ignore();
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) || defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+    auto thrustExecPolicy = thrust::device;
+#else
+    auto thrustExecPolicy = thrust::host;
+#endif
```
krasznaa, line 82: Why not provide these types through the same type traits that provide the vecmem type names at the moment? At which point those traits could really be moved out of the "vecmem specific" sub-namespace. 🤔

device/alpaka/src/utils/thread_id.hpp
```diff
@@ -21,49 +21,50 @@
...
-    unsigned int inline TRACCC_HOST_DEVICE getGridDimX() const {
+    unsigned int inline ALPAKA_FN_ACC getGridDimX() const {
         return static_cast<unsigned int>(
-            ::alpaka::getWorkDiv<::alpaka::Grid, ::alpaka::Blocks>(m_acc)[0u]);
+            ::alpaka::getWorkDiv<::alpaka::Grid, ::alpaka::Blocks>(*m_acc)[0u]);
     }
 
     private:
-    const Acc& m_acc;
+    const TAcc* m_acc;
```
krasznaa, line 63: I'm not happy about this change. 🤔

If the struct needs to be "assignable", please replace the current reference with `std::reference_wrapper<TAcc>` instead. I don't like that the struct now receives a pointer, which itself is not checked anywhere for being a `nullptr` or not.

device/common/include/traccc/finding/device/impl/apply_interaction.ipp
krasznaa: I'm not worried about this all too much. We can declare these functions as `TRACCC_HOST_DEVICE`, that should not have negative effects on the native CUDA and HIP builds as far as I'm aware. 🤔

krasznaa (changes requested):
Given the size of some of my recent PRs, this seems pretty hypocritical of me for sure. But indeed, I think this PR could be split a few ways.
  - The updates to the existing Alpaka algorithms/code;
  - The introduction of the fitting and finding algorithms;
  - The update of most of the test executables.

The second and third shouldn't be completely separated, as some test executables should be updated in the PR that introduces in the finding and fitting algorithms. But some of the executables could at least be updated in a separate PR. 🤔

That way the comments I'll have on the algorithms, which I expect will be grounds for debate, will not hold up all of the updates of this PR at least...

device/alpaka/include/traccc/alpaka/utils/get_vecmem_resource.hpp
CrossR: Possibly!

I just wasn't a fan of the `traccc::alpaka::vecmem` namespace clobbering the `vecmem::` namespace in certain cases, so I changed the name. I'll test some bits out and see what feels more natural.

device/alpaka/include/traccc/alpaka/seeding/track_params_estimation.hpp
CrossR: Slightly over-zealous refactoring when I moved from `traccc::alpaka::vecmem` to `traccc::alpaka::vecmem_resources`.

I'll revert.

CrossR:
(For context, I'm just using this draft to fix CI errors, tidy stuff up and get the changes requested above in, before splitting things up once everything is happy).

CrossR:
Split into #918, #925, #934

PR #902, opened 2025-03-03
Add debugging output to the propagation kernel

stephenswat:
This commit adds some optional debugging output to the propagation kernel, allowing it to output more precise statistics on why tracks were killed.

### DIFF
--- a/core/include/traccc/finding/details/debug_output.hpp
+++ b/core/include/traccc/finding/details/debug_output.hpp
@@ -1,0 +1,28 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2023-2025 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include "traccc/utils/logging.hpp"
+
+namespace traccc {
+enum class propagate_to_next_surface_exit_mode {
+    HOLE_LIMIT_REACHED,
+    BRANCH_LIMIT_REACHED,
+    MAX
+};
+
+struct propagate_to_next_surface_debug {
+    void reset();
+
+    unsigned int m_failure_count[static_cast<std::size_t>(
+        propagate_to_next_surface_exit_mode::MAX)];
+};
+
+void log_propagate_to_next_surface_debug(
+    const propagate_to_next_surface_debug& obj, const Logger& logger);
+}  // namespace traccc
--- a/core/include/traccc/finding/finding_config.hpp
+++ b/core/include/traccc/finding/finding_config.hpp
@@ -48,6 +48,9 @@ struct finding_config {
     /// Particle hypothesis
     detray::pdg_particle<traccc::scalar> ptc_hypothesis =
         detray::muon<traccc::scalar>();
+
+    /// Provide more verbose output for debugging; may reduce performance
+    bool verbose = false;
 };
 
 }  // namespace traccc
--- a/core/src/finding/details/debug_output.cpp
+++ b/core/src/finding/details/debug_output.cpp
@@ -1,0 +1,36 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2023-2025 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include "traccc/finding/details/debug_output.hpp"
+
+#include "traccc/utils/logging.hpp"
+
+namespace traccc {
+void propagate_to_next_surface_debug::reset() {
+    for (std::size_t i = 0;
+         i < static_cast<std::size_t>(propagate_to_next_surface_exit_mode::MAX);
+         ++i) {
+        m_failure_count[i] = 0;
+    }
+}
+
+void log_propagate_to_next_surface_debug(
+    const propagate_to_next_surface_debug& obj, const Logger& logger) {
+    if (unsigned int n = obj.m_failure_count[static_cast<std::size_t>(
+            propagate_to_next_surface_exit_mode::HOLE_LIMIT_REACHED)];
+        n > 0) {
+        TRACCC_DEBUG(n << " tracks were killed due to reaching the hole limit");
+    }
+
+    if (unsigned int n = obj.m_failure_count[static_cast<std::size_t>(
+            propagate_to_next_surface_exit_mode::BRANCH_LIMIT_REACHED)];
+        n > 0) {
+        TRACCC_DEBUG(
+            n << " tracks were killed due to reaching the branch limit");
+    }
+}
+}  // namespace traccc
--- a/device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
+++ b/device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
@@ -50,6 +50,14 @@ TRACCC_DEVICE inline void propagate_to_next_surface(
 
     if (s_pos >= cfg.max_num_branches_per_seed) {
         params_liveness[param_id] = 0u;
+
+        if (payload.debug) {
+            vecmem::device_atomic_ref<unsigned int>(
+                payload.debug->m_failure_count[static_cast<std::size_t>(
+                    propagate_to_next_surface_exit_mode::BRANCH_LIMIT_REACHED)])
+                .fetch_add(1u);
+        }
+
         return;
     }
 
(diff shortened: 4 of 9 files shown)


device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
```diff
@@ -124,6 +142,13 @@ TRACCC_DEVICE inline void propagate_to_next_surface(
     } else {
         params_liveness[param_id] = 0u;
 
+        if (payload.debug) {
+            vecmem::device_atomic_ref<unsigned int>(
+                payload.debug->m_failure_count[static_cast<std::size_t>(
+                    propagate_to_next_surface_failure_mode::DETRAY_FAILURE)])
```
beomki-yeo, line 148: Should this be tagged as a failure? I think this is where CKF propagation is terminated successfully

beomki-yeo (commented):
I like the PR in general except the naming of `failure_mode`. Couldn't this be `exit_mode` or sth like that?

device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
stephenswat: Good point, I'll remove this state for now.

stephenswat:
I've renamed it to exit mode as you requested!

core/include/traccc/finding/details/debug_output.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+struct propagate_to_next_surface_debug {
+    void reset();
+
+    unsigned int m_failure_count[static_cast<std::size_t>(
+        propagate_to_next_surface_exit_mode::MAX)];
+};
+
+void log_propagate_to_next_surface_debug(
+    const propagate_to_next_surface_debug& obj, const Logger& logger);
```
krasznaa, line 27: I wonder what our convention should be for printing custom types. 🤔

I believe if you define

```c++
std::ostream& operator<<(std::ostream&, const propagate_to_next_surface_debug&);
```

, that could be used to print such objects to a `traccc::Logger` (`Acts::Logger`) object. Like we do here for instance:
  - https://github.com/acts-project/traccc/blob/main/performance/include/traccc/performance/timing_info.hpp#L39
  - https://github.com/acts-project/traccc/blob/main/performance/include/traccc/performance/throughput.hpp#L37
  - https://github.com/acts-project/traccc/blob/main/examples/run/common/throughput_st.ipp#L214-L225

core/include/traccc/finding/details/debug_output.hpp
```diff
@@ -0,0 +1,28 @@
...
+    BRANCH_LIMIT_REACHED,
+    MAX
+};
+
+struct propagate_to_next_surface_debug {
+    void reset();
+
+    unsigned int m_failure_count[static_cast<std::size_t>(
+        propagate_to_next_surface_exit_mode::MAX)];
+};
```
krasznaa, line 24: This is fairly subjective, but I'd prefer to do this:

```c++
namespace traccc::details {
struct propagate_to_next_surface_debug_info {
    enum class exit_mode {...};
    unsigned int m_failure_count...;
    void reset();
};
}
```

While talking about this... I assume `reset()` is not meant to be called in device code. But if we explicitly want to prevent users from doing it, it might be helpful to use `TRACCC_HOST` on it. If not, if technically it may be usable in device code in some circumstances, I would put `TRACCC_HOST_DEVICE` on it right away. 🤔

core/include/traccc/finding/finding_config.hpp
```diff
@@ -48,6 +48,9 @@ struct finding_config {
     /// Particle hypothesis
     detray::pdg_particle<traccc::scalar> ptc_hypothesis =
         detray::muon<traccc::scalar>();
+
+    /// Provide more verbose output for debugging; may reduce performance
+    bool verbose = false;
```
krasznaa, line 53: I wonder if `debug` may be a better name here. But I'm really not sure...

core/src/finding/details/debug_output.cpp
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+#include "traccc/finding/details/debug_output.hpp"
+
+#include "traccc/utils/logging.hpp"
+
+namespace traccc {
+void propagate_to_next_surface_debug::reset() {
+    for (std::size_t i = 0;
+         i < static_cast<std::size_t>(propagate_to_next_surface_exit_mode::MAX);
```
krasznaa, line 15: Wouldn't `std::size(m_failure_count)` be more robust here?

core/src/finding/details/debug_output.cpp
```diff
@@ -0,0 +1,36 @@
...
+        m_failure_count[i] = 0;
+    }
+}
+
+void log_propagate_to_next_surface_debug(
+    const propagate_to_next_surface_debug& obj, const Logger& logger) {
+    if (unsigned int n = obj.m_failure_count[static_cast<std::size_t>(
+            propagate_to_next_surface_exit_mode::HOLE_LIMIT_REACHED)];
+        n > 0) {
+        TRACCC_DEBUG(n << " tracks were killed due to reaching the hole limit");
```
krasznaa, line 26: Using an output operator would also allow us to not hardcode the output level here... 🤔

device/common/include/traccc/finding/device/propagate_to_next_surface.hpp
```diff
@@ -76,6 +76,11 @@ struct propagate_to_next_surface_payload {
      * input seed
      */
     vecmem::data::vector_view<unsigned int> n_tracks_per_seed_view;
+
+    /**
+     * @brief Optional debug object
+     */
+    propagate_to_next_surface_debug* debug;
```
krasznaa, line 83:
```suggestion
    propagate_to_next_surface_debug* debug = nullptr;
```

Then you wouldn't need to touch the SYCL code if you don't want to.

examples/options/src/track_finding.cpp
```diff
@@ -102,6 +107,8 @@ std::unique_ptr<configuration_printable> track_finding::as_printable() const {
         std::to_string(m_config.max_num_skipping_per_cand)));
     cat->add_child(std::make_unique<configuration_kv_pair>(
         "PDG number", std::to_string(m_pdg_number)));
+    cat->add_child(std::make_unique<configuration_kv_pair>(
+        "Debug output", std::format("{}", m_config.verbose)));
```
krasznaa, line 111: If you call it "debug" here as well, then let's make the variable name `debug`...

krasznaa (changes requested):
I could imagine a setup in which some shared helper functions would be used by both the CUDA and SYCL code for allocating the debug object in memory, getting it back, and printing it. Since I'm a bit worried about only adding this feature to the CUDA code at the moment. (Not even the host code...?)

core/include/traccc/finding/details/debug_output.hpp
stephenswat: In general, defining `operator<<` for a type is an antipattern unless there is a single, ubiquitous way to turn a value of the type into a string. In this case, there are many different wordings and ways to print this data, so `operator<<` is a no-go. Of course, we can debate that there might be better names for this function.

core/src/finding/details/debug_output.cpp
krasznaa: It's mainly to reduce the number of characters needed to express the same. I agree that it's not more robust code. But it's certainly shorter.

stephenswat: Okay sure, I can use `std::size`. :+1:

core/include/traccc/finding/details/debug_output.hpp
krasznaa: I think it's pretty safe to define an operator for the "standard way" of printing the object. Which will also allow clients to use the code more flexibly. Printing not just to a logger, but to any kind of stream. And also not just at a `DEBUG` level, but also any other that they want. (Very realistically we may end up using such a printout in a warning or error message as well. In case the debug information points at some issue.)

So this I feel fairly strongly about.

stephenswat: Consider what happens if we want to print some exit modes at a different severity level than others; then your approach breaks down. I think it's much better to have a function which calls the logger itself, rather than producing a string for the whole object.

krasznaa:
It's a pity to leave this PR dead. Could we move ahead with it?

PR #904, opened 2025-03-05
Remove inconsistent CKF options

beomki-yeo:
`nmax-per-seed` options overlaps with `max-num-branches-per-seed`, and "Max branches per step" is different from `max_num_branches_per_seed` I guess


stephenswat (approved):
LGTM.

PR #905, opened 2025-03-06
Fix rocprim version to 6.1.1

StewMH:
Required for https://github.com/acts-project/traccc/pull/901 and fixes the compilation error seen there:
```/bld/_deps/rocthrust-build/deps/rocprim/include/rocprim/device/detail/config_types.hpp:324:78: error: use of undeclared identifier 'hipStreamLegacy'
  324 |     if(stream == default_stream || stream == hipStreamPerThread || stream == hipStreamLegacy)
```

After investigation I opened a rocThrust bug (https://github.com/ROCm/rocThrust/issues/527) - we are picking up the tip of the develop branch for rocprim (a rocThrust dependency), which is no longer compatible with rocm-6.1.1. Fix this in the patch.

Additionally, so this can be tested, start building rocThrust and related test code in HIP Alpaka flavour.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -81,7 +81,7 @@ jobs:
           - platform:
               name: "ALPAKA_HIP_SYCL"
               container: ghcr.io/acts-project/ubuntu2404_rocm_oneapi:69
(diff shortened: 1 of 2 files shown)


StewMH:
cc @CrossR

stephenswat (approved):
LGTM; the fact that RocThrust depends on a develop branch is pretty worrying!

stephenswat:
By the way, your git client is still not set up correctly; the user is set to `Stewart Martin-Haugh stewart.martin-haugh@stfc.ac.uk <smh@cern.ch>` rather than either `Stewart Martin-Haugh <smh@cern.ch>` or `Stewart Martin-Haugh <stewart.martin-haugh@stfc.ac.uk>`. Doesn't affect your ability to make PRs but it does affect GitHub's ability to credit you for them.

PR #906, opened 2025-03-07
Add executable to compare truth vs. CUDA seeding

stephenswat:
This commit adds a new executable that compares the truth seeding to the CUDA seeding, emitting a CSV file with matched seeds allowing further analysis with e.g. Python.

### DIFF
--- a/examples/run/cuda/seeding_comparison_example_cuda.cpp
+++ b/examples/run/cuda/seeding_comparison_example_cuda.cpp
@@ -1,0 +1,456 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2023-2025 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+// Project include(s).
+#include "traccc/cuda/seeding/seeding_algorithm.hpp"
+#include "traccc/cuda/utils/stream.hpp"
+#include "traccc/definitions/common.hpp"
+#include "traccc/definitions/primitives.hpp"
+#include "traccc/device/container_d2h_copy_alg.hpp"
+#include "traccc/geometry/detector.hpp"
+#include "traccc/io/read_detector.hpp"
+#include "traccc/io/read_detector_description.hpp"
+#include "traccc/io/read_measurements.hpp"
+#include "traccc/io/read_spacepoints.hpp"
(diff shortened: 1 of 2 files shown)


examples/run/cuda/seeding_comparison_example_cuda.cpp
```diff
@@ -0,0 +1,456 @@
...
+                    truth_track_candidates.at(i_trk).items.at(1);
+                const measurement& top_meas =
+                    truth_track_candidates.at(i_trk).items.at(2);
+
+                const std::map<particle, std::size_t>& bot_ptc_map =
+                    evt_data.m_meas_to_ptc_map.at(bot_meas);
+                const std::map<particle, std::size_t>& mid_ptc_map =
+                    evt_data.m_meas_to_ptc_map.at(mid_meas);
+                const std::map<particle, std::size_t>& top_ptc_map =
+                    evt_data.m_meas_to_ptc_map.at(top_meas);
```
stephenswat, line 201: @beomki-yeo I've noticed that if I get the bottom, middle, and top spacepoint positions like this, I get non-sensical $r$ progression on $79.2\%$ of seeds, i.e. the bottom $r$ isn't lower than the middle $r$ or the middle $r$ isn't lower than the top $r$. Is this expected? Do I need to sort the track candidate states?

beomki-yeo: It is not expected and the track candidates do not have to be sorted.

How did you get the spacepoint positions from `bot_meas`, `mid_meas` and `top_meas`?

beomki-yeo: After going through `event_data.cpp`, it is possible that we may need to sort the track candidate if `csv_measurements` is not sorted in the timing order:
https://github.com/acts-project/traccc/blob/6fee5ca3638f7a4572719a670047a48f05b4907c/performance/src/utils/event_data.cpp#L230

beomki-yeo: I found that `measurement.csv` of geant4 odd data does not have truth timing information (all of them are set to zero).
We will have to find the truth time from the corresponding hit object in `hits.csv`, which has truth timing information, via `measurement-simhit-map.csv`

Then you can run sorting on measurement (track candidate) vectors w.r.t. truth timing

PR #908, opened 2025-03-12
Improve memory usage in track finding postamble

stephenswat:
The code which turns the tips of our track finding into actual tracks uses an excessive amount of memory, as it massively overallocates. Indeed, it allocates memory as though all tips have the maximum number of track states, which is unrealistic. This commit makes it so that the number of valid track states is counted instead, making the allocation more precise. In my measurements, this more than halves the memory usage of traccc on $\langle\mu\rangle = 200$ ttbar events in the ODD.

### Before:

6fee5ca3

```
Reconstructed track parameters: 884796
Time totals:
                  File reading  5202 ms
              Event processing  4388 ms
Throughput:
              Event processing  438.829 ms/event, 2.27879 events/s
```

### After:

a91e94d0

```
Reconstructed track parameters: 884784
Time totals:
                  File reading  5128 ms
              Event processing  4352 ms
Throughput:
              Event processing  435.297 ms/event, 2.29728 events/s
```

Note: The SYCL changes are best-effort but I haven't tested them.

### DIFF
--- a/device/common/include/traccc/finding/device/impl/build_tracks.ipp
+++ b/device/common/include/traccc/finding/device/impl/build_tracks.ipp
@@ -41,16 +41,9 @@ TRACCC_HOST_DEVICE inline void build_tracks(
     auto L = links.at(tip);
     const unsigned int n_meas = measurements.size();
 
-    const unsigned int n_cands = L.step + 1 - L.n_skipped;
-
-    // Resize the candidates with the exact size
-    cands_per_track.resize(n_cands);
-
     // Track summary variables
     scalar ndf_sum = 0.f;
     scalar chi2_sum = 0.f;
-
-    [[maybe_unused]] std::size_t num_inserted = 0;
 
     // Reversely iterate to fill the track candidates
     for (auto it = cands_per_track.rbegin(); it != cands_per_track.rend();
@@ -64,7 +57,6 @@ TRACCC_HOST_DEVICE inline void build_tracks(
         assert(L.meas_idx < n_meas);
 
         *it = {measurements.at(L.meas_idx)};
-        num_inserted++;
 
         // Sanity check on chi2
         assert(L.chi2 < std::numeric_limits<traccc::scalar>::max());
@@ -87,10 +79,6 @@ TRACCC_HOST_DEVICE inline void build_tracks(
     }
 
 #ifndef NDEBUG
-    // Assert that we inserted exactly as many elements as we reserved
-    // space for.
-    assert(num_inserted == cands_per_track.size());
-
     // Assert that we did not make any duplicate track states.
     for (unsigned int i = 0; i < cands_per_track.size(); ++i) {
         for (unsigned int j = 0; j < cands_per_track.size(); ++j) {
--- a/device/cuda/src/finding/finding_algorithm.cu
+++ b/device/cuda/src/finding/finding_algorithm.cu
@@ -176,6 +176,9 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
         m_cfg.max_num_branches_per_seed * n_seeds, m_mr.main,
         vecmem::data::buffer_type::resizable};
     m_copy.setup(tips_buffer)->wait();
+    vecmem::data::vector_buffer<unsigned int> tip_length_buffer{
+        m_cfg.max_num_branches_per_seed * n_seeds, m_mr.main};
+    m_copy.setup(tip_length_buffer)->wait();
 
     std::map<unsigned int, unsigned int> step_to_link_idx_map;
     step_to_link_idx_map[0] = 0;
@@ -279,6 +282,7 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
                     .out_params_view = updated_params_buffer,
                     .out_params_liveness_view = updated_liveness_buffer,
                     .tips_view = tips_buffer,
+                    .tip_lengths_view = tip_length_buffer,
                     .n_tracks_per_seed_view = n_tracks_per_seed_buffer});
             TRACCC_CUDA_ERROR_CHECK(cudaGetLastError());
 
@@ -359,7 +363,8 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
                         .prev_links_idx = step_to_link_idx_map[step],
                         .step = step,
                         .n_in_params = n_candidates,
-                        .tips_view = tips_buffer});
+                        .tips_view = tips_buffer,
+                        .tip_lengths_view = tip_length_buffer});
                 TRACCC_CUDA_ERROR_CHECK(cudaGetLastError());
 
                 m_stream.synchronize();
@@ -384,12 +389,16 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
     // Get the number of tips
     auto n_tips_total = m_copy.get_size(tips_buffer);
 
+    std::vector<unsigned int> tips_length_host;
+
+    if (n_tips_total > 0) {
+        m_copy(tip_length_buffer, tips_length_host)->wait();
+        tips_length_host.resize(n_tips_total);
+    }
+
     // Create track candidate buffer
     track_candidate_container_types::buffer track_candidates_buffer{
-        {n_tips_total, m_mr.main},
-        {std::vector<std::size_t>(n_tips_total,
-                                  m_cfg.max_track_candidates_per_track),
-         m_mr.main, m_mr.host, vecmem::data::buffer_type::resizable}};
+        {n_tips_total, m_mr.main}, {tips_length_host, m_mr.main, m_mr.host}};
 
     m_copy.setup(track_candidates_buffer.headers)->ignore();
     m_copy.setup(track_candidates_buffer.items)->ignore();
--- a/device/alpaka/src/finding/finding_algorithm.cpp
+++ b/device/alpaka/src/finding/finding_algorithm.cpp
@@ -168,6 +168,9 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
         m_cfg.max_num_branches_per_seed * n_seeds, m_mr.main,
         vecmem::data::buffer_type::resizable};
     m_copy.setup(tips_buffer)->wait();
+    vecmem::data::vector_buffer<unsigned int> tip_length_buffer{
+        m_cfg.max_num_branches_per_seed * n_seeds, m_mr.main};
+    m_copy.setup(tip_length_buffer)->wait();
 
     std::map<unsigned int, unsigned int> step_to_link_idx_map;
     step_to_link_idx_map[0] = 0;
@@ -276,6 +279,7 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
                             .out_params_view = updated_params_buffer,
                             .out_params_liveness_view = updated_liveness_buffer,
                             .tips_view = tips_buffer,
+                            .tip_lengths_view = tip_length_buffer,
                             .n_tracks_per_seed_view = n_tracks_per_seed_buffer};
 
             auto bufAcc_payload =
@@ -363,7 +367,8 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
                                 .prev_links_idx = step_to_link_idx_map[step],
                                 .step = step,
                                 .n_in_params = n_candidates,
-                                .tips_view = tips_buffer};
+                                .tips_view = tips_buffer,
+                                .tip_lengths_view = tip_length_buffer};
 
                 auto bufAcc_payload =
                     ::alpaka::allocBuf<PayloadType, Idx>(devAcc, 1u);
@@ -397,12 +402,16 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
     // Get the number of tips
     auto n_tips_total = m_copy.get_size(tips_buffer);
 
+    std::vector<unsigned int> tips_length_host;
+
+    if (n_tips_total > 0) {
+        m_copy(tip_length_buffer, tips_length_host)->wait();
+        tips_length_host.resize(n_tips_total);
+    }
+
     // Create track candidate buffer
     track_candidate_container_types::buffer track_candidates_buffer{
-        {n_tips_total, m_mr.main},
-        {std::vector<std::size_t>(n_tips_total,
-                                  m_cfg.max_track_candidates_per_track),
-         m_mr.main, m_mr.host, vecmem::data::buffer_type::resizable}};
+        {n_tips_total, m_mr.main}, {tips_length_host, m_mr.main, m_mr.host}};
 
     m_copy.setup(track_candidates_buffer.headers)->ignore();
     m_copy.setup(track_candidates_buffer.items)->ignore();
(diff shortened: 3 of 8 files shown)


device/common/CMakeLists.txt
```diff
@@ -64,14 +64,12 @@ traccc_add_library( traccc_device_common device_common TYPE INTERFACE
    "include/traccc/finding/device/fill_sort_keys.hpp"
    "include/traccc/finding/device/make_barcode_sequence.hpp"
    "include/traccc/finding/device/propagate_to_next_surface.hpp"
-   "include/traccc/finding/device/prune_tracks.hpp"
```
krasznaa, line 67: Please list the new `count_tracks` files instead of the removed `prune_tracks` ones.

device/common/include/traccc/finding/device/build_tracks.hpp
```diff
@@ -49,21 +49,14 @@ struct build_tracks_payload {
...
-     */
-    track_candidate_container_types::view track_candidates_view;
+    vecmem::data::vector_view<const unsigned int> valid_tip_idx_view;
 
-    /**
-     * @brief View object to the vector of indices meeting the selection
-     * criteria
-     */
-    vecmem::data::vector_view<unsigned int> valid_indices_view;
+    const unsigned int* num_valid_tips;
```
krasznaa, line 54: 😕 If `valid_tip_idx_view` points at a resizable buffer, then why would you need a separate variable for the size?

device/common/include/traccc/finding/device/impl/count_tracks.ipp
```diff
@@ -0,0 +1,101 @@
...
+
+    // Criteria for valid tracks
+    if (num_meas >= cfg.min_track_candidates_per_track &&
+        num_meas <= cfg.max_track_candidates_per_track && success) {
+
+        vecmem::device_atomic_ref<unsigned int> num_valid_tips(
+            *payload.num_valid_tips);
+        const unsigned int trk_pos = num_valid_tips.fetch_add(1);
+        valid_tip_idx.at(trk_pos) = globalIndex;
+        valid_tip_length.at(trk_pos) = num_meas;
```
krasznaa, line 97: Instead of using two separate vectors with such a by-hand indexing, would it not be better/easier to use a `vecmem::data::vector_buffer<traccc::tuple<unsigned int, unsigned int> >` type instead, which could be filled with VecMem's built-in features? 🤔

One of the things that I'd like to achieve with VecMem is to provide a toolset that allows us not to always use low level code in our algorithms. If we need to add more helpers for that in VecMem, let's do that. But until then we should at least make use of the features that it already has.

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -344,71 +345,74 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
 
-        // Global counter object: Device -> Host
+        unsigned int n_valid_tips = 0;
+
         TRACCC_CUDA_ERROR_CHECK(cudaMemcpyAsync(
-            &n_valid_tracks, n_valid_tracks_device.get(), sizeof(unsigned int),
+            &n_valid_tips, n_valid_tips_device.get(), sizeof(unsigned int),
             cudaMemcpyDeviceToHost, stream));
 
         m_stream.synchronize();
```
krasznaa, line 382: If we use a resizable vector, all this could become simply:
```c++
const unsigned int n_valid_tips = m_copy.get_size(valid_tips_buffer);
```
(Assuming that we merge `valid_tip_idx_buffer` and `valid_tip_length_buffer` into a single resizable buffer called `valid_tips_buffer`.)

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -344,71 +345,74 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
+                                            sizeof(unsigned int), stream));
 
-    // @Note: nBlocks can be zero in case there is no tip. This happens when
-    // chi2_max config is set tightly and no tips are found
-    if (n_tips_total > 0) {
-        vecmem::unique_alloc_ptr<unsigned int> n_valid_tracks_device =
-            vecmem::make_unique_alloc<unsigned int>(m_mr.main);
-        TRACCC_CUDA_ERROR_CHECK(cudaMemsetAsync(n_valid_tracks_device.get(), 0,
-                                                sizeof(unsigned int), stream));
+    std::vector<unsigned int> valid_tips_length_host;
```
krasznaa, line 363: Asynchronous copying will not work into an `std::vector`. 😦 You'll need to allocate the host memory using `vecmem::cuda::host_memory_resource` for that.

device/cuda/src/finding/finding_algorithm.cu
```diff
@@ -344,71 +345,74 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
...
         m_stream.synchronize();
+
+        valid_tips_length_host.resize(n_valid_tips);
+
+        TRACCC_CUDA_ERROR_CHECK(cudaMemcpyAsync(
+            valid_tips_length_host.data(), valid_tip_length_buffer.ptr(),
+            n_valid_tips * sizeof(unsigned int), cudaMemcpyDeviceToHost,
+            stream));
+
+        m_stream.synchronize();
```
krasznaa, line 391: Since you only need one of the vectors on the host, maybe the buffer with `tuple<uint, uint>` elements is not the best after all. But we could still make an SoA for this purpose. 🤔 In which case the low level code would still be done by VecMem, which the algorithmic code just "pushing back" into the container.

krasznaa (changes requested):
So, count+fill it ended up being after all. 🤔 Good! I'm very on board with doing this consistently with what we found to be the most efficient during seed finding.

But I'd really like to reduce the number of individual memory allocations that we need for this. 🤔 With a single (SoA?) container for this new payload we could get away with allocating all needed GPU memory for the counting in one go.

device/common/include/traccc/finding/device/impl/count_tracks.ipp
stephenswat: Having it in one array allows me to pass the vector straight into the jagged vector constructor, which is really nice. No need to then transform the data on the host.

device/cuda/src/finding/finding_algorithm.cu
stephenswat: Asynchronous memory copies are well defined on non-pinned memory! See the last note on the documentation for `cudaMemcpyAsync`.

device/common/include/traccc/finding/device/build_tracks.hpp
stephenswat: No real reason, array + pointer is just more clear and ergonomic in my opinion.

beomki-yeo (commented):
I am happy with the changes as long as the CI passes

device/cuda/src/finding/finding_algorithm.cu
krasznaa: 😕 Is that happening with some new CUDA / driver version? Because in the past NSys was definitely not reporting any asynchronicity without using pinned host memory. I'll need to see a profile to believe this...

stephenswat:
I was able to one-up myself, removing the counting kernel entirely, reducing the memory allocations, keeping the memory usage low, and improving performance:

```
Reconstructed track parameters: 884809
Time totals:
                  File reading  5144 ms
              Event processing  4230 ms
Throughput:
              Event processing  423.022 ms/event, 2.36394 events/s
```

device/common/CMakeLists.txt
stephenswat: This no longer exists in the new version.

device/cuda/src/finding/finding_algorithm.cu
stephenswat: This is no longer relevant in the updated version.

device/cuda/src/finding/finding_algorithm.cu
krasznaa: 🤔 Maybe not right now. But once we'll "give up" the CPU thread during synchronization points, it will be advantageous to set up the transfer, then go and start doing other things, and finally return to the thread once the copy (and everything before it) is done.

So we might as well do it with future proofing in mind already now.

device/common/include/traccc/finding/device/impl/build_tracks.ipp
```diff
@@ -86,64 +60,45 @@ TRACCC_DEVICE inline void build_tracks(const global_index_t globalIndex,
...
         } else {
             L = links[L.previous.first][L.previous.second];
         }
     }
 
+    final_candidates.at(globalIndex).header.seed_params = seeds.at(seed_idx);
+    final_candidates.at(globalIndex).header.trk_quality.chi2 = chi2_sum;
+    final_candidates.at(globalIndex).header.trk_quality.ndf = ndf_sum - 5.f;
+    final_candidates.at(globalIndex).header.trk_quality.n_holes =
+        final_n_skipped;
```
krasznaa, line 89: I honestly don't know this. Since the compiler might already do this as efficiently as possible.

But would it not be more efficient to write this as:

```c++
final_candidates.at(globalIndex).header = {...};
```

? Note that I'm very aware that I'm also talking against my SoA efforts with this. 😛 Since if this formalism is indeed more efficient, the SoA migration will make this part worse. 😦

krasznaa (commented):
I'm still very much on board with the direction.

Could you post a bit of a description of what you did though? I had to realize that it's a bit too much code for me to interpret from scratch. 😦 Just some main bullet points about how the algorithm finds the correct sizes now.

device/common/include/traccc/finding/device/impl/build_tracks.ipp
stephenswat: Mwah, in this case the efficiency difference is going to be so small that it is not worth caring about. That said I do prefer the syntax with aggregate notation, so I'll use that. :+1:

stephenswat:
So the responsibility of the kernels in the old version of the algorithm was as follows:

* `propagate_to_next_surface` propagates the tracks and creates _tips_, i.e. points where a track might have ended. Each tracks corresponds 1:1 with a surface link, which describes the head of a reversed linked list of candidates.
* `build_tracks` follows the linked list of candidates starting at each tip, computes some properties about that track, and places _all_ tracks (regardless of whether they make sense or not) in a very large jagged array
* `prune_tracks` looks at the reified tracks in the big jagged vector and copies the good ones (based on the properties determined by the `build_tracks` kernel) into the final jagged vector.

Now, the core insight is that the checks that `build_tracks` performs are extremely rudimentary. It's literally just that the number of non-hole candidates is between a lower and an upper bound. All other cuts are already performed by the `propagate_to_next_surface`. Thus, the core of this PR is to move the checks that `build_tracks` does to `propagate_to_next_surface`, which has all the necessary information already. The core idea is that we create only tips for valid tracks, rather than tips for _all_ possible tracks. This, of course, also makes it trivial to count the number of valid tracks and how long those tracks will be. So in the new model, the tasks are distributed as follows:

* `propagate_to_next_surface` propagates the tracks and creates tips _only for valid tracks_, where the validity is defined by some very simple set of metrics. The definition of tips remain the same, but we create far fewer of them, and we record the length of each track.
* `build_tracks` no longer does any filtering, it just follows the linked list and places track states into a much smaller jagged vector.

Much simpler, and it has far fewer kernels. Note that in this new model `build_tracks` basically does what `prune_tracks` did in the previous model.

beomki-yeo:
> The core idea is that we create only tips for valid tracks, rather than tips for all possible tracks. 

Yeah I think this is a right way to improve :thinking:

device/cuda/src/finding/finding_algorithm.cu
krasznaa: I would still prefer this to use `m_mr.host`.

krasznaa (changes requested):
Thanks for the description!

Let's just make the internal vector use the correct memory resource, and this one can go ahead as far as I'm concerned.

device/cuda/src/finding/finding_algorithm.cu
stephenswat: This isn't currently possible because the jagged vector constructor in vecmem only accepts default-allocated `std::vector` capacities. If I try passing a PMR vector with pinned host memory, vecmem will not accept it.

stephenswat:
This PR also fixes a bug in the efficiency measurements by the way! Wink wink nudge nudge :wink:

stephenswat:
# Performance summary
## Graphical

## Tabular
| Kernel | 348de1a2 | 921c8768 | Delta |
| --- | --: | --: | --: |
| `fit` | 277.88 ms | 256.77 ms | -7.6% |
| `propagate_to_next_surface` | 118.17 ms | 117.00 ms | -1.0% |
| `find_tracks` | 26.65 ms | 23.00 ms | -13.7% |
| `count_triplets` | 14.16 ms | 14.16 ms | -0.0% |
| `find_triplets` | 5.98 ms | 5.98 ms | 0.0% |
| `find_doublets` | 911.31 μs | 893.99 μs | -1.9% |
| `Thrust::sort` | 757.86 μs | 758.34 μs | 0.1% |
| `ccl_kernel` | 682.50 μs | 682.54 μs | 0.0% |
| `count_doublets` | 627.27 μs | 628.01 μs | 0.1% |
| `build_tracks` | 1.10 ms | 445.35 μs | -59.6% |
| `prune_tracks` | 1.23 ms | &mdash; | &mdash; |
| `select_seeds` | 362.74 μs | 358.26 μs | -1.2% |
| `apply_interaction` | 132.45 μs | 131.94 μs | -0.4% |
| `update_triplet_weights` | 96.16 μs | 96.23 μs | 0.1% |
| `fill_sort_keys` | 68.25 μs | 67.72 μs | -0.8% |
| `estimate_track_params` | 34.31 μs | 34.31 μs | 0.0% |
| `populate_grid` | 30.40 μs | 30.45 μs | 0.2% |
| `count_grid_capacities` | 29.21 μs | 29.18 μs | -0.1% |
| `form_spacepoints` | 14.24 μs | 14.20 μs | -0.3% |
| `reduce_triplet_counts` | 6.71 μs | 6.71 μs | 0.0% |
| `static_kernel` | 1.78 μs | 1.78 μs | 0.1% |
| `make_barcode_sequence` | 992.11 ns | 1.00 μs | 1.3% |
| `fill_prefix_sum` | 165.57 ns | 165.51 ns | -0.0% |
| **Total** | **448.93 ms** | **421.08 ms** | **-6.2%** |
> [!NOTE]

stephenswat:
Reminder that this important PR has been blocked for almost a month now. :wink:

beomki-yeo:
@krasznaa THere are indeed many PRs pending :thinking:

krasznaa (approved):
After quite some delay (sorry about that :frowning:), let's indeed finally get this in.

After resolving the (hopefully trivial... :crossed_fingers:) merge conflicts.

krasznaa:
Any movement here?

stephenswat:
@krasznaa this can be unblocked now. :slightly_smiling_face:

krasznaa (approved):
👍

PR #911, opened 2025-03-14
Fix: uninitialized value when the first step is skipped

fiedlerp:
The counting loop assumed the first step is never skipped. The counting loop can be avoided altogether, as the number of skipped steps is already present in the link.


stephenswat:
Mind you this is very similar to what we're doing in #908; please take a look at that. But in the meanwhile I am very happy to approve this if it passes the CI.

stephenswat (approved):

fiedlerp:
This is just a bug fix for an issue that stumped me for 2 days. This is not an attempt to improve the preamble.

PR #914, opened 2025-03-16
Update detray to v0.92.0

beomki-yeo:
This PR updates detray to v0.92.0.

As the propagator state sets the volume in the constructor, we don't need to call `set_volume` anymore


stephenswat (approved):

stephenswat:
CI failures seem related to #911.

PR #918, opened 2025-03-18
Alpaka: Algorithm Updates (PR 1 of 3)

CrossR:
This splits out the updates to existing code from #901 into its own PR.

I then have a branch for the new finding + fitting code in a second branch, and finally the updated + new examples benefiting from that finding + fitting code in a final branch.

What is the most appropriate way to PR that?

Right now, I have them all branching off each other, i.e.

main -> CrossR/AlpakaUpdates branch
CrossR/AlpakaUpdates -> CrossR/AlpakaFindingAndFitting branch
CrossR/AlpakaFindingAndFitting -> CrossR/AlpakaFittingExamples branch

Since the branches build on each other, ideally we would PR each in sequence.
That is a bit awkward, as the finding + fitting code will be going in without any examples to use it for a bit, but hopefully the delay between each isn't so large to be a pain.

I will almost certainly need some follow up work to fix up bugs and known limitations in the current finding + fitting code, but this set of work is large enough as is, so would be good to start the process of getting things in.

Hopefully...this branch and the others should have been made in such a way that it builds etc without issue.

### DIFF
--- a/device/alpaka/src/clusterization/measurement_sorting_algorithm.cpp
+++ b/device/alpaka/src/clusterization/measurement_sorting_algorithm.cpp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2024 CERN for the benefit of the ACTS project
+ * (c) 2024-2025 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -8,9 +8,8 @@
 // Library include(s).
 #include "traccc/alpaka/clusterization/measurement_sorting_algorithm.hpp"
 
-#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+#include <thrust/execution_policy.h>
 #include <thrust/sort.h>
-#endif
 
 namespace traccc::alpaka {
 
@@ -29,16 +28,15 @@ measurement_sorting_algorithm::operator()(
         m_copy.get().get_size(measurements_view);
 
     // Sort the measurements in place
-    // TODO: This obviously won't work for HIP, but we don't have a HIP
-    // equivalent of Thrust yet.
-#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
-    thrust::sort(thrust::device, measurements_view.ptr(),
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) || defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+    auto thrustExecPolicy = thrust::device;
+#else
+    auto thrustExecPolicy = thrust::host;
+#endif
+
+    thrust::sort(thrustExecPolicy, measurements_view.ptr(),
                  measurements_view.ptr() + n_measurements,
                  measurement_sort_comp());
-#elif !defined(ALPAKA_ACC_GPU_HIP_ENABLED)
-    std::sort(measurements_view.ptr(), measurements_view.ptr() + n_measurements,
-              measurement_sort_comp());
-#endif
 
     // Return the view of the sorted measurements.
     return measurements_view;
--- a/device/alpaka/src/seeding/seed_finding.cpp
+++ b/device/alpaka/src/seeding/seed_finding.cpp
@@ -210,9 +210,7 @@ edm::seed_collection::buffer seed_finding::operator()(
     auto queue = Queue{devAcc};
     auto const deviceProperties = ::alpaka::getAccDevProps<Acc>(devAcc);
     auto maxThreads = deviceProperties.m_blockThreadExtentMax[0];
-    Idx threadsPerBlock = getWarpSize<Acc>() * 2 < maxThreads
-                              ? getWarpSize<Acc>() * 2
-                              : maxThreads;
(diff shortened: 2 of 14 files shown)


device/alpaka/src/clusterization/measurement_sorting_algorithm.cpp
```diff
@@ -29,16 +28,15 @@ measurement_sorting_algorithm::operator()(
...
-    // equivalent of Thrust yet.
-#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
-    thrust::sort(thrust::device, measurements_view.ptr(),
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) || defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+    auto thrustExecPolicy = thrust::device;
+#else
+    auto thrustExecPolicy = thrust::host;
+#endif
+
+    thrust::sort(thrustExecPolicy, measurements_view.ptr(),
```
CrossR, line 37: Attila has already pointed out that this could / should be extracted out to somewhere more general (likely swapping the `alpaka::vecmem_resource` namespace to something more generic, and putting it there).

My preference would be to tackle that in a follow up, just because there is some other slightly larger work that is also needed around making the code generic for all backends (mostly, making the interface for getting vecmem resources truly unified, as currently SYCL is still treated as an edge case with an `ifdef`, as shown in the example updates below).

device/alpaka/src/seeding/seed_finding.cpp
```diff
@@ -210,7 +210,9 @@ edm::seed_collection::buffer seed_finding::operator()(
     auto queue = Queue{devAcc};
     auto const deviceProperties = ::alpaka::getAccDevProps<Acc>(devAcc);
     auto maxThreads = deviceProperties.m_blockThreadExtentMax[0];
-    auto threadsPerBlock = maxThreads;
+    Idx threadsPerBlock = getWarpSize<Acc>() * 2 < maxThreads
+                              ? getWarpSize<Acc>() * 2
+                              : maxThreads;
```
stephenswat, line 215: This is just `std::min`.

device/alpaka/src/seeding/spacepoint_binning.cpp
```diff
@@ -90,11 +90,8 @@ traccc::details::spacepoint_grid_types::buffer spacepoint_binning::operator()(
         grid_capacities_buff;
 
     // Now define the Alpaka Work division
-    auto const deviceProperties = ::alpaka::getAccDevProps<Acc>(devAcc);
-    auto const maxThreadsPerBlock = deviceProperties.m_blockThreadExtentMax[0];
-    auto const threadsPerBlock = maxThreadsPerBlock;
-    auto const blocksPerGrid =
-        (sp_size + threadsPerBlock - 1) / threadsPerBlock;
+    const Idx threadsPerBlock = getWarpSize<Acc>() * 8;
+    const Idx blocksPerGrid = (sp_size + threadsPerBlock - 1) / threadsPerBlock;
```
stephenswat, line 94: Is there a reason why you are not performing the check to see if the block size exceeds the device maximum?

device/alpaka/src/seeding/track_params_estimation.cpp
```diff
@@ -59,10 +59,9 @@ track_params_estimation::output_type track_params_estimation::operator()(
 
     auto devAcc = ::alpaka::getDevByIdx(::alpaka::Platform<Acc>{}, 0u);
     auto queue = Queue{devAcc};
-    auto const deviceProperties = ::alpaka::getAccDevProps<Acc>(devAcc);
-    auto const threadsPerBlock = deviceProperties.m_blockThreadExtentMax[0];
+    const Idx threadsPerBlock = getWarpSize<Acc>() * 2;
 
-    auto blocksPerGrid = (seeds_size + threadsPerBlock - 1) / threadsPerBlock;
+    Idx blocksPerGrid = (seeds_size + threadsPerBlock - 1) / threadsPerBlock;
```
stephenswat, line 64: Same comment as above. Seems inconsistent.

device/alpaka/src/seeding/spacepoint_binning.cpp
CrossR: It mirrors the lack of checking in the CUDA version (actually, CUDA doesn't check the threads either I don't think?)....which isn't really a strong argument, more just that I mirrored the style there.

Perhaps a more repo wide discussion we need to have?

It seems right now we assume we are so far under the limits, it doesn't really matter....which is reasonable until it isn't. I've certainly hit some issues when running in FP64 mode, as some of the thread variables don't fit when using `double` vs `float` etc.

stephenswat: Well, the reason for this is that the limit in CUDA is very clearly defined at 1024. I don't know how this limit is set in Alpaka, especially for e.g. CPU devices. Do you know more about this?

StewMH: I didn't find anything about the limit in Alpaka documentation so far

stephenswat: Okay, I'm sure we can figure this out later. Let's move forward with this PR.

PR #920, opened 2025-03-19
Fix `TRACCC_USE_SPACK_LIBS` CMake variable

stephenswat:
Due to an error on my part this was previously not doing anything. This PR fixes the behaviour of this flag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,7 @@ option( TRACCC_BUILD_EXAMPLES "Build the examples of traccc" TRUE )
 
 # Flags controlling what traccc should use.
 option( TRACCC_USE_SYSTEM_LIBS "Use system libraries be default" FALSE )
-option( TRACCC_USE_SPACK_LIBS "Use pack libraries by default" FALSE )
+option( TRACCC_USE_SPACK_LIBS "Use Spack libraries by default" FALSE )
 option( TRACCC_USE_ROOT "Use ROOT in the build (if needed)" TRUE )
 
 # Check CUDA and SYCL C++ standards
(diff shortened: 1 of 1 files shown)


krasznaa (commented):
I was clearly not paying full attention here before. 🤔 What does `TRACCC_USE_SPACK_LIBS` do exactly, at least in this file, that `TRACCC_USE_SYSTEM_LIBS` doesn't do? I.e. why do we need this variable actually?

I would argue that the logic of what dependencies traccc builds itself, and what dependencies come from Spack, should live in Spack code. 🤔 I would leave it up to that code to turn the individual cache variables on or off, depending on how it wants to build traccc.

So, what's the actual purpose of this variable in our current code exactly? 😕

stephenswat:
> I was clearly not paying full attention here before. 🤔 What does `TRACCC_USE_SPACK_LIBS` do exactly, at least in this file, that `TRACCC_USE_SYSTEM_LIBS` doesn't do? I.e. why do we need this variable actually?

`TRACCC_USE_SYSTEM_LIBS` is for fully offline builds, i.e. you have all dependencies installed locally. `TRACCC_USE_SPACK_LIBS` is in sync with `spack.yaml`, i.e. it ensures that all the libraries installed by Spack are not redundantly built. For now, dependencies like detray and vecmem are not yet in `spack.yaml`, so they need to be handled differently by `TRACCC_USE_SYSTEM_LIBS` and `TRACCC_USE_SPACK_LIBS`.

krasznaa (approved):
I see.

The code repetition could be helped by introducing a small macro for defining all these cache variables, but fine, we can have them for now like this as well.

PR #921, opened 2025-03-20
Make CKF types and initializers more readable

stephenswat:
This commit aims to improve the readability of the CKF by replacing some pairs (with their non-descriptive `.first` and `.second` accessors) with dedicated structs and by adding explicit named member initialization to some larger POD structures.


beomki-yeo (approved):
Makes sense

PR #923, opened 2025-03-22
Add alt greedy ambiguity solver

beomki-yeo:
Still need to make bunch of tests and compare with existing algorithm

### DIFF
--- a/core/include/traccc/ambiguity_resolution/alt_greedy_ambiguity_resolution_algorithm.hpp
+++ b/core/include/traccc/ambiguity_resolution/alt_greedy_ambiguity_resolution_algorithm.hpp
@@ -19,7 +19,7 @@ namespace traccc {
 /// Evicts tracks that seem to be duplicates or fakes. This algorithm takes a
 /// greedy approach in the sense that it will remove the track which looks "most
 /// duplicate/fake"
-class greedy_ambiguity_resolution_algorithm
+class alt_greedy_ambiguity_resolution_algorithm
     : public algorithm<track_state_container_types::host(
           const typename track_state_container_types::host&)>,
       public messaging {
--- a/core/include/traccc/ambiguity_resolution/resolution_config.hpp
+++ b/core/include/traccc/ambiguity_resolution/resolution_config.hpp
@@ -11,17 +11,23 @@
 #include "traccc/definitions/common.hpp"
 #include "traccc/definitions/primitives.hpp"
 
+// System include(s).
+#include <cstdint>
+
 namespace traccc {
+
+enum class resolution_status : uint32_t { UNKNOWN, REJECT, ACCEPT, MAX_STATUS };
 
 /// Configuration struct for ambiguity resolution
 struct resolution_config {
 
     /// Minimum number of measurement to form a track.
-    unsigned int min_measurements_per_track = 3;
+    unsigned int min_meas_per_track = 3;
 
-    /// Maximum amount of shared hits per track. One (1) means "no shared
-    /// hit allowed".
-    unsigned int maximum_shared_hits = 1;
+    /// Minimum number of shared measurement to be sent for competition with
+    /// other tracks. If the number of shared measurement is smaller, the track
+    /// is accepted right away
+    unsigned int min_shared_meas_for_competition = 1;
 };
 
 }  // namespace traccc
--- a/core/include/traccc/ambiguity_resolution/resolution_config.hpp
+++ b/core/include/traccc/ambiguity_resolution/resolution_config.hpp
@@ -1,0 +1,30 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2025 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Local include(s).
+#include "traccc/definitions/common.hpp"
+#include "traccc/definitions/primitives.hpp"
+
+namespace traccc {
+
+/// Configuration struct for ambiguity resolution
+struct resolution_config {
+
+    /// Minimum number of measurement to form a track.
+    unsigned int min_measurements_per_track = 3;
+
+    /// Maximum amount of shared hits per track. One (1) means "no shared
+    /// hit allowed".
+    unsigned int maximum_shared_hits = 1;
+
+    /// Maximum number of iterations.
+    unsigned int maximum_iterations = 1000000;
+};
+
+}  // namespace traccc
--- a/core/include/traccc/ambiguity_resolution/resolution_config.hpp
+++ b/core/include/traccc/ambiguity_resolution/resolution_config.hpp
@@ -22,9 +22,6 @@ struct resolution_config {
     /// Maximum amount of shared hits per track. One (1) means "no shared
     /// hit allowed".
     unsigned int maximum_shared_hits = 1;
-
-    /// Maximum number of iterations.
-    unsigned int maximum_iterations = 1000000;
 };
 
 }  // namespace traccc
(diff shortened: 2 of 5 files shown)


core/include/traccc/ambiguity_resolution/alt_greedy_ambiguity_resolution_algorithm.hpp
```diff
@@ -0,0 +1,50 @@
...
+#pragma once
+
+// Project include(s).
+#include "traccc/edm/track_state.hpp"
+#include "traccc/utils/algorithm.hpp"
+#include "traccc/utils/messaging.hpp"
+
+// Greedy ambiguity resolution adapted from ACTS code
+
+namespace traccc {
```
krasznaa, line 17: Please use the `traccc::host` namespace.

core/include/traccc/ambiguity_resolution/alt_greedy_ambiguity_resolution_algorithm.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+    public:
+    using config_type = resolution_config;
+
+    /// Constructor for the greedy ambiguity resolution algorithm
+    ///
+    /// @param cfg  Configuration object
+    greedy_ambiguity_resolution_algorithm(
+        const config_type& cfg,
+        std::unique_ptr<const Logger> logger = getDummyLogger().clone())
```
krasznaa, line 35: The constructor will need a `vecmem::memory_resource` as well, to create its output container with.

core/include/traccc/ambiguity_resolution/alt_greedy_ambiguity_resolution_algorithm.hpp
```diff
@@ -0,0 +1,50 @@
...
+        : messaging(std::move(logger)), m_config{cfg} {}
+
+    /// Run the algorithm
+    ///
+    /// @param track_states the container of the fitted track parameters
+    /// @return the container without ambiguous tracks
+    track_state_container_types::host operator()(
+        const typename track_state_container_types::host& track_states)
```
krasznaa, line 43: Let's make it receive

```suggestion
        const track_state_container_types::const_view& track_states)
```

instead. So that buffers could also be given to it.

P.S. You shouldn't use `typename` in such situations. You would only need that keyword if `track_state_container_types` were a template type. But that is a concrete type. So the compiler will know that it's a typename.

core/include/traccc/ambiguity_resolution/alt_greedy_ambiguity_resolution_algorithm.hpp
```diff
@@ -0,0 +1,50 @@
...
+        const override;
+
+    /// Algorithm configuration
+    config_type m_config;
```
krasznaa, line 47: Not super important, but you might as well make it into a private member.

core/include/traccc/ambiguity_resolution/resolution_config.hpp
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+#pragma once
+
+// System include(s).
+#include <cstdint>
+
+namespace traccc {
+
+enum class resolution_status : uint32_t { UNKNOWN, REJECT, ACCEPT, MAX_STATUS };
```
krasznaa, line 15: Is this type meant to be given back to the user? (Currently it's only used inside the algorithm, with the user not seeing it.)

If it's an internal type, you might as well move its declaration into the `.cpp` file. 🤔

krasznaa (changes requested):
Since this seems to be a different logic to our existing algorithm, let's come up with a better name for it. One with which it could co-exist with @SylvainJoube's algorithm. (I **really** don't like these `alt_` prefixes...)

I.e. let's replace "greedy" with some other appropriate word. 🤔 Since I admit I didn't try to understand how the code actually works, I'd leave this "artistic" step to you. 😉

beomki-yeo:
It may look different but I am trying make the same greedy ambiguity resolution algorithm.
I will just replace the algorithm if the difference is reasonably small

core/include/traccc/ambiguity_resolution/alt_greedy_ambiguity_resolution_algorithm.hpp
beomki-yeo: Fixed

core/include/traccc/ambiguity_resolution/alt_greedy_ambiguity_resolution_algorithm.hpp
beomki-yeo: Fixed as suggested

core/include/traccc/ambiguity_resolution/alt_greedy_ambiguity_resolution_algorithm.hpp
beomki-yeo: Right. Fixed

core/include/traccc/ambiguity_resolution/alt_greedy_ambiguity_resolution_algorithm.hpp
beomki-yeo: Fixed as suggested

core/include/traccc/ambiguity_resolution/resolution_config.hpp
beomki-yeo: Moved to .cpp now

beomki-yeo:
Sonar cloud warning on `std::mt19937` can be neglected

core/include/traccc/ambiguity_resolution/resolution_config.hpp
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+
+// System include(s).
+#include <cstdint>
+#include <limits>
+
+namespace traccc {
+
+/// Configuration struct for ambiguity resolution
+struct resolution_config {
```
stephenswat, line 17: Nit: Should this be called `ambiguiy_resolution_config` to avoid confusion with configuration for e.g. resolution plots?

core/include/traccc/ambiguity_resolution/legacy/greedy_ambiguity_resolution_algorithm.hpp
```diff
@@ -0,0 +1,140 @@
...
+        /// Maximum amount of shared hits per track. One (1) means "no shared
+        /// hit allowed".
+        std::uint32_t maximum_shared_hits = 1;
+
+        /// Maximum number of iterations.
+        std::uint32_t maximum_iterations = 1000000;
+
+        /// Minimum number of measurement to form a track.
+        std::size_t n_measurements_min = 3;
+    };
```
stephenswat, line 66: Isn't this basically the same as `resolution_config` below?

stephenswat (approved):
Looks good to me; some minor comments but feel free to go ahead as is.

core/include/traccc/ambiguity_resolution/resolution_config.hpp
beomki-yeo: Makes sense. I changed to `ambiguity_resolution_config`

core/include/traccc/ambiguity_resolution/legacy/greedy_ambiguity_resolution_algorithm.hpp
beomki-yeo: Right but I just want to leave the legacy algorithm as is

beomki-yeo:
@krasznaa PR still needs your approval as you requested the changes

beomki-yeo:
> If I remember correctly, you'll want to remove it completely later on, right?

Correct

PR #924, opened 2025-03-25
Allocate CKF links in a flat allocation

stephenswat:
This commit migrates the link allocation in the combinatorial Kálmán filter from using a jagged buffer to using a contiguous, flat buffer. This has several advantages:

1. It generally reduces the complexity of the code.
2. It reduces the number of allocations and copies in the code.
3. It removes the usage of awkward jagged vectors.
4. It increases memory efficiency.
5. It provides kernels with the ability to view links from all steps, not just the current and previous one.


device/common/include/traccc/finding/device/impl/build_tracks.ipp
```diff
@@ -105,8 +102,8 @@ TRACCC_DEVICE inline void build_tracks(const global_index_t globalIndex,
         for (unsigned int i = 0; i < cands_per_track.size(); ++i) {
             for (unsigned int j = 0; j < cands_per_track.size(); ++j) {
                 if (i != j) {
-                    assert(cands_per_track.at(i).measurement_id !=
-                           cands_per_track.at(j).measurement_id);
+                    // assert(cands_per_track.at(i).measurement_id !=
+                    //       cands_per_track.at(j).measurement_id);
```
krasznaa, line 106: Is this test meant to be revived in some way later on? If not, let's remove the whole thing.

While the compilers were probably smart enough to completely remove this double for-loop from the code in optimized code already, such tests would generally be nicer to do through some lambda. With the entire body of the code sitting inside of the `assert(...)` parentheses.

But for now just removing this entire block would be fine. 🤔

krasznaa (commented):
Apart from our SoA discussion, which I still don't feel super strongly about (so the PR's code may just as well go ahead), do you have some results for what it does to the performance of the code? Apart from testing that it would still find the exact same tracks that it did so far... 🤔

stephenswat:
> Apart from our SoA discussion, which I still don't feel super strongly about (so the PR's code may just as well go ahead), do you have some results for what it does to the performance of the code? Apart from testing that it would still find the exact same tracks that it did so far... 🤔

Let me see if pcadp04 is free and I'll make some!

beomki-yeo (approved):
Looks good to me

stephenswat:
### Performance changes



Note that this doesn't take into account performance changes outside of kernels. Throughput example results:

### Before

```
$ ./build/bin/traccc_throughput_st_cuda --input-directory=/data/Acts/odd-simulations-20240506/geant4_ttbar_mu200 --use-detray-detector=on --input-events=10 --cold-run-events 10 --processed-events 100 --use-acts-geom-source=on --deterministic
Reconstructed track parameters: 8846209
Time totals:
                  File reading  5142 ms
            Warm-up processing  8292 ms
              Event processing  50901 ms
Throughput:
            Warm-up processing  829.291 ms/event, 1.20585 events/s
              Event processing  509.015 ms/event, 1.96458 events/s
```

### After

```
$ ./build/bin/traccc_throughput_st_cuda --input-directory=/data/Acts/odd-simulations-20240506/geant4_ttbar_mu200 --use-detray-detector=on --input-events=10 --cold-run-events 10 --processed-events 100 --use-acts-geom-source=on --deterministic
Reconstructed track parameters: 8846324
Time totals:
                  File reading  5456 ms
            Warm-up processing  181880 ms
              Event processing  50846 ms
Throughput:
            Warm-up processing  18188.1 ms/event, 0.0549811 events/s
              Event processing  508.465 ms/event, 1.9667 events/s
```

krasznaa (approved):
It's a pity that no performance improvement comes from this. 😦 But at least things are not getting worse either.

PR #925, opened 2025-03-26
Alpaka: Finding + Fitting Algorithms (PR 2 of 3)

CrossR:
This follows on from #918, and adds the new finding + fitting algorithms for the Alpaka implementation.

The final PR after this, should add the final updates to all the examples to enable finding + fitting, to allow testing of the Alpaka implementation.

Running total of follow up work:

 - "Unify Everything" : I.e. work to condense any / all accelerator specific code to a single place. The two main places that is still needed is the `thrust` device selection and also a unified interface for vecmem (we are almost there on that, but we need to work out the best way to optionally take a SYCL queue in, as we have to special case SYCL for that reason at the mo).
 - Fix any remaining bugs that exist in this implementation. When running all the work together, I'm seeing some instability, especially at higher values of mu, where the alpaka implementation will crash. That needs investigating, but potentially after....
 - Async-ify. Right now, we using a blocking queue and lots of `::alpaka::wait(queue)` calls. That is causing issues in at least one place (part of the instability mentioned above is due to the seeding code, and the fix there seems to be relaxing the synchronisation to only the required points, so the points around copies to host memory, rather than after every kernel), so I'm interested to see if that could be causing the other instability I'm seeing (which I've seemingly isolated to this new track finding code). That would bring closer parity to the CUDA code to, would just need to work out an interface to build an alpaka queue and then pass it everywhere.
 - Test on an Intel GPU. Still not tested there, and there is a few spots that may not actually be functional, but it is difficult to test without hardware.
 - Fix the MT throughput code. There is an old Alpaka issue that implies some of the accelerator back ends aren't / weren't thread safe, which could explain why the MT examples don't work....but the issue is from 2014 and hasn't been updated in a few years. Need to chase that up, and once all three PRs and other fixes are in, start to tackle exactly why the MT example doesn't run.


device/common/include/traccc/fitting/device/impl/fit.ipp
```diff
@@ -57,7 +57,8 @@ TRACCC_HOST_DEVICE inline void fit(
                                           *(seqs_view.ptr() + param_id));
 
     // Run fitting
-    kalman_fitter_status fit_status = fitter.fit(seed_param, fitter_state);
+    [[maybe_unused]] kalman_fitter_status fit_status =
+        fitter.fit(seed_param, fitter_state);
```
CrossR, line 61: I believe this is because for some backends (HIP if I'm remembering correctly? @StewMH ?) `assert` isn't defined, so we don't actually end up using this `fit_status` variable.

StewMH: In an opt build assert isn't called so arguably it is unused here. There is a different weird thing with HIP assert: https://github.com/acts-project/algebra-plugins/pull/145

StewMH:
Pinging @krasznaa and @stephenswat for a review

device/alpaka/src/fitting/fitting_algorithm.cpp
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * @brief View object to the output track states
+     */
+    track_state_container_types::view track_states_view;
+
+    /**
+     * @brief View object to the output barcode sequence
+     */
+    vecmem::data::jagged_vector_view<detray::geometry::barcode> barcodes_view;
+};
```
stephenswat, line 84: Can't this be reused from the CUDA/SYCL code?

stephenswat (approved):
Let's move ahead with this; please rebase this once more and we can move forward.

CrossR:
CI failure seems to just be a random machine issue, rather than my code specifically...
Is there a way to restart it, without needing to commit again?

(Assuming it is a random failure, and I've not missed anything).

PR #927, opened 2025-03-27
Accelerate and update commit benchmarking script

stephenswat:
This commit updates the commit benchmarking code with the following new properties:

1. The benchmark now respects the changes to boolean flags introduced in 380fc78.
2. The benchmark now properly uses Spack libraries as introduced in 069cc80.
3. The benchmark now uses a much smaller set of metrics to speed up data collection.
4. The plotting script now adapts the plot size to the number of commits.

All in all, these changes make the script run much faster and make it compatible with more recent versions of traccc.


krasznaa (approved):
Since you continue being the main user of this code, and nothing of these changes jumps out as crazy-looking, this can all just go ahead from my side.

PR #930, opened 2025-03-28
Allow CUDA CCA algorithm to output cluster info

stephenswat:
This commit allows the CUDA CCA algorithm (and, in the future, the SYCL and Alpaka equivalents) to output cluster information in the form of a disjoint set array. The design is such that the existing API is unchanged, but that extra function calls become available.

### DIFF
--- a/device/common/include/traccc/clusterization/device/aggregate_cluster.hpp
+++ b/device/common/include/traccc/clusterization/device/aggregate_cluster.hpp
@@ -30,6 +30,12 @@ namespace traccc::device {
 /// @param[in] end       partition end point this cell belongs to
 /// @param[in] cid       current cell id
 /// @param[out] out      cluster to fill
+/// @param[out] disjoint_set_ptr Pointer to an array of unsigned integers of
+///                      length $|cells|$ to which an integer is written
+///                      identifying the measurement index to which each cell
+///                      belongs.
+/// @param[out] cluster_size Optional integer which is filled with the size of
+///                      the measurement that is created.
 ///
 TRACCC_HOST_DEVICE
 inline void aggregate_cluster(
@@ -37,7 +43,9 @@ inline void aggregate_cluster(
     const silicon_detector_description::const_device& det_descr,
     const vecmem::device_vector<details::index_t>& f, unsigned int start,
     unsigned int end, unsigned short cid, measurement& out,
-    vecmem::data::vector_view<unsigned int> cell_links, unsigned int link);
+    vecmem::data::vector_view<unsigned int> cell_links, unsigned int link,
+    unsigned int* disjoint_set_ptr,
+    std::optional<std::reference_wrapper<unsigned int>> cluster_size);
 
 }  // namespace traccc::device
 
--- a/device/common/include/traccc/clusterization/device/ccl_kernel.hpp
+++ b/device/common/include/traccc/clusterization/device/ccl_kernel.hpp
@@ -49,6 +49,13 @@ namespace traccc::device {
 ///     matrix fragment storage
 /// @param backup_mutex mutex lock to mediate control over the backup global
 ///     memory data structures.
+/// @param[out] disjoint_set_ptr Pointer to array of unsigned integers of
+///     length $|cells|$ to which an integer is written identifying the
+///     measurement index to which each cell belongs.
+/// @param[out] cluster_size_ptr Pointer to an array of unsigned integers of
+///     length $|cells|$; the first $N$ elements - where $N$ is the number of
+///     output measurements - will have their value set to the size of the
+///     corresponding measurement.
 /// @param barrier  A generic object for block-wide synchronisation
 /// @param[out] measurements_view collection of measurements
 /// @param[out] cell_links    collection of links to measurements each cell is
@@ -66,7 +73,9 @@ TRACCC_DEVICE inline void ccl_kernel(
     vecmem::data::vector_view<details::index_t> gf_backup_view,
     vecmem::data::vector_view<unsigned char> adjc_backup_view,
     vecmem::data::vector_view<details::index_t> adjv_backup_view,
-    vecmem::device_atomic_ref<uint32_t> backup_mutex, const barrier_t& barrier,
+    vecmem::device_atomic_ref<uint32_t> backup_mutex,
+    unsigned int* disjoint_set_ptr, unsigned int* cluster_size_ptr,
+    const barrier_t& barrier,
     measurement_collection_types::view measurements_view,
     vecmem::data::vector_view<unsigned int> cell_links);
 
--- a/device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
+++ b/device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
@@ -18,8 +18,9 @@ inline void aggregate_cluster(
     const silicon_detector_description::const_device& det_descr,
     const vecmem::device_vector<details::index_t>& f, const unsigned int start,
     const unsigned int end, const unsigned short cid, measurement& out,
-    vecmem::data::vector_view<unsigned int> cell_links,
-    const unsigned int link) {
+    vecmem::data::vector_view<unsigned int> cell_links, const unsigned int link,
+    unsigned int* disjoint_set_ptr,
+    std::optional<std::reference_wrapper<unsigned int>> cluster_size) {
     vecmem::device_vector<unsigned int> cell_links_device(cell_links);
 
     /*
@@ -57,6 +58,7 @@ inline void aggregate_cluster(
     const unsigned int module_idx = cells.module_index().at(cid + start);
     const auto module_descr = det_descr.at(module_idx);
     const auto partition_size = static_cast<unsigned short>(end - start);
+    unsigned int tmp_cluster_size = 0;
 
     bool first_processed = false;
 
@@ -114,6 +116,12 @@ inline void aggregate_cluster(
             }
 
             cell_links_device.at(pos) = link;
+
+            tmp_cluster_size++;
+
+            if (disjoint_set_ptr != nullptr) {
+                disjoint_set_ptr[pos] = link;
+            }
         }
 
         /*
@@ -122,6 +130,10 @@ inline void aggregate_cluster(
          */
         if (cell.channel1() > maxChannel1 + 1) {
             break;
+        }
+
+        if (cluster_size.has_value()) {
+            (*cluster_size).get() = tmp_cluster_size;
         }
     }
 
--- a/device/common/include/traccc/clusterization/device/reify_cluster_data.hpp
+++ b/device/common/include/traccc/clusterization/device/reify_cluster_data.hpp
@@ -1,0 +1,23 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2025 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include "traccc/definitions/qualifiers.hpp"
+#include "traccc/device/concepts/thread_id.hpp"
+#include "traccc/edm/silicon_cluster_collection.hpp"
+
+namespace traccc::device {
+template <device::concepts::thread_id1 thread_id_t>
+TRACCC_HOST_DEVICE void reify_cluster_data(
+    const thread_id_t& thread_id, unsigned int* disjoint_set_ptr,
+    unsigned int num_cells,
+    traccc::edm::silicon_cluster_collection::view cluster_view) {
+    traccc::edm::silicon_cluster_collection::device clusters(cluster_view);
+    if (auto idx = thread_id.getGlobalThreadIdX(); idx < num_cells) {
+        clusters.cell_indices().at(disjoint_set_ptr[idx]).push_back(idx);
+    }
+}
+}  // namespace traccc::device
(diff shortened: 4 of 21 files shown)


stephenswat:
Tagging @paradajzblond as an interested party.

device/cuda/include/traccc/cuda/clusterization/clusterization_algorithm.hpp
```diff
@@ -67,12 +78,31 @@ class clusterization_algorithm
...
+    /// @{
+    measurement_collection_types::buffer operator()(
         const edm::silicon_cell_collection::const_view& cells,
         const silicon_detector_description::const_view& det_descr)
         const override;
 
+    measurement_collection_types::buffer operator()(
+        const edm::silicon_cell_collection::const_view& cells,
+        const silicon_detector_description::const_view& det_descr,
+        device::clustering_discard_disjoint_set&&) const override;
```
krasznaa, line 90: Do we need this version of the function?

device/cuda/include/traccc/cuda/clusterization/clusterization_algorithm.hpp
```diff
@@ -67,12 +78,31 @@ class clusterization_algorithm
...
+
+    std::pair<measurement_collection_types::buffer,
+              vecmem::data::vector_buffer<unsigned int>>
+    operator()(const edm::silicon_cell_collection::const_view& cells,
+               const silicon_detector_description::const_view& det_descr,
+               device::clustering_keep_disjoint_set&&) const override;
```
krasznaa, line 96: Using a `struct` like this is an interesting idea. I would not have gone this way, but I kind of like it. As it results in pretty nicely readable client code.

However the return value **must** be:

```c++
std::pair<traccc::measurement_collection_types::buffer, traccc::edm::silicon_cluster_collection::buffer>
```

The algorithm itself must take care of creating the cluster EDM, not the client code.

device/cuda/include/traccc/cuda/clusterization/clusterization_algorithm.hpp
```diff
@@ -67,12 +78,31 @@ class clusterization_algorithm
...
+    /// @}
+
     private:
+    std::pair<measurement_collection_types::buffer,
+              std::optional<vecmem::data::vector_buffer<unsigned int>>>
+    execute_impl(const edm::silicon_cell_collection::const_view& cells,
+                 const silicon_detector_description::const_view& det_descr,
+                 bool keep_disjoint_set) const;
```
krasznaa, line 104: `std::optional` is a possibility, but wouldn't it be easier to have the following interface for this internal function instead?

```c++
void execute_impl(const edm::silicon_cell_collection::const_view& cells,
                  const silicon_detector_description::const_view& det_descr,
                  measurement_collection_types::buffer& meas,
                  edm::silicon_cluster_collection::buffer* clusters = nullptr) const;
```

Though maybe not... 🤔 Still, something to think about a little.

krasznaa (changes requested):
The internal implementation seems to make sense. But as I wrote earlier, I have some pretty concrete ideas about the API.

device/cuda/include/traccc/cuda/clusterization/clusterization_algorithm.hpp
stephenswat: No, I really do not want to add output arguments unless absolutely necessary!

device/cuda/include/traccc/cuda/clusterization/clusterization_algorithm.hpp
stephenswat: This makes it slightly easier to use the function with `std::conditional` if that is ever desired.

device/cuda/include/traccc/cuda/clusterization/clusterization_algorithm.hpp
stephenswat: Sure, we can do that. :+1:

tests/cuda/test_cca.cpp
```diff
@@ -52,16 +56,21 @@ cca_function_t get_f_with(traccc::clustering_config cfg) {
...
 
-        auto measurements_buffer = cc(cells_buffer, dd_buffer);
+        auto [measurements_buffer, cluster_buffer] =
+            cc(cells_buffer, dd_buffer,
+               traccc::device::clustering_keep_disjoint_set{});
         traccc::measurement_collection_types::host measurements{&host_mr};
         copy(measurements_buffer, measurements)->wait();
 
+        traccc::edm::silicon_cluster_collection::host clusters{host_mr};
+        copy(cluster_buffer, clusters)->wait();
```
stephenswat, line 66: @krasznaa trying to get your EDM code to work and this is segfaulting; any ideas?

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -92,16 +123,51 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
...
+
+        std::vector<unsigned int> cluster_sizes_host;
+        cluster_sizes_host.resize(num_measurements);
+
+        TRACCC_CUDA_ERROR_CHECK(cudaMemcpy(
+            cluster_sizes_host.data(), cluster_sizes.get(),
+            num_measurements * sizeof(unsigned int), cudaMemcpyDeviceToHost));
+
+        cluster_data.emplace(cluster_sizes_host, m_mr.main, m_mr.host,
+                             vecmem::data::buffer_type::resizable);
```
krasznaa, line 164: I don't understand what's going on here. 😦 How is this meant to fill the container?

I imagined that we would want to have the cluster information on the device. I was imagining that the `ccl_kernel` would fill this SoA container directly. Or possibly we would launch one more kernel that fills this SoA.

Right now you create the SoA container, don't set it up (with `vecmem::copy::setup`), and just return it. It's very understandable that it would not work right now. 🤔

stephenswat: Perfect, adding the `setup` call fixes the problem. :+1:

device/common/include/traccc/clusterization/device/aggregate_cluster.hpp
```diff
@@ -37,7 +37,8 @@ inline void aggregate_cluster(
     const silicon_detector_description::const_device& det_descr,
     const vecmem::device_vector<details::index_t>& f, unsigned int start,
     unsigned int end, unsigned short cid, measurement& out,
-    vecmem::data::vector_view<unsigned int> cell_links, unsigned int link);
+    vecmem::data::vector_view<unsigned int> cell_links, unsigned int link,
+    unsigned int* disjoint_set_ptr, unsigned int* cluster_size_ptr);
```
krasznaa, line 41: Please provide Doxygen descriptions of the new parameters.

device/common/include/traccc/clusterization/device/ccl_kernel.hpp
```diff
@@ -66,7 +66,9 @@ TRACCC_DEVICE inline void ccl_kernel(
     vecmem::data::vector_view<details::index_t> gf_backup_view,
     vecmem::data::vector_view<unsigned char> adjc_backup_view,
     vecmem::data::vector_view<details::index_t> adjv_backup_view,
-    vecmem::device_atomic_ref<uint32_t> backup_mutex, const barrier_t& barrier,
+    vecmem::device_atomic_ref<uint32_t> backup_mutex,
+    unsigned int* disjoint_set_ptr, unsigned int* cluster_size_ptr,
```
krasznaa, line 70: Please provide Doxygen descriptions of the new parameters.

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
```diff
@@ -18,8 +18,8 @@ inline void aggregate_cluster(
     const silicon_detector_description::const_device& det_descr,
     const vecmem::device_vector<details::index_t>& f, const unsigned int start,
     const unsigned int end, const unsigned short cid, measurement& out,
-    vecmem::data::vector_view<unsigned int> cell_links,
-    const unsigned int link) {
+    vecmem::data::vector_view<unsigned int> cell_links, const unsigned int link,
+    unsigned int* disjoint_set_ptr, unsigned int* cluster_size_ptr) {
```
krasznaa, line 22: I wonder if it wouldn't be better to:

```suggestion
    unsigned int* disjoint_set_ptr, unsigned int& cluster_size) {
```

🤔 The function could still receive a dummy reference even when we don't have a global memory location for this info. I just wonder if the one less `if` statement could help us a little. 🤔

device/common/include/traccc/clusterization/device/tags.hpp
```diff
@@ -0,0 +1,13 @@
...
+ * (c) 2022-2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+namespace traccc::device {
+struct clustering_discard_disjoint_set {};
+struct clustering_keep_disjoint_set {};
```
krasznaa, line 12: Please add some minimal documentation for the new types.

device/cuda/src/clusterization/clusterization_algorithm.cu
```diff
@@ -65,7 +96,9 @@ clusterization_algorithm::output_type clusterization_algorithm::operator()(
 
     // If there are no cells, return right away.
     if (num_cells == 0) {
-        return measurements;
+        return {std::move(measurements),
+                traccc::edm::silicon_cluster_collection::buffer{
+                    std::vector<unsigned int>{}, m_mr.main}};
```
krasznaa, line 101: `vecmem::edm::buffer` has a default constructor. You can use that here.

krasznaa (changes requested):
To put it here as well, I personally would be a lot happier if the new function arguments would also be given to the functions via vecmem views. And in fact I plan to update the code myself to that formalism when upgrading the Alpaka and SYCL algorithms.

But other than that, only some small requests for updating the documentation. And don't forget to add the missing `.cu` file!

stephenswat:
> And in fact I plan to update the code myself to that formalism when upgrading the Alpaka and SYCL algorithms.

Please feel free to try, but I will just warn you _again_ that you are wasting your time. If you try this, you will notice that using vecmem views here is a lot of effort, adds a lot of complication, and yields _no_ benefits.

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
stephenswat: One of the points of references is explicitly that they should not point to `nullptr`, so no - this is not a good idea.

krasznaa: I did not mean to point at `nullptr`. What I was wondering about was the following:
  - In `ccl_kernel.ipp` you currently do the following:

```c++
                              cluster_size_ptr != nullptr
                                  ? &cluster_size_ptr[meas_pos]
                                  : nullptr
```

  - Since you already have an `if` statement right there, why don't we make that the only `if` statement. Like:

```c++
unsigned int dummy;
...
                              cluster_size_ptr != nullptr
                                  ? cluster_size_ptr[meas_pos]
                                  : dummy
...
```

My main dislike here is that this `cluster_size_ptr` is just the same as the variable received by `ccl_kernel`. Right now we need to write it into our Doxygen documentation that one must receive an array of a given size, while the other must receive a pointer to a single element in an array.

It's exactly for this reason that I feel quite partial towards using clearer types with vecmem views. 🤔

stephenswat: I'm really struggling to understand this aversion to pointers... Surely you do not really want me add a dummy integer somewhere so that we can avoid passing a pointer? If you feel _so_ strongly about this I can turn the argument into a `std::optional<std::reference_wrapper<unsigned int>>`?

device/cuda/src/clusterization/clusterization_algorithm.cu
stephenswat: :+1:

device/common/include/traccc/clusterization/device/tags.hpp
stephenswat: Added.

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
stephenswat: I changed it to be `std::optional<std::reference_wrapper<unsigned int>>`; hopefully that will be a satisfactory solution to everyone!

device/common/include/traccc/clusterization/device/reify_cluster_data.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include "traccc/definitions/qualifiers.hpp"
+#include "traccc/device/concepts/thread_id.hpp"
+#include "traccc/edm/silicon_cluster_collection.hpp"
+
+namespace traccc::device {
+template <device::concepts::thread_id1 thread_id_t>
+TRACCC_HOST_DEVICE void reify_cluster_data(
+    const thread_id_t& thread_id, unsigned int* disjoint_set_ptr,
+    unsigned int num_cells,
+    traccc::edm::silicon_cluster_collection::view cluster_view) {
```
krasznaa, line 17: Please provide some function documentation.

krasznaa (changes requested):
First time that I would ever see this word. 🤔

https://www.merriam-webster.com/dictionary/reify

stephenswat:
...

PR #932, opened 2025-03-29
CPU implementation of Triplet Fitting Algorithm

Abhirikshma:
This pull request adds the first CPU implementation of the General Triplet Track Fit (GTTF) algorithm.

To switch the fitting algorithm from Kalman filter to GTTF, one can uncomment the lines in `full_chain_algorithm` and `seq_example`.

The (almost current) results from running GTTF on ODD samples were reported in a TDAQ Open Meeting: [here](https://indico.cern.ch/event/1491865/contributions/6311243/attachments/3006006/5298604/TDAQ_Open_30Jan.pdf). I have made some improvements to the throughput reported by using a custom Cholesky inversion (I have not added that to this PR, can add that in a next).

Please of course let me know if you need some information from me!

### DIFF
--- a/core/include/traccc/fitting/triplet_fit/triplet_fitter.hpp
+++ b/core/include/traccc/fitting/triplet_fit/triplet_fitter.hpp
@@ -17,13 +17,6 @@
 #include "traccc/fitting/fitting_config.hpp"
 
 // detray include(s).
-#include "detray/propagator/actor_chain.hpp"
-#include "detray/propagator/actors/aborters.hpp"
-#include "detray/propagator/actors/parameter_resetter.hpp"
-#include "detray/propagator/actors/parameter_transporter.hpp"
-#include "detray/propagator/actors/pointwise_material_interactor.hpp"
-#include "detray/propagator/propagator.hpp"
-
 #include "detray/tracks/bound_track_parameters.hpp"
 
 // System include(s).
@@ -35,18 +28,10 @@ namespace traccc {
 
     /// Triplet fitting algorithm to fit a single track
 
-    // we dont need the stepper/navigator now but
-    // the detector and bfield types are obtained from these,
-    // as in the original Kalman filter code
-
-    // Try with explicitly passing the actually required types
     template <typename detector_t, typename bfield_t>
     class triplet_fitter {
 
         public:
-        // Detector type
-        // using detector_t = typename navigator_t::detector_t;
-
         // Algebra type
         using algebra_type = typename detector_t::algebra_type;
 
@@ -56,9 +41,6 @@ namespace traccc {
 
         // Configuration type
         using config_type = fitting_config;
-
-        // Field type
-        // using bfield_t = typename stepper_t::magnetic_field_type;
 
         // Matrix types
         using matrix_operator = detray::dmatrix_operator<algebra_type>;
@@ -132,6 +114,8 @@ namespace traccc {
         ///
         TRACCC_HOST_DEVICE 
         void make_triplets() {
+
+            // std::cout << "Making triplets\n";
             
             std::size_t n_triplets = m_track_states.size() - 2;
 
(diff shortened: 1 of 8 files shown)


beomki-yeo:
Thanks for the work. It will take some time to review this PR though. 
Meanwhile could you add a test as well? You can start with telescope geometry (Take a look at [this test](https://github.com/acts-project/traccc/blob/main/tests/cpu/test_kalman_fitter_telescope.cpp) for reference)

examples/run/cpu/full_chain_algorithm.hpp
```diff
@@ -62,6 +63,8 @@ class full_chain_algorithm : public algorithm<track_state_container_types::host(
         traccc::host::combinatorial_kalman_filter_algorithm;
     /// Track fitting algorithm type
     using fitting_algorithm = traccc::host::kalman_fitting_algorithm;
+    // Uncomment (and comment line above) to use Triplet Fitting Algorithm
+    // using fitting_algorithm = traccc::host::triplet_fitting_algorithm;
```
beomki-yeo, line 67: Let' not have this in the example like this. We might need to find a better way to select an algorithm later.
Adding tests would be enough for this PR.

examples/run/cpu/seq_example.cpp
```diff
@@ -112,6 +113,8 @@ int seq_run(const traccc::opts::input_data& input_opts,
     using finding_algorithm =
         traccc::host::combinatorial_kalman_filter_algorithm;
     using fitting_algorithm = traccc::host::kalman_fitting_algorithm;
+    // Uncomment (and comment the line above) to use Triplet Fitting Algorithm
```
beomki-yeo, line 116: Same here

core/include/traccc/fitting/triplet_fit/triplet_fitter.hpp
```diff
@@ -0,0 +1,931 @@
...
+            const measurement& meas_0 = state_0.get_measurement();
+            const measurement& meas_1 = state_1.get_measurement();
+            const measurement& meas_2 = state_2.get_measurement();
+
+            // Get surfaces
+            detray::tracking_surface meas_0_sf(m_detector, meas_0.surface_link);
+            detray::tracking_surface meas_1_sf(m_detector, meas_1.surface_link);
+            detray::tracking_surface meas_2_sf(m_detector, meas_2.surface_link);
+
+            point2 loc_2d_0{meas_0.local[0], meas_0.local[1]};
```
beomki-yeo, line 148: Do you guarantee somewhere that all three measurements are from pixels? This will not work on 1D measurements (strip, drift chamber, straw tube).

beomki-yeo (commented):
just a couple of comments for now. Also please remove commented `printf` lines.

core/include/traccc/fitting/triplet_fit/triplet_fitter.hpp
Abhirikshma: I tested this for also strip measurements, and it worked fine. From my experience I saw always two dimensions. Also the `local` member variable in the structure is two dimensional

beomki-yeo: Code may run w/o error but a "single" strip measurement is not supposed to be used for `bound_to_global`. It is just going to produce wrong result

It is correct that measurement has two dimensional local member variable but one of the two elements is not used for strip measurement - that's why there is a `meas_dim` member variable.

Abhirikshma:
@beomki-yeo I made the changes you wanted and added a test for the Triplet Fitting algorithm with the Telescope geometry. Running the test however errors out with:
```
unknown file: Failure
C++ exception with description "map::at" thrown in the test body.
```
I could not find what is causing this failure! The sequential example runs fine.

stephenswat:
> I could not find what is causing this failure! The sequential example runs fine.

Please build the code in debug mode and run it through GDB. That will tell you where the issue is.

Abhirikshma:
> > I could not find what is causing this failure! The sequential example runs fine.
> 
> Please build the code in debug mode and run it through GDB. That will tell you where the issue is.

I don't have experience using GDB. Can you please tell me how to do it?

stephenswat:
Nota bene (but not for this PR): it would also be very interesting to see how this fitter compares to the MBF in terms of physics and compute.

PR #933, opened 2025-03-30
Do not store links which will be rejected

fiedlerp:


fiedlerp:
Should I also make the changes to the CPU version?

stephenswat:
Could you please provide us a bit more information about what you are doing here? And how does this interact with #908?

No need to update the CPU code if the physics performance doesn't change.

stephenswat:
Having read through it I really like the changes. Can we un-draft this? I am currently running some performance tests, if those turn out alright I am happy to merge.

stephenswat:
Be aware that this change seems to significantly impact the compute performance.



My initial guess would be that this is due to the fact that we might somehow be keeping more tracks? This would explain the increase in fitting time, even if you didn't touch the fitting kernel at all.

beomki-yeo:
It will be likely that cuda tests might fail if the finding results change. Have you run `traccc_test_cuda`?

stephenswat:
> It will be likely that cuda tests that compare with cpu finding results might fail if the finding results change. Have you run `traccc_test_cuda`?

Ah, well spotted! The Gitlab CI doesn't run for Petr.

stephenswat:
Confirmed the tests fail:

```
[ RUN      ] CUDACkfCombinatoricsTelescopeValidation/CudaCkfCombinatoricsTelescopeTests.Run/1

Detector check: OK
Using hit time
/mnt/ssd1/sswatman/traccc/tests/cuda/test_ckf_combinatorics_telescope.cpp:225: Failure
Expected equality of these values:
  track_candidates_cuda.size()
    Which is: 52488
  std::pow(n_truth_tracks, std::get<11>(GetParam()) + 1)
    Which is: 59049
```

Here's the number of tracks over these commits:

a9cf4fbf:

```
Using CUDA device: NVIDIA RTX A5000 [id: 0, bus: 1, device: 0]
Reconstructed track parameters: 1102959
Time totals:
                  File reading  561 ms
            Warm-up processing  10038 ms
              Event processing  10002 ms
Throughput:
            Warm-up processing  1003.89 ms/event, 0.996124 events/s
              Event processing  1000.27 ms/event, 0.999735 events/s
```
              
dcb8d32c:

```
Using CUDA device: NVIDIA RTX A5000 [id: 0, bus: 1, device: 0]
Reconstructed track parameters: 1102962
Time totals:
                  File reading  558 ms
            Warm-up processing  8985 ms
              Event processing  8935 ms
Throughput:
            Warm-up processing  898.504 ms/event, 1.11296 events/s
              Event processing  893.565 ms/event, 1.11911 events/s
```

348de1a2 (main):

```
Using CUDA device: NVIDIA RTX A5000 [id: 0, bus: 1, device: 0]
Reconstructed track parameters: 1034853
Time totals:
                  File reading  557 ms
            Warm-up processing  8677 ms
              Event processing  8613 ms
Throughput:
            Warm-up processing  867.758 ms/event, 1.15239 events/s
              Event processing  861.315 ms/event, 1.16102 events/s
```

fiedlerp:
I see, I will run it locally before the next push. However, I think that I found the discrepancy. It inserts holes when the branching is stopped due to reaching maximum candidates per seed. Also, the hole inserting does not apply this cut. This is one of the reasons why I did not want to undraft it yet.

paulgessinger:

_This is an automated message_

stephenswat:

_This is an automated message_

stephenswat:
I'd really like to understand the performance regression that we have going on in the fit kernel. Since you haven't touched the fitting kernel itself, it would seem to me that you are either fitting more tracks or that your tracks are somehow longer. Have you looked into this?

fiedlerp:
I have not looked into the fitting. I see that the increase is related to the max skipping. The number of tracks is the same, as proven by `traccc_test_cuda`. The tracks are shorter by one hole, as my change creates the tip instead of creating a candidate with too many holes.

stephenswat:
I see the following change in the number of tracks being passed to the fitter:

e31cdb87:
```
Fitting 103495 tracks with 572400 states
```

974a3299:
```
Fitting 104126 tracks with 577516 states
```

So the number of tracks increases by 0.6%, the number of track states increases by 0.9%, and the average number of states per track increases by 0.2%. By itself I don't think this is a problem (and I'm not sure why the fitting time increases by 8.7% while the number of track states increases by only 0.9%) but you claim that tracks should be _shorter_ because you remove holes. :thinking:

device/common/include/traccc/finding/device/impl/build_tracks.ipp
```diff
@@ -13,7 +13,8 @@
 namespace traccc::device {
 
 TRACCC_HOST_DEVICE inline void build_tracks(
-    const global_index_t globalIndex, const finding_config& cfg,
+    const global_index_t globalIndex,
+    [[maybe_unused]] const finding_config& cfg,
```
stephenswat, line 17: If this is no longer necessary, let's remove it entirely as an argument.

device/common/include/traccc/finding/device/impl/find_tracks.ipp
```diff
@@ -270,37 +307,47 @@ TRACCC_HOST_DEVICE inline void find_tracks(
...
-    if (in_param_id < payload.n_in_params &&
-        in_params_liveness.at(in_param_id) > 0u &&
-        shared_payload.shared_num_candidates[thread_id.getLocalThreadIdX()] ==
+    if (in_param_id >= payload.n_in_params ||
+        in_params_liveness.at(in_param_id) == 0u ||
+        shared_payload.shared_num_candidates[thread_id.getLocalThreadIdX()] >
             0u) {
-        // Add measurement candidates to link
-        const unsigned int l_pos = links.bulk_append_implicit(1);
+        return;
```
stephenswat, line 314: I'd like to avoid all these early returns if possible, as any kernel using early returns increases significantly the mental load involved in determining where barriers can be used without resulting in deadlock, and that mental load increases with the number of early returns.

device/common/include/traccc/finding/device/impl/build_tracks.ipp
fiedlerp: The config is used in asserts

device/common/include/traccc/finding/device/impl/find_tracks.ipp
fiedlerp: I did that to minimize the nesting, as SonarCloud complains otherwise

device/common/include/traccc/finding/device/impl/build_tracks.ipp
stephenswat: Could we move the asserts to another kernel, such as prop. to next surf. perhaps?

stephenswat:
The difference in track counts does away if we increase `--max-num-branches-per-seed`:

```
Before with --max-num-branches-per-seed=1000000
Fitting 117770 tracks with 630243 states

After with --max-num-branches-per-seed=1000000
Fitting 117784 tracks with 630450 states
```

@beomki-yeo what's your take on this? I think it is okay, and I think we might want to eventually get rid of the `--max-num-branches-per-seed` flag entirely as it is completely non-deterministic, you have no control over which tracks you keep and which you throw away.

beomki-yeo:
`--max-num-branches-per-seed` is (1) to apply a upper bound for the CKF output of very similar tracks which can explode exponentially and (2) to allocate the memory for tips. I am afraid that removing it may cause a bottleneck even though it won't happen very frequently. 

It is correct that it is non-deterministic but we may need to find alternative if we want to get rid of it :thinking:

stephenswat:
> It is correct that it is non-deterministic but we may need to find alternative if we want to get rid of it 🤔

Indeed this is a good point; ideally I'd like to replace it with some of deterministic mechanism that achieves the same thing... I'll think about it.

fiedlerp:
> @fiedlerp Could you write a description with some details so we can review the PR? Increased fitting time reported by Stephen is definitely weird to me

The PR consists of 3 commits, each handling one of the conditions which kill or reject candidates.

The first commit moves the seed branching cut from `propage_to_next_surface` to `find_tracks`. This is done by counting the seed branches while building the candidates. When the seed branching limit is hit, the computation of candidates with the same branch is stopped as fast as possible to avoid wasting computation time.

The second commit moves the step skipping (hole counting) condition also from `propage_to_next_surface` to `find_tracks`. This is the part that makes the 10% speedup of the track finding. If the candidate has already reached the maximum number of holes, it is added to tips instead of creating a new hole.

The third commit adjusts the candidate length condition in `propagate_to_next_surface` to account for holes, so if the candidate is too short, it is not added to the tips. The same condition is also added to `find_track`'s hole counting part. Also, when reaching the last step, the tip adding is moved from `propage_to_next_surface` to `find_tracks`, and the computation is stopped after the `find_tracks` kernel, as no propagation to the next surface is needed.

beomki-yeo (approved):
The list of changes makes sense to me. Please resolve the @stephenswat 's comments and update the branch to merge the PR

stephenswat:
Please revert the change introducing the early returns, rebase and we can move towards merging this.

fiedlerp:
The build fails on a detray-related issue.

device/sycl/src/finding/find_tracks.hpp
```diff
@@ -26,7 +26,6 @@
 #include "traccc/finding/device/find_tracks.hpp"
 #include "traccc/finding/device/make_barcode_sequence.hpp"
 #include "traccc/finding/device/propagate_to_next_surface.hpp"
-#include "traccc/finding/device/prune_tracks.hpp"
```
stephenswat, line 29: Please add this:

```cpp
// TODO: Remove when updating to future detray version.
#include "detray/core/concepts.hpp"
#include "traccc/finding/device/propagate_to_next_surface.hpp"
```

To see if it removes the issue.

niermann999: This is weird, because I think the `constrained_step` header has the required includes (which would be `#include "detray/definitions/algebra.hpp"`, because it is an algebra concept). However, the constrained step is itself only included in `propagate_to_next_surface.ipp`. So, could you try including 
```
#include <detray/propagator/constrained_step.hpp>
```
in this file https://github.com/acts-project/traccc/blob/45f1cc8c1e041accb78b261e5181b220bbbdb4f1/core/include/traccc/utils/propagation.hpp#L1 ?

stephenswat:
# Performance summary
## Graphical

## Tabular
| Kernel | 369caa90 | f12c0e2b | Delta |
| --- | --: | --: | --: |
| `DeviceRadixSortExclusiveSumKernel` | 508.05 ns | &mdash; | &mdash; |
| `DeviceRadixSortHistogramKernel` | 114.24 μs | &mdash; | &mdash; |
| `DeviceRadixSortOnesweepKernel` | 209.14 μs | &mdash; | &mdash; |
| `Thrust::sort` | 450.09 μs | &mdash; | &mdash; |
| `apply_interaction` | 150.25 μs | &mdash; | &mdash; |
| `build_tracks` | 1.08 ms | &mdash; | &mdash; |
| `ccl_kernel` | 681.42 μs | &mdash; | &mdash; |
| `count_doublets` | 627.08 μs | &mdash; | &mdash; |
| `count_grid_capacities` | 29.13 μs | &mdash; | &mdash; |
| `count_triplets` | 14.18 ms | &mdash; | &mdash; |
| `estimate_track_params` | 34.16 μs | &mdash; | &mdash; |
| `fill_prefix_sum` | 165.41 ns | &mdash; | &mdash; |
| `fill_sort_keys` | 68.22 μs | &mdash; | &mdash; |
| `find_doublets` | 899.17 μs | &mdash; | &mdash; |
| `find_tracks` | 26.42 ms | &mdash; | &mdash; |
| `find_triplets` | 5.98 ms | &mdash; | &mdash; |
| `fit` | 533.81 ms | &mdash; | &mdash; |
| `form_spacepoints` | 13.48 μs | &mdash; | &mdash; |
| `make_barcode_sequence` | 1.00 μs | &mdash; | &mdash; |
| `populate_grid` | 30.48 μs | &mdash; | &mdash; |
| `propagate_to_next_surface` | 126.41 ms | &mdash; | &mdash; |
| `prune_tracks` | 670.34 μs | &mdash; | &mdash; |
| `reduce_triplet_counts` | 6.68 μs | &mdash; | &mdash; |
| `select_seeds` | 360.89 μs | &mdash; | &mdash; |
| `static_kernel` | 1.76 μs | &mdash; | &mdash; |
| `update_triplet_weights` | 95.93 μs | &mdash; | &mdash; |
| **Total** | **712.32 ms** | **None** | **-100.0%** |
> [!IMPORTANT]

> [!NOTE]

core/include/traccc/utils/propagation.hpp
```diff
@@ -14,6 +14,8 @@
 #include <detray/navigation/direct_navigator.hpp>
 #include <detray/navigation/navigator.hpp>
 #include <detray/propagator/actors.hpp>
+// TODO: Remove when updating to future detray version.
```
niermann999, line 17: I think this include can stay that way, since the constrained step size is actually used in multiple other places in the propagation in traccc 
```suggestion
```

device/common/include/traccc/finding/device/impl/find_tracks.ipp
```diff
@@ -271,41 +308,48 @@ TRACCC_HOST_DEVICE inline void find_tracks(
...
+                    in_params.at(in_param_id);
+                out_params_liveness.at(l_pos - payload.curr_links_idx) = 1u;
+            }
         }
-
-        out_params.at(l_pos - payload.curr_links_idx) =
-            in_params.at(in_param_id);
-        out_params_liveness.at(l_pos - payload.curr_links_idx) = 1u;
     }
-}
```
stephenswat, line 309: Found the problem! It was this bracket that caused all the namespace problems :laughing:

PR #935, opened 2025-04-02
Alpaka: Fitting Examples (PR 3 of 3)

CrossR:
This builds on #925 and #918, adding in the updates to the examples to actual use this new code.

Here is the running total of follow up work. I've stuck the ones that changed first, then the rest directly copied over from the last PR.

 - ~~Test on an Intel GPU. Still not tested there, and there is a few spots that may not actually be functional, but it is difficult to test without hardware.~~ Intel hardware has been tested now, and we are getting somewhere! The Alpaka code functions up to, but not including the new finding code. I did a replacement of some of the thrust code there with `ifdef SYCL: oneDPL::XXX` and got a bit further into the finding, but I get a hang when attempting to launch one of the kernels (prop to next surface I believe). Going to follow up with later, but I suspect it is due to the difference in kernel writing between CUDA/HIP + SYCL (as referenced by the register spills I get when compiling for SYCL etc). May need to play with register sizes or see about splitting things to get it happy.

Copied over: 

 - "Unify Everything" : I.e. work to condense any / all accelerator specific code to a single place. The two main places that is still needed is the thrust device selection and also a unified interface for vecmem (we are almost there on that, but we need to work out the best way to optionally take a SYCL queue in, as we have to special case SYCL for that reason at the mo).

 - Fix any remaining bugs that exist in this implementation. When running all the work together, I'm seeing some instability, especially at higher values of mu, where the alpaka implementation will crash. That needs investigating, but potentially after....

 - Async-ify. Right now, we using a blocking queue and lots of `::alpaka::wait(queue)` calls. That is causing issues in at least one place (part of the instability mentioned above is due to the seeding code, and the fix there seems to be relaxing the synchronisation to only the required points, so the points around copies to host memory, rather than after every kernel), so I'm interested to see if that could be causing the other instability I'm seeing (which I've seemingly isolated to this new track finding code). That would bring closer parity to the CUDA code to, would just need to work out an interface to build an alpaka queue and then pass it everywhere.

 - Fix the MT throughput code. There is an old Alpaka issue that implies some of the accelerator back ends aren't / weren't thread safe, which could explain why the MT examples don't work....but the issue is from 2014 and hasn't been updated in a few years. Need to chase that up, and once all three PRs and other fixes are in, start to tackle exactly why the MT example doesn't run.

### DIFF
--- a/device/alpaka/src/finding/finding_algorithm.cpp
+++ b/device/alpaka/src/finding/finding_algorithm.cpp
@@ -68,6 +68,11 @@ finding_algorithm<stepper_t, navigator_t>::operator()(
     const typename measurement_collection_types::view& measurements,
     const bound_track_parameters_collection_types::buffer& seeds_buffer) const {
 
+    assert(m_cfg.min_step_length_for_next_surface >
+               math::fabs(m_cfg.propagation.navigation.overstep_tolerance) &&
+           "Min step length for the next surface should be higher than the "
+           "overstep tolerance");
+
     // Setup alpaka
     auto devHost = ::alpaka::getDevByIdx(::alpaka::Platform<Host>{}, 0u);
     auto devAcc = ::alpaka::getDevByIdx(::alpaka::Platform<Acc>{}, 0u);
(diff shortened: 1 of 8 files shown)


CrossR:
Rebase-d and seems happy, but the original PR (#934) didn't seem to like the force push, so I've just remade the PR.

stephenswat (approved):

PR #937, opened 2025-04-04
New CPU example to demonstrate the context-aware transform store usage

tsulaiav:
This is a simple demonstration of how the new context-aware interface of `detray::single_store` can be used in CPU examples. It is based on the example that @beomki-yeo prepared for me several months ago.

The transform store has two contexts. The "misaligned" context is just a copy of the default context. The event loop is split into two parts. The first half of the events runs with context 0, and the second half with context 1. I checked that for the second half, all calls to `event_store::at()` are made with the context argument equal to 1. I.e., the context value is correctly propagated down the call chain to the `event_store`.

No other checks have been done with this example.

CC: @niermann999


examples/run/cpu/misaligned_truth_fitting_example.cpp
```diff
@@ -0,0 +1,219 @@
...
+    // Iterate over events
+    for (auto event = input_opts.skip;
+         event < input_opts.events + input_opts.skip; ++event) {
+
+        // Truth Track Candidates
+        traccc::event_data evt_data(input_opts.directory, event, host_mr,
+                                    input_opts.use_acts_geom_source, &host_det,
+                                    input_opts.format, false);
+
+        // For the first half of events run Alg0
```
beomki-yeo, line 158: Doesn't it make more sense run entire events for each context so we can compare the results between aligned and misaligned?

examples/run/cpu/misaligned_truth_fitting_example.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    /// Create a "misaligned" context in the transform store
+    using xf_container = host_detector_type::transform_container;
+    using xf_vector = xf_container::base_type;
+
+    const xf_container& default_xfs = host_det.transform_store();
+    xf_vector misaligned_xfs;
+    misaligned_xfs.reserve(default_xfs.size());
+    for (const auto& xf : default_xfs) {
+        misaligned_xfs.push_back(xf);
```
beomki-yeo, line 111: Can we actually misalign the modules to see the impact of misalignment?

examples/run/cpu/misaligned_truth_fitting_example.cpp
```diff
@@ -0,0 +1,219 @@
...
+    /// Type declarations
+    using host_detector_type = traccc::default_detector::host;
+
+    // Memory resources used by the application.
+    vecmem::host_memory_resource host_mr;
+    // Copy obejct
+    vecmem::copy copy;
+
+    // Performance writer
+    traccc::fitting_performance_writer fit_performance_writer(
```
beomki-yeo, line 76: We can make a writer for each context

beomki-yeo (commented):
Looks good in general

examples/run/cpu/misaligned_truth_fitting_example.cpp
beomki-yeo: Probably the current design makes more sense if we resolve the alignment with the first half and apply the aligned context to the rest of tracks.

examples/run/cpu/misaligned_truth_fitting_example.cpp
```diff
@@ -0,0 +1,219 @@
...
+    using xf_vector = xf_container::base_type;
+
+    const xf_container& default_xfs = host_det.transform_store();
+    xf_vector misaligned_xfs;
+    misaligned_xfs.reserve(default_xfs.size());
+    for (const auto& xf : default_xfs) {
+        misaligned_xfs.push_back(xf);
+    }
+    xf_container* ptr_default_xfs = const_cast<xf_container*>(&default_xfs);
+    ptr_default_xfs->add_context(misaligned_xfs);
```
niermann999, line 114: I think for now this is fine, but let me consider how we can do this without a const cast

niermann999 (approved):

examples/run/cpu/misaligned_truth_fitting_example.cpp
tsulaiav: Yes, it would be good to move away from the `const_cast` eventually. Thanks

PR #938, opened 2025-04-11
Migrate fitting kernels to use a payload struct

stephenswat:
This commit migrates the CUDA and SYCL fitting kernels to use payloads like the Alpaka algorithm does, thereby harmonizing the interface between these different implementations.


stephenswat:
Pinging @StewMH and @CrossR as this affects the Alpaka code.

beomki-yeo (approved):
looks good to me once CI passes

PR #940, opened 2025-04-15
Add cluster link back to the measurement EDM

beomki-yeo:
I am adding it back as @krasznaa requested
But #930 does not include any cluster info in the measurement EDM. @stephenswat do we need it or not?

### DIFF
--- a/core/include/traccc/edm/measurement.hpp
+++ b/core/include/traccc/edm/measurement.hpp
@@ -51,6 +51,9 @@ struct measurement {
 
     /// subspace
     subspace<default_algebra, e_bound_size, 2u> subs{{0u, 1u}};
+
+    /// Cluster link
+    std::size_t cluster_link = std::numeric_limits<std::size_t>::max();
 };
 
 /// Comparison / ordering operator for measurements


krasznaa:
I believe #930 will output clusters and measurements in the same order. In fact, if their order were always the same, we wouldn't need this index. But even with #930 in place, we may still do things that will require this info:
  - We may still re-order measurements in some processing steps possibly;
  - And this is the one that I've been wondering about since a little while, we may create a single measurement collection from multiple cluster collections. Athena at the moment deals with pixel and strip clusters in separate containers. We may very well need to do the same at one point. And then we'll not only need to save an index for the clusters, but a "container identifier" as well. Like what Acts does. Which itself is very much following what Athena does with `ElementLink<T>`.

@beomki-yeo, since **right now** this additional variable is of not much use, we may just put this PR aside as a draft for a while. We should probably think a bit more about the "final EDM design" first. 🤔

krasznaa:
Let's discuss about this at our next meeting. As there may be some misunderstandings here about the purpose of these indices.

As I wrote before, this PR should probably just be put aside for now.

PR #941, opened 2025-04-15
Use pvalue for resolver

beomki-yeo:
Based on #923

This PR replace chi2 criteria with p-value criteria because p-value includes both chi2 and NDF for fair comparison between tracks


stephenswat:
I'll mark this as draft until #923 is merged.

beomki-yeo:
It's good to go now

beomki-yeo:
That's a bit controversial but yeah I think synchronizing them is necessary for maintenance even though it meant that the legacy is not the legacy anymore.

PR #943, opened 2025-04-17
CUDA greedy resolver

beomki-yeo:
PR description proofread by ChatGPT

### 1. Overview

As many might already know, the greedy ambiguity resolver is inherently not GPU-friendly. In each iteration where a track is removed from the pool, only a small number of tracks—typically fewer than ten—require updates, and they must be re-sorted based on the new number of shared measurements. This low degree of parallelism is far from sufficient to fully utilize GPU resources.

In short, despite the significant optimizations I've implemented over the past several months, the GPU version still performs worse than the single-threaded CPU implementation. To compensate for the GPU inefficiency, we may need to either (1) reduce the number of duplicate and fake tracks generated during seeding and CKF, or (2) combine the greedy approach with other ambiguity resolution strategies, such as score-based or ML-based algorithms.

Avoiding the greedy resolver altogether is also a valid option at this point, given its current performance. Still, this is just a start so there must be room for further optimization.

### 2. Benchmark

Following is the output from `seq_example_cuda` with ttbar<200> where I fed the same `track_candidate_container` to CPU and CUDA greedy resolver for fair comparison. (I replaced the CPU CKF output with CUDA CKF output)

**Command**

```shell
    ./bin/traccc_seq_example_cuda \
    --use-detray-detector=true \
    --use-acts-geom-source=true \
    --input-directory=odd/odd-simulations-20240509/geant4_ttbar_mu200/ \
    --input-events=1 \
    --compare-with-cpu
```

**Output**

```
Number of resolved track candidates (header): 5984 (host), 5984 (device)
  Matching rate(s):
    - 100% at 0.01% uncertainty
    - 100% at 0.1% uncertainty
    - 100% at 1% uncertainty
    - 100% at 5% uncertainty
22:27:44    CudaSeqExample                INFO        Reolved track candidates (item) matching rate: 100%
Number of track states: 5067 (host), 5984 (device)
  Matching rate(s):
    - 39.1377% at 0.01% uncertainty
    - 60.9793% at 0.1% uncertainty
    - 67.7306% at 1% uncertainty
    - 68.9505% at 5% uncertainty
22:27:45    CudaSeqExample                INFO      ==> Statistics ... 
22:27:45    CudaSeqExample                INFO      - read    438151 cells
22:27:45    CudaSeqExample                INFO      - created (cpu)  117907 measurements     
22:27:45    CudaSeqExample                INFO      - created (cuda)  117907 measurements     
22:27:45    CudaSeqExample                INFO      - created (cpu)  92892 spacepoints     
22:27:45    CudaSeqExample                INFO      - created (cuda) 92892 spacepoints     
22:27:45    CudaSeqExample                INFO      - created  (cpu) 42372 seeds
22:27:45    CudaSeqExample                INFO      - created (cuda) 42372 seeds
22:27:45    CudaSeqExample                INFO      - found (cpu)    114567 tracks
22:27:45    CudaSeqExample                INFO      - found (cuda)   114984 tracks
22:27:45    CudaSeqExample                INFO      - resolved (cpu)    5984 tracks
22:27:45    CudaSeqExample                INFO      - resolved (cuda)   5984 tracks
22:27:45    CudaSeqExample                INFO      - fitted (cpu)   5067 tracks
22:27:45    CudaSeqExample                INFO      - fitted (cuda)  5984 tracks
22:27:45    CudaSeqExample                INFO      ==>Elapsed times...            File reading  (cpu)  915 ms
22:27:45    CudaSeqExample                INFO               Clusterization (cuda)  4 ms
22:27:45    CudaSeqExample                INFO               Clusterization  (cpu)  29 ms
22:27:45    CudaSeqExample                INFO         Spacepoint formation (cuda)  1 ms
22:27:45    CudaSeqExample                INFO         Spacepoint formation  (cpu)  2 ms
22:27:45    CudaSeqExample                INFO                      Seeding (cuda)  52 ms
22:27:45    CudaSeqExample                INFO                      Seeding  (cpu)  7319 ms
22:27:45    CudaSeqExample                INFO                 Track params (cuda)  0 ms
22:27:45    CudaSeqExample                INFO                 Track params  (cpu)  12 ms
22:27:45    CudaSeqExample                INFO                Track finding (cuda)  340 ms
22:27:45    CudaSeqExample                INFO                 Track finding (cpu)  12998 ms
22:27:45    CudaSeqExample                INFO         Ambiguity resolution (cuda)  4314 ms
22:27:45    CudaSeqExample                INFO          Ambiguity resolution (cpu)  3382 ms
22:27:45    CudaSeqExample                INFO                Track fitting (cuda)  213 ms
22:27:45    CudaSeqExample                INFO                 Track fitting (cpu)  590 ms
22:27:45    CudaSeqExample                INFO                           Wall time  31374 ms
```

~~The source of discrepancy is under investigation - I don't see the discrepancy below ttbar<140>~~

The bug is fixed now - works perfectly well for ttbar<200>

### 3. Implementation

People might wonder why there are so many kernels used inside the iteration loop
`for (unsigned int iter = 0; iter < n_max_it; iter++)`
in `greedy_ambiguity_resolution_algorithm.cu`.

The very first working version only used 2–3 functions—namely `thrust::max_element`, `kernels::update_vector`, and `thrust::sort`. However, that version was significantly slower than the current implementation, mainly due to two reasons:
(1) reliance on Thrust functions, and
(2) overhead from `cudaMemcpyAsync`, which was used to retrieve values from `update_result` in order to decide whether to terminate the loop.

Among these, `thrust::sort` was found to be the biggest bottleneck according to nsight compute. It performs memory allocations and deallocations internally on every call, and sorting all tracks in each iteration was clearly overkill for our use case (see the strategy discussed in #923). To address this, seven dedicated kernels—from `kernels::sort_update_tracks` to `kernels::gather_tracks`—were implemented as a replacement for `thrust::sort`. These kernels brought major performance improvements by maintaining O(log N) complexity when determining the correct location to swap elements in the sorted vector.

Furthermore, `cudaMemcpyAsync` was eliminated by moving the termination check into the GPU kernel itself. This change not only removed unnecessary host-device transactions but also enabled the use of cudaGraph, allowing multiple kernels to be chained together with minimal launch overhead. This brought another noticeable performance gain.

### 4. Bottlenecks

`kernels::update_vectors` and `kernels::rearrange_tracks` are the bottlenecks.

image

image

Note that cudaGraph is removed in the above figures to profile the individual kernels (w/ cudaGraph, the kernels are merged into one block)

### 5. Caveats

Have not encountered yet, but the algorithm might fail for events with the number of tracks larger than like a million.


device/common/include/traccc/device/impl/container_d2h_copy_alg.ipp
```diff
@@ -75,6 +75,11 @@ container_d2h_copy_alg<CONTAINER_TYPES>::operator()(input_type input) const {
     host_header_copy_event->wait();
     host_item_copy_event->wait();
 
+    const auto sizes = m_deviceCopy.get_sizes(input.items);
+    for (std::size_t i = 0; i < result.size(); ++i) {
+        result[i].items.resize(sizes[i]);
+    }
```
krasznaa, line 81: :confused: Did you observe some issue? Why did you add this?

The vecmem code already resizes the output containers during the copy. These lines here should be at best unnecessary, at worst hurtful.

device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu
```diff
@@ -0,0 +1,487 @@
...
+                });
+            TRACCC_CUDA_ERROR_CHECK(cudaGetLastError());
+
+            kernels::gather_tracks<<<nBlocks, nThreads, 0, stream>>>(
+                device::gather_tracks_payload{
+                    .update_res = update_res_device.get(),
+                    .temp_sorted_ids_view = temp_sorted_ids_buffer,
+                    .sorted_ids_view = sorted_ids_buffer,
+                    .is_updated_view = is_updated_buffer});
+            TRACCC_CUDA_ERROR_CHECK(cudaGetLastError());
```
krasznaa, line 448: :thinking: These kernels could be merged into a single one, no? Many of them are launched in the exact same way. Some of them are a bit different, both shared memory and thread-count-wise, but it would still seem possible to do this in one go, no?

krasznaa (commented):
I think you're making your development harder than necessary. If I were you, I would try to do a first pass of this "in pure CUDA". I.e. not bother about making the code modular, shareable with the other languages. Since that's anyway super hard to do while developing just in a single language. (I already found a few places where the current code is CUDA-only.)

I also very seriously wonder whether the EDM update of #971 might make your life a little easier. I see that you copy data out of the current AoS EDM into separate simple vectors. That would be a matter of a few memcopies in the new EDM, you wouldn't need to start a kernel to make it happen.

It could make sense to have a dedicated discussion about this at one point. :thinking: I'd like to understand the design a bit more, and possibly help you with some ideas.

device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu
beomki-yeo: Not in one go - mostly because of synchronization issue between blocks. 
BTW, we don't have to worry about the launch overhead now - I tied them inside the cudagraph loop

device/common/include/traccc/device/impl/container_d2h_copy_alg.ipp
beomki-yeo: I don't remember exactly why I made this change. 
Unfortunately I just tested again without this change and the tests fail :/

```
byeo@hermes:/mnt/nvme0n1/byeo/projects/traccc/traccc_build$ ./bin/traccc_test_cuda --gtest_filter=*Ambi*Comp*
Note: Google Test filter = *Ambi*Comp*
[----------] 1 test from CudaAmbiguitySolverTests
[ RUN      ] CudaAmbiguitySolverTests.CPU_Comparison
Event: 0 Seed: 42
 Time for the cpu method 50 ms
 Time for the cuda method 373 ms
/mnt/nvme0n1/byeo/projects/traccc/traccc/tests/cuda/test_ambiguity_resolution.cpp:371: Failure
Value of: find_pattern(res_trk_cands_cuda, get_pattern(res_trk_cands_cpu, i))
  Actual: false
Expected: true

[  FAILED  ] CudaAmbiguitySolverTests.CPU_Comparison (617 ms)
[----------] 1 test from CudaAmbiguitySolverTests (617 ms total)

[==========] 1 test from 1 test suite ran. (617 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] CudaAmbiguitySolverTests.CPU_Comparison
```

beomki-yeo: I think `TEST(CudaAmbiguitySolverTests, CPU_Comparison)` in `test_ambiguity_resolution.cpp` is the only place I use the d2h copy for the container. Could you take a look and find if there is any mistake?

beomki-yeo:
>  If I were you, I would try to do a first pass of this "in pure CUDA". I.e. not bother about making the code modular, shareable with the other languages.

Writing shareable codes was easier because I could just follow the existing algorithm structures. I will get rid of the CUDA-specific primitives in the next week. Sth like `__shfl_down_sync` may not be removable right now, but let's just keep them as they are :thinking: We can add common functions for them later

beomki-yeo:
> It could make sense to have a dedicated discussion about this at one point. 

Sure thing - Shall we talk maybe on Friday right after the usual detray chat?

examples/run/cuda/seq_example_cuda.cpp
```diff
@@ -350,12 +372,33 @@ int seq_run(const traccc::opts::detector& detector_opts,
...
+                // CPU
+                auto track_candidates_cuda_tmp =
+                    copy_track_candidates(track_candidates_buffer);
+
+                if (accelerator_opts.compare_with_cpu) {
+                    traccc::performance::timer timer{
+                        "Ambiguity resolution (cpu)", elapsedTimes};
+                    res_track_candidates = resolution_alg_cpu(
+                        traccc::get_data(track_candidates_cuda_tmp));
+                    // resolution_alg_cpu(traccc::get_data(track_candidates));
```
beomki-yeo, line 392: I will uncomment this before the PR is merged. Currently I am using cuda CKF output for the comparison

beomki-yeo:
The cudaGraph  crashes with the debug build but let's not care for the moment :man_shrugging: 
```
========= Program hit cudaErrorLaunchOutOfResources (error 701) due to "too many resources requested for launch" on CUDA API call to cudaGraphInstantiate.
=========     Saved host backtrace up to driver entry point at error
=========     Host Frame: [0x442065]
=========                in /usr/lib/x86_64-linux-gnu/libcuda.so.1
=========     Host Frame:cudaGraphInstantiate [0x6b4d0]
=========                in /usr/local/cuda-12.4/targets/x86_64-linux/lib/libcudart.so.12
=========     Host Frame:cudaGraphInstantiate(CUgraphExec_st**, CUgraph_st*, CUgraphNode_st**, char*, unsigned long) in /usr/local/cuda/targets/x86_64-linux/include/cuda_runtime.h:439 [0x1b0af3]
=========                in /mnt/nvme0n1/byeo/projects/traccc/traccc_build/lib/libtraccc_cuda.so.0
=========     Host Frame:traccc::cuda::greedy_ambiguity_resolution_algorithm::operator()(traccc::container_view<traccc::finding_result const, traccc::measurement const> const&) const in /mnt/nvme0n1/byeo/projects/traccc/traccc/device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu:496 [0x1b3ff5]
=========                in /mnt/nvme0n1/byeo/projects/traccc/traccc_build/lib/libtraccc_cuda.so.0
=========     Host Frame:CudaAmbiguitySolverTests_CPU_Comparison_Test::TestBody() in /mnt/nvme0n1/byeo/projects/traccc/traccc/tests/cuda/test_ambiguity_resolution.cpp:340 [0x544419]
```

beomki-yeo:
@krasznaa As a follow up of our discussion, I think we can still parallelize multiple iterations by taking out "multiple" worst tracks at the end of vector instead of doing it one-by-one :thinking:

beomki-yeo:
@krasznaa I will probably make the further optimizations in the separate PRs - I will appreciate if you review the PR as it is now

device/common/include/traccc/device/impl/container_d2h_copy_alg.ipp
krasznaa: I'll have a look at this part (now that the code is ready for review), as I still find this suspicious. :thinking:

device/cuda/include/traccc/cuda/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <vecmem/utils/copy.hpp>
+
+namespace traccc::cuda {
+
+/// Evicts tracks that seem to be duplicates or fakes. This algorithm takes a
+/// greedy approach in the sense that it will remove the track which looks "most
+/// duplicate/fake"
+class greedy_ambiguity_resolution_algorithm
+    : public algorithm<track_candidate_container_types::buffer(
+          const typename track_candidate_container_types::const_view&)>,
```
krasznaa, line 30: You're a bit too liberal with `typename`. :stuck_out_tongue: It's needed when referring to a "sub-type" of a template type. But here `track_candidate_container_types::const_view` is a concrete type.

I.e.:

```suggestion
          const track_candidate_container_types::const_view&)>,
```

Notice how you also didn't use `typename` on line 53. :wink:

examples/run/cuda/seq_example_cuda.cpp
```diff
@@ -174,6 +188,8 @@ int seq_run(const traccc::opts::detector& detector_opts,
         host_mr, logger().clone("HostTrackParEstAlg"));
     host_finding_algorithm finding_alg(finding_cfg,
                                        logger().clone("HostFindingAlg"));
+    host_ambiguity_resolution_algorithm resolution_alg_cpu(
+        resolution_config, hmr, logger().clone("AmbiguityResolutionAlg"));
```
krasznaa, line 192:
```suggestion
        resolution_config, hmr, logger().clone("HostAmbiguityResolutionAlg"));
```

examples/run/cuda/seq_example_cuda.cpp
```diff
@@ -192,6 +208,9 @@ int seq_run(const traccc::opts::detector& detector_opts,
         mr, copy, stream, logger().clone("CudaTrackParEstAlg"));
     device_finding_algorithm finding_alg_cuda(finding_cfg, mr, copy, stream,
                                               logger().clone("CudaFindingAlg"));
+    device_ambiguity_resolution_algorithm resolution_alg_cuda(
+        resolution_config, mr, copy, stream,
+        logger().clone("AmbiguityResolutionAlg"));
```
krasznaa, line 213:
```suggestion
        logger().clone("CudaAmbiguityResolutionAlg"));
```

examples/run/cuda/seq_example_cuda.cpp
```diff
@@ -143,6 +150,10 @@ int seq_run(const traccc::opts::detector& detector_opts,
         traccc::host::combinatorial_kalman_filter_algorithm;
     using device_finding_algorithm =
         traccc::cuda::finding_algorithm<stepper_type, device_navigator_type>;
+    using host_ambiguity_resolution_algorithm =
+        traccc::host::greedy_ambiguity_resolution_algorithm;
+    using device_ambiguity_resolution_algorithm =
+        traccc::cuda::greedy_ambiguity_resolution_algorithm;
```
krasznaa, line 156: These typedefs are probably a bit overkill. The algorithms are not templated, so we might as well use the names directly.

You could argue that this setup would make experimentation with different resolution algorithms earlier in the future, but since on line 168 then you use the typename directly anyway, not sure that that's a very strong argument. :stuck_out_tongue:

examples/run/cuda/seq_example_cuda.cpp
krasznaa: Don't forget about undoing/fixing this!

krasznaa (commented):
Boy... is this a lot of new code... :thinking:

I only scraped through it. Since it's all new, I think it's fine that we'll fix performance issues / imperfections later on. We anyway will want to re-arrange the code for SYCL/Alpaka compatibility as well.

I'll just double-check what's going on with the container copy, you should fix the one actual issue in the sequence example (and possibly modify the other non-important things that I pointed out), and then we can get this in. :wink:

device/cuda/src/ambiguity_resolution/kernels/scan_block_offsets.cu
```diff
@@ -0,0 +1,61 @@
...
+
+// VecMem include(s).
+#include <vecmem/containers/device_vector.hpp>
+
+namespace traccc::cuda::kernels {
+
+__global__ void scan_block_offsets(device::scan_block_offsets_payload payload) {
+
+    if (*(payload.terminate) == 1 || *(payload.n_updated_tracks) == 0) {
+        return;
```
krasznaa, line 21: I noticed this only now. Many of these kernels are implemented with undefined behaviour at the moment. :frowning: (One should not early return from kernels that require synchronization between the threads.)

PR #945, opened 2025-04-22
Remove Indicators-based progress bars

stephenswat:
This commit removes the Indicators-based progress bars which were cute but ultimately interfere with the logging of traccc, and which add an unmaintained dependency to the program.


examples/run/common/throughput_mt.ipp
```diff
@@ -212,19 +209,12 @@ int throughput_mt(std::string_view description, int argc, char* argv[],
...
-            indicators::option::ShowPercentage{true},
-            indicators::option::ShowRemainingTime{true},
-            indicators::option::MaxProgress{throughput_opts.cold_run_events}};
-
         // Measure the time of execution.
         performance::timer t{"Warm-up processing", times};
 
         // Process the requested number of events.
         for (std::size_t i = 0; i < throughput_opts.cold_run_events; ++i) {
+            TRACCC_INFO("Processing warm-up event #" << i);
```
krasznaa, line 217: I'm on board with using a different user feedback mechanism. But if we use something of our own, that needs to be smarter than this.
  - There should be a printout with some regularity about the program still working, and not being stuck;
  - There should be an ETA estimate printed if at all possible.

Both of those are **big** upsides in interactive use. So if you prefer not to implement them (which I wouldn't either), then just:
  - Make the applications detect whether their output is being redirected or not;
    * The trick I used during LHC Run-1 (https://sourceforge.net/p/sframe/code/HEAD/tree/SFrame/trunk/core/src/SLogWriter.cxx#l73) should still work I believe;
  - Use `indicators::ProgressBar` in an interactive session, and use simple printouts in a non-interactive one.

krasznaa (changes requested):
I'm not married to that external, but the feedback from it made my life a **lot** easier. So let's not remove all that goodness too hastily.

examples/run/common/throughput_mt.ipp
stephenswat: > So if you prefer not to implement them (which I wouldn't either), then just:
> 
>   * Make the applications detect whether their output is being redirected or not;
>       
>     * The trick I used during LHC Run-1 (https://sourceforge.net/p/sframe/code/HEAD/tree/SFrame/trunk/core/src/SLogWriter.cxx#l73) should still work I believe;
> 
>   * Use `indicators::ProgressBar` in an interactive session, and use simple printouts in a non-interactive one.

This misses the point of this PR entirely. :laughing: 

Anyway if you really feel so strongly that this stuff is useful I'll add an ETA estimate.

krasznaa: Isn't the point that interactive and non-interactive use has different needs? :confused:

I agree that these indicators are pretty bad when you want to run jobs in a batch. But I can tell you that interactively they are great. So:
  - Let's use them in interactive sessions;
  - Use some absolutely trivial printout (no ETA, no fanciness, no nothing) in batch jobs.

PR #946, opened 2025-04-22
fix: IO sorting

niermann999:
This is a hotfix for the csv io (there are probably better solutions to this), where after sorting the measurement container, the hit to measurement indices are incorrect. This has lead to seeds in the toy detector benchmark that point inwards to the IP.  Now, `read_measurement` keeps track of the measurement ids throughout sorting and returns a vector that maps the initial index to the one after sorting. However, in the current implementation it sorts the collection twice for that.

### DIFF
--- a/io/include/traccc/io/read_measurements.hpp
+++ b/io/include/traccc/io/read_measurements.hpp
@@ -17,6 +17,7 @@
 // System include(s).
 #include <cstddef>
 #include <string_view>
+#include <vector>
 
 namespace traccc::io {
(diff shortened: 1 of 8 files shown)


stephenswat (approved):
Good stuff. :smile:

niermann999:
The performance drop is expected. The bug prevented the generation of correct seeds, which lead to most tracks being aborted during track finding. Now more tracks should be reconstructed, which lowers the throughput.

PR #947, opened 2025-04-23
ref: detray includes

niermann999:
Move some detray includes to their own traccc headers, so that changes are confined to these headers. Also defines ```detray::pdg_particle``` as ```traccc::pdg_particle```.


stephenswat (approved):
LGTM. :+1:

PR #951, opened 2025-04-26
fix: toy detector benchmark

niermann999:
Fix the segmentation fault in the CKF by limiting the minimum track momentum during reconstruction. An excessively large `qop` value can lead the stepper to issue invalid track parameters (containing `nan`) during `evaluate_dqopds` and `evaluate_dtds` leading to a crash of the navigator. 
 TODO: Another source of such a track state seems to be from the Kalman Fitter in single precision with covariance inflation. Double precision seems to be fine, as is setting the inflation factor to 1 in single precision. Running in a debug build triggers an exception in the material_interactor that `qop` is zero. The source of this will need to be followed up


core/include/traccc/finding/actors/ckf_aborter.hpp
```diff
@@ -61,6 +58,16 @@ struct ckf_aborter : detray::actor {
                 "CKF: Maximum number of steps to reach next sensitive surface "
                 "exceeded");
         }
+
+        const auto &track = stepping();
+        const scalar q{stepping.particle_hypothesis().charge()};
+        const scalar mag{track.pT(q)};
+
+        if (mag <= 500.f * traccc::unit<scalar>::MeV) {
```
beomki-yeo, line 66: Spotted this while randomly partolling the open PRs.
We should not abort the navigator just because its momentum is low

 Could you please find another way to handle low momentum? For example, I can suggest to add a configuration parameter (min_pT) to  finding config, propagation config or fitting config. Of course, the default value should be much lower (e.g. 0) - 500 MeV/c is unacceptably high.

benchmarks/common/benchmarks/toy_detector_benchmark.hpp
```diff
@@ -83,6 +83,11 @@ class ToyDetectorBenchmark : public benchmark::Fixture {
         apply_propagation_config(finding_cfg.propagation);
         apply_propagation_config(fitting_cfg.propagation);
 
+        // Prevent segmentation fault in the Kalman Fitter
+        fitting_cfg.covariance_inflation_factor = 1.f;
```
beomki-yeo, line 87: Why is this necessary?

benchmarks/common/benchmarks/toy_detector_benchmark.hpp
```diff
@@ -83,6 +83,11 @@ class ToyDetectorBenchmark : public benchmark::Fixture {
...
+
+        finding_cfg.chi2_max = 30.f;
```
beomki-yeo, line 89: Same here

benchmarks/common/benchmarks/toy_detector_benchmark.hpp
niermann999: I am not sure anymore, because this is from a while back, but I think with this I could recover the tracks that were lost during the fit. I don't want to keep it like that, however, the covariance inflation seems to have a big effect on the the floating point stability. Before it was added, if I remember correctly, the benchmarks did not segfault with the full number of tracks being reconstructed (i.e. setting the grid window to 3 3).

benchmarks/common/benchmarks/toy_detector_benchmark.hpp
niermann999: I had to set it that big to recover all tracks in a single muon, no CKF branching scenario.

niermann999:
Closed after https://github.com/acts-project/traccc/pull/1190

PR #956, opened 2025-04-30
Accelerate and fix bug in performance code

stephenswat:
This commit accelerates some of the performance code and fixes a bug in it. The bug fix concerns very low eta tracks which would be put into the non-existant bin -1. The optimizations are simple (using references and precomputing some mappings which cause the performance comparison code to speed up by a factor 20 for $\mu = 80$ events.

### DIFF
--- a/performance/src/utils/event_data.cpp
+++ b/performance/src/utils/event_data.cpp
@@ -300,6 +300,16 @@ void event_data::fill_cca_result(
         found_meas_to_cluster_map[meas] = iocells;
     }
 
+    // Construct a mapping of geometry IDs to cells and the associated
+    // particles, in order to get much faster lookups later.
+    std::map<geometry_id, std::vector<std::reference_wrapper<const std::decay_t<
+                              decltype(m_cell_to_particle_map)>::value_type>>>
+        geo_id_to_cell_to_particle_map;
+
+    for (auto const& v : m_cell_to_particle_map) {
+        geo_id_to_cell_to_particle_map[v.first.geometry_id].emplace_back(v);
+    }
+
     for (auto const& [ms, cluster] : found_meas_to_cluster_map) {
 
         std::map<uint64_t, std::size_t> meas_counts;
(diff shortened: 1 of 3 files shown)


performance/src/utils/event_data.cpp
```diff
@@ -300,6 +300,16 @@ void event_data::fill_cca_result(
         found_meas_to_cluster_map[meas] = iocells;
     }
 
+    // Construct a mapping of geometry IDs to cells and the associated
+    // particles, in order to get much faster lookups later.
+    std::map<geometry_id, std::vector<std::reference_wrapper<const std::decay_t<
```
beomki-yeo, line 305: Would `std::unordered_map` even accelerate the performance? (only if it does not break the logic)

krasznaa: I believe `std::unordered_map` is really only relevant for speed when (with quasi-code) `operator<(std::hash(TYPE), std::hash(TYPE))` is faster than `operator<(TYPE, TYPE)`. Since the key in this map is an integer, I believe `std::unordered_map` would only slow things down.

Though I admit I'm on a bit shaky ground with this. 🤔

krasznaa (commented):
As far as I'm concerned, this all looks reasonable. Though I really don't have a good understanding of the innards of this code.

As long as @beomki-yeo is on board, I'm definitely on board.

performance/src/utils/event_data.cpp
stephenswat: > I believe `std::unordered_map` is really only relevant for speed when (with quasi-code) `operator<(std::hash(TYPE), std::hash(TYPE))` is faster than `operator<(TYPE, TYPE)`. Since the key in this map is an integer, I believe `std::unordered_map` would only slow things down.

No, not quite. :stuck_out_tongue:  An unordered map doesn't use `operator<` at all, and an ordered map doesn't use a hash function at all. Thus, an ordered map is an obvious choice if `operator<` is available but a hash function is not, and an unordered map is a good choice if a hash function is available but `operator<` is not. This is only a superficial difference though. The real trick is that with a properly distributed hash function, an unordered map achieves a $\mathcal{O}(1)$ best-case lookup time but a $\mathcal{O}(n)$ worst-case lookup time while the ordered map obviously achieves a $\Theta(\log n)$ lookup time.

Given that the hash function on integers is just the identity function, the distribution of the hash function of these geometry identifiers depends on the distribution of the geometry identifiers themselves which is obviously non-uniform. As such, the intuitive answer is that a tree-based ordered map will provide more consistent performance than an ordered map.

beomki-yeo (approved):
The PR looks good to me

PR #958, opened 2025-05-01
Alpaka: Unified Vecmem Interface

CrossR:
This is a draft PR, based on the comment in #857.

Any input would be welcome!

Essentially, this is supposed to achieve two things:

 - Unify the interface finally, so zero `ifdef SYCL` in user code or whatever, and intelligently pulling in `vecmem::[cuda|hip|sycl]` as required.
 - Because of the above, also removes any dependence on Alpaka in the `traccc::alpaka` public interface, as now there shouldn't be any dependence on Alpaka.

Whilst doing this, I also added an `alpaka::Queue` wrapper, which is needed for the SYCL vecmem objects, and will also be required for the upcoming async Alpaka code. It is based on the CUDA stream wrapper, with an additional method to get the underlying stream/queue too (i.e. from the `alpaka::Queue` get the underlying `cudaStream_t` or equivalent).

There is a few places I'm not 100% happy with....so I'll comment on them below if anyone has any input.


device/alpaka/src/utils/queue.cpp
```diff
@@ -0,0 +1,76 @@
...
+#elif defined(TRACCC_BUILD_SYCL)
+    auto nativeQueue = ::alpaka::getNativeHandle(*(m_queue->m_queue));
+    return static_cast<void*>(&nativeQueue);
+#else
+    // TODO: What is the best way to handle this?
+    //       Not having the method is a pain for other parts of the code,
+    //       but having it and throwing an error is not great either.
+    throw std::runtime_error("Native queue not available for this device");
+#endif
+}
```
CrossR, line 74: I'm not 100% happy with this bit either.

Essentially, we need to deal with 3 bits here:

 - CUDA + HIP use a pointer type for their stream.
 - SYCL does not use a pointer.
 - CPU outright can't call `getNativeHandle` as it is a compile error, as there is no equivalent there.

I ended up going this way (throwing an error) as it means in other places, like the queue, I can use a `if constexpr` to make sure it isn't called, but if I just don't define this method, I end up needing an `#if` in there to deal with it (and anywhere else I call this method).

examples/run/alpaka/throughput_st.cpp
```diff
@@ -16,9 +16,8 @@ int main(int argc, char* argv[]) {
 
     // Execute the throughput test.
     static const bool use_host_caching = true;
-    return traccc::throughput_st<
-        traccc::alpaka::full_chain_algorithm,
-        traccc::alpaka::vecmem_resources::host_memory_resource>(
+    return traccc::throughput_st<traccc::alpaka::full_chain_algorithm,
+                                 vecmem::host_memory_resource>(
```
CrossR, line 20: The only caveat of the Pimpl implementation is that this is now using the default `vecmem::host_memory_resource` rather than in the older case it was using the most appropriate vecmem resource (i.e. `cuda|hip|sycl`).

Am I missing something to enable this?

tests/alpaka/alpaka_basic.cpp
```diff
@@ -131,17 +131,37 @@ GTEST_TEST(AlpakaBasic, VecMemOp) {
...
+#elif defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+    vecmem::hip::copy vm_copy;
+    vecmem::hip::host_memory_resource host_mr;
+    vecmem::hip::device_memory_resource device_mr;
 #else
-    traccc::alpaka::vecmem_resources::device_copy vm_copy;
+    vecmem::copy vm_copy;
+    vecmem::host_memory_resource host_mr;
+    vecmem::host_memory_resource device_mr;
 #endif
```
CrossR, line 163: This is the single place using the older `ifdef` blocks...are we okay with that?

There isn't an easy way to ensure that `Acc` here matches the `Acc` picked in `traccc::alpaka`, so I figured that instead of trying to pass over and build around an existing Alpaka queue, it was easier in this single test to just ignore the new code.

device/alpaka/src/utils/queue.cpp
CrossR: Updated, and now uses the Alpaka flags not the generic ones I added since this isn't in a header etc...but the issue still remains that this could / should be neater.

device/alpaka/src/utils/get_vecmem_resource.cpp
```diff
@@ -0,0 +1,137 @@
...
+    using device_copy = vecmem::cuda::copy;
+    using device_async_copy = vecmem::cuda::async_copy;
+};
+template <>
+struct host_device_types<::alpaka::TagGpuHipRt> {
+    using device_memory_resource = vecmem::hip::device_memory_resource;
+    using host_memory_resource = vecmem::hip::host_memory_resource;
+    using managed_memory_resource = vecmem::hip::managed_memory_resource;
+    using device_copy = vecmem::hip::copy;
+    using device_async_copy = vecmem::hip::copy;
```
CrossR, line 43: When we upgrade vecmem to pull in the new version, I can update this to actually use `async` for HIP too.

My follow up PR will be Async-Alpaka, so zero rush / can be made then.

CrossR:
Closed in favour of #960.

PR #959, opened 2025-05-01
Alpaka VecMem Interface, main branch (2025.05.01.)

krasznaa:
This is a proposal for how I'd imagine using Alpaka in the longer term. This was triggered by #958, whose code I didn't really like. But since I have very little experience with Alpaka, I wanted to try out for myself what sort of code could be written.

@CrossR, @StewMH, this is how I imagine that things could look like. :thinking:
  - `traccc::alpaka::queue` is the same sort of thing that you tried to make as well Ryan. I just simplified it so that it would only expose the underlying Alpaka queue, and not try to be too smart with the native devices/queues.
  - `traccc::alpaka::vecmem_objects` has the interface that Ryan made, just with a (in my mind) much simpler implementation.

You see, once you hide the choice of VecMem objects behind such a non-specific interface, you can just use simple pre-processor statements in the background. There's no need to mix pre-processor formalism with traits in this case.

Please check if this sort of a setup could be made to work with the existing algorithms. Since that I didn't really try...

### DIFF
--- a/device/alpaka/CMakeLists.txt
+++ b/device/alpaka/CMakeLists.txt
@@ -30,6 +30,12 @@ traccc_add_alpaka_library( traccc_alpaka alpaka TYPE SHARED
   "include/traccc/alpaka/utils/make_prefix_sum_buff.hpp"
   "src/utils/make_prefix_sum_buff.cpp"
   "src/utils/get_device_info.cpp"
+  "include/traccc/alpaka/utils/queue.hpp"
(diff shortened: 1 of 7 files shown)


CrossR:
Thanks for this! I'll take a look today and look at integrating it back into the Alpaka algorithms, and see how I get on.

CrossR:
PR #960 is now opened with this commit as the base.

krasznaa:
Implemented in #960.

PR #960, opened 2025-05-02
Alpaka: Vecmem Interface

CrossR:
Based on the code in #959, just with a few small changes:

 - Test + examples updated (where appropriate) to use the new code.
 - ~~`ifdef` guard around `async_copy` init in the `CUDA || HIP` bit, just whilst we are still on an older vecmem release. Can be updated when we update that, though no rush as the async copy is not used in Alpaka yet.~~
 - I swapped the `alpaka::Queue` to default to device 0 in cases where there is no device given. Any opinions on this welcomed. I didn't / couldn't go down the CUDA route (which is `if device == INVALID_DEVICE get_current_device()`), as Alpaka has no concept of the current device. Could instead just remove the default and force a device ID to be given? I think either is preferable to just silently failing.

I've deleted the previous code now as well, which removes Alpaka from the `traccc::alpaka` public headers once again.


examples/run/alpaka/throughput_mt.cpp
```diff
@@ -5,20 +5,17 @@
...
 
 int main(int argc, char* argv[]) {
 
     // Execute the throughput test.
     static const bool use_host_caching = true;
-    return traccc::throughput_mt<
-        traccc::alpaka::full_chain_algorithm,
-        traccc::alpaka::vecmem_resources::host_memory_resource>(
+    return traccc::throughput_mt<traccc::alpaka::full_chain_algorithm,
+                                 vecmem::host_memory_resource>(
```
CrossR, line 18: This is the only bit I'm still not 100% sure on the impact of.

I.e. we now always use the default `vecmem::host_memory_resource`, rather than `vecmem::cuda::host_memory_resource` etc as appropriate.

I guess there is some performance impact around missing out on the pinning the going via `cudaMallocHost` gives? Not sure on the most appropriate way to surface this type though, or if it is easier to change the interface instead.

tests/alpaka/alpaka_basic.cpp
```diff
@@ -15,7 +15,24 @@
 #include <vecmem/memory/host_memory_resource.hpp>
 #include <vecmem/utils/copy.hpp>
 
-#include "traccc/alpaka/utils/get_vecmem_resource.hpp"
+// VecMem include(s).
```
krasznaa, line 18: I'm not sure why you went this way here. :confused:

This source file does need to remain an "Alpaka source file", but you could still make use of the new class for figuring out which VecMem types to use. :thinking:

krasznaa (changes requested):
I like how things look, but let's do one other kind of cleanup as part of this PR!

Now that `traccc::alpaka` takes care of everything internally, practically all executables should become "simple C\+\+ code". I.e.

https://github.com/acts-project/traccc/blob/main/examples/run/alpaka/CMakeLists.txt#L12-L13

should no longer be needed **I believe**. (But you'll need to test this.)

tests/alpaka/alpaka_basic.cpp
CrossR: The bit I'm struggling with here, is keeping all the types consistent.

All the Alpaka types used here (Dim, Idx, Acc, Queue, WorkDiv) are private, so they could technically be different to the ones used in `traccc::alpaka`.

That means whilst I can totally setup a wrapped queue and a `vecmem_object` object to pull vecmem types out of, I couldn't think of a good way of keeping things in sync between the two things. The chosen accelerator and queue type (blocking vs not) could be made different in `traccc::alpaka` and here.

It is entirely possible that I am missing some smart way of getting around that though... since this was a one-off test, I decided to go with the dumb way for now.

examples/run/alpaka/full_chain_algorithm.hpp
```diff
@@ -120,19 +121,17 @@ class full_chain_algorithm
...
-    ::sycl::queue m_queue;
-    vecmem::sycl::queue_wrapper m_queue_wrapper;
-#endif
-
     /// Device memory resource
-    traccc::alpaka::vecmem_resources::device_memory_resource m_device_mr;
+    ::vecmem::memory_resource& m_device_mr;
     /// Memory copy object
-    traccc::alpaka::vecmem_resources::device_copy m_copy;
+    mutable std::reference_wrapper<::vecmem::copy> m_copy;
```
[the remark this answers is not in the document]
CrossR: Done for all but `host_mr` (as it is passed into the algorithm and may be cached or not), and `m_cached_device_mr`.

examples/run/alpaka/CMakeLists.txt
```diff
@@ -38,7 +35,7 @@ add_library( traccc_examples_alpaka STATIC
    "full_chain_algorithm.hpp"
    "full_chain_algorithm.cpp" )
 target_link_libraries( traccc_examples_alpaka
-   PUBLIC alpaka::alpaka vecmem::core detray::core detray::detectors
+   PUBLIC vecmem::core detray::core detray::detectors
```
CrossR, line 38: `alpaka` calls now removed from the `examples` CMakeLists.

tests/alpaka/alpaka_basic.cpp
krasznaa: I don't quite understand your points about `Dim`, etc., but I think the problem here is that `traccc::alpaka::queue` needs to become a little smarter. :thinking: I think it will need to be like [vecmem::sycl::queue_wrapper](https://github.com/acts-project/vecmem/blob/main/sycl/include/vecmem/utils/sycl/queue_wrapper.hpp). (At the same time, at one point we should figure out if we want/need these `_wrapper` postfixes. I'm really not sure...)

The main thing being that `traccc::alpaka::queue` should be able to both own its personal `::alpaka::Queue` object, or be given one from the outside. So that this test's code could become:

```c++
auto alpakaQueue = Queue{devAcc};
traccc::alpaka::queue tracccQueue{&alpakaQueue};
traccc::alpaka::vecmem_objects vo{tracccQueue};
```

This way the logic of choosing the VecMem objects could still be taken from `traccc::alpaka` all the time. Even when the client code wants to use its own Alpaka queue/code.

CrossR: I was over-complicating things, and worrying about what happens if the `Acc` definition in this file and `traccc::alpaka` become out of sync, so what would happen if one decided they were using a CUDA device and one decided they were using a CPU device.

I think in reality, that couldn't end up happening, so it shouldn't be an issue.
I've added the sort of wrapper you linked to now, and that seems happy.

device/alpaka/include/traccc/alpaka/utils/queue.hpp
```diff
@@ -0,0 +1,63 @@
...
+        std::numeric_limits<std::size_t>::max();
+
+    /// Construct a new stream (possibly for a specified device)
+    explicit queue(std::size_t device = INVALID_DEVICE);
+
+    /// Wrap an existing ::alpaka::Queue object
+    ///
+    /// Without taking ownership of it!
+    ///
+    queue(void* input_queue);
```
krasznaa, line 33: To be **very** safe, you could add `explicit` to this constructor. Just as how SonarCloud recommends.

device/alpaka/src/utils/queue.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+};  // struct queue::impl
+
+queue::queue(std::size_t device) : m_impl{std::make_unique<impl>(device)} {}
+
+queue::queue(void* input_queue)
+    : m_impl{std::make_unique<impl>(INVALID_DEVICE)} {
+
+    // Copy the queue pointer
+    m_impl->m_queue = *static_cast<Queue*>(input_queue);
```
krasznaa, line 41: :thinking: This is not exactly how I imagined it...

This interface, with a `void*` pointer suggests that it just wraps around an existing `alpaka::Queue`. Not make a copy of it, and then use the copy from there on.

I rather imagined making the `impl` struct something like this:

https://github.com/acts-project/vecmem/blob/main/sycl/src/utils/sycl/queue_wrapper.sycl#L20-L25

**Not** with a `shared_ptr`, but rather with a `unique_ptr`. (Since this class, unlike `vecmem::sycl::queue_wrapper`, cannot be copied.) So the main thing is that there would be both a bare pointer, and a unique pointer inside of `impl`. Which would be set up according to how the main class was created.

examples/run/alpaka/CMakeLists.txt
krasznaa: :+1:

krasznaa (changes requested):
Sorry to continue the unpleasantness like this. :frowning: But I'd really like to get the behaviour of `traccc::alpaka::queue` "correct" from the start. :wink:

krasznaa (approved):
I like it. Hopefully the CI does too. 🤞

device/alpaka/src/utils/queue.cpp
```diff
@@ -0,0 +1,73 @@
...
+        ::alpaka::getDevByIdx(::alpaka::Platform<Acc>{}, m_impl->m_device));
+    m_impl->m_queue = m_impl->m_managedQueue.get();
+}
+
+queue::queue(void* input_queue) : m_impl{std::make_unique<impl>()} {
+
+    assert(input_queue != nullptr);
+    m_impl->m_queue = static_cast<Queue*>(input_queue);
+    m_impl->m_device =
+        ::alpaka::getNativeHandle(::alpaka::getDev(*m_impl->m_queue));
```
krasznaa, line 41: Argh... I didn't check, I thought the failure was because of an infrastructure issue.

But this is a new bug. I believe `::alpaka::getDev(...)` gets exactly the integer that you're after here. The `::alpaka::getNativeHandle(...)` "surrounding" should be removed.

But of course test it before pushing my suggestion. 😉

krasznaa (changes requested):

device/alpaka/src/utils/queue.cpp
CrossR: Looking into this, I think the way I am doing it is correct (`getDev` returns the actual `alpaka::DevUniformCudaHipRt` or `alpaka::DevGenericSycl` etc, rather than the device num)...but only for CUDA + HIP.

For SYCL, it returns the `sycl::device` + `sycl::context` (I guess because the "device" in use is a lot harder to distinguish, with the larger set of devices it could be), so I may just need to special case SYCL to pull out the `sycl::info::device::vendor_id` or whatever the appropriate ID is there from the `sycl::device`.

device/alpaka/src/utils/queue.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+#if defined(ALPAKA_ACC_SYCL_ENABLED)
+    auto sycl_device =
+        ::alpaka::getNativeHandle(::alpaka::getDev(*m_impl->m_queue)).first;
+    m_impl->m_device = static_cast<std::size_t>(
+        sycl_device.get_info<::sycl::info::device::vendor_id>());
+#else
+    m_impl->m_device = static_cast<std::size_t>(
+        ::alpaka::getNativeHandle(::alpaka::getDev(*m_impl->m_queue)));
+#endif
```
krasznaa, line 49: 🤔 What's the point of this `m_device` variable in the end? Why do we need it? It seems to me that Alpaka itself doesn't care about this variable a whole lot either.

To step back a bit: I think we're doing something silly with both CUDA(HIP) and SYCL in this code. The `std::size_t` index that Alpaka uses to choose a concrete device from all the available devices on a given platform is fairly close to how CUDA(HIP) defines a device. But it's still not the same. Alpaka, I believe on purpose, defined its own device identifier as an `std::size_t`, while CUDA(HIP) defines it as an `unsigned int`.

But the `vendor_id` from SYCL absolutely has nothing to do with the way that "device 0" (as an example) would be chosen from the available SYCL devices.

So to get back to my first question: Why do we need to provide this info from `traccc::alpaka::queue`, if clearly `::alpaka::Queue` doesn't provide such an ID itself in any convenient way? Even if we want to make a copy of an `::alpaka::Queue` (which we currently don't), we'd have better options for doing that than to use this integer. 🤔

CrossR: It was mostly to have some way to programatically surface "This specific device is in use" (alongside the more user facing "get device name as a string" method elsewhere).

That said, I think as you've said if a user does want more than just a basic understanding of the device, we can assume they'll include Alpaka and get things that way, and be in control of it themselves. I'll push up with it removed in a mo.

krasznaa (approved):
Let's get it in finally.

PR #961, opened 2025-05-02
make sycl measurement sorting algo interface match cuda's

flg:
Not sure it's worth it but I'll let traccc folks decide, since it was suggested by @krasznaa. It would save G-200 one template specialization: great success :)

### DIFF
--- a/device/sycl/include/traccc/sycl/clusterization/measurement_sorting_algorithm.hpp
+++ b/device/sycl/include/traccc/sycl/clusterization/measurement_sorting_algorithm.hpp
@@ -21,6 +21,10 @@
 // System include(s).
 #include <functional>
 
+namespace traccc {
+struct memory_resource;
+}
+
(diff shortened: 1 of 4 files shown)


device/sycl/include/traccc/sycl/clusterization/measurement_sorting_algorithm.hpp
```diff
@@ -21,6 +21,10 @@
 // System include(s).
 #include <functional>
 
+namespace traccc {
+struct memory_resource;
```
krasznaa, line 25: No need to be too clever here. Let's just include the appropriate header for this type.

krasznaa (changes requested):
I'm on board. It's probably good to add a memory resource to the API. Even if the algorithm doesn't need it.

krasznaa (approved):
:+1:

PR #965, opened 2025-05-05
Strip Spacepoints, main branch (2025.05.05.)

krasznaa:
After a bit of trial-and-error, I think this would be the way to go with the spacepoint EDM to enable the usage of strip spacepoints. :thinking:

After all the trial-and-error I'm still not sure though whether this is the correct way or not. :thinking: I might still give jagged vectors another go. (As I just now realised, while writing down what would be the deal-breaker for jagged vectors, that that's not really a deal-breaker after all...)

Still, I wanted to send this to @benjaminhuth for info. To see if this API may be a possibility for his needs.

### DIFF
--- a/core/include/traccc/edm/spacepoint_collection.hpp
+++ b/core/include/traccc/edm/spacepoint_collection.hpp
@@ -39,30 +39,47 @@ class spacepoint : public BASE {
     /// @name Spacepoint Information
     /// @{
 
+    /// Invalid measurement index
+    static constexpr unsigned int INVALID_MEASUREMENT_INDEX =
+        std::numeric_limits<unsigned int>::max();
+
     /// The index of the measurement producing this spacepoint (non-const)
     ///
     /// @return A (non-const) vector of <tt>unsigned int</tt> values
     ///
     TRACCC_HOST_DEVICE
-    auto& measurement_index() { return BASE::template get<0>(); }
+    auto& measurement_index_1() { return BASE::template get<0>(); }
     /// The index of the measurement producing this spacepoint (const)
     ///
     /// @return A (const) vector of <tt>unsigned int</tt> values
     ///
     TRACCC_HOST_DEVICE
-    const auto& measurement_index() const { return BASE::template get<0>(); }
+    const auto& measurement_index_1() const { return BASE::template get<0>(); }
+
+    /// The index of the measurement producing this spacepoint (non-const)
+    ///
+    /// @return A (non-const) vector of <tt>unsigned int</tt> values
+    ///
+    TRACCC_HOST_DEVICE
+    auto& measurement_index_2() { return BASE::template get<1>(); }
+    /// The index of the measurement producing this spacepoint (const)
+    ///
+    /// @return A (const) vector of <tt>unsigned int</tt> values
+    ///
+    TRACCC_HOST_DEVICE
+    const auto& measurement_index_2() const { return BASE::template get<1>(); }
 
     /// Global / 3D position of the spacepoint
     ///
     /// @return A (non-const) vector of @c traccc::point3 values
     ///
-    TRACCC_HOST_DEVICE auto& global() { return BASE::template get<1>(); }
+    TRACCC_HOST_DEVICE auto& global() { return BASE::template get<2>(); }
     /// Global / 3D position of the spacepoint
     ///
     /// @return A (const) vector of @c traccc::point3 values
     ///
     TRACCC_HOST_DEVICE const auto& global() const {
-        return BASE::template get<1>();
+        return BASE::template get<2>();
     }
 
     /// The X position of the spacepoint (non-const)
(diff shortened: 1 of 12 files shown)


core/include/traccc/edm/spacepoint_collection.hpp
```diff
@@ -39,30 +39,47 @@ class spacepoint : public BASE {
...
     TRACCC_HOST_DEVICE
-    const auto& measurement_index() const { return BASE::template get<0>(); }
+    const auto& measurement_index_1() const { return BASE::template get<0>(); }
+
+    /// The index of the measurement producing this spacepoint (non-const)
+    ///
+    /// @return A (non-const) vector of <tt>unsigned int</tt> values
+    ///
+    TRACCC_HOST_DEVICE
+    auto& measurement_index_2() { return BASE::template get<1>(); }
```
stephenswat, line 64: Why do we need to keep two indices?

krasznaa:
Never mind, jagged vectors are not a good fit for the current code after all...

At the moment at least we create spacepoints using resizable buffers. Ones that would have a size equal the number of measurements, and we would fill them up with as many spacepoints, as many **pixel measurements** we have.
  - https://github.com/acts-project/traccc/blob/main/device/cuda/src/seeding/spacepoint_formation_algorithm.cu#L51
  - https://github.com/acts-project/traccc/blob/main/device/common/include/traccc/seeding/device/impl/form_spacepoints.ipp#L42

With jagged vectors the whole thing would need to be re-thought a bit. Since we cannot have resizable jagged vectors quite like this. :thinking:

So on first order, let's see what everyone thinks about this simple proposal, with separate variables for the indices.

core/include/traccc/edm/spacepoint_collection.hpp
benjaminhuth: Strip spacepoints consist of the combination of two 1D measurements

krasznaa: In case / once we form spacepoints from strip measurements, we need at least / exactly two measurements to form one spacepoint.

This is something that @benjaminhuth needs for his code. The ML model would give us all the spacepoints belonging to the same track, but that then needs to be translated back into measurements, so that we could perform a final Kálmán fit using all those measurements.

stephenswat: Ah yes okay, good.

krasznaa: > E.g., the ActsExamples SimSpacepoint stores them in a `boost::containers::static_vector<SourceLink, 2>`.

:thinking: At one point I should really make [vecmem::static_vector](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/static_vector.hpp) work properly in device code. It could be a good candidate here indeed, similar to what Acts does.

benjaminhuth:
Thanks @krasznaa, I think that API would fullfill the needs. One thought: If the additional index creates a performance/memory issue for e.g. pixel seeding, I think it would be possible to have separate types as well, like `pixel_spacepoint_collection` and `strip_spacepoint_collection`.
I would anyways try to process them seperately if possible, to avoid the branch with the index-check in the kernel...

core/include/traccc/edm/spacepoint_collection.hpp
stephenswat: Can't we just use `std::array<measurement_index, 2>` or just `measurement_index[2]`?

Let's absolutely not do this with jagged vectors though, that sounds awful in terms of ergonomics!

krasznaa:
I **think** most seeding algorithms would want to see the different types of spacepoints in the same container. :thinking: Or at least some of them... So having a single type would still be my very clear preference.

core/include/traccc/edm/spacepoint_collection.hpp
krasznaa: The reason I didn't do a simple array was to avoid the extra 32-bit memory movement. :thinking: Though a static vector would have the same overhead as well. (Actually, even a bit more.)

But other than possible performance reasons, an array could technically be used, yes.

benjaminhuth: One could expose the `2` as an integer template parameter, and keep the code flexible in a way, that e.g., for a fast tracking chain with pixel only seeding there could be zero overhead in terms of memory...

krasznaa: Technically that wouldn't make things different from having 2 separate types. The device code needs to use a specific count, so we'd still need to define a separate pixel and strip spacepoint typedef for the two cases.

No, a fixed array in which the second element is either valid or not, may indeed be the best middle ground here. :thinking:

benjaminhuth: Okay, any way is good for me I think.

krasznaa:
Having looked back through the code, I think there aren't any other additions for now after all. :thinking:

I'd propose to stay with the two separate variables for now. This should have the smallest overhead over the current code. (Using an array would have a negative effect on our memory access pattern.)

PR #970, opened 2025-05-08
Clean Device Track Finding, main branch (2025.05.08.)

krasznaa:
While working on the EDM, I found that the CUDA and Alpaka track finding algorithms are currently implemented in a bit of a sketchy way. Providing a not-too-perfect API.

While fixing that, I also simplified the algorithm implementations a little. (More so for Alpaka than for CUDA...)

Pinging @CrossR as well.

### DIFF
--- a/device/alpaka/include/traccc/alpaka/finding/finding_algorithm.hpp
+++ b/device/alpaka/include/traccc/alpaka/finding/finding_algorithm.hpp
@@ -30,8 +30,8 @@ class finding_algorithm
     : public algorithm<track_candidate_container_types::buffer(
           const typename navigator_t::detector_type::view_type&,
           const typename stepper_t::magnetic_field_type&,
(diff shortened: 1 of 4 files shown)


CrossR:
Alpaka side of things looks good to me, and a quick test locally shows everything is happy!

krasznaa:
> Alpaka side of things looks good to me, and a quick test locally shows everything is happy!

To make it clear, in the past this verbose usage of `vecmem::get_data(...)` was necessary. But when @stephenswat introduced the "payload structs", the calling convention could be simplified a whole lot. We just never did so.

I also saw a few cases since then in the SYCL code where some simplification will be possible. But since the API of the SYCL code is to my liking, I didn't want to mix those changes into this PR. :wink:

stephenswat (approved):

PR #974, opened 2025-05-13
CMake 4 Updates, main branch (2025.05.13.)

krasznaa:
These are some changes that make the build warning-free with CMake 4+.

The Alpaka warning about `find_package(Boost)` I just had to "patch away". 😦 As setting policies "across file boundaries" doesn't really work.

The rest of the updates simply use [CMAKE_POLICY_VERSION_MINIMUM](https://cmake.org/cmake/help/v4.0/variable/CMAKE_POLICY_VERSION_MINIMUM.html), which became available with CMake 4.0. Exactly to be able to handle these sort of situations. (For the externals that already require something newer than CMake 3.10, I didn't need to do any of this.)


beomki-yeo (approved):
Looks good to me

PR #976, opened 2025-05-13
Make Greedy solver handel duplicate measurements in patterns

beomki-yeo:
This is from PR #943

One problem of the current and legacy greedy resolver is that the algorithm fails when there is a duplicate measurement in the pattern. This happens occasionally when CKF unexpectedly adds duplicate measurement in the pattern.

While that's mainly due to the bug in the CKF but it does not hurt to make the greedy resolver to handle such cases correctly.  

I also added a couple of tests for duplicate measurements and complicated the patterns of the existing tests.


stephenswat (approved):
LGTM.

PR #977, opened 2025-05-14
Refactor the cell data initialization in test

beomki-yeo:
Using `push_back` in the clusterization test code


stephenswat (approved):
Yes, this is much nicer. Thanks!

PR #979, opened 2025-05-15
Fix dangling reference in full chain algorithms

stephenswat:
A segmentation fault spotted by @CrossR has exposed a few problems in the way we handle full chain algorithms in our examples code. Fundamentally, the problem is that we create a full chain algorithm `A` which contains a copy object `A.copy` and a subalgorithm `A.subalg`. This subalgorithm has a copy object reference `A.subalg.copy = A.copy`. We then move `A` into `A'` which properly moves `A.copy` into `A'.copy` and `A.subalg` into `A'.subalg`, but the reference `A'.subalg.copy` still points to `A.copy` and not `A'.copy`, leading to a dangling reference.

I fixed this problem by using smart pointers, which avoid having the copy object embedded in the algorithm. I also had to make some further changes to make the algorithms properly movable, but this allowed me to remove some of the very verbose custom move- and copy constructors we had!

### DIFF
--- a/examples/run/alpaka/full_chain_algorithm.hpp
+++ b/examples/run/alpaka/full_chain_algorithm.hpp
@@ -99,15 +99,11 @@ class full_chain_algorithm
                          host_detector_type* detector,
                          std::unique_ptr<const traccc::Logger> logger);
 
-    /// Copy constructor
-    ///
-    /// An explicit copy constructor is necessary because in the MT tests
-    /// we do want to copy such objects, but a default copy-constructor can
-    /// not be generated for them.
-    ///
-    /// @param parent The parent algorithm chain to copy
-    ///
-    full_chain_algorithm(const full_chain_algorithm& parent);
+    full_chain_algorithm() = delete;
+    full_chain_algorithm(const full_chain_algorithm& other) = delete;
+    full_chain_algorithm(full_chain_algorithm&& other) noexcept;
+    full_chain_algorithm& operator=(const full_chain_algorithm& other) = delete;
+    full_chain_algorithm& operator=(full_chain_algorithm&& other) = delete;
 
     /// Algorithm destructor
     ~full_chain_algorithm();
@@ -122,9 +118,9 @@ class full_chain_algorithm
 
     private:
     /// Alpaka Queue
-    traccc::alpaka::queue m_queue;
+    std::shared_ptr<traccc::alpaka::queue> m_queue;
     /// Alpaka Vecmem objects, to get the memory resources
-    traccc::alpaka::vecmem_objects m_vecmem_objects;
+    std::shared_ptr<traccc::alpaka::vecmem_objects> m_vecmem_objects;
 
     /// Host memory resource
     ::vecmem::memory_resource& m_host_mr;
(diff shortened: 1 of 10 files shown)


stephenswat:
I put back the reference wrappers. :wink:

examples/run/alpaka/full_chain_algorithm.hpp
```diff
@@ -122,9 +118,9 @@ class full_chain_algorithm
 
     private:
     /// Alpaka Queue
-    traccc::alpaka::queue m_queue;
+    std::shared_ptr<traccc::alpaka::queue> m_queue;
```
krasznaa, line 121: Why shared pointer? You explicitly disallow the copy of the class anyway. Then why not a unique pointer?

examples/run/alpaka/full_chain_algorithm.hpp
```diff
@@ -99,15 +99,11 @@ class full_chain_algorithm
...
-    /// not be generated for them.
-    ///
-    /// @param parent The parent algorithm chain to copy
-    ///
-    full_chain_algorithm(const full_chain_algorithm& parent);
+    full_chain_algorithm() = delete;
+    full_chain_algorithm(const full_chain_algorithm& other) = delete;
+    full_chain_algorithm(full_chain_algorithm&& other) noexcept;
+    full_chain_algorithm& operator=(const full_chain_algorithm& other) = delete;
+    full_chain_algorithm& operator=(full_chain_algorithm&& other) = delete;
```
krasznaa, line 106: I still feel fairly strongly about this, that if a class is allowed to be move-constructed, then it should be allowed to be moved-assigned as well. I don't like introducing such a non-intuitive behaviour.

examples/run/alpaka/full_chain_algorithm.hpp
```diff
@@ -122,9 +118,9 @@ class full_chain_algorithm
...
     private:
     /// Alpaka Queue
-    traccc::alpaka::queue m_queue;
+    std::shared_ptr<traccc::alpaka::queue> m_queue;
     /// Alpaka Vecmem objects, to get the memory resources
-    traccc::alpaka::vecmem_objects m_vecmem_objects;
+    std::shared_ptr<traccc::alpaka::vecmem_objects> m_vecmem_objects;
 
     /// Host memory resource
     ::vecmem::memory_resource& m_host_mr;
```
krasznaa, line 126: Since we have the bare references, move-assignment is actually disabled at the moment. :frowning: Fine, let's not change that in this PR, but then why do we need to even make it explicit? The code was already implicitly not allowing any assignments anyway. :thinking:

examples/run/cpu/full_chain_algorithm.hpp
```diff
@@ -90,8 +97,10 @@ class full_chain_algorithm : public algorithm<track_state_container_types::host(
         const edm::silicon_cell_collection::host& cells) const override;
 
     private:
+    /// The host memory resource
+    vecmem::memory_resource& m_mr;
```
krasznaa, line 101: You correctly removed the configuration objects. I think that's correct, it was just making things more obscure that they were kept in the class.

But then why add `m_mr`? The class doesn't need it. :confused:

examples/run/cpu/full_chain_algorithm.hpp
```diff
@@ -90,8 +97,10 @@ class full_chain_algorithm : public algorithm<track_state_container_types::host(
...
     /// Vecmem copy object
-    vecmem::copy m_copy;
+    std::shared_ptr<vecmem::copy> m_copy;
```
krasznaa, line 103: Also here, I believe `std::unique_ptr` should be used.

examples/run/cuda/full_chain_algorithm.hpp
```diff
@@ -125,13 +121,13 @@ class full_chain_algorithm
...
+    std::shared_ptr<stream> m_stream;
     /// Device memory resource
-    vecmem::cuda::device_memory_resource m_device_mr;
+    std::shared_ptr<vecmem::cuda::device_memory_resource> m_device_mr;
     /// Device caching memory resource
-    std::unique_ptr<vecmem::binary_page_memory_resource> m_cached_device_mr;
+    std::shared_ptr<vecmem::binary_page_memory_resource> m_cached_device_mr;
     /// (Asynchronous) Memory copy object
-    mutable vecmem::cuda::async_copy m_copy;
+    std::shared_ptr<vecmem::cuda::async_copy> m_copy;
```
krasznaa, line 130: What's with all the shared pointers? :confused:

krasznaa (changes requested):
In general I'm on board.

PR #980, opened 2025-05-15
Alpaka: Asynchronous Alpaka

CrossR:
I've put this as a draft whilst #979 is in review, as that should go in first, and this touches the same spots in the Alpaka full chain.

This moves the Alpaka algorithm chain over to being truly asynchronous (at least, when the backend allows it, if not for a CPU backend, the Queue and `vecmem::copy` objects both become synchronous once again).

This is achieved with a few mains changes:

 - Swap to a `NonBlocking` queue.
 - Pass around the new `traccc::alpaka::queue` object to each algorithm, and use a single, consistent queue.
 - Swap to using `vecmem::[cuda|hip|sycl]::async_copy`.
 - Update the examples for this new approach.
 - Removing unnecessary sync calls.
   -  That is, instead of sync-ing after every single Alpaka call, just let things run and only sync when stuff is needed on the host. This is better for performance, but also stability.

Apologies for the number of files touched...there isn't a clean way to split this up, as both the code and examples needs updating simultaneously.

### DIFF
--- a/device/alpaka/src/finding/combinatorial_kalman_filter.hpp
+++ b/device/alpaka/src/finding/combinatorial_kalman_filter.hpp
@@ -554,7 +554,7 @@ struct BlockSharedMemDynSizeBytes<
         TArgs const&... /* args */
         ) -> std::size_t {
         return static_cast<std::size_t>(blockThreadExtent.prod()) *
-                   sizeof(unsigned int) +
+                   sizeof(unsigned long long int) +
                2 * static_cast<std::size_t>(blockThreadExtent.prod()) *
                    sizeof(std::pair<unsigned int, unsigned int>);
     }
--- a/device/cuda/src/finding/combinatorial_kalman_filter.cuh
+++ b/device/cuda/src/finding/combinatorial_kalman_filter.cuh
@@ -289,7 +289,7 @@ combinatorial_kalman_filter(
             const unsigned int nBlocks =
                 (n_in_params + nThreads - 1) / nThreads;
             const std::size_t shared_size =
-                nThreads * sizeof(unsigned int) +
+                nThreads * sizeof(unsigned long long int) +
                 2 * nThreads * sizeof(std::pair<unsigned int, unsigned int>);
 
             // Run the kernel.


device/alpaka/include/traccc/alpaka/clusterization/clusterization_algorithm.hpp
```diff
@@ -67,12 +71,14 @@ class clusterization_algorithm
...
 
     private:
-    /// The average number of cells in each partition
-    config_type m_config;
     /// The memory resource(s) to use
     traccc::memory_resource m_mr;
     /// The copy object to use
     std::reference_wrapper<vecmem::copy> m_copy;
+    /// The Alpaka queue to use
+    std::reference_wrapper<queue> m_queue;
```
CrossR, line 79: This is done inconsistently throughout the code... both here and in the CUDA.

I.e. some places use `vecmem::copy&` and `stream&`/`queue&` and other places use `std::reference_wrapper`.

I don't think this is an "issue" per se, but maybe something we should standardise eventually.

device/alpaka/src/clusterization/measurement_sorting_algorithm.cpp
```diff
@@ -11,30 +11,44 @@
...
+            .on(stream);
+#elif defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+    auto stream = ::alpaka::getNativeHandle(queue);
+    auto execPolicy = thrust::hip_rocprim::par_nosync(
+                          std::pmr::polymorphic_allocator(&(m_mr.main)))
+                          .on(stream);
 #else
-    auto thrustExecPolicy = thrust::host;
+    auto execPolicy = thrust::host;
 #endif
```
CrossR, line 49: This is not very ideal still.

I'm going to follow up this PR with a "Thrust Unification" (name to be determined...) that gets the most appropriate execution policy from `thrust` / `rocThrust` / `oneDPL`, and then calls the relevant algorithm, since we need to do that to start to push forward SYCL support.

CrossR:
Not hitting the HIP build issue locally annoyingly...so a bit confused as to why it is failing to deduce things.

PR #982, opened 2025-05-20
Add translation to silicon detector description

stephenswat:
In order to model the Lorentz shift and other effects, we will need to be able to apply a module-specific translation to measurements created in the clustering algorithm. This commit adds the infrastructure to do that.

Note that properly modelling the Lorentz shift will allow us to achieve a 1:1 match between the Athena and traccc clustering.

Tagging @paradajzblond.

### DIFF
--- a/core/include/traccc/geometry/silicon_detector_description.hpp
+++ b/core/include/traccc/geometry/silicon_detector_description.hpp
@@ -191,6 +191,20 @@ class silicon_detector_description_interface : public BASE {
     TRACCC_HOST_DEVICE
     const auto& dimensions() const { return BASE::template get<7>(); }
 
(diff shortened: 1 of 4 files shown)


krasznaa (approved):
Is this correction such a very constant value? Curious, I thought it might be a bit more elaborate of a calculation. 🤔

Sure, this works. The name of the variable (`measurement_translation`) is a bit long, but I also don't have a good alternative. This is why we have CoPilot after all... 😛

stephenswat:
> Is this correction such a very constant value?

I was also surprised to learn this, but @paradajzblond confirmed that it really is that simple. The $\Omega$ factor is less constant, but apparently that is no longer used by Athena and we should really only need the Lorentz shift.

PR #983, opened 2025-05-20
Add cluster diameter computation to clustering

stephenswat:
This commit allows the clustering algorithm to estimate the diameter of a cluster using four different strategies. The ability to compute this metric is required for the integration of GBTS into traccc.


device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
```diff
@@ -140,6 +151,21 @@ inline void aggregate_cluster(
...
+        out.diameter = delta0;
+    } else if (cfg.diameter_strategy ==
+               clustering_diameter_strategy::CHANNEL1) {
+        out.diameter = delta1;
+    } else if (cfg.diameter_strategy == clustering_diameter_strategy::MAXIMUM) {
+        out.diameter = std::max(delta0, delta1);
+    } else if (cfg.diameter_strategy ==
+               clustering_diameter_strategy::DIAGONAL) {
+        out.diameter = math::sqrt(delta0 * delta0 + delta1 * delta1);
+    }
```
krasznaa, line 168: A `switch` expression may be a bit easier for the compiler to optimize here. 🤔 (I like `switch` statements...)

krasznaa (commented):
It looks reasonable to me. I believe it will be fairly easy to adjust the algorithms to use this code.

As long as nobody complains soon, I think you should go forward like this.

device/common/include/traccc/clusterization/device/impl/aggregate_cluster.ipp
stephenswat: Hmm, why? I find the if-else construct a lot easier to read and there is definitely no performance difference.

krasznaa (approved):
Let's put it into the merge queue...

PR #969, opened 2025-05-08
feat: Update to detray v0.96.0

niermann999:
Fix some new assertions and make the host side algebra print operators available for debugging

PR #984, opened 2025-05-21
Track Finding Tweaks, main branch (2025.05.21.)

krasznaa:
Updates to help with understanding the current assertion failures coming from Detray.

There was a missing protection in the host code against considering material interaction on surfaces that have no material attached. #969 protected the device code against this, but the host code didn't receive this update.

Then, since I added a bunch of verbose printouts in the code to try to understand the next set of assertions, I thought I might as well request these printouts to be added to the repository. 🤔 The printed messages are pretty rough, but they are already better than nothing in my mind.

### DIFF
--- a/core/include/traccc/finding/details/find_tracks.hpp
+++ b/core/include/traccc/finding/details/find_tracks.hpp
@@ -17,6 +17,7 @@
 #include "traccc/fitting/kalman_filter/gain_matrix_updater.hpp"
 #include "traccc/fitting/status_codes.hpp"
 #include "traccc/sanity/contiguous_on.hpp"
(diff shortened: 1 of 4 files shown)


beomki-yeo:
Is that assertion from the detector without any material mapping? Because track parameters in the CKF loop are always supposed to be on sensitive surfaces.

niermann999:
> Is that assertion from the detector without any material mapping? Because track parameters in the CKF loop are always supposed to be on sensitive surfaces.

I think this is expected, because in the ODD and ITk there is no material mapped to the sensitive surfaces.

niermann999 (approved):

PR #991, opened 2025-05-27
Refactor Kálmán fitter forward-backward logic

stephenswat:
This commit slightly refactors the logic in the Kálmán fitter, splitting it up into a step function and separating the filter function from the smooth function. The idea here is to split the CUDA Kálmán fitter function in two later.

PR #994, opened 2025-05-30
Split KF kernel into forward and backward steps

stephenswat:
This PR splits the Kálmán filter fitting kernel into forwards and backwards steps in order to shrink the kernels and allow better future optimization. Depends on #991.


stephenswat:
This now compiles on my machine, so I consider it ready for review.

beomki-yeo (approved):
Looks good to me. We might need to discuss @noyaboy next week as #1001 seems to have a similar idea :thinking:

krasznaa (approved):
Eventually we'll want to write additional documentation for the code, and maybe even introduce some `.ipp` files. But since this is pretty urgent to get in, I'm happy with its current state.

PR #996, opened 2025-05-30
VecMem 1.18.0 Update, main branch (2025.05.30.)

krasznaa:
Updated the project to [vecmem-1.18.0](https://github.com/acts-project/vecmem/releases/tag/v1.18.0).

The main thing the new version brings is a fix to some (more or less) corner cases with memory copies. Like the one that @beomki-yeo ran into in #943. With those fixed, I was able to simplify the container copy algorithms a bit. (Even if these will disappear once the SoA migration is done, this cleanup still seemed like a good idea.)


beomki-yeo (approved):
Looks good to me

PR #999, opened 2025-05-31
feat: Sort measurements to be able to use fewer branches

niermann999:
This is something I started a while back that is not yet working, but may be a starting point. I comes from PR #761, which already does this


niermann999:
Nvm, seems to have been solved already

PR #989, opened 2025-05-23
Reduce CKF branching factor from 10 to 2

stephenswat:
Following #929, we can now reduce the branching factor without incurring a decrease in physics performance. This commit drops the branching dramatically from 10 to 2.

PR #929, opened 2025-03-28
Keep states ordered by quality in CKF

stephenswat:
This commit allows the `find_tracks` kernel in the CKF to keep the best measurements instead of keeping random ones, which will hopefully allow us to significantly reduce the branching factor of the CKF. The logic here works by employing a series of mutexes, allowing threads to lock a critical section and insert their own measurement, overwriting the previous ones.

PR #1000, opened 2025-05-31
Find Best Host Measurements, main branch (2025.05.31.)

krasznaa:
In order to fix the currently failing unit tests, I added measurement sorting to the host track finding algorithm. That should implement a "similar enough" logic as what was added in #929.

With that added, I spent an embarrassing amount of time trying to understand why `CpuCkfCombinatoricsTelescopeTests` broke. :frowning: That turned out to be due to an oversight in #989. The test implicitly relied on the per surface branching fraction being large. :thinking:

### DIFF
--- a/tests/cpu/test_ckf_combinatorics_telescope.cpp
+++ b/tests/cpu/test_ckf_combinatorics_telescope.cpp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2023-2024 CERN for the benefit of the ACTS project
+ * (c) 2023-2025 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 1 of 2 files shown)


tests/cpu/test_ckf_combinatorics_telescope.cpp
```diff
@@ -115,10 +115,12 @@ TEST_P(CpuCkfCombinatoricsTelescopeTests, Run) {
     traccc::finding_config cfg_no_limit;
     cfg_no_limit.max_num_branches_per_seed =
         std::numeric_limits<unsigned int>::max();
+    cfg_no_limit.max_num_branches_per_surface = 10;
     cfg_no_limit.chi2_max = 30.f;
 
     traccc::finding_config cfg_limit;
     cfg_limit.max_num_branches_per_seed = 500;
+    cfg_limit.max_num_branches_per_surface = 10;
```
stephenswat, line 123: I see you went through the same fun experience I had with #989. :smile:

stephenswat (approved):

PR #1001, opened 2025-06-02
Kalman Filter Acceleration via Simple Block-Level Barrier: Achieving 7.25% Throughput Gain

noyaboy:
Partitioning the Kalman Filter into Three Stages with __syncthreads(), Allowing Early Register Release to Mitigate Spilling and Cache Pollution. I use the following full reconstruction chain on the NVIDIA RTX 2080 Ti. 
```
./build/bin/traccc_throughput_mt_cuda \
  --use-detray-detector=on \
  --use-acts-geom-source=on \
  --input-directory=odd/geant4_10muon_10GeV/ \
  --input-events=10 \
  --processed-events=1000 \
  --threads=1
```
Applying this procedure resulted in a throughput improvement, achieving 23.1102 ms/event. The baseline measurement, taken from traccc upstream (upstream/main, commit 1febe21), was 24.7872 ms/event (fix: Preserve hit-to-meas and hit-to-id mapping after measurement sorting (#946)).


beomki-yeo:
My understanding is that `__syncthreads()`  is usually used to avoid the race condition rather than achieving the better computing performance 🤔 

> Allowing Early Register Release to Mitigate Spilling and Cache Pollution

I don't know how this is related to `__syncthreads()` but it is interesting

beomki-yeo:
My humble opinion is that the `return` is called earlier in this PR, which can free the memory quicker. I don't think `__syncthreads()` can be used for such compiler optimization

noyaboy:
@beomki-yeo: Thank you for your precise question — it motivated me to run an ablation study examining the performance impact of synchronization and function splitting.

I still use the following full reconstruction chain on the NVIDIA RTX 2080 Ti.
```
./build/bin/traccc_throughput_mt_cuda \
  --use-detray-detector=on \
  --use-acts-geom-source=on \
  --input-directory=odd/geant4_10muon_10GeV/ \
  --input-events=10 \
  --processed-events=1000 \
  --threads=1
  ```

Version A incorporates `__syncthreads()` and separates the filter logic into multiple functions. Its inverse throughput, averaged over six runs, was `22.2607 ms/event`, with a standard deviation of `0.0829 ms/event`.

Version B removes `__syncthreads()` but maintains the same function-splitting structure. Its average inverse throughput over six runs was significantly slower, at `42.5906 ms/event`, with a standard deviation of `0.2454 ms/event`.

From these results, it seems that although Version B adopts function splitting, its performance is not only inferior to Version A, but also lags behind the baseline `(24.7872 ms/event)`. A possible explanation is that, without synchronization, the compiler may retain live registers for each stage across the split functions, leading to prolonged register lifetimes and increased spilling. In comparison, the baseline—despite not splitting the stages—might allow the compiler to optimize register usage more effectively across the entire three-stage pipeline.

More importantly, when compared with the baseline, detailed profiling shows a drop in the number of registers per thread from 255 to 143 in Version A. This indicates that `__syncthreads()` may have played a beneficial role in alleviating register pressure.

**Comment by beomki-yeo** [comment]
I tested the same command and I got 23.649 ms/event for version A and 23.8617 for version B. (Device: NVIDIA A30)

version B diff
```
byeo@hermes:/mnt/nvme0n1/byeo/projects/traccc/traccc$ git diff
diff --git a/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp b/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
index 18726bd0..ecbe3d12 100644
--- a/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
+++ b/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
@@ -245,7 +245,7 @@ class kalman_fitter {
 
         //  Block-level同步以提早釋放暫存器
         #ifdef __CUDA_ARCH__
-            __syncthreads();
+            //__syncthreads();
         #endif
         return kalman_fitter_status::SUCCESS;
     }
@@ -263,7 +263,7 @@ class kalman_fitter {
         }
 
         #ifdef __CUDA_ARCH__
-            __syncthreads();
+            //__syncthreads();
         #endif
         return kalman_fitter_status::SUCCESS;
     }
```

File: .gitignore [hunk 1]
```diff
@@ -40,3 +40,18 @@
 # Build system related files.
 CMakeUserPresets.json
 out/
+
```
ytchoutw, line 43: You may want to avoid including personal gitignore settings here.

stephenswat: Protip: you can put personal gitignore preferences in `.git/info/exclude` and they will work without polluting your working tree.

noyaboy:
> I tested the same command and I got 23.649 ms/event for version A, i.e. this PR, and 23.8617 ms/event for version B. (Device: NVIDIA A30)
> 
> version B diff:
> 
> ```diff
> byeo@hermes:/mnt/nvme0n1/byeo/projects/traccc/traccc$ git diff
> diff --git a/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp b/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
> index 18726bd0..ecbe3d12 100644
> --- a/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
> +++ b/core/include/traccc/fitting/kalman_filter/kalman_fitter.hpp
> @@ -245,7 +245,7 @@ class kalman_fitter {
>  
>          //  Block-level同步以提早釋放暫存器
>          #ifdef __CUDA_ARCH__
> -            __syncthreads();
> +            //__syncthreads();
>          #endif
>          return kalman_fitter_status::SUCCESS;
>      }
> @@ -263,7 +263,7 @@ class kalman_fitter {
>          }
>  
>          #ifdef __CUDA_ARCH__
> -            __syncthreads();
> +            //__syncthreads();
>          #endif
>          return kalman_fitter_status::SUCCESS;
>      }
> ```

noyaboy:
@beomki-yeo It seems to depend on the device and the computing capability we use.

beomki-yeo:
That's a possible explanation but your argument a bit sounds like "`__syncthreads()` alleviates the register pressure because `__syncthreads()` alleviates the register pressure." Not saying you are wrong but I would have better understanding on this 🤔

noyaboy:
@beomki-yeo Inserting two __syncthreads() statements effectively divides a CUDA kernel into three distinct execution phases: before the first synchronization point, between the two synchronization points, and after the second one. From the compiler's perspective, each __syncthreads() serves as a barrier across all threads. This means that any computation result produced before the barrier that is not used afterward can be considered "no longer live," and the associated registers can be reclaimed early.

In other words, before generating the final machine code (SASS), the compiler first translates the program into PTX intermediate representation and constructs a Control Flow Graph (CFG). It then performs liveness analysis on the CFG. When the compiler encounters a synchronization barrier, it segments the analysis scope, evaluating variable liveness independently within each of the three phases. If variables used in the first phase are not accessed in the second or third phases, they can be marked as "dead" at the first __syncthreads() and their registers can be repurposed for new variables in the second phase. The same principle applies at the end of the second phase: if registers used between the two __syncthreads() are no longer needed in the third phase, they are released and made available to subsequent code.

This segmentation helps significantly shorten the live range of variables within each phase. For example, in a simple three-phase kernel, if registers r0, r1, and r2 are used only in the first phase and are not read afterward, they can be retired at the first synchronization point. When the second phase begins, the compiler can reuse r0–r2 or allocate other available registers. Furthermore, if registers such as r3 and r4 are only used during the second phase, they can be released at the second synchronization point and reused in the third phase—for instance, by registers like r5. Without synchronization points, the compiler may conservatively assume that registers allocated early on may be used at any later point, making it difficult to release them mid-way. However, with well-placed barriers, the compiler gains a clearer understanding of when values are used exclusively within specific phases, enabling significant compression of register usage per phase.

asalzburger:
That was a very interesting thread - looking forward to the meeting where this is discussed.

PR #1002, opened 2025-06-02
Decoupling Memory- and Compute-Bound Characteristics in Kalman Filter Prediction via Cooperative Kernel: 29.67% Throughput Gain

noyaboy:
Decoupling the prediction in Kalman Filter into two stages with grid-level barrier, allowing separated optimization to memory- and compute-bound characteristics. I use the following full reconstruction chain on the NVIDIA RTX 2080 Ti.

```
./build/bin/traccc_throughput_mt_cuda \
  --use-detray-detector=on \
  --use-acts-geom-source=on \
  --input-directory=odd/geant4_10muon_10GeV/ \
  --input-events=10 \
  --processed-events=1000 \
  --threads=1
```
Applying this procedure resulted in a throughput improvement, achieving 19.1168 ms/event. The baseline measurement, taken from traccc upstream (upstream/main, commit https://github.com/acts-project/traccc/commit/1febe2147a2af0dedd9e5660449c78af149d2d50), was 24.7872 ms/event (fix: Preserve hit-to-meas and hit-to-id mapping after measurement sorting (https://github.com/acts-project/traccc/pull/946)).


stephenswat:
I'll take a more detailed look at this later, but FYI it's a generally accepted good practice to keep commit and PR titles under 50 characters.

device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
```diff
@@ -126,3 +126,34 @@ TRACCC_HOST_DEVICE inline void propagate_to_next_surface(
...
+
+    finding_config cfg = cfg_in;
+    cfg.propagation.stepping.step_constraint *= 5.0f;  // 粗精度
+
+    propagate_to_next_surface(globalIndex, cfg, payload);
+}
+
+/// Stage-2 ── 以原本高精度設定進行 covariance / gain 更新
+template <typename propagator_t, typename bfield_t>
+TRACCC_HOST_DEVICE inline void propagate_stage2(
```
beomki-yeo, line 151: What is the difference between `propagate_stage1` and `propagate_stage2` ? They look the same except that `propagate_stage1` has a tighter constraint in step size? I wounder How this could be used for the performance optimization

noyaboy: Apologies — I'll review the code carefully and submit an updated PR shortly.

beomki-yeo:
If you haven't, I suggest running `./bin/traccc_test_cuda` to check if the physical results are OK.

device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
beomki-yeo: No need to apologize - I just need to get convinced

noyaboy:
@beomki-yeo I have 5 fail test: 
image

noyaboy:
You’re absolutely right. The performance improvement comes from disregarding the surface, not the grid-level barrier. In the future, I will personally perform the GPU test and rule out any external factors to enable a more accurate analysis.

PR #1003, opened 2025-06-03
Add Compute Capability benchmarks to extras

stephenswat:
This commit expands on the existing benchmarking code in the extras directory by adding new scripts for benchmarking and plotting the performance of traccc across different NVIDIA Compute Capabilities.

### DIFF
--- a/extras/benchmark/benchmark_cc.py
+++ b/extras/benchmark/benchmark_cc.py
@@ -17,6 +17,22 @@ import time
 from traccc_bench_tools import parse_profile, types, build, profile, benchmark
 
 
+ALL_COMPUTE_CAPABILITIES = [
+    "50",
+    "52",
+    "60",
+    "61",
+    "70",
+    "75",
+    "80",
+    "86",
+    "89",
+    "90",
+    "100",
+    "120",
+]
+
+
 log = logging.getLogger("traccc_benchmark")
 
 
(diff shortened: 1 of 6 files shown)


extras/benchmark/benchmark_cc.py
```diff
@@ -0,0 +1,254 @@
...
+        getattr(args, "num_threads_per_sm"),
+        getattr(args, "num_sm") * getattr(args, "num_threads_per_sm"),
+    )
+
+    repo = git.Repo(args.repo)
+
+    log.info("Using git repository at %s", repo.git_dir)
+
+    if repo.is_dirty():
+        log.fatal("Repository is dirty; please clean it before use!")
```
niermann999, line 133: Use error string `e` like below?

extras/benchmark/plot_cc.py
```diff
@@ -0,0 +1,102 @@
...
+from datetime import datetime
+
+log = logging.getLogger("traccc_plot")
+
+
+def main():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "input",
```
niermann999, line 19: Is this intended?
```suggestion
        "--input",
```

extras/benchmark/traccc_bench_tools/profile.py
```diff
@@ -0,0 +1,72 @@
+import subprocess
```
niermann999, line 1: copyright?

extras/benchmark/benchmark_cc.py
stephenswat: Well spotted!

PR #1007, opened 2025-06-06
Fix traccc::full_chain_algorithm Copying, main branch (2025.06.06.)

krasznaa:
While we sort out some deeper questions in #979, here I'd like to fix the actual issue breaking `traccc_throughput_mt` since a while. With the minimum change necessary.

It is not to say though that memory management in all the algorithms should not be cleaned up more thoroughly. But for a change I'd like to make a small PR...

### DIFF
--- a/examples/run/cpu/full_chain_algorithm.hpp
+++ b/examples/run/cpu/full_chain_algorithm.hpp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2022-2024 CERN for the benefit of the ACTS project
+ * (c) 2022-2025 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -29,6 +29,7 @@
 
 // System include(s).
 #include <functional>
+#include <memory>
 
 namespace traccc {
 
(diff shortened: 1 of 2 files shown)


examples/run/cpu/full_chain_algorithm.hpp
```diff
@@ -91,7 +92,7 @@ class full_chain_algorithm : public algorithm<track_state_container_types::host(
 
     private:
     /// Vecmem copy object
-    vecmem::copy m_copy;
+    std::shared_ptr<vecmem::copy> m_copy;
```
beomki-yeo, line 95: Should this be `shared_ptr`?

krasznaa: Using `std::shared_ptr` requires the least number of lines of change. 🤔

If I make it into an `std::unique_ptr` then I also need to provide a hand-written copy constructor and copy assignment operator. But with `std::shared_ptr` the auto-generated ones will do the correct thing.

Note that `vecmem::copy` has no internal state. So all its functions are by definition thread safe. Some of the device specific copy types are not thread safe, but the base / host class is.

With all this being said... Just relying on move construction/assignment may be enough... 🤔 I'll give that a try...

beomki-yeo (approved):

PR #1011, opened 2025-06-10
Remove dependency on dfelibs

paulgessinger:
The ACTS project is deprecating its fork of dfelibs, which we use only for CSV reading. Thus, this commit removes our dependency on dfelibs and simply places the relevant code into the traccc repository.


stephenswat:
Closing in favour of #893.

PR #1012, opened 2025-06-10
Fix the soa comparator typo in seq_example_cuda

beomki-yeo:
Fix the soa comparator

### DIFF
--- a/examples/run/cuda/seq_example_cuda.cpp
+++ b/examples/run/cuda/seq_example_cuda.cpp
@@ -493,7 +493,7 @@ int seq_run(const traccc::opts::detector& detector_opts,
                             const_proxy_type>{
                         vecmem::get_data(measurements_per_event),
                         vecmem::get_data(measurements_per_event_cuda)}};
-            compare_track_candidates(
+            compare_resolved_track_candidates(
                 vecmem::get_data(res_track_candidates),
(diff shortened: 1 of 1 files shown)


examples/run/cuda/seq_example_cuda.cpp
```diff
@@ -493,7 +493,7 @@ int seq_run(const traccc::opts::detector& detector_opts,
                             const_proxy_type>{
                         vecmem::get_data(measurements_per_event),
                         vecmem::get_data(measurements_per_event_cuda)}};
-            compare_track_candidates(
+            compare_resolved_track_candidates(
```
krasznaa, line 496: Hmm... The only difference between `compare_track_candidates` and `compare_resolved_track_candidates` is the string printout that they make.

For clear readability this update should be done indeed. But I don't see how this change should affect the calculated efficiencies. 😕 (For context, that's the discussion that we started from with Beomki.) Since both comparator functors operate with the same measurements. 🤔

krasznaa (approved):

examples/run/cuda/seq_example_cuda.cpp
beomki-yeo: You need to check the codes in my DM again and understand why it did not work for me

PR #1014, opened 2025-06-12
Synchronize threads to update shared measurements correctly

beomki-yeo:
Currently `n_shared` array in `remove_tracks.cu` is not synchronized correctly, resulting into a wrong (and non-deterministic) calculation of relative shared measurements 

Before

```
Number of resolved track candidates: 5860 (host), 5858 (device)
  Matching rate(s):
    - 99.8805% at 0.01% uncertainty
    - 99.8805% at 0.1% uncertainty
    - 99.8805% at 1% uncertainty
    - 99.8805% at 5% uncertainty
```

After

```
Number of resolved track candidates: 5860 (host), 5860 (device)
  Matching rate(s):
    - 100% at 0.01% uncertainty
    - 100% at 0.1% uncertainty
    - 100% at 1% uncertainty
    - 100% at 5% uncertainty
```

Also added tests which can catch the current bug


asalzburger (approved):
Looks good, makes also sense to to move the synchronization point.

PR #1019, opened 2025-06-17
Make B-field converter delimiter configurable

stephenswat:
The magnetic field converter is currently hardcoded to use a space as the delimiter between fields and values, but some field files use commas. This commit adds a new CLI argument that allows users to configure their own delimiters.

### DIFF
--- a/examples/tools/convert_csv_bfield.cpp
+++ b/examples/tools/convert_csv_bfield.cpp
@@ -41,7 +41,10 @@ void parse_opts(int argc, char* argv[],
         "input,i", boost::program_options::value<std::string>()->required(),
         "input magnetic field to read")(
         "output,o", boost::program_options::value<std::string>()->required(),
-        "output magnetic field to write");
+        "output magnetic field to write")(
+        "delimiter,d",
+        boost::program_options::value<char>()->default_value(' ')->required(),
+        "delimiter character");
 
     boost::program_options::parsed_options parsed =
         boost::program_options::command_line_parser(argc, argv)
@@ -63,7 +66,7 @@ void parse_opts(int argc, char* argv[],
     }
 }
 
-field_t read_bfield(const std::string& fn) {
+field_t read_bfield(const std::string& fn, char delimiter) {
     std::ifstream f;
 
     float minx = std::numeric_limits<float>::max();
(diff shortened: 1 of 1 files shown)


examples/tools/convert_csv_bfield.cpp
```diff
@@ -41,7 +41,10 @@ void parse_opts(int argc, char* argv[],
         "input,i", boost::program_options::value<std::string>()->required(),
         "input magnetic field to read")(
         "output,o", boost::program_options::value<std::string>()->required(),
-        "output magnetic field to write");
+        "output magnetic field to write")(
+        "delimiter,d",
+        boost::program_options::value<char>()->default_value(' ')->required(),
```
krasznaa, line 46: Could it not be an optional argument? You already provide a default value. (I believe this was just a typo probably.)

krasznaa (approved):
Looks simple enough. Though the CLI should probably be tweaked.

examples/tools/convert_csv_bfield.cpp
stephenswat: No, this is intentional. Because we access this variable unconditionally, we should mark it as required.

krasznaa: But then what's the purpose of the default value?

stephenswat: I think you're misunderstanding what `required` means in Boost. The key idea here is that `required` simply requires a value to be set for a specific option, not that the user must specifically supply it. Thus, a default value will also satisfy the requirement.

krasznaa: How sure are you about this? :thinking:

Take the following example:

```c++
#include <boost/program_options.hpp>
#include <iostream>

namespace po = boost::program_options;

   po::options_description desc("Optional test");
   desc.add_options()
      ("delimiter,d",
       boost::program_options::value<char>()->default_value(' '));

   po::variables_map vm;
   po::store(po::parse_command_line(argc, argv, desc), vm);
   po::notify(vm);

   std::cout << "Delimeter: \"" << vm["delimiter"].as<char>() << "\"" << std::endl;
}
```

It happily works for me without specifying anything from the command line. But can also receive an argument.

```
[bash][Legolas]:Development > ./build/options 
Delimeter: " "
[bash][Legolas]:Development > ./build/options -d f
Delimeter: "f"
[bash][Legolas]:Development >
```

I strongly believe that `required()` doesn't make a difference here. Since there's already a default value, the requirement is always satisfied.

But as such, it also doesn't hurt. I was afraid that it would require users to always provide something, but I was wrong on that. The default value seemingly already satisfies the requirement. :thinking:

PR #1016, opened 2025-06-13
Fix oneDPL Usage, main branch (2025.06.13.)

krasznaa:
It turns out / I was reminded once again, that we can't use `SYSTEM` with the oneDPL external. :cry: It's because the oneAPI compiler makes use of its own oneDPL headers when we provide our own oneDPL headers using `-isystem`.

I noticed this through the following errors when trying to build the project using the latest, 2025.1.3 version of oneAPI.

```
[build] In file included from /home/krasznaa/ATLAS/projects/traccc/traccc/device/sycl/src/clusterization/measurement_sorting_algorithm.sycl:21:
[build] In file included from /home/krasznaa/software/intel/oneapi-2025.1.3/dpl/2022.8/include/oneapi/dpl/execution:34:
[build] In file included from /home/krasznaa/software/intel/oneapi-2025.1.3/dpl/2022.8/include/oneapi/dpl/pstl/glue_algorithm_impl.h:27:
[build] In file included from /home/krasznaa/software/intel/oneapi-2025.1.3/dpl/2022.8/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h:22:
[build] In file included from /home/krasznaa/software/intel/oneapi-2025.1.3/dpl/2022.8/include/oneapi/dpl/pstl/hetero/../parallel_backend.h:32:
[build] In file included from /home/krasznaa/software/intel/oneapi-2025.1.3/dpl/2022.8/include/oneapi/dpl/pstl/hetero/../hetero/dpcpp/parallel_backend_sycl.h:41:
[build] /home/krasznaa/software/intel/oneapi-2025.1.3/dpl/2022.8/include/oneapi/dpl/pstl/hetero/../hetero/dpcpp/parallel_backend_sycl_merge.h:182:99: error: '&&' within '||' [-Werror,-Wlogical-op-parentheses]
[build]   182 |             __rng3[__rng3_idx] = (!__rng1_idx_less_n1 || __rng1_idx_less_n1 && __rng2_idx_less_n2 &&
[build]       |                                                       ~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
[build]   183 |                                                              __comp(__rng2[__rng2_idx], __rng1[__rng1_idx]))
[build]       |                                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #1021, opened 2025-06-17
Diagnostic Pragma Cleanup, main branch (2025.06.17.)

krasznaa:
I was not all too happy about how I worked around oneDPL warnings in #1016. Here I try to simplify things a little.

For oneDPL I introduced a helper header, which I then mark as a "system header" for Clang with:

```c++
#pragma clang system_header
```

Instead of separately listing all the warnings that should be ignored.

While at it, I also removed the warning suppression pragmas from `throughput_mt.ipp`, as these should not be necessary since #792. :thinking:

Note that I left the pragmas as they were in: https://github.com/acts-project/traccc/blob/main/tests/sycl/test_dpl.sycl In this test I didn't want to complicate things further by making it pick up a private header from `traccc::sycl`.


stephenswat (approved):
Seems harmless enough.

PR #1023, opened 2025-06-17
Device Cluster Reconstruction, main branch (2025.06.17.)

krasznaa:
It's time to re-open some old wounds...

This is an attempt at bringing #930 over the finish line, as we really-really need this code for the EFTracking report. :thinking:

The main things:
  - I left the `std::optional` in `traccc::device::aggregate_cluster`. As API-wise probably this is the most descriptive in the end for what argument the function should be receiving.
  - I switched all new arguments to vecmem views/types. So that their sizes would be less implicit. Making the code a bit easier to debug later on hopefully.

Based on ideas I gathered from the EDM updates, I now have some slightly different ideas for what the final algorithm API should be, but didn't want to slow things down with that in this PR. I'll leave that to a follow-up PR, in which I'll teach the Alpaka and SYCL algorithms as well how to output cluster collections.

### DIFF
--- a/device/common/include/traccc/clusterization/device/aggregate_cluster.hpp
+++ b/device/common/include/traccc/clusterization/device/aggregate_cluster.hpp
@@ -30,6 +30,12 @@ namespace traccc::device {
 /// @param[in] end       partition end point this cell belongs to
 /// @param[in] cid       current cell id
 /// @param[out] out      cluster to fill
+/// @param[out] disjoint_set_ptr Pointer to an array of unsigned integers of
+///                      length $|cells|$ to which an integer is written
+///                      identifying the measurement index to which each cell
+///                      belongs.
+/// @param[out] cluster_size Optional integer which is filled with the size of
+///                      the measurement that is created.
 ///
 TRACCC_HOST_DEVICE
 inline void aggregate_cluster(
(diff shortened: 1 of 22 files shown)


device/common/include/traccc/clusterization/device/aggregate_cluster.hpp
```diff
@@ -38,7 +44,9 @@ inline void aggregate_cluster(
     const silicon_detector_description::const_device& det_descr,
     const vecmem::device_vector<details::index_t>& f, unsigned int start,
     unsigned int end, unsigned short cid, measurement& out,
-    vecmem::data::vector_view<unsigned int> cell_links, unsigned int link);
+    vecmem::data::vector_view<unsigned int> cell_links, unsigned int link,
+    vecmem::device_vector<unsigned int> disjoint_set,
```
krasznaa, line 48: Crud... I just now see that it should rather be

```suggestion
    const vecmem::device_vector<unsigned int>& disjoint_set,
```

:frowning:

device/common/include/traccc/clusterization/device/impl/ccl_kernel.ipp
```diff
@@ -145,7 +145,8 @@ TRACCC_DEVICE inline void ccl_core(
     const edm::silicon_cell_collection::const_device& cells_device,
     const silicon_detector_description::const_device& det_descr,
     measurement_collection_types::device measurements_device,
-    const barrier_t& barrier) {
+    const barrier_t& barrier, vecmem::device_vector<unsigned int> disjoint_set,
+    vecmem::device_vector<unsigned int> cluster_size) {
```
krasznaa, line 149: These as well...

stephenswat (approved):
So, as I had told you back in March, shoehorning vecmem into this PR has made it less performant, more verbose, and harder to read. :wink:

stephenswat:
Closes #930.

stephenswat:
> I'm very confused. Why are there all these changes in the performance plot? 😕
> 
> I didn't expect much of a change for the `ccl_kernel`, and indeed there doesn't seem to be any real difference. But what are all the other kernels doing?

You're seeing the effects of the merge commit.

krasznaa:
@Debottam, this is the update that will help G-100 and G-200. Though I'll also want to add some more changes before we start updating the Athena code...

PR #1025, opened 2025-06-18
Update detray to version 0.99.0

stephenswat:
This commit updates detray to version 0.99.0, and makes the appropriate changes as provided by Joana.

### DIFF
--- a/extern/detray/CMakeLists.txt
+++ b/extern/detray/CMakeLists.txt
@@ -13,7 +13,7 @@ message( STATUS "Building Detray as part of the TRACCC project" )
 
 # Declare where to get Detray from.
 set( TRACCC_DETRAY_SOURCE
-   "URL;https://github.com/acts-project/detray/archive/refs/tags/v0.98.0.tar.gz;URL_MD5;0de44d53f8d5fb31a2d74071f4333abc"
+   "URL;https://github.com/acts-project/detray/archive/refs/tags/v0.99.0.tar.gz;URL_MD5;ee3bf6a61e923d364ee5111f237f1fe2"
    CACHE STRING "Source for Detray, when built as part of this project" )
 mark_as_advanced( TRACCC_DETRAY_SOURCE )
 FetchContent_Declare( Detray SYSTEM ${TRACCC_DETRAY_SOURCE} )
--- a/simulation/include/traccc/simulation/smearing_writer.hpp
+++ b/simulation/include/traccc/simulation/smearing_writer.hpp
@@ -19,6 +19,7 @@
 
 // Detray core include(s).
 #include <detray/propagator/base_actor.hpp>
+#include <detray/utils/concepts.hpp>
 
 // DFE include(s).
 #include <dfe/dfe_io_dsv.hpp>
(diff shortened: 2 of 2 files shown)


stephenswat:
# Performance summary
## Graphical

## Tabular
| Kernel | 779853a4 | e258b39b | Delta |
| --- | --: | --: | --: |
| `fit_forward` | 107.08 ms | 109.46 ms | 2.2% |
| `propagate_to_next_surface` | 72.00 ms | 69.87 ms | -3.0% |
| `fit_backward` | 53.44 ms | 53.81 ms | 0.7% |
| `count_triplets` | 14.23 ms | 14.25 ms | 0.1% |
| `find_triplets` | 6.00 ms | 6.00 ms | 0.0% |
| `find_tracks` | 4.29 ms | 4.29 ms | -0.1% |
| `fit_prelude` | 1.12 ms | 1.13 ms | 0.5% |
| `find_doublets` | 907.86 μs | 893.28 μs | -1.6% |
| `ccl_kernel` | 860.32 μs | 862.00 μs | 0.2% |
| `count_doublets` | 618.40 μs | 617.80 μs | -0.1% |
| `select_seeds` | 363.35 μs | 361.30 μs | -0.6% |
| `Thrust::sort` | 272.37 μs | 272.61 μs | 0.1% |
| `build_tracks` | 192.62 μs | 193.65 μs | 0.5% |
| `update_triplet_weights` | 99.93 μs | 100.35 μs | 0.4% |
| `DeviceRadixSortHistogramKernel` | 97.98 μs | 98.04 μs | 0.1% |
| `apply_interaction` | 63.72 μs | 63.84 μs | 0.2% |
| `DeviceRadixSortOnesweepKernel` | 54.74 μs | 54.76 μs | 0.0% |
| `fill_sort_keys` | 44.68 μs | 44.63 μs | -0.1% |
| `estimate_track_params` | 34.39 μs | 34.40 μs | 0.0% |
| `populate_grid` | 30.45 μs | 30.48 μs | 0.1% |
| `count_grid_capacities` | 29.12 μs | 29.15 μs | 0.1% |
| `unknown` | 20.45 μs | 20.64 μs | 0.9% |
| `form_spacepoints` | 12.39 μs | 12.50 μs | 0.9% |
| `reduce_triplet_counts` | 6.71 μs | 6.69 μs | -0.4% |
| `make_barcode_sequence` | 1.02 μs | 1.02 μs | 0.5% |
| `DeviceRadixSortExclusiveSumKernel` | 513.07 ns | 513.35 ns | 0.1% |
| `fill_prefix_sum` | 165.42 ns | 165.44 ns | 0.0% |
| **Total** | **261.88 ms** | **262.51 ms** | **0.2%** |
> [!IMPORTANT]

> [!NOTE]

krasznaa:
Did we expect / do we understand these performance differences? It seems to not just be statistical. Though it's hard to tell...

niermann999 (approved):

krasznaa:
Mmmmkay... :thinking: Then it's pretty interesting that in total we pretty much come to a net zero effect. Was the expectation not that this would speed things up a bit?

I guess the question is why fitting seems to slow down a bit. Since I'm happy to take the small apparent speedup in the propagation.

niermann999:
> Mmmmkay... 🤔 Then it's pretty interesting that in total we pretty much come to a net zero effect. Was the expectation not that this would speed things up a bit?
> 
> I guess the question is why fitting seems to slow down a bit. Since I'm happy to take the small apparent speedup in the propagation.

I am glad that it did not get slower actually, because the mask iteration could induce thread divergence (if two portals have a different number of masks). It could give more of a speedup when I get around to intersect the cylinder only once and then check the masks, instead of re-intersecting the same cylinder for every mask.

PR #1028, opened 2025-06-19
Device Cluster Reconstruction, main branch (2025.06.19.)

krasznaa:
Following up from #1023, this is how I believe the clusterization algorithm should provide its results. And I'm very open to a discussion on this.

While I generally liked the idea of having different functions for reconstructing just the measurements or the measurements plus the clusters, I think in the end that just makes the code more complicated than it needs to be. I propose that we just have a single interface for the device clusterization algorithms. One which returns an optional cluster collection buffer. Behind the scenes this was the implementation in `cuda::clusterization_algorithm` already. And I think this is a functional interface towards the users as well. :thinking:

After updating the CUDA algorithm, I added the functionality of reconstructing cell clusters to the Alpaka and SYCL algorithms as well.

Unfortunately however the Alpaka algorithm, on an NVIDIA backend, is not behaving well at the moment. :sob: The unit test reports that a random number of the tests fail to return the correct (number of) clusters from the algorithm. And so far I didn't manage to find yet what is causing this issue. :frowning:

Still, even with that open issue, I wanted to let people see my proposal already. As there may be a discussion about the user API...

As before, pinging @CrossR and @StewMH for their opinions.

P.S. Almost forgot... I also modified `traccc::algorithm` to allow fundamental arguments to be passed by value to an algorithm. As I **really** wanted to be able to pass a `bool` to the clusterization algorithms...


device/alpaka/include/traccc/alpaka/clusterization/clusterization_algorithm.hpp
```diff
@@ -73,13 +82,16 @@ class clusterization_algorithm
...
+    /// The Alpaka queue to use
+    std::reference_wrapper<queue> m_queue;
     /// Memory reserved for edge cases
+    /// @{
     vecmem::data::vector_buffer<device::details::index_t> m_f_backup,
         m_gf_backup;
     vecmem::data::vector_buffer<unsigned char> m_adjc_backup;
     vecmem::data::vector_buffer<device::details::index_t> m_adjv_backup;
     vecmem::unique_alloc_ptr<unsigned int> m_backup_mutex;
-    mutable std::once_flag m_setup_once;
```
stephenswat, line 82: Unrelated change?

device/alpaka/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -91,45 +103,63 @@ clusterization_algorithm::clusterization_algorithm(
...
+        return {
+            .measurements = {},
+            .clusters =
+                (reconstruct_clusters
+                     ? std::optional<
+                           edm::silicon_cluster_collection::
+                               buffer>{edm::silicon_cluster_collection::
+                                           buffer{}}
+                     : std::optional<edm::silicon_cluster_collection::buffer>{
+                           std::nullopt})};
```
stephenswat, line 137: There's got to be a way to make this more readable. :worried:

stephenswat (changes requested):
I didn't really bother reading past the Alpaka code, but could you elaborate why exactly you think this makes the code "simpler"? The net PR delta is almost +100 lines and this change makes the API for users more obtuse to use, who in the existing implementation know exactly that if they pass the right tag, they get the desired result. Returning an optional value leaves open the semantic to return an non-extant value even if the user requests the cluster data to be delivered.

In my eyes, this just makes the code less understandable for no good reason, so I'd like to understand what the benefit of this approach is.

stephenswat:
> These changes are needed relatively urgently.

Why? The API currently in the repository is perfectly functional already... I don't see the urgency in changing it.

krasznaa:
There is urgency in providing this for Alpaka and SYCL as well. As those will be needed for our tests.

I'll walk back on the interface change a little then. As implementing the feature in all backends is fairly urgent.

PR #1017, opened 2025-06-14
Add cuda inhomogeneous bfield type

beomki-yeo:
This PR currently adds instantiations for (1) Host KF with inhomogeneous field (2) CUDA CKF with inhomogeneous field

To-do list
- [ ] **Add CUDA KF with inhomogeneous field** - Any help?!
- [ ] Add examples with inhomogeneous field
- [ ] Remove old algorithm classes (This can be done later...)
- [ ] Add tests with inhomogeneous fields (This can be done later....)

PR #1029, opened 2025-06-20
CUDA Templating Rewrite, main branch (2025.06.20.)

krasznaa:
Similar to #1026, in preparation for introducing inhomogeneous magnetic fields as discussed in #1017, this PR attempts to harmonize the setup of the CUDA CFK and KF algorithms with how all the other algorithms are implemented by now.

This is of course a big change, but we do need it relatively urgently.

A couple of things to point out:
  - The template specializations for the KF were in a relatively easily extensible state already, there I just needed to separate out the building of the `fill_sort_keys` kernel. This was not an issue previously, where only a single template specialization existed for the KF algorithm.
    * Note though that the magnetic field type is a bit more deeply woven into the helper typedefs in `traccc::core`, but I decided to leave that alone in this PR.
  - The template specializations for the CKF were however pretty rigidly written. Not allowing for the introduction of the telescope geometry in an easy way.
    * It was the specializations for `propagate_to_next_surface` that I modified a bit more deeply. In such a way that (hopefully) introducing non-constant magnetic fields should now be a bit easier in a follow-up PR.

The clients (in this repository) were all easy to update, but it's worth pointing out to people like @paradajzblond (I couldn't easily find Fabrice's GitHub handle...) that client code will need some changes after this PR.


stephenswat (changes requested):
Looks mostly harmless assuming you didn't change any of the actual algorithm code, but let's get all the B-field specialisation done in one go. Plus could do with some refactoring to reduce code volume and improve maintainability.

krasznaa (commented):
I'd prefer to attack `traccc/utils/detector_type_utils.hpp` in a separate PR. One that mainly deals with the magnetic field. I was on purpose trying to limit the number of changes in this PR. :thinking:

krasznaa:
I think I managed to find a way to make the templating int he CKF functions a bit more user friendly. Making the "main functions" specifically templated on the detector and magnetic field types. And then using some newly introduced typedefs to figure out the correct Detray types based on just the detector and magnetic field types.

Now on to do the same for the KF algorithms...

krasznaa:
This PR ballooned into a much bigger thing by now than I originally intended. But the good news is that after this, adding non-const magnetic field versions of the finding and fitting functions will take a very finite amount of new lines of code. :thinking:

stephenswat:
FYI traccc no longer compiles for me with this change. :upside_down_face:

PR #1030, opened 2025-06-21
New example for demonstrating how to run reconstruction on GPU with time-varying misalignments

tsulaiav:
For the time being, this example only demonstrates how to use multiple detray detector views from the traccc code. Several other improvements should be made to the code. For instance, multiple contexts are ignored in the CPU-based fitting.

That's why the PR is marked as draft for the time being.

### DIFF
--- a/examples/run/cuda/CMakeLists.txt
+++ b/examples/run/cuda/CMakeLists.txt
@@ -27,6 +27,10 @@ traccc_add_executable( truth_fitting_example_cuda "truth_fitting_example_cuda.cp
    LINK_LIBRARIES vecmem::core vecmem::cuda traccc::io traccc::performance
                   traccc::core traccc::device_common traccc::cuda
                   traccc::options traccc_examples_common )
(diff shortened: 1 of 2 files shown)


krasznaa (commented):
Would it make sense to rather than introducing a new executable, add misalignment as a new option to the existing `traccc_truth_fitting_esample_cuda` executable? :thinking: 

I started playing a little while ago with merging the example applications of all the languages into a single codebase, but did not find a good design for that yet. :frowning: Still, cutting down a bit on the number of executables could be helpful for that effort as well.

tsulaiav:
> Would it make sense to rather than introducing a new executable, add misalignment as a new option to the existing `traccc_truth_fitting_esample_cuda` executable? 🤔
> 
> I started playing a little while ago with merging the example applications of all the languages into a single codebase, but did not find a good design for that yet. 😦 Still, cutting down a bit on the number of executables could be helpful for that effort as well.

I'm with you on this. I will see over the next days if/how this can be merged with `truth_fitting_example_cuda.cpp`. By the way, the same should be done for the similar CPU-based example.

PR #1031, opened 2025-06-21
Limit iterations and improve state tracking

noyaboy:
1. Use is_updated_buffer to correctly track per-iteration track ID updates
2. Introduce scanned_block_offsets_buffer to avoid reusing original block offsets
3. Add iter < m_config.max_iterations to bound the loop and prevent infinite runs
4. Launch CUDA graph once per iteration, then stream.synchronize() for proper data sync

**Example**
```
./traccc_throughput_mt_cuda
              --use-detray-detector=on \
              --use-acts-geom-source=on \
              --input-directory=odd/geant4_10muon_10GeV \
              --input-events=10 \
              --processed-events=1000 \
              --threads=1
```
**Result with relative improvement 0.4553%±0.1716%**
image

**Test**
image

### DIFF
--- a/device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu
+++ b/device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu
@@ -295,7 +295,7 @@ greedy_ambiguity_resolution_algorithm::operator()(
 
     // Whether track id is updated after an iteration
     vecmem::data::vector_buffer<int> is_updated_buffer{n_tracks, m_mr.main};
-    m_copy.get().setup(inverted_ids_buffer)->ignore();
+    m_copy.get().setup(is_updated_buffer)->ignore();
 
     // Prefix sum buffer
     vecmem::data::vector_buffer<int> prefix_sums_buffer{n_tracks, m_mr.main};
@@ -364,9 +364,10 @@ greedy_ambiguity_resolution_algorithm::operator()(
     m_copy.get().setup(block_offsets_buffer)->ignore();
     vecmem::data::vector_buffer<int> scanned_block_offsets_buffer{nBlocks_scan,
                                                                   m_mr.main};
-    m_copy.get().setup(block_offsets_buffer)->ignore();
+    m_copy.get().setup(scanned_block_offsets_buffer)->ignore();
 
-    while (!terminate && n_accepted > 0) {
+    unsigned int iter = 0;
+    while (!terminate && n_accepted > 0 && iter < m_config.max_iterations) {
         nBlocks_adaptive = (n_accepted + 1023) / 1024;
         nBlocks_warp = (n_accepted + nThreads_warp - 1) / nThreads_warp;
         nBlocks_scan = (n_accepted + 1023) / 1024;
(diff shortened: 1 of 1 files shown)


device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu
```diff
@@ -364,9 +364,10 @@ greedy_ambiguity_resolution_algorithm::operator()(
     m_copy.get().setup(block_offsets_buffer)->ignore();
     vecmem::data::vector_buffer<int> scanned_block_offsets_buffer{nBlocks_scan,
                                                                   m_mr.main};
-    m_copy.get().setup(block_offsets_buffer)->ignore();
+    m_copy.get().setup(scanned_block_offsets_buffer)->ignore();
 
-    while (!terminate && n_accepted > 0) {
+    unsigned int iter = 0;
+    while (!terminate && n_accepted > 0 && iter < m_config.max_iterations) {
```
krasznaa, line 370: Hmm... If we want to make a big loop "out here", then the setup of the CUDA graph needs to be taken outside of this loop. Since as long as we keep re-making the graph on each iteration, the graph code will just slow us down.

So I'd re-think this part of the code even more fundamentally. 🤔

device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu
```diff
@@ -295,7 +295,7 @@ greedy_ambiguity_resolution_algorithm::operator()(
 
     // Whether track id is updated after an iteration
     vecmem::data::vector_buffer<int> is_updated_buffer{n_tracks, m_mr.main};
-    m_copy.get().setup(inverted_ids_buffer)->ignore();
+    m_copy.get().setup(is_updated_buffer)->ignore();
```
krasznaa, line 298: Good catch with these typos! 😄

device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu
beomki-yeo: @krasznaa is correct. The graph is not supposed to be created every iteration. Probably you are not seeing the impact due to small size of data.

Even though it is ideal to never re-created the cudaGraph, currently the cudaGraph is recreated every 100 iterations to update the block and grid size based on the number of remaining tracks.

device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu
```diff
@@ -542,19 +543,19 @@ greedy_ambiguity_resolution_algorithm::operator()(
...
-        for (unsigned int iter = 0; iter < n_it; iter++) {
-            cudaGraphLaunch(graphExec, stream);
-        }
+        cudaGraphLaunch(graphExec, stream);
 
         cudaMemcpyAsync(&terminate, terminate_device.get(), sizeof(int),
                         cudaMemcpyDeviceToHost, stream);
         cudaMemcpyAsync(&n_accepted, n_accepted_device.get(),
                         sizeof(unsigned int), cudaMemcpyDeviceToHost, stream);
+        m_stream.get().synchronize();
```
beomki-yeo, line 552: I also don't use a synchronize function as its overhead was pretty big. Have you encountered any issue regarding it?

krasznaa: It's good that you pointed this out Beomki.

The deeper issue here is that using `cudaMemcpyAsync` suggests that something would be happening asynchronously here. But since you're copying into variables that are on the stack, these copies will very much happen synchronously. (It's a long story with how SYCL will happily set up asynchronous copies into/from temporary memory areas, but CUDA is only willing to do so with "pinned" host memory. I believe CUDA is the one behaving in the expected way here.)

So indeed, the explicit synchronization should not be doing anything here. But at the same time, it would also probably be worth making the memory copies into synchronous ones as well, to avoid this type of confusion in the future.

stephenswat:
Hi @noyaboy, could you let us know what you are working on these days? Thanks!

noyaboy:
Hi @stephenswat ! I’m currently occupied with a paper submission in my lab (July 07 - present). Before that, I had been working on the ML-based component of FastML (June 15 - July 01). I expect to finish that soon and then begin exploring improvements for traccc !

PR #1033, opened 2025-06-23
Inhomogeneous Magnetic Field, main branch (2025.06.23.)

krasznaa:
I did not realize up until now how much work we still have with this... 😦

As I discussed earlier today with @stephenswat, I had to realize that I don't want to use the CUDA specific magnetic field type that @beomki-yeo introduced in #1017, in the public interface of `traccc::cuda`. Since that would mean a public dependency on `CUDA::cudart` in that library. Which I'd really like to avoid. Not to mention that I really don't want to force users of `traccc::sycl` to use a SYCL compiler to be able to use tools/algorithms from that library.

After this discussion with @stephenswat, I came up with the `traccc::bfield` type introduced by this PR. 🤔 Which allows us to pass around specific Covfie field objects in a type-erased way, while still allowing for relatively safe runtime behaviour. (Using [std::any](https://en.cppreference.com/w/cpp/utility/any.html) for the heave lifting.)

But of course, since I'm now modifying existing interfaces, this PR ends up touching a ton of files. 😦 Though luckily in most cases the necessary changes are quite small.

I still need to update the SYCL and Alpaka algorithms, and I'll also want to make at least some of the example algorithms a little smarter. (So that they could load the ODD magnetic field.) But I wanted to expose everyone to this idea of `traccc::bfield` already. Since if this turns out to be a viable road for us, I may very well try to introduce a similar concept for the detector objects as well.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -12,6 +12,10 @@ stages:
 # Base job template.
 .base_template: &base_job
   before_script:
+    - apt install -y zstd
+    - curl --retry 5 --retry-delay 10 --output deps.tar.zst https://acts.web.cern.ch/ACTS/ci/ubuntu-24.04/deps.v6.tar.zst
+    - tar -xf deps.tar.zst -C /usr/local --strip-components=1
+    - rm deps.tar.zst
     - git clone $CLONE_URL src
     - git -C src checkout $HEAD_SHA
     - source ./src/.github/ci_setup.sh ${TRACCC_BUILD_TYPE}
@@ -54,7 +58,7 @@ stages:
 # CUDA build job.
 build:cuda:
   <<: *build_job
-  image: ghcr.io/acts-project/ubuntu2204_cuda:56
+  image: ghcr.io/acts-project/ubuntu2404_cuda:69
   variables:
     TRACCC_BUILD_TYPE: CUDA
     TRACCC_BUILD_PRESET: cuda-fp32
@@ -62,7 +66,7 @@ build:cuda:
 # CUDA test job.
 test:cuda:
   <<: *nvidia_test_job
-  image: ghcr.io/acts-project/ubuntu2204_cuda:56
+  image: ghcr.io/acts-project/ubuntu2404_cuda:69
   variables:
     TRACCC_BUILD_TYPE: CUDA
   dependencies:
@@ -71,7 +75,7 @@ test:cuda:
 # SYCL build job (with an Intel backend).
 build:sycl_intel:
   <<: *build_job
-  image: ghcr.io/acts-project/ubuntu2404_oneapi:56
+  image: ghcr.io/acts-project/ubuntu2404_oneapi:69
   variables:
     TRACCC_BUILD_TYPE: SYCL
     TRACCC_BUILD_PRESET: sycl-fp32
@@ -81,7 +85,7 @@ build:sycl_intel:
 # SYCL test job (with an Intel backend).
 test:sycl_intel:
   <<: *intel_test_job
-  image: ghcr.io/acts-project/ubuntu2404_oneapi:56
+  image: ghcr.io/acts-project/ubuntu2404_oneapi:69
   variables:
     TRACCC_BUILD_TYPE: SYCL
     ONEAPI_DEVICE_SELECTOR: opencl:*
@@ -91,7 +95,7 @@ test:sycl_intel:
 # SYCL build job (with an NVIDIA backend).
 build:sycl_nvidia:
   <<: *build_job
-  image: ghcr.io/acts-project/ubuntu2204_cuda_oneapi:56
+  image: ghcr.io/acts-project/ubuntu2404_cuda_oneapi:69
   variables:
     TRACCC_BUILD_TYPE: SYCL
     TRACCC_BUILD_PRESET: sycl-fp32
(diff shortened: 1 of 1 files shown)


krasznaa:
Multiple things came to light (at least to me...):
  - The GitLab jobs have been / are using Ubuntu 22.04 CUDA jobs still;
  - The reason that we haven't seen problems coming from the use of `std::format` in these jobs, is that GitLab CI jobs have not been building the "example code". And we've only been using `std::format` in the example code so far.

So I just switched the GitLab CI jobs to the same images that we use in GitHub CI. And now let's see why they are starting to fail... 😕

core/src/finding/combinatorial_kalman_filter_algorithm_default_detector.cpp
```diff
@@ -0,0 +1,39 @@
...
+            seeds, m_config, m_mr.get(), logger());
+    } else if (field.is<inhom_bfield_backend_t<scalar>>()) {
+        return details::combinatorial_kalman_filter(
+            det, field.as<inhom_bfield_backend_t<scalar>>(), measurements,
+            seeds, m_config, m_mr.get(), logger());
+    } else {
+        throw std::invalid_argument(
+            "Unsupported b-field type received in "
+            "traccc::host::combinatorial_kalman_filter_algorithm");
+    }
```
stephenswat, line 36: This pattern is fine for now, but it is already fairly repetitious and if we want to include detector type polymorphism in the future (which I think would be a good idea), we should automate this. I would propose the following sort of two-dimensional compile-time visitor pattern (implemented here using `std::any` but can of course be adapted to `traccc::bfield` and `traccc::detector` in the future):

```c++
#include <iostream>
#include <optional>
#include <any>

struct my_visitor {
    template<typename bfield, typename detector>
    int operator()(bfield b, detector d) const {
        if constexpr (std::is_same_v<bfield, float>) {
            std::cout << "float " << b;
        } else if constexpr (std::is_same_v<bfield, int>) {
            std::cout << "int " << b;
        } else if constexpr (std::is_same_v<bfield, long>) {
            std::cout << "long " << b;
        }

        std::cout << " ";

        if constexpr (std::is_same_v<detector, unsigned>) {
            std::cout << "unsigned " << d;
        } else if constexpr (std::is_same_v<detector, double>) {
            std::cout << "double " << d;
        }

        std::cout << std::endl;

        return 0;
    }
};

struct field_detector_visitor {
    using valid_bfields = std::tuple<float, int, long>;
    using valid_detectors = std::tuple<unsigned, double>;

    template<typename bfield_type, typename callable, typename detector_type, typename ... detector_types>
    auto visit_helper2(std::any bfield, std::any detector, callable && c, std::tuple<detector_type, detector_types...> &&) const {
        using rv = std::decay_t<decltype(c(std::any_cast<bfield_type>(bfield), std::any_cast<detector_type>(detector)))>;

        try {
            return std::optional<rv>(c(std::any_cast<bfield_type>(bfield), std::any_cast<detector_type>(detector)));
        } catch (std::bad_any_cast) {
            if constexpr (sizeof...(detector_types) == 0) {
                return std::optional<rv>{};
            } else {
                return visit_helper2<bfield_type>(bfield, detector, std::forward<callable>(c), std::tuple<detector_types...>{});
            }
        }
    }

    template<typename callable, typename bfield_type, typename ... bfield_types>
    auto visit_helper1(std::any bfield, std::any detector, callable && c, std::tuple<bfield_type, bfield_types...> &&) const {
        if (auto rv = visit_helper2<bfield_type>(bfield, detector, c, valid_detectors{}); rv) {
            return rv;
        } else {
            if constexpr (sizeof...(bfield_types) == 0) {
                return rv;
            } else {
                return visit_helper1(bfield, detector, std::forward<callable>(c), std::tuple<bfield_types...>{});
            }
        }
    }

    template<typename callable>
    auto visit(std::any bfield, std::any detector, callable && c) const {
        return visit_helper1(bfield, detector, std::forward<callable>(c), valid_bfields{});
    }
};

int main() {
    std::any bfield((long) 2.f);
    std::any detector((double) 5u);

    field_detector_visitor().visit(bfield, detector, my_visitor{});
}
```

Godbolt for reference: https://godbolt.org/z/h5sK1dbYd

This will allow us to make the code much cleaner and more maintainable, not to mention future-proof.

device/alpaka/src/finding/combinatorial_kalman_filter.hpp
```diff
@@ -553,7 +553,7 @@ struct BlockSharedMemDynSizeBytes<
         TVec const& blockThreadExtent, TVec const& /* threadElemExtent */,
         TArgs const&... /* args */
         ) -> std::size_t {
-        return static_cast<std::size_t>(blockThreadExtent.prod()) *
+        return 3 * static_cast<std::size_t>(blockThreadExtent.prod()) *
```
stephenswat, line 556: Shouldn't this be part of #1034?

io/src/read_bfield.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    // Check the magic header, to see that it's really a "covfie file".
+    if (auto hdr = covfie::utility::read_binary<std::uint32_t>(ifile);
+        hdr != covfie::utility::MAGIC_HEADER) {
+        throw std::invalid_argument(std::format(
+            "Input file ({}) is not a valid covfie file", filename));
+    }
+
+    // Reset the read position to the beginning of the file.
+    ifile.seekg(0);
```
stephenswat, line 44: This check is also done by the IO code in covfie, so it's somewhat redundant here.

stephenswat (commented):
This looks great as a first attempt, although I would make a forward-looking move and add a visitor pattern for B-field types, ideally a two-dimensional Cartesian visitor for B-field _and_ detector types.

PR #1034, opened 2025-06-24
Alpaka Fix, main branch (2025.06.24.)

krasznaa:
While working on #1033, I believe I managed to find the issue that I believe was hurting me in #1029 already. It turns out that the `find_tracks` kernel was not getting enough shared memory allocated. Though don't ask me why I couldn't diagnose this successfully earlier, and managed to do so now. 😕

~~It turns out that the CUDA code is allocating too much dynamic shared memory itself. (To be fixed in a separate PR.) But since Alpaka allocates **everything** as dynamic shared memory, it needs to account for the 3 `unsigned int` variables (https://github.com/acts-project/traccc/blob/main/device/alpaka/src/finding/combinatorial_kalman_filter.hpp#L83-L88) explicitly here.~~

It turns out that I was misunderstanding things even when writing this PR. 😦 In the end it turned out to be a case of incorrect types being used in calculating how much dynamic shared memory to allocate for Alpaka (and CUDA). Not a failure in adding a correct multiplier...

### DIFF
--- a/device/cuda/src/finding/combinatorial_kalman_filter.cuh
+++ b/device/cuda/src/finding/combinatorial_kalman_filter.cuh
@@ -289,7 +289,7 @@ combinatorial_kalman_filter(
             const unsigned int nBlocks =
                 (n_in_params + nThreads - 1) / nThreads;
             const std::size_t shared_size =
-                nThreads * sizeof(unsigned int) +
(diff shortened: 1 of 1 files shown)


krasznaa:
Pinging @CrossR as well.

stephenswat (approved):

PR #1037, opened 2025-06-25
Update CUDA propagation and fit kernel launch

stephenswat:
This commit updates the CUDA propagation and fitting kernel launch parameters in two ways. Firstly it increases the block size from 64 threads to 128 threads. The reason for this is that some Compute Capabilities (specificially 8.6 and 8.7) cannot achieve optimal occupancy with 64 threads, as the resident block limit per SM is too small. Secondly, this commit adds launch bounds to the kernels, requesting a minimum of 8 blocks of 128 threads per SM. This will increase the theoretical occupancy to 66% on CC 8.6 and to at least 50% on all other Compute Capabilities.


krasznaa (approved):
Pretty un-controversial, sure! 👍

PR #1038, opened 2025-06-25
Switch from synchronized propagator to default

stephenswat:
This commit switches our usage of the detray propagator from the synchronized version to the default version, which has some upcoming optimizations in detray and will yield higher propagation performance.


stephenswat:
Note: the relevant PR is https://github.com/acts-project/detray/pull/1006.

beomki-yeo:
Makes sense as the performance gets better

PR #1039, opened 2025-06-25
Update algebra plugins to v0.28.0

stephenswat:
This commit updates the algebra plugins external to v0.28.0 and uses its new transposed product function in the Kálmán filter code.


krasznaa (approved):
I was at first waiting for some more comments. But generally I don't see any problem with including this. Even if it's not as helpful as we hoped.

PR #1044, opened 2025-06-26
Use inhomogeneous magnetic field in benchmark

stephenswat:
This commit updates the commit benchmarking script to automatically use the inhomogeneous magnetic field whenever it considers commits that have 673d408 as a parent. Also expands the list of kernel names that are considered part of `Thrust::sort`.


stephenswat:
This is now ready.

PR #1046, opened 2025-06-27
Deduplicate sorting key filling CUDA kernels

stephenswat:
Pull request #1029 inadvertently breaks the debug build by introducing a new CUDA file with the same base name as another, previously existing, file. This causes a race condition during compilation. This commit renames the relevant files to avoid the naming collision and thereby fixes the build bug.


krasznaa (approved):
👍

PR #1052, opened 2025-07-02
Skip forward fitting pass by keeping CKF states

stephenswat:
This (extremely experimental) PR allows us to skip the first forward fitting pass by keeping the track states from the CKF.


stephenswat:
Closing in favour of https://github.com/acts-project/traccc/pull/1149.

PR #1057, opened 2025-07-03
Update covfie external hash

beomki-yeo:
It seems covfie hash has not been updated?

### DIFF
--- a/extern/covfie/CMakeLists.txt
+++ b/extern/covfie/CMakeLists.txt
@@ -13,7 +13,7 @@ message( STATUS "Fetching covfie as part of the traccc project" )
 
 # Declare where to get covfie from.
 set( TRACCC_COVFIE_SOURCE
-   "URL;https://github.com/acts-project/covfie/archive/refs/tags/v0.15.1.tar.gz;URL_MD5;aa54f7e7e1a83edd02660e2778d29109"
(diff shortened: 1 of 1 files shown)


stephenswat:
@sethrj Could this be a case of the GitHub tarball instability you mentioned? The MD5 sum of this release seems to have changed well within the one year stability window defined by GitHub.

sethrj:
@stephenswat I have a repository of scripts for doing github releaes, including downloading and re-uploading artifacts: https://github.com/celeritas-project/release-scripts . The line to do this is:
```
artifact_url, artifact_tgz = release_notes.get_or_upload_tarball(cached, gh_release)
```

@beomki-yeo I doin't know about ROOT, but for tools like [versioneer](https://github.com/python-versioneer) and [CGV](https://github.com/sethrj/cmake-git-version) that have `$Format:%(describe:tags)$` or `$Format:%d$`, the underlying issue is that the current release can be listed as `v0.1.23, HEAD` immediately after the release, but `v0.1.23` after another commit is added to the main branch. The `git export`-ed versions are then different. (Oh it's also an issue as the repository gains more objects in its tree, so a describe with the hash summarised as `-g123abc` can later become `-g123abcd`.)

Github *says* it guarantees stable hashes for a year (see https://github.com/orgs/community/discussions/46034), but maybe that's the guarantee only if the export contents don't change. See https://github.com/golang/go/issues/27153, https://github.com/NixOS/nix/issues/7596, and many more related threads.

PR #1064, opened 2025-07-09
Enable fast math library in CUDA compilation

stephenswat:
One of the biggest performance bottlenecks of our code right now are the math functions we use from the standard library. To illustrate the problem, these functions have the problem that they contain a _lot_ of edge case handling code for NaN values, infinities, etc. which are not necessary for our use case. This causes an incredible amount of code to be added and causes a huge amount of instruction cache misses when though CUDA contains native intrinsics of high accurary which model the same behaviour. This PR serves to illustrate the massive performance gain that we can achieve by using intrinsic math rather than the math provided by the C++ standard library.

### DIFF
--- a/cmake/traccc-compiler-options-cuda.cmake
+++ b/cmake/traccc-compiler-options-cuda.cmake
@@ -25,7 +25,7 @@ endif()
 
 # Allow to use functions in device code that are constexpr, even if they are
 # not marked with __device__.
-traccc_add_flag( CMAKE_CUDA_FLAGS "--expt-relaxed-constexpr" )
+traccc_add_flag( CMAKE_CUDA_FLAGS "--expt-relaxed-constexpr --use_fast_math" )
 
 # Make CUDA generate debug symbols for the device code as well in a debug
 # build.


stephenswat:
It's clear that using fast math has no meaningful effect on the physics performance, so we can move ahead with this.

beomki-yeo (approved):
Looks good to me!

krasznaa:
I'm generally on board. The Intel compiler is anyway doing "aggressive maths" by default as I understand it, so this will probably just make the CUDA<->SYCL comparisons more fair.

But @beomki-yeo, the failure in the ambiguity resolution test will need to be understood first. I''m pretty sure that it's something that can be fixed by making some requirement in the test a bit more loose, but we can of course not make tests fail in the `main` branch. :thinking:

stephenswat:
> Not sure why they fail. One possible scenario is the calculation of relative shared have different values with this PR
> 
> in `remove_tracks.cu`:
> 
> ```c++
>             rel_shared.at(tid) = static_cast<traccc::scalar>(n_shared.at(tid)) /
>                                  static_cast<traccc::scalar>(n_meas.at(tid));
> ```
> 
> and in `greedy_ambiguity_resolution.cu`
> 
> ```c++
>     thrust::transform(thrust_policy, n_shared_buffer.ptr(),
>                       n_shared_buffer.ptr() + n_tracks, n_meas_buffer.ptr(),
>                       rel_shared_buffer.ptr(), devide_op{});
> ```
> 
> (This division is the only floating number math operation used in the greedy ambiguity solver)
> 
> We can check if the test passes after forcing this operation executed with original non-fast-math.

Thanks Beomki, this was exactly correct. I enforced slow division for these functions and everything works now. :smile:

stephenswat:
# Performance summary
## Graphical

## Tabular
| Kernel | 9457b4f9 | a2705074 | Delta |
| --- | --: | --: | --: |
| `propagate_to_next_surface` | 25.22 ms | 19.86 ms | -21.3% |
| `fit_forward` | 31.06 ms | 19.25 ms | -38.0% |
| `fit_backward` | 12.87 ms | 10.04 ms | -22.0% |
| `find_tracks` | 2.35 ms | 2.05 ms | -12.7% |
| `count_triplets` | 1.87 ms | 1.30 ms | -30.4% |
| `ccl_kernel` | 872.55 μs | 826.62 μs | -5.3% |
| `count_doublets` | 970.86 μs | 664.93 μs | -31.5% |
| `find_doublets` | 779.33 μs | 495.01 μs | -36.5% |
| `Thrust::sort` | 451.59 μs | 450.41 μs | -0.3% |
| `find_triplets` | 463.10 μs | 347.43 μs | -25.0% |
| `fit_prelude` | 187.00 μs | 185.97 μs | -0.6% |
| `remove_duplicates` | 129.49 μs | 110.38 μs | -14.8% |
| `select_seeds` | 77.61 μs | 70.43 μs | -9.3% |
| `build_tracks` | 52.62 μs | 47.80 μs | -9.2% |
| `apply_interaction` | 37.61 μs | 33.19 μs | -11.8% |
| `estimate_track_params` | 31.09 μs | 30.40 μs | -2.2% |
| `populate_grid` | 23.63 μs | 23.36 μs | -1.1% |
| `fill_finding_propagation_sort_keys` | 23.09 μs | 23.07 μs | -0.1% |
| `count_grid_capacities` | 22.43 μs | 22.13 μs | -1.3% |
| `update_triplet_weights` | 21.61 μs | 21.27 μs | -1.6% |
| `unknown` | 20.47 μs | 20.55 μs | 0.4% |
| `form_spacepoints` | 12.46 μs | 12.20 μs | -2.1% |
| `fill_finding_duplicate_removal_sort_keys` | 10.11 μs | 10.12 μs | 0.1% |
| `reduce_triplet_counts` | 6.68 μs | 6.70 μs | 0.3% |
| `fill_fitting_sort_keys` | 1.10 μs | 1.09 μs | -0.2% |
| `make_barcode_sequence` | 1.00 μs | 1.01 μs | 0.3% |
| `fill_prefix_sum` | 171.85 ns | 171.79 ns | -0.0% |
| **Total** | **77.57 ms** | **55.90 ms** | **-27.9%** |
> [!IMPORTANT]

> [!NOTE]

PR #1065, opened 2025-07-09
Implement visitor for polymorphic magnetic field

stephenswat:
This commit implements the visitor logic that I proposed for #1033, drastically reducing the amount of boilerplate we need to handle different magnetic field types and increasing the maintainability of the code significantly. This also paves the way for polymorphic detector types which we will need in the future.

### DIFF
--- a/core/include/traccc/bfield/magnetic_field.hpp
+++ b/core/include/traccc/bfield/magnetic_field.hpp
@@ -13,6 +13,7 @@
 
 // System include(s).
 #include <any>
+#include <sstream>
 
 namespace traccc {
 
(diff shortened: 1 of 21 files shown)


beomki-yeo (approved):
Looks good to me

core/include/traccc/bfield/magnetic_field.hpp
```diff
@@ -64,12 +65,55 @@ class magnetic_field {
...
+///
+/// This function takes a list of supported magnetic field types and checks
+/// if the provided field is one of them. If it is, it will call the provided
+/// callable on a view of it and otherwise it will throw an exception.
+template <typename bfield_list_t, typename callable_t>
+auto magnetic_field_visitor(const magnetic_field& bfield,
+                            callable_t&& callable) {
+    return magnetic_field_visitor_helper(bfield,
+                                         std::forward<callable_t>(callable),
+                                         static_cast<bfield_list_t*>(nullptr));
```
krasznaa, line 114: Would [std::declval](https://en.cppreference.com/w/cpp/utility/declval.html) not be more elegant here?

krasznaa (approved):
Sure, let's start with this cleanup finally.

core/include/traccc/bfield/magnetic_field.hpp
krasznaa: Never mind, apparently not. 🤔

PR #1055, opened 2025-07-03
Switch CUDA B-field to use texture memory

stephenswat:
This commit changes the CUDA magnetic field representation from a classical array-based field to a field in texture memory.

PR #1069, opened 2025-07-10
B-Field Rewrite / Cleanup, main branch (2025.07.10.)

krasznaa:
This is the change that I was foretelling about in #1055 already. Where I would right some of the wrongs introduced in #1033.

There were multiple things about `traccc::bfield` that I didn't like:
  - The name. I really wanted to call it `traccc::magnetic_field` instead. While calling instances/variables `bfield`.
  - I didn't like how `"traccc/utils/bfield.hpp"` contained pretty much everything.
  - I was not happy about the naming I gave to some of its functions.

I tried to address all of these in this PR. While also resolving the `bfield` <-> `magnetic_field` inconsistencies in the names of various functions.

It ended up touching way more files than I expected, so in the end I decided not to add the CUDA texture memory changes from #1055 in this PR. That would come in a follow-up instead.

I know that this will mess up multiple open PRs from @stephenswat, but as I told him already, I'd really like to clean this up before we make even more code depend on the naming that I didn't like...


stephenswat (approved):
Looks good!

PR #1074, opened 2025-07-14
Alpaka: Parallel Algorithm Unification

CrossR:
The Alpaka code needs to be able to call:

 - Thrust with CPU backend
 - Thrust with CUDA backend
 - rocThrust with a HIP backend
 - oneDPL with a SYCL backend

with a consistent interface.

This is a very basic outline of what that could look like, though I'm open to any suggestions as I think this way, whilst simple, is potentially a pain if we need to keep adding to it.

I've tested this on an A100, MI210 and GPU Max 1100 and it works across every back end at least.


stephenswat (approved):
Let's move forward on this.

device/alpaka/src/utils/parallel_algorithms.hpp
```diff
@@ -0,0 +1,132 @@
...
+#include "traccc/utils/memory_resource.hpp"
+
+// Thrust include(s).
+#include <thrust/binary_search.h>
+#include <thrust/copy.h>
+#include <thrust/execution_policy.h>
+#include <thrust/fill.h>
+#include <thrust/scan.h>
+#include <thrust/sort.h>
+#include <thrust/unique.h>
```
krasznaa, line 24: Should these not be hidden when a SYCL build is performed? I know that right now Thrust is always available in our builds, but that may not remain the case always.

Since in the same way we could technically make oneDPL just always be available as well, and could remove the preprocessor statements around those includes as well. (But in general I don't think that would be a good idea.)

device/alpaka/src/utils/parallel_algorithms.hpp
```diff
@@ -0,0 +1,132 @@
...
+#pragma clang diagnostic ignored "-Wshadow"
+#pragma clang diagnostic ignored "-Wunused-parameter"
+#pragma clang diagnostic ignored "-Wshorten-64-to-32"
+#pragma clang diagnostic ignored "-Wsign-conversion"
+#pragma clang diagnostic ignored "-Wimplicit-int-conversion"
+#pragma clang diagnostic ignored "-Wsign-compare"
+#pragma clang diagnostic ignored "-Wold-style-cast"
+#include <oneapi/dpl/algorithm>
+#include <oneapi/dpl/execution>
+#pragma clang diagnostic pop
```
krasznaa, line 38: You should do something like this instead:

https://github.com/acts-project/traccc/blob/main/device/sycl/src/utils/oneDPL.hpp

Since listing the warnings individually is pretty fragile. (But we do need to turn them off in pragmas unfortunately. 😦)

krasznaa (commented):
Good effort, I like it, let's bring it across the finish line this week!

CrossR:
Hm...I did a test build on all 3 platforms before pushing, but SYCL isn't happy.

I'll check tomorrow morning.

PR #1077, opened 2025-07-15
Add truth finding cuts and efficiency output

stephenswat:
This commit updates the truth matching code to have additional cuts on r, z, and the number of measurements on the track. It also adds some printouts to the finding efficiency writer to give the overall efficiency, fake rate, etc. without the need to use ROOT.


stephenswat:
This is a prerequisite for #1082 and useful in general.

niermann999 (approved):
Looks good to me

krasznaa (approved):
Looks harmless enough.

PR #1079, opened 2025-07-16
Increase the shared memory allocation

beomki-yeo:
After the update of #1076, the following test fails, which is probably due to the insufficient memory allocation.

```
INSTANTIATE_TEST_SUITE_P(
    Dense, GreedyResolutionCompareToCPU,
    ::testing::Values(std::make_tuple(3u, 5000u,
                                      std::array<std::size_t, 2u>{3u, 10u},
                                      100u, true),
                      std::make_tuple(3u, 5000u,
                                      std::array<std::size_t, 2u>{3u, 10u},
                                      100u, false)));
```

This PR increases the size enough to avoid the failure.
I also found that the tests of ambiguity resolution is not included in the gitlab CI. Anyone has a clue on it?

### DIFF
--- a/device/cuda/src/ambiguity_resolution/kernels/count_removable_tracks.cu
+++ b/device/cuda/src/ambiguity_resolution/kernels/count_removable_tracks.cu
@@ -67,10 +67,10 @@ __launch_bounds__(512) __global__ void count_removable_tracks(
         return;
     }
 
-    __shared__ int shared_n_meas[512];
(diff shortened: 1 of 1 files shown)


beomki-yeo:
@krasznaa Right. It turns out that `count_removable_tracks` has nothing to do with the current bug, the size of 512 is always enough for the current algorithm by its logic. The instability comes from other cuda kernels

The fix is handled in #1083 instead, so closing the PR

PR #1080, opened 2025-07-17
Add black-box cut optimisation tool to extras

stephenswat:
This commit adds a new black-box optimisation tool written in Python which can be used to find better configuration parameters for seeding and finding algorithms in traccc. The code is split up into a benchmarking tool which explores a design space and finds efficiencies, duplicate rates, fake rates, and throughput numbers. Then, another tool is capable of finding the Pareto set.


stephenswat:
Depends on #1077.

PR #1086, opened 2025-07-23
Remove the unused variables

beomki-yeo:
Remove the unused variables


krasznaa (approved):
A trivial modification for a change! 😁

PR #1087, opened 2025-07-24
Add fast math function library

stephenswat:
This commit adds a new `math::fast` namespace alternative to `math` which will always attempt to use fast instrinsics when possible. This allows much finer control over what is and isn't using fast math to allow us to manage precision. Also migrate some `std` math functions to call into `math` or `math::fast` instead.


krasznaa:
Also, with the CUDA measurements not showing any improvement, there might be something going wrong with the CUDA build as well, not just the SYCL one. :thinking:

stephenswat:
Closing in favour of #1064.

PR #1088, opened 2025-07-24
CMDL Options Improvements, main branch (2025.07.24.)

krasznaa:
Added command line options for all of the configuration flags that are being overridden from their default values in the EFTracking code. Took the changes implemented by @flg in his fork as a guide for all this.

This finally adds at least some configurable options for track seeding. For which I also changed the API of `traccc::opts::track_seeding` to be more in line with all the other options classes. So that it would provide the underlying configuration objects through conversion operators, which ended up simplifying the code using `traccc::opts::track_seeding` a little. (Though I do fear that the usage of the class may have become a bit more opaque like this. :frowning:)

Plus added some additional flags to `traccc::opts::track_propagation`. Again, just ones that were missing for EFTracking's benefit.

### DIFF
--- a/examples/options/include/traccc/options/track_seeding.hpp
+++ b/examples/options/include/traccc/options/track_seeding.hpp
@@ -30,13 +30,13 @@ class track_seeding : public interface,
     /// @{
 
     /// Configuration provider for the seed-finder
-    operator seedfinder_config() const override;
+    explicit operator seedfinder_config() const override;
     /// Configuration provider for the seed-filter
-    operator seedfilter_config() const override;
+    explicit operator seedfilter_config() const override;
     /// Configuration provider for the spacepoint grid
-    operator spacepoint_grid_config() const override;
+    explicit operator spacepoint_grid_config() const override;
     /// Configuration provider for the constant magnetic field assumed
-    operator vector3() const override;
+    explicit operator vector3() const override;
 
     /// @}
 
--- a/examples/options/include/traccc/options/track_seeding.hpp
+++ b/examples/options/include/traccc/options/track_seeding.hpp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2022-2024 CERN for the benefit of the ACTS project
+ * (c) 2022-2025 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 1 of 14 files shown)


examples/options/include/traccc/options/track_seeding.hpp
```diff
@@ -1,39 +1,75 @@
...
-    /// Configuration for the seed filtering
-    traccc::seedfilter_config seedfilter;
+    /// Configuration provider for the seed-finder
+    operator seedfinder_config() const override;
+    /// Configuration provider for the seed-filter
+    operator seedfilter_config() const override;
+    /// Configuration provider for the spacepoint grid
+    operator spacepoint_grid_config() const override;
+    /// Configuration provider for the constant magnetic field assumed
+    operator vector3() const override;
```
stephenswat, line 39: With this many conversion operators, let's make them explicit.

examples/options/src/track_propagation.cpp
```diff
@@ -52,11 +52,45 @@ track_propagation::track_propagation()
...
+    m_desc.add_options()("min-step-size-mm",
+                         po::value(&(m_config.stepping.min_stepsize))
+                             ->default_value(m_config.stepping.min_stepsize /
+                                             traccc::unit<float>::mm),
+                         "The minimum step size [mm]");
+    m_desc.add_options()("stepping-path-limit",
+                         po::value(&(m_config.stepping.path_limit))
+                             ->default_value(m_config.stepping.path_limit /
+                                             traccc::unit<float>::m),
+                         "The maximum path length for the stepper [m]");
```
stephenswat, line 69: Is there a reason for the top one to have the unit but the bottom one not?

examples/options/src/track_propagation.cpp
```diff
@@ -52,11 +52,45 @@ track_propagation::track_propagation()
         "search-window",
         po::value(&m_search_window)->default_value(m_search_window),
         "Size of the grid surface search window");
-    m_desc.add_options()("rk-tolerance-mm [mm]",
+    m_desc.add_options()("rk-tolerance-mm",
```
stephenswat, line 55: Should we use the `stepping-` namespace for all these options?

examples/options/src/track_seeding.cpp
```diff
@@ -10,12 +10,154 @@
...
+    m_seedfinder.rMax = m_r_range[1] * unit<float>::mm;
+    m_seedfinder.collisionRegionMin = m_vertex_range[0] * unit<float>::mm;
+    m_seedfinder.collisionRegionMax = m_vertex_range[1] * unit<float>::mm;
+    m_seedfinder.deltaRMin = m_delta_r_range[0] * unit<float>::mm;
+    m_seedfinder.deltaRMax = m_delta_r_range[1] * unit<float>::mm;
+
+    m_seedfinder.minPt *= unit<float>::GeV;
+    m_seedfinder.impactMax *= unit<float>::mm;
+    m_seedfinder.maxPtScattering *= unit<float>::GeV;
+    m_seedfinder.bFieldInZ *= unit<float>::T;
```
stephenswat, line 112: Don't forget to call `.setup()` here!

stephenswat (changes requested):
Great initiative but at the very least we need to properly `setup()` the configs.

examples/options/src/track_propagation.cpp
krasznaa: :thinking: The naming was pretty all over the place already, I didn't notice that my AI helper was doing this.

Since I personally don't like putting `-mm` into the option itself, I'll make all new options skip having units in their names.

stephenswat (approved):
This I can live with. 👍

PR #1089, opened 2025-07-24
Deduplicate max seeds per middle SP parameter

stephenswat:
Currently, we have three seeding parameters which all control the maximum number of seeds per middle spacepoint. This is unnecessary and makes life much more complicated for everyone. Thus, this commit deduplicates those parameters, leaving a single parameter in the seed finder configuration.

### DIFF
--- a/core/include/traccc/seeding/detail/seeding_config.hpp
+++ b/core/include/traccc/seeding/detail/seeding_config.hpp
@@ -62,7 +62,7 @@ struct seedfinder_config {
     float maxPtScattering = 10.f * unit<float>::GeV;
 
     // for how many seeds can one SpacePoint be the middle SpacePoint?
-    unsigned int maxSeedsPerSpM = 10;
+    unsigned int maxSeedsPerSpM = 5;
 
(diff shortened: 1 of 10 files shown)


stephenswat:
Tagging @paradajzblond as the person who identified this issue.

paradajzblond:
Hi! Thank you for making this MR. Related to this, there is rMin in seed finder, but also spB_min_radius in seed filter config, I assume these are also the same or similar?

stephenswat:
> I guess the seed filter value could be removed. 🤔 "rMin" is for all of the spacepoints. But indeed, that includes the bottom ones.
> 
> And it would be silly to select middle or top spacepoints closer to the beamline than the bottom ones. 😛

Unfortunately the world is not so simple... The `spB_min_radius` is not to be read as "this is the minimum radius for the bottom spacepoint" but rather "any seed whose bottom spacepoint is further from the origin than this minimum radius gets an increase in its weight equal to the configuration parameter `good_spB_weight_increase`". :frowning_face:

krasznaa (approved):
Looks reasonable.

PR #1095, opened 2025-07-26
Reduce the unnecessary iterations in `remove_tracks`

beomki-yeo:
Similar to what was done in #1093.

The iterations of `for (unsigned int i = 0; i < threadIndex; ++i) {}` can be very massive because `threadIndex` can be large. We can minimize the iteration by storing the track ids to `shared_tids` from the first index sequentially.

This PR gives around 5% speed improvement in the greedy resolver.


beomki-yeo:
As a bonus, `count_removable_tracks` takes the most of top time-consuming kernels now.

image

PR #1082, opened 2025-07-21
Implement spacepoint voting seed filtering

stephenswat:

PR #1098, opened 2025-07-28
Implement a spacepoint weight cut on seeds 

stephenswat:
This PR implements a cut that requires a seed to be the best seed for at least $n$ of its constituent spacepoints, which mirrors the cut that is used in ACTS. This cut is a specialisation of the cut in https://github.com/acts-project/traccc/pull/1082, implemented in a way that should allow it to run faster.

### DIFF
--- a/core/include/traccc/definitions/common.hpp
+++ b/core/include/traccc/definitions/common.hpp
@@ -21,6 +21,10 @@ using unit = detray::unit<scalar_t>;
 template <typename scalar_t>
 using constant = detray::constant<scalar_t>;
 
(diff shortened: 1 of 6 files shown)


stephenswat:
As it stands this cut is _not_ beneficial to seeding performance.

test

stephenswat:
I discussed this with @andiwand and I found out that the cut that ACTS implements isn't actually quite this, but rather a greedy algorithm that updates after every step. That means that the ACTS implementation is essentially impossible to efficiently reproduce on GPU.

PR #1103, opened 2025-07-29
ITk Throughput, main branch (2025.07.27.)

krasznaa:
Made a script based on [extras/traccc_odd_throughput_mt_profiler.sh](extras/traccc_odd_throughput_mt_profiler.sh), which would run throughput tests using the ITk data prepared by @paradajzblond.

Nothing too fancy, and I'm also a bit suspicious about the numbers, but adding the script should not be too harmful.

@flg, I'll still run a couple of tests, comparing the results coming out of this script to what your modified code does. Once Neza is done with her Athena jobs. 😄


flg:
Quick comment to mention that we have adapted this script on our side as well:
https://gitlab.cern.ch/atlas-tdaq-ph2upgrades/atlas-tdaq-eftracking/traccc-integration/G-200/-/blob/master/power/traccc_ITk_throughput_mt_profiler.sh?ref_type=heads

stephenswat (approved):
Lives in `extras/` so is fine. :+1:

krasznaa:
> Quick comment to mention that we have adapted this script on our side as well: https://gitlab.cern.ch/atlas-tdaq-ph2upgrades/atlas-tdaq-eftracking/traccc-integration/G-200/-/blob/master/power/traccc_ITk_throughput_mt_profiler.sh?ref_type=heads

That is indeed good to know. 🤔 This script is a bit more general though, so it still makes sense to have it here as well.

PR #1109, opened 2025-08-01
fix: Refined hole counting in KF

niermann999:
If the navigation in the forward KF missed a surface, advance the track state iterator correctly (for example if the propagation config between CKF and KF is different or we are running truth fitting). This is only needed for navigators that are not "guided" or "direct". 
It also adds an option to the KF to do the full hole count, which is only possible if the propagation starts at the IP and continues through the whole detector, otherwise the first and last holes are missed and the hole count remains an underestimation. Since this will not help in increasing the fit accuracy, but eat more compute resources it will remain optional.

### DIFF
--- a/core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
+++ b/core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
@@ -20,6 +20,7 @@
 #include "traccc/utils/particle.hpp"
 
 // detray include(s).
+#include <detray/definitions/navigation.hpp>
 #include <detray/propagator/base_actor.hpp>
 
 // vecmem include(s)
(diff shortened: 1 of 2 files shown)


core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
```diff
@@ -71,12 +77,107 @@ struct kalman_actor_state {
...
+                if (trk_state.smoothed().is_invalid()) {
+                    ++n_missed;
+                }
+            }
+        } else {
+            for (const auto& trk_state : m_track_states) {
+                if (trk_state.filtered().is_invalid()) {
+                    ++n_missed;
+                }
+            }
```
stephenswat, line 101: For the sake of preventing divergence (impact probably minor):

```suggestion
        for (const auto& trk_state : m_track_states) {
            bool valid = backward_mode ? trk_state.smoothed().is_invalid() : trk_state.filtered().is_invalid();
            if (valid) {
                ++n_missed;
            }
        }
```

PR #1111, opened 2025-08-05
chore: Update clang-format to v18.1.8

paulgessinger:


krasznaa (approved):
It's only the "weird stuff" that gets modified, so let's just do it. I was afraid that an upgrade would change almost all of our sources...

PR #1112, opened 2025-08-05
Track Fit/State SoA, main branch (2025.08.05.)

krasznaa:
It's time for the next iteration of the SoA migration. :thinking:

I replaced the current `traccc::track_state_container_types` types with 2 new collection, and a "helper type".
  - `traccc::edm::track_fit_collection<ALGEBRA>` is a collection that stores overall information about the fitted tracks. Similar to what the "header" of `traccc::track_state_container_types` does;
    * It has one jagged vector variable (`state_indices`) which points at which track states belong to the tracks;
  - `traccc::edm::track_state_collection<ALGEBRA>` is a 1D collection of track states. Every element describes one track state, with an index (`measurement_index`) to the measurement that the state is defined on top of;
  - `traccc::edm::track_fit_container<ALGEBRA>` is a type similar to `traccc::edm::track_candidate_container<ALGEBRA>`. It defines simple structs that collect objects that need to be used together to be able to make sense of them.

The code by now is in a semi-functional state. All of the code "runs", but seemingly the host code is producing buggy output. (Two host unit tests fail, while the device unit tests all succeed. Plus the comparisons with `traccc_seq_example_cuda` and `traccc_seq_example_alpaka` show differences between the host and device results.) This issue absolutely needs to be figured out before this could go in. But I anyway wanted to open a PR already, to make everyone start getting used to the new classes.

Once/if this PR is sorted out, I plan the following developments:
  - Migrating `traccc::measurement` to an SoA layout should be fairly trivial once this goes in, so I will do that next;
  - Once everything is an SoA, the track finding and fitting EDM will need to be merged. So that the CKF would produce the same type of objects as the KF, just with less information attached.
    * I have some ideas for this already, but this will still be a bit complicated I fear.

I didn't do any performance checks yet, since the code's behaviour is questionable at the moment. But hopefully with these changes we will both need less memory (no copies of the measurement properties) and will speed up a little.

### DIFF
--- a/performance/src/resolution/stat_plot_tool.ipp
+++ b/performance/src/resolution/stat_plot_tool.ipp
@@ -12,9 +12,10 @@
 
 namespace traccc {
 
-template <typename T>
-void stat_plot_tool::fill(stat_plot_cache& cache,
-                          const edm::track_candidate<T>& find_res) const {
+template <typename track_candidate_backend_t>
+void stat_plot_tool::fill(
+    stat_plot_cache& cache,
+    const edm::track_candidate<track_candidate_backend_t>& find_res) const {
 
     // Avoid unused variable warnings when building the code without ROOT.
     (void)cache;
@@ -30,9 +31,10 @@ void stat_plot_tool::fill(stat_plot_cache& cache,
 #endif  // TRACCC_HAVE_ROOT
 }
 
-template <typename T>
-void stat_plot_tool::fill(stat_plot_cache& cache,
-                          const edm::track_fit<T>& fit_res) const {
+template <typename track_fit_backend_t>
+void stat_plot_tool::fill(
+    stat_plot_cache& cache,
+    const edm::track_fit<track_fit_backend_t>& fit_res) const {
 
     // Avoid unused variable warnings when building the code without ROOT.
     (void)cache;
@@ -48,9 +50,10 @@ void stat_plot_tool::fill(stat_plot_cache& cache,
 #endif  // TRACCC_HAVE_ROOT
 }
 
-template <typename T>
+template <typename track_state_backend_t>
 void stat_plot_tool::fill(
-    stat_plot_cache& cache, const edm::track_state<T>& trk_state,
+    stat_plot_cache& cache,
+    const edm::track_state<track_state_backend_t>& trk_state,
     const measurement_collection_types::host& measurements) const {
 
     // Avoid unused variable warnings when building the code without ROOT.
--- a/performance/src/resolution/stat_plot_tool.hpp
+++ b/performance/src/resolution/stat_plot_tool.hpp
@@ -59,25 +59,34 @@ class stat_plot_tool {
 
     /// @brief fill the cache
     ///
+    /// @tparam track_candidate_backend_t the backend type for @c find_res
+    ///
     /// @param cache the cache for statistics plots
     /// @param find_res track finding result
-    template <typename T>
-    void fill(stat_plot_cache& cache,
-              const edm::track_candidate<T>& find_res) const;
+    template <typename track_candidate_backend_t>
+    void fill(
+        stat_plot_cache& cache,
+        const edm::track_candidate<track_candidate_backend_t>& find_res) const;
 
     /// @brief fill the cache
+    ///
+    /// @tparam track_fit_backend_t the backend type for @c fit_res
     ///
     /// @param cache the cache for statistics plots
     /// @param fit_res fitting information that contains statistics
-    template <typename T>
-    void fill(stat_plot_cache& cache, const edm::track_fit<T>& fit_res) const;
+    template <typename track_fit_backend_t>
+    void fill(stat_plot_cache& cache,
+              const edm::track_fit<track_fit_backend_t>& fit_res) const;
 
     /// @brief fill the cache
     ///
+    /// @tparam track_state_backend_t the backend type for @c trk_state
+    ///
     /// @param cache the cache for statistics plots
     /// @param trk_state track state at local measurements
-    template <typename T>
-    void fill(stat_plot_cache& cache, const edm::track_state<T>& trk_state,
+    template <typename track_state_backend_t>
+    void fill(stat_plot_cache& cache,
+              const edm::track_state<track_state_backend_t>& trk_state,
               const measurement_collection_types::host& measurements) const;
 
     /// @brief fill the cache


core/include/traccc/edm/impl/track_fit_collection.ipp
```diff
@@ -0,0 +1,21 @@
...
+namespace traccc::edm {
+
+template <typename BASE>
+TRACCC_HOST_DEVICE void track_fit<BASE>::reset_quality() {
+
+    ndf() = {};
+    chi2() = {};
+    pval() = {};
+    nholes() = {};
+}
```
stephenswat, line 19: Please don't make an `.ipp` file for this; fold it into the `.hpp` file.

core/include/traccc/edm/impl/track_state_collection.ipp
```diff
@@ -0,0 +1,122 @@
...
+            if constexpr (D == 2u) {
+                getter::element(pos, 1, 0) = local[0];
+            }
+            break;
+        default:
+#if defined(__GNUC__)
+            __builtin_unreachable();
+#endif
+    }
+}
```
stephenswat, line 79: For this whole file: let's not embed too much business logic into data classes. We should lift these member functions into free functions.

benchmarks/cuda/toy_detector_cuda.cpp
```diff
@@ -129,7 +125,7 @@ BENCHMARK_DEFINE_F(ToyDetectorBenchmark, CUDA)(benchmark::State& state) {
                                params_cuda_buffer);
 
             // Run track fitting
-            traccc::track_state_container_types::buffer
+            traccc::edm::track_fit_container<traccc::default_algebra>::buffer
```
stephenswat, line 128: More future-proof to use the algorithm's output type here.

core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp
```diff
@@ -37,20 +38,24 @@ struct gain_matrix_updater {
     /// @param bound_params bound parameter
     ///
     /// @return true if the update succeeds
+    template <typename track_state_backend_t>
```
stephenswat, line 41: What is this `track_state_backend_t` here?

performance/include/traccc/performance/details/projector.hpp
```diff
@@ -47,12 +47,4 @@ struct projector<traccc::bound_track_parameters<algebra_t>> {
...
 };
 
-template <>
-struct projector<fitting_result<traccc::default_algebra>> {
-    static constexpr bool exists = true;
-
-    float operator()(const fitting_result<traccc::default_algebra>& i) {
-        return static_cast<float>(i.trk_quality.ndf);
-    }
-};
```
stephenswat, line 57: Where did this go?

performance/src/efficiency/finding_performance_writer.cpp
```diff
@@ -129,23 +129,33 @@ std::vector<std::vector<measurement>> prepare_data(
...
 
-    // Iterate over the tracks.
-    track_state_container_types::const_device track_states(track_states_view);
+    // Set up the input containers.
+    const edm::track_fit_collection<default_algebra>::const_device track_fits(
+        track_fit_view);
+    const edm::track_state_collection<default_algebra>::const_device
+        track_states(track_states_view);
+    const measurement_collection_types::const_device measurements{
+        measurements_view};
```
stephenswat, line 145: Is there really no nicer way of doing this than tripling the already extreme amount of vecmem boilerplate we have?

stephenswat (commented):
This is a _very_ concerning amount of added complexity... I really hope you manage to make this a bit simpler.

benchmarks/cuda/toy_detector_cuda.cpp
krasznaa: The applications are a bit all over the place with their formalism. In this PR I was going to keep the current idioms everywhere. Not wanting to change more things than necessary.

core/include/traccc/edm/impl/track_fit_collection.ipp
krasznaa: Since I already made the file, it would be silly to remove it. Even if you don't like them, I myself do like `.ipp` files.

core/include/traccc/edm/impl/track_state_collection.ipp
krasznaa: Yeah, I agree with this. 🤔 Again, here I was keeping the idioms that we currently have in the code.

https://github.com/acts-project/traccc/blob/main/core/include/traccc/edm/track_state.hpp#L76-L135

I'm open to changing this aspect of our code in this PR, I just want to point out that most things here are not new. They are just re-packaged.

core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp
krasznaa: It is the `BASE` type from `track_state_collection.hpp`. "Backend" seems like an appropriate term for it.

The templating is needed here because in the CKF this code gets called on a "standalone" `traccc::edm::track_state` object. While in the KF it is called on an element of an SoA container. The "backend" describes that in one case the `trk_state` is operating on values, and in the other it operates on references.

performance/include/traccc/performance/details/projector.hpp
krasznaa: The `traccc::fitting_result` object is no more.

And this class is fundamentally not needed with the SoA data. We don't need to do anything to "project" the `ndf` variable into a vector. We can just access the array holding the `ndf` variable "natively", without any transformation.

performance/src/efficiency/finding_performance_writer.cpp
krasznaa: This is a good find. 🤔 

I wrote this code long ago, before introducing `traccc::edm::track_fit_container`. These could now be replaced by a single `traccc::edm::track_fit_container<traccc::default_algebra>::const_device` type. (Which is to be added to that container type.)

core/include/traccc/edm/impl/track_fit_collection.ipp
andiwand: lol

core/include/traccc/edm/impl/track_state_collection.ipp
krasznaa: I moved these into standalone helper functions.

krasznaa:
The code should now be working correctly. :thinking:

It took me a bit embarrassing amount of time to debug it, but the reason that there were two tests failing earlier is that with the new EDM I changed the behaviour of `traccc::host::kalman_fitting_algorithm` a bit. Right now, with the current code, `host::kalman_fitting_algorithm` throws away tracks that could not be fitted correctly. It never returns failed fits to the user.

https://github.com/acts-project/traccc/blob/main/core/include/traccc/fitting/details/kalman_fitting.hpp#L88-L94

While the device code has always produced tracks with failed fits as well. (Since it's much easier to just allocate the memory for all found tracks, and then just flag the ones that failed the fit.)

Since I believe the device code's behaviour is actually more correct, during the update I switched the host code to do the same. But the host tests were written with the assumption that all tracks received from the fitting algorithm would've been successfully fitted. Even though some of the test fits are known to fail.

Long story short, I disabled some of the detailed checks in the tests for the failed fits.

I think I also finished with every other part of the update at this point, so this is generally ready for serious review.

stephenswat:
# Physics performance summary
Here is a summary of the physics performance effects of this PR. Command used:

```
traccc_seeding_example_cuda --input-directory=/data/Acts/odd-simulations-20240506/geant4_ttbar_mu200 --digitization-file=geometries/odd/odd-digi-geometric-config.json --detector-file=geometries/odd/odd-detray_geometry_detray.json --grid-file=geometries/odd/odd-detray_surface_grids_detray.json --use-detray-detector=on --input-events=10 --use-acts-geom-source=on --check-performance --truth-finding-min-track-candidates=5 --truth-finding-min-pt=1.0 --truth-finding-min-z=-150 --truth-finding-max-z=150 --truth-finding-max-r=10 --seed-matching-ratio=0.99 --track-matching-ratio=0.5 --track-candidates-range=5:100 --seedfinder-vertex-range=-150:150
```

## Seeding performance
Total number of seeds went from **336240** to **336240** **(+0.0%)**




## Track finding performance
Total number of found tracks went from **153828** to **153825** **(-0.0%)**










> [!NOTE]

krasznaa:
Curious. :thinking: Are these physics efficiency plots generally "stable"? The following worries me a bit:

image

Since the physics performance should really not have changed at all by these changes. :thinking:

What I'll be even more curious about is the compute performance of these changes. Since I still didn't test that seriously myself. :frowning:

krasznaa:
> Curious. 🤔 Are these physics efficiency plots generally "stable"? The following worries me a bit:
> 
> image
> Since the physics performance should really not have changed at all by these changes. 🤔
> 
> What I'll be even more curious about is the compute performance of these changes. Since I still didn't test that seriously myself. 😦

Though then again, that plot is about the track finding performance. While this PR changes how the track fitting (and not the finding) works. :thinking: So probably there's a bit of uncertainty in these plots then.

Which will hopefully reduce a whole lot in the future, with improved maths/formalisms for the Kalman fitting. :wink:

stephenswat (changes requested):
Let me just slap on one of these to make sure we don't make any hasty decisions.

In the meanwhile, could you help me understand how this change affects #1052? I.e. what kinds of changes will be necessary following this to get that PR in order?

stephenswat:
# Performance summary
## Graphical

## Tabular
| Kernel | a2dbda9e | d8b94dc5 | Delta |
| --- | --: | --: | --: |
| `fit_forward` | 32.73 ms | 32.68 ms | -0.2% |
| `propagate_to_next_surface` | 28.62 ms | 28.67 ms | 0.2% |
| `fit_backward` | 13.54 ms | 13.91 ms | 2.7% |
| `find_tracks` | 2.69 ms | 2.77 ms | 3.0% |
| `count_triplets` | 2.06 ms | 2.06 ms | -0.2% |
| `count_doublets` | 1.01 ms | 996.79 μs | -1.5% |
| `ccl_kernel` | 872.42 μs | 871.71 μs | -0.1% |
| `find_doublets` | 797.52 μs | 804.14 μs | 0.8% |
| `find_triplets` | 571.07 μs | 570.79 μs | -0.0% |
| `Thrust::sort` | 457.36 μs | 456.78 μs | -0.1% |
| `remove_duplicates` | 169.45 μs | 169.64 μs | 0.1% |
| `fit_prelude` | 197.31 μs | 163.50 μs | -17.1% |
| `select_seeds` | 102.32 μs | 102.51 μs | 0.2% |
| `build_tracks` | 54.64 μs | 55.27 μs | 1.2% |
| `apply_interaction` | 42.22 μs | 42.26 μs | 0.1% |
| `estimate_track_params` | 38.11 μs | 38.05 μs | -0.1% |
| `update_triplet_weights` | 29.84 μs | 30.02 μs | 0.6% |
| `fill_finding_propagation_sort_keys` | 26.00 μs | 25.99 μs | -0.0% |
| `populate_grid` | 23.81 μs | 23.81 μs | -0.0% |
| `count_grid_capacities` | 22.58 μs | 22.61 μs | 0.1% |
| `unknown` | 20.52 μs | 20.43 μs | -0.4% |
| `form_spacepoints` | 12.52 μs | 12.61 μs | 0.7% |
| `fill_finding_duplicate_removal_sort_keys` | 10.77 μs | 10.76 μs | -0.0% |
| `reduce_triplet_counts` | 6.84 μs | 6.85 μs | 0.2% |
| `fill_fitting_sort_keys` | 1.15 μs | 1.15 μs | 0.2% |
| `make_barcode_sequence` | 1.01 μs | 1.01 μs | -0.1% |
| `fill_prefix_sum` | 171.98 ns | 171.93 ns | -0.0% |
| **Total** | **84.11 ms** | **84.51 ms** | **0.5%** |
> [!IMPORTANT]

> [!NOTE]

krasznaa:
> Let me just slap on one of these to make sure we don't make any hasty decisions.
> 
> In the meanwhile, could you help me understand how this change affects #1052? I.e. what kinds of changes will be necessary following this to get that PR in order?

I looked at that PR just now. I believe the changes to it will be pretty mechanical.

This PR doesn't fundamentally change the data model (yet). The current track state "container" (composed of a 1D + a jagged vector) is split into 2 SoA collections. But the type is still different from the type used as the output of the track finding. So the changes needed for #1052 will follow a recipe.

My eventual goal here is to come up with a `traccc::edm::track_container` type. Which would be used as the output of both track finding and track fitting. With some information just at runtime missing from the objects. Like it is done in Acts, where the fitter receives and returns fundamentally the same type of object.

https://github.com/acts-project/acts/blob/main/Core/include/Acts/TrackFitting/KalmanFitter.hpp#L1062

(The fit there rather modifies the track container in place, which I'm not super keen on, but we'll see.)

But I believe your changes with track finding will need to go in before that fundamental of a rewrite of the GPU EDM.

krasznaa:
> # Performance summary
> Here is a summary of the performance effects of this PR:
> 
> ## Graphical
> 
> 
> ## Tabular
> Kernel	[a2dbda9](https://github.com/acts-project/traccc/commit/a2dbda9e9c97f360d46badd8bdbc7d4036cb203f)	[d8b94dc](https://github.com/acts-project/traccc/commit/d8b94dc5d541a1cc6a69d24f20930f99e7aa0f81)	Delta
> `fit_forward`	32.73 ms	32.68 ms	-0.2%
> `propagate_to_next_surface`	28.62 ms	28.67 ms	0.2%
> `fit_backward`	13.54 ms	13.91 ms	2.7%
> `find_tracks`	2.69 ms	2.77 ms	3.0%
> `count_triplets`	2.06 ms	2.06 ms	-0.2%
> `count_doublets`	1.01 ms	996.79 μs	-1.5%
> `ccl_kernel`	872.42 μs	871.71 μs	-0.1%
> `find_doublets`	797.52 μs	804.14 μs	0.8%
> `find_triplets`	571.07 μs	570.79 μs	-0.0%
> `Thrust::sort`	457.36 μs	456.78 μs	-0.1%
> `remove_duplicates`	169.45 μs	169.64 μs	0.1%
> `fit_prelude`	197.31 μs	163.50 μs	-17.1%
> `select_seeds`	102.32 μs	102.51 μs	0.2%
> `build_tracks`	54.64 μs	55.27 μs	1.2%
> `apply_interaction`	42.22 μs	42.26 μs	0.1%
> `estimate_track_params`	38.11 μs	38.05 μs	-0.1%
> `update_triplet_weights`	29.84 μs	30.02 μs	0.6%
> `fill_finding_propagation_sort_keys`	26.00 μs	25.99 μs	-0.0%
> `populate_grid`	23.81 μs	23.81 μs	-0.0%
> `count_grid_capacities`	22.58 μs	22.61 μs	0.1%
> `unknown`	20.52 μs	20.43 μs	-0.4%
> `form_spacepoints`	12.52 μs	12.61 μs	0.7%
> `fill_finding_duplicate_removal_sort_keys`	10.77 μs	10.76 μs	-0.0%
> `reduce_triplet_counts`	6.84 μs	6.85 μs	0.2%
> `fill_fitting_sort_keys`	1.15 μs	1.15 μs	0.2%
> `make_barcode_sequence`	1.01 μs	1.01 μs	-0.1%
> `fill_prefix_sum`	171.98 ns	171.93 ns	-0.0%
> **Total**	**84.11 ms**	**84.51 ms**	**0.5%**
> Important
> 
> 
> Note
> 

This is a bit disappointing. :frowning: `fit_prelude` became faster, though not by as much as I expected. And I was hoping that `fit_forward` and `fit_backward` would at worst just keep the same speed. I did not expect them to become slightly slower.

As normal, there's a noticeable wiggle on kernels that this code doesn't touch, but it still seems that the fitting functions are a little slower. :confused: I'll do some tests of my own as well then.

krasznaa:
Unfortunately the situation is quite a bit worse than I thought. :sob:

Running `traccc_throughput_mt_cuda` tests with ODD mu200 files using this PR's code is producing miserable results. :confused: Some of which I understand to some level (creating the output host object with the SoA layout is expensive), but some of it I don't quite. The fitting kernels seem less eager to overlap with each other (or just any other thing).

So, this is definitely no bueno like this as is.

stephenswat:
> This PR doesn't fundamentally change the data model (yet).

Wait so this does not yet deduplicate the data model between finding and fitting? Will that be a follow-up PR?

krasznaa:
> > This PR doesn't fundamentally change the data model (yet).
> 
> Wait so this does not yet deduplicate the data model between finding and fitting? Will that be a follow-up PR?

Yes, it will be. Now... to merge `traccc::edm::track_candidate_container` and `traccc::edm::track_fit_container`, as they are in this PR, will be a lot easier to follow for any reviewer than if I'd change everything absolutely at the same time. Since many of the "collections" will remain mostly unchanged. It's the "top most" classes that would be merged such that a single interface would describe tracks at various stages of reconstruction. (Again, similar to what is done in Acts.)

Note though that there is a whole lot less of data duplication with this PR than is the current status quo. However the interfaces for the output of track finding and track fitting stay separated in this PR.

krasznaa:
> Unfortunately the situation is quite a bit worse than I thought. 😭
> 
> Running `traccc_throughput_mt_cuda` tests with ODD mu200 files using this PR's code is producing miserable results. 😕 Some of which I understand to some level (creating the output host object with the SoA layout is expensive), but some of it I don't quite. The fitting kernels seem less eager to overlap with each other (or just any other thing).
> 
> So, this is definitely no bueno like this as is.

I think I more or less understand the situation by now. But maybe let me start with things that were, as it turns out, not clear to me previously:
  - Trying to understand kernel run times in multi-threaded applications is quite futile.
    * As kernels overlap with each other in parallel streams, their run time can change a fair bit.
    * So it's really only `traccc_throughput_st_<flavour>` that's useful for understanding just what the kernels are doing.
  - The throughput applications are still a bit inconsistent with their memory use.

So, once I use just `traccc_throughput_st_cuda` between the main branch and this PR's branch, I see pretty much what your previous measurement was Stephen. That the kernel times are largely unchanged. (Using `traccc_throughput_mt_cuda` was giving me false hope that the fitting kernels would've actually sped up.)

The reason that the overall application ridiculously slows down is something that may be best to fix in a follow-up PR. :thinking:  Since there are actually multiple problems with how copies are made by the "full chain algorithms" at the moment. :frowning:
  - On the input we copy `traccc::edm::silicon_cell_collection::host` directly to the device. (https://github.com/acts-project/traccc/blob/main/examples/run/cuda/full_chain_algorithm.cpp#L165) This results in 5 copies being made instead of just 1. This is not the end of the world, but still not perfect.
  - On the output the code now creates a `traccc::edm::track_fit_collection<ALGEBRA>::host` object, using the cached host memory resource that the application uses. This host object has one big `std::vector<std::vector<unsigned int>>` in it (beside a number of 1D vectors), which `vecmem::binary_page_memory_resource` has a hell of a time with. :frowning: I see in my profile that the CPU is spending a ridiculous amount of time with allocating the space for that output object.
    * I have a number of ideas for how to sidestep these many small allocations. But they will all require changes orthogonal to what this PR is doing.

So, as bad as this is, now that at least I think I understand what the situation is, I think it may be best to power on with the PR as it is. :thinking:

core/include/traccc/edm/track_fit_collection.hpp
```diff
@@ -0,0 +1,175 @@
...
+    // params
+    vecmem::edm::type::vector<bound_track_parameters<ALGEBRA>>,
+    // ndf
+    vecmem::edm::type::vector<detray::dscalar<ALGEBRA>>,
+    // chi2
+    vecmem::edm::type::vector<detray::dscalar<ALGEBRA>>,
+    // pval
+    vecmem::edm::type::vector<detray::dscalar<ALGEBRA>>,
+    // nholes
+    vecmem::edm::type::vector<unsigned int>,
```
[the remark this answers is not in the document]
krasznaa: It's really not because of the access pattern that @beomki-yeo created `traccc::fit_quality` back in the day. The reason that this struct was added was because the same quality variables are used in the output of track finding and it track fitting.

So if you want to complain about something, you should've asked why we have duplication between these variable definitions between `track_fit_collection` and `track_candidate_collection`.

https://github.com/acts-project/traccc/blob/main/core/include/traccc/edm/track_candidate_collection.hpp#L150

This duplication is one of the main reasons that we'll need to merge these two data models together.

But I'm really not convinced that using a struct for those variables instead of a fully SoA layout, makes much of a difference. With detailed profiling we may decide in the future that some of these individual `float`-s should be merged into small structs. But I don't think that even then we will group the exact same variables together that are in `traccc::fit_quality` at the moment.

core/include/traccc/edm/track_fit_container.hpp
```diff
@@ -0,0 +1,106 @@
...
+    };
+
+    struct const_view {
+        /// The fitted tracks
+        track_fit_collection<ALGEBRA>::const_view tracks;
+        /// The track states used for the fit
+        track_state_collection<ALGEBRA>::const_view states;
+        /// The measurements used for the fit
+        measurement_collection_types::const_view measurements;
+    };
```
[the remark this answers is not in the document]
krasznaa: Right now no part of the code needs `track_candidate_container` and `track_fit_container` at the same time.

For the longer future, this is something that the merging of the data models will take care of. With that in place, we shouldn't have duplication anymore.

I do very strongly believe though that having such simple structs for grouping together collections that belong together, is a pretty good way of organising the data.

core/include/traccc/edm/track_state_collection.hpp
```diff
@@ -0,0 +1,223 @@
...
+    // filtered_chi2
+    vecmem::edm::type::vector<detray::dscalar<ALGEBRA>>,
+    // smoothed_chi2
+    vecmem::edm::type::vector<detray::dscalar<ALGEBRA>>,
+    // backward_chi2
+    vecmem::edm::type::vector<detray::dscalar<ALGEBRA>>,
+    // filtered_params
+    vecmem::edm::type::vector<bound_track_parameters<ALGEBRA>>,
+    // smoothed_params
+    vecmem::edm::type::vector<bound_track_parameters<ALGEBRA>>,
```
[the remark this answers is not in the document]
krasznaa: :thinking: Are you saying that the "state" should just be "chi2" and "params", with the main track container associating as many states to the different measurements as many are needed?

That does sound interesting. But in this specific PR I would really prefer to do just the AoS->SoA conversion. If we mix in more fundamental re-designs as well, things will be even harder to review. :thinking:

core/include/traccc/fitting/details/kalman_fitting.hpp
```diff
@@ -50,46 +54,54 @@ track_state_container_types::host kalman_fitting(
...
+            result_tracks_device{result_tracks_view};
+        typename edm::track_fit_collection<algebra_t>::device::proxy_type
+            fitted_track_device =
+                result_tracks_device.at(result_tracks_device.size() - 1);
+        auto result_states_view = vecmem::get_data(result.states);
+        typename fitter_t::state fitter_state(
+            fitted_track_device,
+            typename edm::track_state_collection<algebra_t>::device{
+                result_states_view},
+            measurements, seqs_buffer);
```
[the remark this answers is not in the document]
krasznaa: One of the long names could be replaced by `auto`, though I fear that it will make it even harder to understand what types are being used exactly.

The code here is largely due to the design decision we/I made many years ago. That "view" and "device" objects would be separate types. So now we need to always create device objects out of the views.

As long as we don't change that (we can have a discussion about it, but it's definitely not a black or white question whether we should), we will need to be explicit about all these conversions. As they are not made implicitly. (Again, to hopefully make it a bit easier to debug code. As auto-conversions could really get out of hand with things like this.)

core/include/traccc/fitting/details/kalman_fitting.hpp
```diff
@@ -50,46 +54,54 @@ track_state_container_types::host kalman_fitting(
...
         // Run the fitter.
         kalman_fitter_status fit_status =
             fitter.fit(track_candidates.params().at(i), fitter_state);
 
-        if (fit_status == kalman_fitter_status::SUCCESS) {
-            // Save the results into the output container.
-            result.push_back(std::move(fitter_state.m_fit_res),
-                             std::move(input_states));
-        } else {
+        if (fit_status != kalman_fitter_status::SUCCESS) {
```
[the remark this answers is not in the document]
krasznaa: Please elaborate. Don't understand the question.

I commented about exactly his in: https://github.com/acts-project/traccc/pull/1112#issuecomment-3178171586

core/include/traccc/fitting/kalman_filter/gain_matrix_updater.hpp
krasznaa: We had this exact discussion on a previous PR already.

If not for this type of templating, then we'd need a concept for `traccc::edm::track_state<T>`. But the intention here is really that the user should only be able to provide `traccc::edm::track_state` objects here. Not something that maybe has the same accessor functions as it does.

performance/src/resolution/stat_plot_tool.ipp
```diff
@@ -29,4 +30,46 @@ void stat_plot_tool::fill(stat_plot_cache& cache,
 #endif  // TRACCC_HAVE_ROOT
 }
 
+template <typename T>
+void stat_plot_tool::fill(stat_plot_cache& cache,
+                          const edm::track_fit<T>& fit_res) const {
```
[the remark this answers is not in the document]
krasznaa: `T` here is the same "backend" that you asked about in  `gain_matric_updater`.

I can of course rename `T` into `track_state_backend_t` here as well. :thinking:

PR #1113, opened 2025-08-06
Use PDF for impact parameter weight

stephenswat:
This commit uses the PDF of a half-Gaussian distribution to determine the weight of the impact factor in the track finding. This change makes it so that weights are strictly positive and negates the "cut-off" effect caused by the confirmation weight.

### DIFF
--- a/core/include/traccc/seeding/detail/seeding_config.hpp
+++ b/core/include/traccc/seeding/detail/seeding_config.hpp
@@ -192,6 +192,9 @@ struct seedfilter_config {
     // the impact parameters (d0) is multiplied by this factor and subtracted
     // from weight
     float impactWeightFactor = 1.f;
+    // assumed variance of the half-Gaussian distribution of the impact
+    // parameter $d_0$ in mm
+    float impactSigma = 5.f;
     // seed weight increased by this value if a compatible seed has been found.
     float compatSeedWeight = 200.f;
     // minimum distance between compatible seeds to be considered for weight
@@ -211,7 +214,7 @@ struct seedfilter_config {
     float good_spB_min_weight = 380.f;
 
     // seed cut
-    float seed_min_weight = 200.f;
+    float seed_min_weight = 400.f;
     float spB_min_radius = 43.f * unit<float>::mm;
 };
 
--- a/device/common/include/traccc/seeding/device/impl/find_triplets.ipp
+++ b/device/common/include/traccc/seeding/device/impl/find_triplets.ipp
@@ -112,11 +112,35 @@ inline void find_triplets(
                 spM, lb, lt, config, iSinTheta2, scatteringInRegion2, curvature,
                 impact_parameter)) {
 
+            // Compute the weight attributed by the impact factor as the PDF of
+            // a half-Gaussian distribution, i.e. we compute:
+            //
+            // w' = \frac{\sqrt{2}}{\sigma\sqrt{\pi}}
+            //      \exp\left(\frac{-x^2}{2\sigma^2}\right)
+            //
+            // However, we would like the weight to be normalized to the range
+            // [0, 1], so we divide by the maximum probability density which
+            // is easily found as:
+            //
+            // w_{max} = \frac{\sqrt{2}}{\sigma\sqrt{\pi}}
+            //
+            // Thus normalizing gives:
+            //
+            // w = \frac{w'}{w_{max}} = \exp\left(\frac{-x^2}{2\sigma^2}\right)
+            //
+            // Conveniently, this normalization also reduces the number of
+            // operations to be performed.
+            scalar impactParameterProb = math::exp(
+                -(impact_parameter * impact_parameter) /
+                (2.f * filter_config.impactSigma * filter_config.impactSigma));
+
+            scalar weight =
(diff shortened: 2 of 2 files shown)


stephenswat:
# Physics performance summary
Here is a summary of the physics performance effects of this PR. Command used:

```
traccc_seeding_example_cuda --input-directory=/data/Acts/odd-simulations-20240506/geant4_ttbar_mu200 --digitization-file=geometries/odd/odd-digi-geometric-config.json --detector-file=geometries/odd/odd-detray_geometry_detray.json --grid-file=geometries/odd/odd-detray_surface_grids_detray.json --use-detray-detector=on --input-events=10 --use-acts-geom-source=on --check-performance --truth-finding-min-track-candidates=5 --truth-finding-min-pt=1.0 --truth-finding-min-z=-150 --truth-finding-max-z=150 --truth-finding-max-r=10 --truth-finding-matching-ratio=0.7 --track-candidates-range=5:100 --seedfinder-vertex-range=-150:150
```

## Seeding performance
Total number of seeds went from **354150** to **709767** **(+100.4%)**




## Track finding performance
Total number of found tracks went from **142155** to **200642** **(+41.1%)**




> [!NOTE]

stephenswat:
> So this PDF is less tighter than the current weight, which makes seeding find more seeds? The increased number of seeds is concerning but I might be onboard with it as the efficiency is significantly increased.

The problem is that we have a minimum seed weight of $200$ and we also have a per-confirmation seed weight of $200$. Imagine that a seed in the old system has an impact parameter of $0.0001$ mm. In the old system, this creates a weight of $200 - 0.0001 = 199.9999$ so it gets cut by the minimum weight of $200$. In this new system, the weights are strictly positive which I think is good; it means that we don't have this "cut-off" effect. But it also means that - for now - the cut is looser. I plan to rework this so we don't end up with way too many seeds.

> Meanwhile do you know why the seeding efficiency is higher than track finding efficiency in your plots above? My understanding is that it is supposed to be the other way around.

It's interesting; I guess that if we kill tracks in the CKF, there is no way to get them back. So we might be killing the wrong tracks? It's hard to say. Also, this assumes a matching ratio of 70% which is pretty tight. For the EF tracking in ATLAS we assume a weighted matching ratio of only 50%.

stephenswat:
# Physics performance summary
Here is a summary of the physics performance effects of this PR. Command used:

```
traccc_seeding_example_cuda --input-directory=/data/Acts/odd-simulations-20240506/geant4_ttbar_mu200 --digitization-file=geometries/odd/odd-digi-geometric-config.json --detector-file=geometries/odd/odd-detray_geometry_detray.json --grid-file=geometries/odd/odd-detray_surface_grids_detray.json --use-detray-detector=on --input-events=10 --use-acts-geom-source=on --check-performance --truth-finding-min-track-candidates=5 --truth-finding-min-pt=1.0 --truth-finding-min-z=-150 --truth-finding-max-z=150 --truth-finding-max-r=10 --truth-finding-matching-ratio=0.7 --track-candidates-range=5:100 --seedfinder-vertex-range=-150:150
```

## Seeding performance
Total number of seeds went from **354150** to **354052** **(-0.0%)**




## Track finding performance
Total number of found tracks went from **142155** to **142126** **(-0.0%)**




> [!NOTE]

stephenswat:
> The problem is that we have a minimum seed weight of 200 and we also have a per-confirmation seed weight of 200 . Imagine that a seed in the old system has an impact parameter of 0.0001 mm. In the old system, this creates a weight of 200 − 0.0001 = 199.9999 so it gets cut by the minimum weight of 200 . In this new system, the weights are strictly positive which I think is good; it means that we don't have this "cut-off" effect. But it also means that - for now - the cut is looser. I plan to rework this so we don't end up with way too many seeds.

Thankfully, this issue is easily resolved by requiring one more confirmation seed, i.e. by increasing the minimum weight by $200$.

stephenswat:
With that fix, this is now ready to go!

stephenswat:
I updated this to use a normalized PDF with range $[0,1]$. This has the benefit of giving a more consistent range (I want to move all weights to have range $[0, 1]$) while also requiring fewer floating point operations.

stephenswat:
# Physics performance summary
Here is a summary of the physics performance effects of this PR. Command used:

```
traccc_seeding_example_cuda --input-directory=/data/Acts/odd-simulations-20240506/geant4_ttbar_mu200 --digitization-file=geometries/odd/odd-digi-geometric-config.json --detector-file=geometries/odd/odd-detray_geometry_detray.json --grid-file=geometries/odd/odd-detray_surface_grids_detray.json --use-detray-detector=on --input-events=10 --use-acts-geom-source=on --check-performance --truth-finding-min-track-candidates=5 --truth-finding-min-pt=1.0 --truth-finding-min-z=-150 --truth-finding-max-z=150 --truth-finding-max-r=10 --seed-matching-ratio=0.99 --track-matching-ratio=0.5 --track-candidates-range=5:100 --seedfinder-vertex-range=-150:150
```

## Seeding performance
Total number of seeds went from **336240** to **336227** **(-0.0%)**




## Track finding performance
Total number of found tracks went from **146077** to **146069** **(-0.0%)**




> [!NOTE]

PR #1123, opened 2025-08-11
Remove unnecessary reset of `is_updated_buffer`

beomki-yeo:
`is_updated_buffer` is already reset in `gather_tracks` so removing the overlapping parts


krasznaa (approved):
Looks sane.

PR #1125, opened 2025-08-13
Optimize gather_tracks

beomki-yeo:
This PR significantly reduces the grid-block dimension of `gather_tracks`. The performance improvement is pretty trivial though


stephenswat (approved):
Nice work. 👍

PR #1126, opened 2025-08-14
Remove the iteration in deduplication

beomki-yeo:
Follow up PR of #1095

The following iteration in `remove_tracks` to deduplicate is removed by introducing a global buffer of `track_count`

```c++
        for (unsigned int i = 0; i < pos1; ++i) {
            if (sh_buffer[i] == tid) {
                already_pushed = true;
                break;
            }
        }
```


beomki-yeo:
`remove_tracks` is a single block kernel. There is no desync issue as there is a `syncthreads()`

PR #1127, opened 2025-08-15
Constant Detector, main branch (2025.08.15.)

krasznaa:
As @niermann999 reminded me earlier today, in all our algorithms we access the tracking geometry in device code with non-const pointers. Which is tangentially related to the non-cached access that @stephenswat floated recently.

As I discussed with Joana, right now [detray](https://github.com/acts-project/detray) doesn't actually provide a way to create a "device detector" object that would use const access. It provides typedefs for constant views, but those constant views can not be used in any shape or form with a device detector object with the existing infrastructure.

But as it turns out, one can create a "constant device detector" object with a relatively minimal amount of effort. See the `traccc::details::device_detector_container_types` type introduced by this PR! Using that, plus some relatively manageable amount of changes, the project should now be accessing the data held by Detray, using constant pointers.

Don't think that this will make any difference for the performance of the compiled code, but at least it should make the code a little safer vis-a-vis const correctness.

Of course some of this will need to find its way back into Detray. Since as we discussed with Joana, `detray::detector` should pretty much only ever offer constant versions of itself. As its API doesn't allow for any modification of an existing detector. Not on the host and even less on a device.

### DIFF
--- a/benchmarks/cuda/toy_detector_cuda.cpp
+++ b/benchmarks/cuda/toy_detector_cuda.cpp
@@ -69,7 +69,7 @@ BENCHMARK_DEFINE_F(ToyDetectorBenchmark, CUDA)(benchmark::State& state) {
                                                           async_copy, stream);
 
     // Copy detector to device
-    auto det_buffer = detray::get_buffer(det, device_mr, copy);
+    const auto det_buffer = detray::get_buffer(det, device_mr, copy);
     // Detector view object
     auto det_view = detray::get_data(det_buffer);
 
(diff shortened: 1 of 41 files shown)


krasznaa:
It's also worth saying that most of the changes in this PR were necessary because a constant detector view cannot be created from a non-const one at the moment. (For purely technical reasons.) So I had to finesse the Detray code into producing the right type of view everywhere.

Once we make some updates in Detray with the lessons learned from this PR, it will be a lot less fragile to write user code around detector objects. 🤔

beomki-yeo (approved):
Looks good to me

PR #1134, opened 2025-08-21
Remove reset_status kernel

beomki-yeo:
This PR removes the `reset_status` kernel by improving the logic. There is a performance improvement by around 1-2%


stephenswat (approved):
Looks good to me. 👍

PR #1139, opened 2025-08-27
Add parameter for HS truth particle selection

paradajzblond:
Hello! I have encapsulated "comes from HS process and is stable particle" into a `is_HS` truth parameter selection to have a coherent selection of truth particles between Athena and Traccc. This is currently read from the `process` value in the input "particles-initial" csv file. Happy to discuss if it should be renamed, placed somewhere else etc.

### DIFF
--- a/performance/src/efficiency/finding_performance_writer.cpp
+++ b/performance/src/efficiency/finding_performance_writer.cpp
@@ -264,6 +264,9 @@ void finding_performance_writer::write_common(
             ptc.vertex[2] < m_cfg.truth_config.z_min ||
             ptc.vertex[2] > m_cfg.truth_config.z_max ||
             vector::perp(ptc.vertex) > m_cfg.truth_config.r_max ||
+            std::abs(vector::eta(ptc.momentum)) > m_cfg.truth_config.eta_max ||
+            (m_cfg.truth_config.process_id >= 0 &&
+             m_cfg.truth_config.process_id != ptc.process) ||
             num_measurements < m_cfg.truth_config.min_track_candidates) {
             continue;
         }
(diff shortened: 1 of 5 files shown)


performance/src/efficiency/finding_performance_writer.cpp
```diff
@@ -257,6 +257,7 @@ void finding_performance_writer::write_common(
             ptc.vertex[2] < m_cfg.truth_config.z_min ||
             ptc.vertex[2] > m_cfg.truth_config.z_max ||
             vector::perp(ptc.vertex) > m_cfg.truth_config.r_max ||
+            ptc.process != m_cfg.truth_config.is_HS ||
```
stephenswat, line 260: The problem with this is that it will permanently affect all of our truth filtering, even if we do not want to select only the HS vertex.

To make this more robust, I'd suggest making this `bool is_HS` into a `int process_id`, and then we can use the logic than any negative value (default) selects _any_ process, and a positive ID selects only that process:


```suggestion
            (m_cfg.truth_config.process_id >= 0 && m_cfg.truth_config.process_id == ptc.process)
```

stephenswat (changes requested):

performance/src/efficiency/finding_performance_writer.cpp
```diff
@@ -257,6 +257,7 @@ void finding_performance_writer::write_common(
             ptc.vertex[2] < m_cfg.truth_config.z_min ||
             ptc.vertex[2] > m_cfg.truth_config.z_max ||
             vector::perp(ptc.vertex) > m_cfg.truth_config.r_max ||
+            vector::eta(ptc.momentum) > m_cfg.truth_config.eta_max ||
```
stephenswat, line 260: Does this need to have an absolute operator on it?

PR #1147, opened 2025-09-03
Automatically generate kernel specializations

stephenswat:
Currently, we have manually written specialization files for our CUDA kernels, but this isn't necessary; it's much easier to let a code generation program produce these for us during the build. This commit adds a very simple dependency-less Python script that can do this for us.


CMakeLists.txt
```diff
@@ -85,6 +85,11 @@ if(${TRACCC_BUILD_HIP} AND ${CMAKE_HIP_STANDARD} LESS 20)
    message(SEND_ERROR "CMAKE_HIP_STANDARD=${CMAKE_HIP_STANDARD}, but traccc requires C++>=20")
 endif()
 
+# In order to generate specializations for our kernels, we need to have both
+# a working Python interpreter, and a list of supported detectors.
+find_package (Python COMPONENTS Interpreter REQUIRED)
+set(TRACCC_SUPPORTED_DETECTORS "default_detector;telescope_detector;toy_detector")
```
niermann999, line 91: Should this be cmdline configurable? In the end experiments won't want to compile anything but their own setup, right?

niermann999 (approved):

CMakeLists.txt
stephenswat: One can set this via `-DTRACCC_SUPPORTED_DETECTORS` I think!

paulgessinger:
I think it could also be done with multiple compilations units and the preprocessor. In fact, I think I'm doing something very similar with the gain matrix updater in ACTS that I split by dimension and reuse the same source file.

PR #1151, opened 2025-09-03
Bug Fix: Correct copy typo.

CrossR:
Tiny change, the new polymorphic detector code was using the host copy function, rather than the device one, causing a seg fault when I tried to test this example. It is correct in the other Alpaka examples.


stephenswat (approved):
Good catch, thank you.

PR #1159, opened 2025-09-17
Free the memory properly in `array-wrapper-test`

beomki-yeo:
Reported by @gagnonlg 
This unfreed memory makes the greedy ambiguity resolution tests fail when we run whole tests with `./bin/traccc_test_cuda`


beomki-yeo:
I am afraid that uncleared memory in the wrapper tests might contaminate the memory allocation in ambuiguity resolution test.

Following is not relevant our discussion: Please make sure that the argumenrs in cudaMemset is being used correctly in your tests.

PR #939, opened 2025-04-14
Remove cluster link in measurement edm

beomki-yeo:
This variable is not used anywhere

PR #1161, opened 2025-09-25
Add Cluster Index, main branch (2025.09.25.)

krasznaa:
(Re-)Introduced a cluster index into measurement objects. To make it possible to freely sort the measurements, and still be able to point back at the cluster that belongs to a given measurement.

Technically this is duplication of information. :frowning: Since we use the exact same index as the unique measurement identifier. But since that is really not guaranteed, I believe it makes sense to have two separate variables in the EDM with the same value. :thinking:

@flg, this should in principle be enough to make the G-200 code a lot more robust. Regardless of how the reconstruction code would re-order the measurements.

This is of course a reversal / follow-up of #939.

### DIFF
--- a/tests/cuda/test_ambiguity_resolution.cpp
+++ b/tests/cuda/test_ambiguity_resolution.cpp
@@ -937,8 +937,10 @@ TEST_P(CUDAGreedyResolutionCompareToCPU, Comparison) {
     }
 };
 
+// The following tests are not working for some not fully understood reason.
+// We are blaming the ambiguity resolution algorithm for now.
 INSTANTIATE_TEST_SUITE_P(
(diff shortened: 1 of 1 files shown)


krasznaa:
Interesting. :thinking: I actually saw these CUDA failures on my laptop as well. But I could not understand how they would come from my change. So I concluded that these would be some weird thing not working specifically on my laptop GPU.

Okay, let's check then what the heck is going on... :thinking:

stephenswat:
Possibly related to #1159.

beomki-yeo:
Could you attach a log of crash or failure?

beomki-yeo:
Thanks have you checked what compute-sanitizer says?

beomki-yeo:
Hmm OK... I might need more time to think about this

PR #1171, opened 2025-10-21
feat: Harmonize acts and traccc parameter smearing

niermann999:
Adds the momentum dependent smearing used in acts


niermann999:
Close for now, as we will be moving towards Fatras eventually

PR #1193, opened 2025-11-13
HIP Library, main branch (2025.11.13.)

krasznaa:
This all started out with me trying to see if CoPilot would be able to do all of this automatically / by itself. Unfortunately that all worked very poorly, so I rather did the whole migration by hand in the end. :frowning:

But still, I wanted to start making a `traccc::hip` library finally.

To start things off, I converted `traccc::cuda::clusterization_algorithm` into a `traccc::hip::clusterization_algorithm`. Code-wise it was pretty much a copy-paste, with a `cuda => hip` replacement applied. The only tricky part was to convince the build system to behave more or less reliably.

Which for me specifically it by now does. But I'll need some additional eyes as well. @StewMH, please have a look. Specifically at what ( did to the setup of rocThrust. As it was giving me a world of hurt. :frowning:

For now I didn't do anything specific to the CI setup. But (for me a bit) surprisingly `TRACCC_BUILD_HIP` is already exercised by the `"ALPAKA HIP SYCL"` build. :confused: @StewMH, should we take things a bit more apart? The Alpaka build should really only test builds through Alpaka. I'd rather test the build of "native HIP code" in a separate CI job. Or rather, ideally we'd add 2 new CI jobs. One building the HIP code with its AMD backend, and another one using the NVIDIA backend. As one can very easily break either one of these without breaking the other. :frowning:


StewMH:
Hi Attila,

Great to see this. `TRACCC_BUILD_HIP` is enabled for Alpaka-HIP so that the rocThrust tests are also built. Didn't seem worth having its own build just for this, but now that we have all this code it definitely makes sense to test it properly.

StewMH:
Alpaka and hip tests work fine for me, without getting into bigger issues.

krasznaa:
My intention was to **really** bring this PR up to date. Including splitting the PR's `traccc::hip` library into a separate `traccc::hip_utils` and `traccc::hip` libraries. (Like we have it for CUDA and SYCL.)

But I'm just tired of HIP's shenanigans for today. :frowning: I'm having a really hard time with making all of CMake's own, ROCm's installed cmake code (from `/opt`) and rocThrust cmake code from the build directory, play along nicely. Since on my test machine I have a bunch of different ROCm versions installed. And the latest one is not compatible anymore with the rocThrust version that our project tries to build. But convincing CMake to consistently use not the latest ROCm version is a nightmare as it turns out. :sob:

So I'm stopping at this point. I'm 99% sure that the PR is good like this. But I'll need to run some tests in a simpler build environment to do the final tests.

stephenswat:
> But I'm just tired of HIP's shenanigans for today. 😦

Shall we then merge it in in its current state? If the skeleton is there, we can always iterate and improve after the ACTS integration - would be a shame for this work to go to waste.

krasznaa:
Yeah, I could live with it going in like this. Let's just wait for the Alpaka tests to come back. (In the past they were fine, but it's been a while since the last time that this PR was checked.)

stephenswat (approved):
Let's go with this if the Alpaka CI runs through, then.

PR #1194, opened 2025-11-14
GBTS in throughput examples

m-brettell:
Adding the option to use GBTS in the full_chain_alg -> throughput_mt
Reads in the needed layer linking scheme configuration information from the file

Also adds placeholder configurations to other backends to keep the signature of full_chain_alg's constructor consistent between backends.


krasznaa (dismissed):
Making it possible to use GBTS in the throughput applications would indeed be super helpful. Please do go ahead with this development!

I would prefer to do it differently though. 🤔 Instead of modifying all the existing full chain algorithms, you should just introduce a new algorithm just for CUDA (`traccc::cuda::full_chain_algorithm_gbts`?). And then build a separate set of executables (`traccc_throughput_st_cuda_gbts`, `traccc_throughput_mt_cuda_gbts`) using that full chain algorithm.

Once every language has GBTS available, we could re-visit this. But for now I think this would work better. 🤔

krasznaa:
> What? 😕 I thought we agreed to _reduce_ the number of executables, not increase it. Mark's approach is perfectly reasonable and a lot more maintainable than what you are suggesting.

I want to reduce the number of individual application source files. The throughput applications are very well synchronized in their code already. The new executables will need this code:

```c++
#include "../common/throughput_mt.hpp"

#include "full_chain_algorithm_gbts.hpp"

    return traccc::throughput_mt<traccc::cuda::full_chain_algorithm_gbts>(
        "Multi-threaded CUDA (GBTS) GPU throughput tests", argc, argv);
}
```

And:

```c++
#include "../common/throughput_st.hpp"

#include "full_chain_algorithm_gbts.hpp"

    return traccc::throughput_st<traccc::cuda::full_chain_algorithm_gbts>(
        "Single-threaded CUDA (GBTS) GPU throughput tests", argc, argv);
}
```

Please stop arguing with absolutely everything that I would suggest.

examples/options/src/track_gbts_seeding.cpp
```diff
@@ -0,0 +1,94 @@
...
+    layerInfoFile >> nLayers;
+    layerInfo.reserve(nLayers);
+    char type = 0;
+    std::array<int, 2> info = {0, 0};
+    std::array<float, 2> geo = {0, 0};
+    for (; nLayers > 0u; --nLayers) {
+        layerInfoFile >> type;
+        layerInfoFile >> info[0] >> info[1];
+        layerInfoFile >> geo[0] >> geo[1];
+        layerInfo.addLayer(type, info[0], info[1], geo[0], geo[1]);
```
mimodekjaer: This will add type of ascii '0' and ascii '1', which have the int value of 48 and 49. Which as far as I can see it not intended.

m-brettell: Added cast to fix, thanks.

flg:
It would be really useful to have this merged into main.

m-brettell:
Hi, it appears that it's not possible to add a GBTS config without modifying the existing full_chain_algorithms. All the constructors must have the same signature for the push_back in common/throughput_mt.ipp. Is there a way around this other than having dummy configs (as I do now)? Thanks. [alg push_back](https://github.com/acts-project/traccc/blob/main/examples/run/common/throughput_mt.ipp#L164-173)

flg:
Mentioning https://github.com/acts-project/traccc/pull/1288 to Mark

stephenswat:
> Hi, it appears that it's not possible to add a GBTS config without modifying the existing full_chain_algorithms. All the constructors must have the same signature for the push_back in common/throughput_mt.ipp. Is there a way around this other than having dummy configs (as I do now)? Thanks. [alg push_back](https://github.com/acts-project/traccc/blob/main/examples/run/common/throughput_mt.ipp#L164-173)

Indeed I don't think this is possible, at least not easily. Let's not waste time making this PR more complex now and let's go ahead with it as is, last Friday we all agreed that this PR was good to go as it is.

stephenswat (approved):

PR #1197, opened 2025-11-14
Acts <--> Traccc Track Container, main branch (2025.11.14.)

krasznaa:
I spent the better half of today experimenting with the following. As I told people multiple times in the past, I very much want to figure out how we can put an [Acts::TrackContainer](https://github.com/acts-project/acts/blob/main/Core/include/Acts/EventData/TrackContainer.hpp) on top of the traccc data collections/containers. So that after we copy the reconstructed tracks back to the host, we would not need to translate them anymore. We could just put such an adapter layer on top of them, and voila.

That's the idea anyway. For now I just fumbled around with the code based on what I could find in code written by @paradajzblond and @CarloVarni. So, this PR is just a super rough first version of how I managed to understand the code so far. But I'll definitely want to discuss next week with @CarloVarni and @paulgessinger about the details. 🤔

Even until then, feel very free to criticise this code already. I cannot imagine that I would not have done some silly stuff in it. 🤔


core/src/edm/DeviceMultiTrajectory.cpp
```diff
@@ -0,0 +1,165 @@
...
+    for (unsigned int i = 0; i < Acts::BoundSquareMatrix::RowsAtCompileTime;
+         ++i) {
+        for (unsigned int j = 0; j < Acts::BoundSquareMatrix::ColsAtCompileTime;
+             ++j) {
+            covariance(i, j) = params.covariance()[i][j];
+        }
+    }
+
+    // Construct the covariance expected by Acts.
+    return ConstTrackStateProxy::ConstCovariance{covariance.data()};
```
paulgessinger: So, this will be dangling immediately, right? This is a bit the weakness of the current design that I chose, where the interface expects to be able to hand out references that the caller can manipulate directly. This probably prevents doing on the fly conversion here.

krasznaa: I didn't look at this in detail. Just assumed that the object probably makes a copy. 

It is a defendable design that it doesn't. But then indeed this code will need to be written differently. 🤔

PR #1198, opened 2025-11-17
Remove Kokkos and Futhark code bases

stephenswat:
This PR is a revival of #652, removing the obsolete Kokkos and Futhark code bases.


krasznaa (approved):
As soon as the CI becomes happy, I'll be happy to remove these.

PR #1202, opened 2025-11-18
Alpaka inhomogeneous magnetic field

StewMH:
Implementation of alpaka on-device magnetic field. This either calls the CUDA or SYCL versions, or a HIP implementation written here. In principle this could be factorised out as well.

Seems to give reasonable results, although not identical track parameters.
1)AMD HIP:
```
Using Alpaka device: AMD Instinct MI210 [id: 0]
Reconstructed track parameters: 7107
Time totals:
                  File reading  5 ms
            Warm-up processing  827 ms
              Event processing  7832 ms
Throughput:
            Warm-up processing  82.7008 ms/event, 12.0918 events/s
              Event processing  78.3286 ms/event, 12.7667 events/s
```
2)NVidia CUDA:
```
Using Alpaka device: NVIDIA A100 80GB PCIe [id: 0]
Reconstructed track parameters: 7373
Time totals:
                  File reading  4 ms
            Warm-up processing  429 ms
              Event processing  1843 ms
Throughput:
            Warm-up processing  42.9128 ms/event, 23.3031 events/s
              Event processing  18.4336 ms/event, 54.2489 events/s
```
3)After some delay, Intel (with FP64 emulation):
```
Using Alpaka device: Intel(R) Data Center GPU Flex 170 [id: 0]
Reconstructed track parameters: 7212
Time totals:
                  File reading  4 ms
            Warm-up processing  1090842 ms
              Event processing  11936 ms
Throughput:
            Warm-up processing  109084 ms/event, 0.00916723 events/s
              Event processing  119.369 ms/event, 8.37741 events/s
````

### CODE REVIEW

File: device/alpaka/src/utils/magnetic_field_types.hpp [hunk 1]
```diff
@@ -8,12 +8,50 @@
...
+#elif defined(ALPAKA_ACC_SYCL_ENABLED)
+#include <covfie/sycl/backend/primitive/sycl_device_array.hpp>
+#endif
 
 namespace traccc::alpaka {
 
+/// Inhomogeneous B-field backend type for Alpaka
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+/// Inhomogeneous B-field backend type using CUDA global memory
```
StewMH: This is all copied from the CUDA and SYCL magnetic_field_types: if these were public headers then could avoid some code there. This would also mean that we could move the HIP magnetic field code to `device/hip`.

StewMH:
Think this is ready to merge.

krasznaa (changes requested):
The feature addition in general is great of course. Let's just arrange the code in a better way. 😉

device/alpaka/src/utils/make_magnetic_field.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+// System include(s).
+#include <stdexcept>
+
+namespace traccc::alpaka {
+
+magnetic_field make_magnetic_field(const magnetic_field& bfield,
+                                   [[maybe_unused]] const queue& queue) {
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+    return traccc::cuda::make_magnetic_field(
```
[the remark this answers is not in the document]
StewMH: I started off by making `traccc_cuda_utils` from just splitting out everything in `device/cuda/utils`. I would like it to be `traccc::cuda_utils` but couldn't get that to work so far.

CMakeLists.txt
```diff
@@ -58,8 +58,10 @@ set( TRACCC_DEVICE_LOG_LVL "NONE" CACHE STRING
 
 # Flags controlling which parts of traccc to build.
 option( TRACCC_BUILD_CUDA "Build the CUDA sources included in traccc" FALSE )
+option( TRACCC_BUILD_CUDA_UTILS "Build the CUDA util sources included in traccc" FALSE )
 option( TRACCC_BUILD_HIP "Build the HIP sources included in traccc" FALSE)
 option( TRACCC_BUILD_SYCL "Build the SYCL sources included in traccc" FALSE )
+option( TRACCC_BUILD_SYCL_UTILS "Build the SYCL util sources included in traccc" FALSE )
```
krasznaa: We haven't done such a thing in traccc yet, but this appears to be a good use-case. Let's use [cmake_dependent_option(...)](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html) for setting up these flags.

I have the following in mind:

```c++
option( TRACCC_BUILD_CUDA_UTILS "Build the CUDA utility sources included in traccc" FALSE )
cmake_dependent_option( TRACCC_BUILD_CUDA "Build the CUDA sources included in traccc"
   FALSE TRACCC_BUILD_CUDA_UTILS FALSE )
```

I **think** this would produce the wanted behaviour, but you should experiment a little. Of course what we want is:
  - In case `TRACCC_BUILD_CUDA_UTILS==FALSE`, then `TRACCC_BUILD_CUDA` **must be** `FALSE`;
  - if `TRACCC_BUILD_CUDA_UTILS==TRUE`, then `TRACCC_BUILD_CUDA` **may be** `TRUE` (but is not by default);

And then of course, to make things really fancy, you should make `TRACCC_BUILD_ALPAKA` depend on the utility flags as well. :thinking:

krasznaa (commented):
Let's discuss in Friday's parallelization meeting about this a little. I'm not absolutely sure what will be the most elegant setup for all this code. :thinking:

PR #1204, opened 2025-11-21
Seeding/Sequence Example Rewrite, main branch (2025.11.21.)

krasznaa:
To try to prepare things a bit for the coming HIP code (#1193), this is a proposal for how we could manage the example applications between the different backends.

The PR introduces "device backend" types as part of the examples. Which would provide enough information such that a common templated code could implement the applications. Similar to how we've done it for the throughput applications since forever.

Since the "device backends" are written a bit verbosely, the number of saved lines of code in the PR is not as good as I would've hoped. But remember that adding applications for the HIP backend should become much simpler after this. Plus the maintenance of these applications should also simplify.

Note:
  - I only tackled the "device" "`seq`" and "`seeding`" applications for now. Since the host applications fundamentally do a different thing, I don't think those would fit under this umbrella.
  - There are a couple of CUDA-only applications at the moment that should be possible to extend to all backends with this code structure later on. But I didn't want to attack that in this PR.

Finally: @stephenswat, I tried to implement the new applications such that your existing scripts could switch to these with minimal effort. But I'll need your input on how that should go exactly.


krasznaa:
Let's just retire this one. We'll need to rewrite the examples for Acts in a quite different way. No reason to put in a significant effort into this sort of code harmonization at this point.

PR #1205, opened 2025-11-21
fix: the indices in the host track parameter estimation

niermann999:
Fixes the indices into the track parameters for the host track parameter estimation. This avoids a segmentation fault.
Also changes the access to the track parameters to the getters of the track parameter class to include their assertions.


krasznaa (approved):
Did you check what it does to the output of `traccc_seq_example_cuda` for instance? Hopefully it makes the agreement between the host and device results a whole lot better...

PR #1209, opened 2025-11-21
feat: Build detectors with custom metadata

niermann999:
Build the ODD and ITk detectors with their custom metadata. Also remove the io roundtrip in the tests for the telescope and toy detector, since the custom metadata of the test detectors is already getting compiled and they do not need to be read into the default detector anymore


krasznaa (changes requested):
Just to make your life more difficult.

krasznaa:
You can find my CMake changes here: https://github.com/krasznaa/traccc/commit/6103b5e5c55407acb9fca1353cd82ebdeb46ba62

PR #1212, opened 2025-11-27
Update to alpaka 2.0

StewMH:
Remove dependency on boost. Remains to be seen if we still need to patch printf. Otherwise appears to build cleanly without changes required. I'd like to see the CI run before any more large-scale tests.


StewMH:
Passes tests on AMD locally, CPU in CI.

krasznaa (approved):
:+1:

PR #1213, opened 2025-11-30
fix: Enable the momentum aborter in the Kalman Fitter

niermann999:
Add the momentum aborter to the Kalman Fitter to prevent a segmentation fault in the truth fitter example where the B-field query position is out of bounds


krasznaa (approved):
As long as you sort out the conflict with #1218, this looks okay as well.

PR #1225, opened 2026-01-06
Algorithm Harmonization #1 Clustering, main branch (2026.01.06.)

krasznaa:
Let's start off the new year with some maintenance work. Since the device algorithms for CUDA, SYCL and Alpaka have gone fairly out of sync in the run-up to our internal report, it's time to clean things up a bit.

As a first step I wanted to synchronize how clusterization would be done in our current 3 languages. While trying to figure out a good way in which code could be shared more between the different GPU languages. The proposal in this PR should serve us well for the algorithms that don't use polymorphic inputs. Though for the polymorphic case something slightly more elaborate will have to be put in place. (I have some ideas...)

Through this synchronization the SYCL and Alpaka algorithms are now also providing measurement to cell associations with the interface written originally just for the CUDA algorithm. Tagging @flg, as someone who has been interested in this since a while. (Though I guess by now it will be up to @paradajzblond and us to put such things into the G-200 code.)

Basically I added 2 new classes:
  - `traccc::device::algorithm_base`:To take care of holding on to the necessary VecMem objects in the "device algorithms" in a uniform way;
  - `traccc::device::clusterization_algorithm`: Serving as a base class for all device clusterization algorithms. Leaving 3 virtual functions for the derived classes to implement.
    * `input_is_valid(...)`: Function meant to check whether the input data looks valid. Used in an assertion in `device::clusterization_algorithm`;
    * `ccl_kernel(...)`: Function launching the kernel that would execute `traccc::device::ccl_kernel(...)`;
    * `cluster_maker_kernel(...)`: Function launching the kernel that would execute `traccc::device::reify_cluster_data(...)`.

I ended up doing some small amount of housekeeping along the way, but this you can check out in the diffs.

### DIFF
--- a/device/common/include/traccc/clusterization/device/clusterization_algorithm.hpp
+++ b/device/common/include/traccc/clusterization/device/clusterization_algorithm.hpp
@@ -110,35 +110,41 @@ class clusterization_algorithm
     virtual bool input_is_valid(
         const edm::silicon_cell_collection::const_view& cells) const = 0;
 
+    /// Payload for the @c ccl_kernel function
+    struct ccl_kernel_payload {
+        /// Number of cells in the event
+        unsigned int n_cells;
+        /// The clustering configuration
+        const config_type& config;
+        /// All cells in an event
+        const edm::silicon_cell_collection::const_view& cells;
+        /// The detector description
+        const silicon_detector_description::const_view& det_descr;
+        /// The measurement collection to fill
+        edm::measurement_collection<default_algebra>::view& measurements;
+        /// Buffer for linking cells to measurements
+        vecmem::data::vector_view<unsigned int>& cell_links;
+        /// Buffer for backup of the first element links
+        vecmem::data::vector_view<details::index_t>& f_backup;
+        /// Buffer for backup of the group first element links
+        vecmem::data::vector_view<details::index_t>& gf_backup;
+        /// Buffer for backup of the adjacency matrix (counts)
+        vecmem::data::vector_view<unsigned char>& adjc_backup;
+        /// Buffer for backup of the adjacency matrix (values)
+        vecmem::data::vector_view<details::index_t>& adjv_backup;
+        /// Mutex for the backup structures
+        unsigned int* backup_mutex;
+        /// Buffer for the disjoint set data structure
+        vecmem::data::vector_view<unsigned int>& disjoint_set;
+        /// Buffer for the sizes of the clusters
+        vecmem::data::vector_view<unsigned int>& cluster_sizes;
+    };
+
     /// Main CCL kernel launcher
     ///
-    /// @param num_cells     Number of cells in the event
-    /// @param config        The clustering configuration
-    /// @param cells         All cells in an event
-    /// @param det_descr     The detector description
-    /// @param measurements  The measurement collection to fill
-    /// @param cell_links    Buffer for linking cells to measurements
-    /// @param f_backup      Buffer for backup of the first element links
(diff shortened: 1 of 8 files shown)


flg:
Thanks for this. I was going to test in G-200. Just commenting to ask for an update here once the SYCL queue issue is fixed so that I can resume since I'm not sure I'll receive notifications from vecmem.

At least I'll have the G-200 code ready.

device/common/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -0,0 +1,169 @@
...
+    // Check the input data in debug mode.
+    assert(input_is_valid(cells));
+
+    // Get the number of cells, in an asynchronous way if possible.
+    edm::silicon_cell_collection::const_view::size_type num_cells = 0u;
+    if (mr().host) {
+        const vecmem::async_size size = copy().get_size(cells, *(mr().host));
+        // Here we could give control back to the caller, once our code allows
+        // for it. (coroutines...)
+        num_cells = size.get();
```
krasznaa: At first I was planning to do this separately, after the harmonization was done. But since the CUDA seeding code already does some asynchronous size retrievals, the harominzation PR will need to do this sort of thing there anyway.

So, this is how I propose to retrieve the sizes of containers asynchronously in a language agnostic way. This will replace the existing `traccc::cuda::get_size(...)` function in code shared between the languages.

Pinging @m-fila because of the coroutine implication. :wink:

device/common/src/clusterization/clusterization_algorithm.cpp
```diff
@@ -0,0 +1,169 @@
...
+    const silicon_detector_description::const_view& det_descr,
+    bool keep_disjoint_set) const {
+
+    // Check the input data in debug mode.
+    assert(input_is_valid(cells));
+
+    // Get the number of cells, in an asynchronous way if possible.
+    edm::silicon_cell_collection::const_view::size_type num_cells = 0u;
+    if (mr().host) {
+        const vecmem::async_size size = copy().get_size(cells, *(mr().host));
```
stephenswat: This is a huge performance footgun (and also hard to debug) if the user passes a non-cached pinned memory resource.

device/common/include/traccc/clusterization/device/clusterization_algorithm.hpp
```diff
@@ -0,0 +1,175 @@
...
+        const silicon_detector_description::const_view& det_descr,
+        edm::measurement_collection<default_algebra>::view& measurements,
+        vecmem::data::vector_view<unsigned int>& cell_links,
+        vecmem::data::vector_view<details::index_t>& f_backup,
+        vecmem::data::vector_view<details::index_t>& gf_backup,
+        vecmem::data::vector_view<unsigned char>& adjc_backup,
+        vecmem::data::vector_view<details::index_t>& adjv_backup,
+        unsigned int* backup_mutex,
+        vecmem::data::vector_view<unsigned int>& disjoint_set,
+        vecmem::data::vector_view<unsigned int>& cluster_sizes) const = 0;
```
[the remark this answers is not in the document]
krasznaa: :+1: 

I was considering adding a `traccc::device::ccl_kernel_payload` struct that could be shared between the `traccc::device::ccl_kernel` device function and this algorithm function. But while that is not impossible to do, it's also not super convenient to do. :thinking: So in the end I just added a helper struct specifically for the algorithm(s).

stephenswat (approved):

PR #1251, opened 2026-02-05
Fix incorrect assertion in `build_tracks` kernel

stephenswat:
This commit fixes an incorrect assertion in the `build_tracks` kernel to account for the fact that this kernel can now produce fitted tracks. It also re-enables the check for duplicate measurements on each track.

### DIFF
--- a/device/common/include/traccc/finding/device/impl/build_tracks.ipp
+++ b/device/common/include/traccc/finding/device/impl/build_tracks.ipp
@@ -220,13 +220,15 @@ TRACCC_HOST_DEVICE inline void build_tracks(
 #ifndef NDEBUG
     // Assert that we did not make any duplicate track states.
     for (const auto& i : track.constituent_links()) {
-        assert(i.type == edm::track_constituent_link::measurement);
+        if (run_mbf) {
+            assert(i.type == edm::track_constituent_link::track_state);
+        } else {
+            assert(i.type == edm::track_constituent_link::measurement);
+        }
         for (const auto& j : track.constituent_links()) {
-            assert(j.type == edm::track_constituent_link::measurement);
             if (i.index != j.index) {
-                // TODO: Re-enable me!
-                // assert(measurements.at(i->index).identifier() !=
-                //       measurement.at(j->index).identifier());
+                assert(measurements.at(i.index).identifier() !=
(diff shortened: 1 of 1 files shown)


device/common/include/traccc/finding/device/impl/build_tracks.ipp
```diff
@@ -220,13 +220,15 @@ TRACCC_HOST_DEVICE inline void build_tracks(
 #ifndef NDEBUG
     // Assert that we did not make any duplicate track states.
     for (const auto& i : track.constituent_links()) {
-        assert(i.type == edm::track_constituent_link::measurement);
+        if (run_mbf) {
+            assert(i.type == edm::track_constituent_link::track_state);
+        } else {
+            assert(i.type == edm::track_constituent_link::measurement);
+        }
```
krasznaa: Curious that none of the compilers were unhappy about this. I could've imagined some of them claiming that in optimized mode this would be "dead code". 🤔 

Doing this all inside of `assert(...)` would be a little bit better.

```c++
assert(((run_mbf == true) && (i.type == ...)) || ((run_mbf == false) && (i.type == ...)));
```

device/common/include/traccc/finding/device/impl/build_tracks.ipp
```diff
@@ -220,13 +220,15 @@ TRACCC_HOST_DEVICE inline void build_tracks(
...
         for (const auto& j : track.constituent_links()) {
-            assert(j.type == edm::track_constituent_link::measurement);
             if (i.index != j.index) {
-                // TODO: Re-enable me!
-                // assert(measurements.at(i->index).identifier() !=
-                //       measurement.at(j->index).identifier());
+                assert(measurements.at(i.index).identifier() !=
+                       measurements.at(j.index).identifier());
```
krasznaa: But if we use MBF, this could potentially lead to invalid memory access, no? 🤔

device/common/include/traccc/finding/device/impl/build_tracks.ipp
stephenswat: Hard disagree, that's _much_ less clear. :stuck_out_tongue:

stephenswat:
Closing in favour of #1311.

PR #1253, opened 2026-02-05
Harmonize theta pole check in Kálmán updater

stephenswat:
The gain matrix updater currently uses two slightly different methods for checking for theta poles. This means that we sometimes fire an assertion even if the code for gracefully exiting in these cases did not fire. This commit harmonizes the pole checks so that they produce identical results. In the process, it deduplicates the checks to save on compute.


stephenswat:
Closing in favour of #1311.

PR #1264, opened 2026-02-20
Cell Sorting, main branch (2026.02.20.)

krasznaa:
After earlier discussions about how fast we can be with sorting cells as part of the throughput measurements, I spent some time in putting up some code for this.

I introduced "cell sorting algorithms" for all backends. In pretty much the same way in which the measurement sorting algorithms are implemented.

Then I taught `traccc::io::read_cells(...)` how to randomize the order of the cells on request. I did it like this because the CSV reading code is fundamentally set up such that it would output a sorted vector of cells. Instead of completely re-thinking the logic of the I/O code, it was easier to add a shuffling step at the end. (When the user asks for it.)

Finally I updated the throughput measurement applications to:
  - shuffle the cells that they read into host memory;
  - make use of the appropriate cell sorting algorithm as part of their data processing.

Unfortunately the result is slightly worse than what I was hoping for. :frowning: With the current `main` branch I see the following (reference) throughput on our trusty ol' A5000:

```
[bash][pcadp04]:traccc > ./build_current/bin/traccc_throughput_mt_cuda --input-directory /data/Acts/odd-simulations-20240509/geant4_ttbar_mu200/ --input-events=20 --track-candidates-range=5:100 --seedfinder-vertex-range=-150:150 --finding-run-mbf-smoother=false --processed-events=500 --deterministic --cpu-threads=8
...
04:34:29 PM ThroughputExample             INFO      Reconstructed track parameters: 2622220
04:34:29 PM ThroughputExample             INFO      Time totals:                   File reading  1249 ms
04:34:29 PM ThroughputExample             INFO                  Warm-up processing  153 ms
04:34:29 PM ThroughputExample             INFO                    Event processing  5087 ms
04:34:29 PM ThroughputExample             INFO      Throughput:            Warm-up processing  15.3186 ms/event, 65.2802 events/s
04:34:29 PM ThroughputExample             INFO                    Event processing  10.1754 ms/event, 98.2765 events/s
[bash][pcadp04]:traccc >
```

While when I add an extra sorting step, I get:

```
[bash][pcadp04]:traccc > ./build_new/bin/traccc_throughput_mt_cuda --input-directory /data/Acts/odd-simulations-20240509/geant4_ttbar_mu200/ --input-events=20 --track-candidates-range=5:100 --seedfinder-vertex-range=-150:150 --finding-run-mbf-smoother=false --processed-events=500 --deterministic --cpu-threads=8
...
04:35:53 PM ThroughputExample             INFO      Reconstructed track parameters: 2622229
04:35:53 PM ThroughputExample             INFO      Time totals:                   File reading  975 ms
04:35:53 PM ThroughputExample             INFO                  Warm-up processing  161 ms
04:35:53 PM ThroughputExample             INFO                    Event processing  5510 ms
04:35:53 PM ThroughputExample             INFO      Throughput:            Warm-up processing  16.1115 ms/event, 62.0676 events/s
04:35:53 PM ThroughputExample             INFO                    Event processing  11.0217 ms/event, 90.7302 events/s
[bash][pcadp04]:traccc >
```

So the cell sorting adds almost an entire millisecond to the event processing. :frowning: Way more than I was expecting...

I didn't do any deeper profiling on the sorting code. It's not impossible that it could still be improved. And it's also worth remembering that the random shuffling of the cells that the code does is a much worse scenario than what we would ever get from real data. Even under the least ideal circumstances.

Still, I was hoping for a quicker sorting, even with all this taken into account. :thinking:

Pinging @flg, @paradajzblond.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -1,6 +1,6 @@
 # TRACCC library, part of the ACTS project (R&D line)
 #
-# (c) 2021-2025 CERN for the benefit of the ACTS project
+# (c) 2021-2026 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
@@ -78,6 +78,8 @@ traccc_add_library( traccc_core core TYPE SHARED
   "src/clusterization/measurement_sorting_algorithm.cpp"
   "include/traccc/clusterization/clusterization_algorithm.hpp"
   "src/clusterization/clusterization_algorithm.cpp"
+  "include/traccc/clusterization/silicon_cell_sorting_algorithm.hpp"
+  "src/clusterization/silicon_cell_sorting_algorithm.cpp"
   # Finding algorithmic code
   "include/traccc/finding/candidate_link.hpp"
   "include/traccc/finding/finding_config.hpp"
--- a/core/include/traccc/clusterization/silicon_cell_sorting_algorithm.hpp
+++ b/core/include/traccc/clusterization/silicon_cell_sorting_algorithm.hpp
@@ -1,0 +1,59 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2026 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Library include(s).
+#include "traccc/edm/silicon_cell_collection.hpp"
+#include "traccc/utils/algorithm.hpp"
+#include "traccc/utils/messaging.hpp"
+
+// VecMem include(s).
+#include <vecmem/memory/memory_resource.hpp>
+
+// System include(s).
+#include <functional>
+#include <memory>
+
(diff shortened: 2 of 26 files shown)


krasznaa:
This latest version of the code, which only shuffles cells within the same module, runs like this:

```
05:24:07 PM ThroughputExample             INFO      Reconstructed track parameters: 2622230
05:24:07 PM ThroughputExample             INFO      Time totals:                   File reading  1034 ms
05:24:07 PM ThroughputExample             INFO                  Warm-up processing  158 ms
05:24:07 PM ThroughputExample             INFO                    Event processing  5237 ms
05:24:07 PM ThroughputExample             INFO      Throughput:            Warm-up processing  15.8187 ms/event, 63.2164 events/s
05:24:07 PM ThroughputExample             INFO                    Event processing  10.4753 ms/event, 95.4628 events/s
```

So Thrust's sorting, as expected, is quite a bit quicker in this case.

PR #1285, opened 2026-04-01
ref: Move Kalman smoother into CKF

niermann999:
Move the Kalman smoother into the CKF and bypass the KF foward fit


niermann999:
This has gone so much out of scope that it will need a reimplementation, which can be done just as well in ACTS

PR #1287, opened 2026-04-08
EDM De-Algebraification #1, main branch (2026.04.08.)

krasznaa:
With this let me start yet another big rewrite / redesign. :thinking:

It's been bugging me for a long time that the data model we use depends on the algebra that we use. Which in my mind is not a workable API. The data itself needs to be fixed somehow.

So I started by making `traccc::edm::silicon_cell` and `traccc::edm::measurement` not be dependent on the build setup of the project. As these seemed technically the easiest to do, and since `edm::measurement` is used pretty much everywhere, I thought that this would be a good stress-test for the rewrite.

Making `edm::silicon_cell` always use `float` values for its "activation" and "time" properties was a trivial change, that didn't actually necessitate any code changes. So that was surprisingly easy.

But of course `edm::measurement` was a bit more tricky. What we do currently is that we use the `detray::point2D` type for storing the local coordinates and the variances of those local coordinates. But I believe it's a lot better idea to just store plain `std::array<float, 2>` objects. Independent of what algebra we want to use for our calculations.

To help with this, I added a couple of additional functions to `edm::measurement`. Which would help with both retrieving and setting the `local_position` and `local_variance` variables using Detray objects directly. Just so that the actual algorithmic code would look a bit more "natural". Like:

```c++
    const point3 global = sf.local_to_global(
        gctx,
        meas.template local_position_in<typename detector_t::algebra_type>(),
        {});
```

(Okay, maybe "natural" is a bit over-selling it. :thinking:)

I see this all as one step towards hiding the whole algebra business behind the scenes. As the API of the GPU tools should really not reveal what algebra is being used underneath. In case we end up using array and Eigen algebras interchangeably in the end after all. :thinking:

I assume there will be plenty of discussion around this. Especially around what we'll do with `edm::track_state` afterwards. (As there I do have some worries about what we actually want to do.) But let's see how this goes...

### DIFF
--- a/core/include/traccc/edm/impl/measurement_collection.ipp
+++ b/core/include/traccc/edm/impl/measurement_collection.ipp
@@ -1,6 +1,6 @@
 /** TRACCC library, part of the ACTS project (R&D line)
  *
- * (c) 2025 CERN for the benefit of the ACTS project
+ * (c) 2025-2026 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -12,6 +12,59 @@
 #include "traccc/definitions/math.hpp"
 
 namespace traccc::edm {
+
+template <typename BASE>
+template <detray::concepts::algebra ALGEBRA_TYPE>
+TRACCC_HOST_DEVICE detray::dpoint2D<ALGEBRA_TYPE>
+measurement<BASE>::local_position_in() const {
+
+    detray::dpoint2D<ALGEBRA_TYPE> result;
+    getter::element(result, 0) =
+        static_cast<typename ALGEBRA_TYPE::value_type>(local_position()[0]);
+    getter::element(result, 1) =
+        static_cast<typename ALGEBRA_TYPE::value_type>(local_position()[1]);
+    return result;
+}
+
+template <typename BASE>
+template <detray::concepts::algebra ALGEBRA_TYPE>
+TRACCC_HOST_DEVICE void measurement<BASE>::set_local_position_in(
+    const detray::dpoint2D<ALGEBRA_TYPE>& pos) {
+
+    local_position()[0] = static_cast<float>(getter::element(pos, 0));
+    local_position()[1] = static_cast<float>(getter::element(pos, 1));
+}
+
+template <typename BASE>
+template <detray::concepts::algebra ALGEBRA_TYPE>
+TRACCC_HOST_DEVICE detray::dpoint2D<ALGEBRA_TYPE>
+measurement<BASE>::local_variance_in() const {
+
+    detray::dpoint2D<ALGEBRA_TYPE> result;
+    getter::element(result, 0) =
+        static_cast<typename ALGEBRA_TYPE::value_type>(local_variance()[0]);
+    getter::element(result, 1) =
+        static_cast<typename ALGEBRA_TYPE::value_type>(local_variance()[1]);
+    return result;
+}
+
+template <typename BASE>
+template <detray::concepts::algebra ALGEBRA_TYPE>
+TRACCC_HOST_DEVICE void measurement<BASE>::set_local_variance_in(
+    const detray::dpoint2D<ALGEBRA_TYPE>& var) {
+
+    local_variance()[0] = static_cast<float>(getter::element(var, 0));
+    local_variance()[1] = static_cast<float>(getter::element(var, 1));
+}
+
+template <typename BASE>
+template <std::integral TYPE>
+TRACCC_HOST_DEVICE void measurement<BASE>::set_subspace(
+    const std::array<TYPE, 2u>& subs) {
+
+    subspace()[0] = static_cast<unsigned int>(subs[0]);
+    subspace()[1] = static_cast<unsigned int>(subs[1]);
+}
 
 template <typename BASE>
 template <typename T>
--- a/core/include/traccc/edm/impl/measurement_collection.ipp
+++ b/core/include/traccc/edm/impl/measurement_collection.ipp
@@ -14,56 +14,12 @@
 namespace traccc::edm {
 
 template <typename BASE>
-template <detray::concepts::algebra ALGEBRA_TYPE>
-TRACCC_HOST_DEVICE detray::dpoint2D<ALGEBRA_TYPE>
-measurement<BASE>::local_position_in() const {
-
-    detray::dpoint2D<ALGEBRA_TYPE> result;
-    getter::element(result, 0) =
-        static_cast<typename ALGEBRA_TYPE::value_type>(local_position()[0]);
-    getter::element(result, 1) =
-        static_cast<typename ALGEBRA_TYPE::value_type>(local_position()[1]);
-    return result;
-}
-
-template <typename BASE>
-template <detray::concepts::algebra ALGEBRA_TYPE>
-TRACCC_HOST_DEVICE void measurement<BASE>::set_local_position_in(
-    const detray::dpoint2D<ALGEBRA_TYPE>& pos) {
-
-    local_position()[0] = static_cast<float>(getter::element(pos, 0));
-    local_position()[1] = static_cast<float>(getter::element(pos, 1));
-}
-
-template <typename BASE>
-template <detray::concepts::algebra ALGEBRA_TYPE>
-TRACCC_HOST_DEVICE detray::dpoint2D<ALGEBRA_TYPE>
-measurement<BASE>::local_variance_in() const {
-
-    detray::dpoint2D<ALGEBRA_TYPE> result;
-    getter::element(result, 0) =
-        static_cast<typename ALGEBRA_TYPE::value_type>(local_variance()[0]);
-    getter::element(result, 1) =
-        static_cast<typename ALGEBRA_TYPE::value_type>(local_variance()[1]);
-    return result;
-}
-
-template <typename BASE>
-template <detray::concepts::algebra ALGEBRA_TYPE>
-TRACCC_HOST_DEVICE void measurement<BASE>::set_local_variance_in(
-    const detray::dpoint2D<ALGEBRA_TYPE>& var) {
-
-    local_variance()[0] = static_cast<float>(getter::element(var, 0));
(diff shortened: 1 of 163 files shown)


core/include/traccc/edm/impl/measurement_helpers.ipp
```diff
@@ -55,7 +55,8 @@ TRACCC_HOST_DEVICE void get_measurement_covariance(
     assert((meas.subspace()[0] == e_bound_loc0) ||
            (meas.subspace()[0] == e_bound_loc1));
 
-    const variance2& variance = meas.local_variance();
+    const detray::dpoint2D<detray::array<float>>& variance =
+        meas.local_variance();
```
krasznaa: Ahh, these I did last week, before introducing `local_position_in` / `local_variance_in`... To be updated.

core/include/traccc/edm/measurement_collection.hpp
```diff
@@ -60,6 +59,28 @@ class measurement : public BASE {
...
 
+    /// Get the local position object in a specific (Detray) algebra
+    ///
+    /// @note This function must only be used on proxy objects, not on
+    ///       containers!
+    ///
+    /// @tparam ALGEBRA_TYPE The algebra to retrieve the local position in
+    /// @return An algebra specific 1D/2D point object
+    ///
+    template <typename ALGEBRA_TYPE>
```
krasznaa: Forgot to use `detray::concepts::algebra` here... 😦

core/include/traccc/utils/subspace.hpp
```diff
@@ -41,13 +41,12 @@ struct subspace {
     /// Construct from a container of axis indices.
     ///
     /// @param indices Unique, ordered indices
-    TRACCC_HOST_DEVICE
-    constexpr subspace(const std::array<size_type, kSize>& indices) {
+    template <typename SIZE_TYPE>
+    TRACCC_HOST_DEVICE constexpr subspace(
+        const std::array<SIZE_TYPE, kSize>& indices) {
```
stephenswat, line 46: Why would we not just use the `size_type` that is already defined in the class?

core/include/traccc/edm/measurement_collection.hpp
```diff
@@ -150,6 +193,18 @@ class measurement : public BASE {
...
+    /// Using a possibly slightly different array than the one used by the
+    /// object itself.
+    ///
+    /// @note This function must only be used on proxy objects, not on
+    ///       containers!
+    ///
+    /// @param subs The subspace to set
+    ///
+    template <std::integral TYPE>
+    TRACCC_HOST_DEVICE void set_subspace(const std::array<TYPE, 2u>& subs);
```
stephenswat, line 163: Rather than templating this function on the size type of subspace, callers of this function should rather ignore the size type of their local algebra and simply use `subspace::size_type`.

core/include/traccc/edm/measurement_collection.hpp
```diff
@@ -73,6 +94,28 @@ class measurement : public BASE {
...
+    ///
+    /// @note This function must only be used on proxy objects, not on
+    ///       containers!
+    ///
+    /// @tparam ALGEBRA_TYPE The algebra to set the local variance in
+    /// @param var The local variance to set
+    ///
+    template <detray::concepts::algebra ALGEBRA_TYPE>
+    TRACCC_HOST_DEVICE void set_local_variance_in(
+        const detray::dpoint2D<ALGEBRA_TYPE>& var);
```
stephenswat: Embedding all of this conversion logic into this type is poor API design. It would be nicer to introduce a few free functions which can convert a `detray::dpoint2D` from any given algebra to a canonical standard type, and it should then be the responsibility of the call site to ensure that they produce a `dpoint2D` of the right type.

core/include/traccc/edm/impl/measurement_collection.ipp
```diff
@@ -13,6 +13,59 @@
...
+    getter::element(result, 0) =
+        static_cast<typename ALGEBRA_TYPE::value_type>(local_variance()[0]);
+    getter::element(result, 1) =
+        static_cast<typename ALGEBRA_TYPE::value_type>(local_variance()[1]);
+    return result;
+}
+
+template <typename BASE>
+template <detray::concepts::algebra ALGEBRA_TYPE>
+TRACCC_HOST_DEVICE void measurement<BASE>::set_local_variance_in(
```
stephenswat: Case in point, this is all just duplicated code.

stephenswat (commented):
Removing the algebra from the EDM is largely uncontroversial, so I'm on board with this. The only real issue with this PR is that it tasks the EDM with the responsibility of type conversions where that should really be the callers' responsibility.

core/include/traccc/utils/subspace.hpp
krasznaa: `size_type` in this class is actually algebra specific. For `std::array` it's `std::size_t`. For some other algebras it would be `unsigned int`.

And I'm here wondering that we're wasting a **lot** of memory on this. For now I went with `unsigned int` in `edm::measurement`. But even `std::uint8_t` should be enough I reckon. :thinking:

core/include/traccc/edm/measurement_collection.hpp
krasznaa: The whole reason that algebras advertise their `size_type` is that we had issues with using `std::size_t` versus `unsigned int` with the various algebras. With compilers freaking out about type conversions.

core/include/traccc/utils/subspace.hpp
stephenswat: Absolutely yes, this could be a smaller integer type!

core/include/traccc/edm/measurement_collection.hpp
krasznaa: I guess we can have a free-standing function that can convert between whatever `dpoint2D` and `std::array<float, 2>`. But I'm not quite as convinced as you that it's hurtful to have the EDM class know exactly what type it itself has. For me this seemed like a quite good place to put such a conversion in.

stephenswat: The free function approach avoids quite a lot of awkwardness at both the call site (no need to call a templated member function) and the EDM (no need to define separate getters and setters, no need to duplicate conversion code, no need to include the algebra in a template variable). 

This EDM design has four functions for each `dpoint2D` member, where only two are really necessary (both of which are _not_ templated):

```cpp
    TRACCC_HOST_DEVICE
    auto& local_position() { return BASE::template get<0>(); }

    TRACCC_HOST_DEVICE
    const auto& local_position() const { return BASE::template get<0>(); }

    template <detray::concepts::algebra ALGEBRA_TYPE>
    TRACCC_HOST_DEVICE detray::dpoint2D<ALGEBRA_TYPE> local_position_in() const;

    template <detray::concepts::algebra ALGEBRA_TYPE>
    TRACCC_HOST_DEVICE void set_local_position_in(
        const detray::dpoint2D<ALGEBRA_TYPE>& pos);
```

krasznaa: I'm curious about your proposal. :confused: What exact arguments would those free-standing functions use?

If we want to make use of `detray::dpoint2D`, I don't think there's a way to avoid specifying explicitly which algebra we're using. As it's not possible to figure out purely from `dpoint2D<T>` what `T` should be. And I also would like to avoid using explicit Eigen, etc. types directly in the traccc code.

So, what did you have in mind exactly? :thinking:

stephenswat: I'm suggesting we could have the following free functions:

```cpp
template<typename T>
std::array<float, 2u> to_canonical_point2d(const T & p) requires detray::concepts::point<T> {
    ...
}

template<typename Algebra>
detray::dpoint2D<Algebra> from_canonical_point2d(const std::array<float, 2u> & p) {
    ...
}
```

krasznaa: Since there were some pre-existing helper functions in `measurement_helpers.hpp`, I added functions there in the same style as those already existing things.  It did not make an amazing API though. 😦 

Still, client code will not need to be exposed to this type of code...

core/include/traccc/edm/measurement_helpers.hpp
```diff
@@ -14,6 +14,18 @@
...
+///
+/// @tparam algebra_t The algebra type used to describe the tracks
+///
+/// @param meas The measurement to extract the local position from
+/// @param pos The 2D point to fill with the local position of the measurement
+///
+template <detray::concepts::algebra algebra_t, typename measurement_backend_t>
+TRACCC_HOST_DEVICE void get_measurement_local(
+    const edm::measurement<measurement_backend_t>& meas,
+    detray::dpoint2D<algebra_t>& pos);
```
stephenswat: Is there a good reason why this cannot return the dpoint2D by value?

core/include/traccc/edm/measurement_helpers.hpp
```diff
@@ -29,6 +41,18 @@ TRACCC_HOST_DEVICE void get_measurement_local(
...
+///
+/// @tparam algebra_t The algebra type used to describe the tracks
+///
+/// @param meas The measurement to extract the local position from
+/// @param pos The 2D vector to fill with the local variance of the measurement
+///
+template <detray::concepts::algebra algebra_t, typename measurement_backend_t>
+TRACCC_HOST_DEVICE void get_measurement_variance(
+    const edm::measurement<measurement_backend_t>& meas,
+    detray::dvector2D<algebra_t>& var);
```
stephenswat: Same here.

stephenswat (commented):
Looks much nicer, thanks.

core/include/traccc/edm/measurement_helpers.hpp
krasznaa: I'm trying to balance between writing code in the style of existing code, and writing it in a way that I would like. 🤔 But indeed, while the existing API makes sense for variable sized matrices, it's indeed not great for fixed sized vectors.

stephenswat (approved):
Looks good, let's go with this. 👍

PR #1288, opened 2026-04-08
Add GBTS command-line options and config reading

stephenswat:
This commit lifts the non-controversial parts out of #1194 so that we can unblock the developments of Fabrice, Mikkel, and Mark.

### DIFF
--- a/examples/options/include/traccc/options/track_gbts_seeding.hpp
+++ b/examples/options/include/traccc/options/track_gbts_seeding.hpp
@@ -1,0 +1,47 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2022-2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/gbts_seeding/gbts_seeding_config.hpp"
+#include "traccc/options/details/interface.hpp"
+
+// System include(s).
+#include <cstddef>
+
+namespace traccc::opts {
+
+/// Option(s) for multi-threaded code execution
+class track_gbts_seeding : public interface {
+
+    public:
+    /// @name Options
+    /// @{
+
+    bool m_use_gbts = false;
+    std::string m_config_dir = "DEFAULT";
+
+    /// @}
+    // config info from file
+    std::vector<std::pair<uint64_t, short>> barcodeBinning;
+    std::vector<std::pair<int, std::vector<int>>> binTables;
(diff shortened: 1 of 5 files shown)


examples/options/include/traccc/options/track_gbts_seeding.hpp
```diff
@@ -0,0 +1,47 @@
...
+    /// @{
+
+    bool useGBTS = false;
+    std::string config_dir = "DEFAULT";
+
+    /// @}
+    // config info from file
+    std::vector<std::pair<uint64_t, short>> barcodeBinning;
+    std::vector<std::pair<int, std::vector<int>>> binTables;
+    traccc::device::gbts_layerInfo layerInfo;
```
krasznaa, line 33: Would it not be better to just have a


```suggestion
    /// Configuration object
    gbts_seedfinder_config config;
```

object in there? Then we could easily add more command line options later on, as needed. In case we'd want to override even more options from the command line. 🤔

Though I don't claim to properly understand the code organization at this point.

examples/options/src/track_gbts_seeding.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+namespace traccc::opts {
+
+track_gbts_seeding::track_gbts_seeding() : interface("GBTS Options") {
+    m_desc.add_options()("useGBTS",
+                         boost::program_options::bool_switch(&useGBTS),
+                         "use gbts algorithm");
+
+    m_desc.add_options()(
+        "gbts_config_dir",
```
krasznaa, line 26: We use dashes and not underscores in the existing command line options.


```suggestion
        "gbts-config-dir",
```

examples/options/src/track_gbts_seeding.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+// System include(s).
+#include <filesystem>
+#include <fstream>
+#include <stdexcept>
+
+namespace traccc::opts {
+
+track_gbts_seeding::track_gbts_seeding() : interface("GBTS Options") {
+    m_desc.add_options()("useGBTS",
```
krasznaa, line 21: And then, to be a bit more consistent with the other options, I would propose:


```suggestion
    m_desc.add_options()("use-gbts",
```

krasznaa (commented):
We will need to have a discussion at one point about how we would handle configuration (files) in the future. This PR can go ahead without that, but I'll be really interested how this is imagined to work in a production system later on. 🤔

stephenswat:
Closing as #1194 is now merged.

PR #1290, opened 2026-04-10
EDM De-Algebraification #2, main branch (2026.04.10.)

krasznaa:
Before turning to any of the actually difficult EDM changes, let's make `edm::spacepoint` independent of the underlying Detray algebra first. Since this one is a fairly simple operation.

This follows on from #1287.


stephenswat (approved):
Looks harmless.

PR #1293, opened 2026-04-13
make examples a shared lib

paradajzblond:
Updating examples to be shared library. This enables the usage of, for example, the full chain algorithm in ACTS examples


examples/run/alpaka/CMakeLists.txt
```diff
@@ -23,23 +23,24 @@ set(LIBRARIES vecmem::core traccc::io traccc::performance
     traccc::options traccc_examples_common ${EXTRA_LIBS})
 set(DETRAY detray::io detray::detectors)
 
-traccc_add_executable( seq_example_alpaka "seq_example_alpaka.cpp"
-    LINK_LIBRARIES ${LIBRARIES} ${DETRAY} )
-traccc_add_executable( seeding_example_alpaka "seeding_example_alpaka.cpp"
-    LINK_LIBRARIES ${LIBRARIES} )
```
krasznaa, line 29: I didn't think that these would need to be moved.

It's not the end of the world that you did. But I thought that `src/` could as well have been kept for the sources that we want to include "in the library". 🤔

That way only the library declaration would need to be updated in the build configuration.

krasznaa (commented):
The code changes all look good. But I could imagine the CMake changes simplified a bit. By leaving the source files of the executables where they are now.

If you really dislike having them there, you could move those sources into a final subdirectory called let's say "`apps/`". But keeping them separate from the library source(s) seems like a good idea. 🤔

examples/run/alpaka/CMakeLists.txt
krasznaa: Having just now lost a pretty long comment, let me give the short answer now... Let's do the following:

```
alpaka/
  |-- include/traccc/examples/alpaka/
  |     |-- full_chain_algorithm.hpp
  |-- src/
  |     |-- full_chain_algorithm.cpp
  |-- apps/
        |-- seeding_example_alpaka.cpp
        |-- seq_example_alpaka.cpp
...
```

paradajzblond: great! thank you.

niermann999 (approved):
Looks good to me. Can we move ahead with this soon (it seems to block the traccc/detray update in ACTS)?

cmake/traccc-compiler-options-cpp.cmake
```diff
@@ -23,8 +23,10 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wextra" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wshadow" )
    traccc_add_flag( CMAKE_CXX_FLAGS "-Wunused-local-typedefs" )
-   traccc_add_flag( CMAKE_CXX_FLAGS "-pedantic" )
+   traccc_add_flag( CMAKE_CXX_FLAGS "-Wpedantic" )
```
krasznaa, line 26: Just to be pedantic myself, `-pedantic` and `-Wpedantic` mean the exact same thing for GCC (https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html), and as far as I can tell, for Clang as well (https://clang.llvm.org/docs/DiagnosticsReference.html#wpedantic). Though the latter's documentation is a bit less clear on this. 🤔

In ATLAS offline we use `-pedantic`.

https://gitlab.cern.ch/atlas/atlasexternals/-/blob/main/Build/AtlasCMake/modules/AtlasCompilerSettings.cmake?ref_type=heads#L150

But since Acts indeed uses `-Wpedantic`, this change is indeed well founded. Even if a bit unnecessary. 😛

examples/run/alpaka/src/full_chain_algorithm.cpp
```diff
@@ -6,7 +6,7 @@
  */
 
 // Local include(s).
-#include "full_chain_algorithm.hpp"
+#include "traccc/examples/full_chain_algorithm.hpp"
```
krasznaa: Unfortunately you didn't read my earlier comment carefully. 😦 This needs to go into an "Alpaka specific" directory.

In the current setup, as soon as we build traccc with "multiple languages", it becomes random which of the implementations ends up as `traccc/examples/full_chain_algorithm.hpp`. This **must** be `traccc/examples/alpaka/full_chain_algorithm.hpp` instead.

That should be easy/quick to fix.

krasznaa (changes requested):

examples/run/alpaka/src/full_chain_algorithm.cpp
paradajzblond: indeed i did not spot this! should be fixed now :)

examples/run/alpaka/src/throughput_mt.cpp
```diff
@@ -6,9 +6,9 @@
  */
 
 // Local include(s).
-#include "../common/throughput_mt.hpp"
+#include "traccc/examples/throughput_mt.hpp"
 
-#include "full_chain_algorithm.hpp"
+#include "traccc/examples/full_chain_algorithm.hpp"
```
krasznaa: The CI should still fail on a few things unfortunately...

examples/run/cpu/apps/misaligned_truth_fitting_example.cpp
```diff
@@ -6,10 +6,10 @@
  */
 
 // Project include(s).
```
krasznaa, line 8: Only sources that "belong to the library" should be in `src/`. What I recommended yesteday was that application sources should be in `apps/`.

This I will not necessarily block the PR over. But this layout is still not what I had in mind...

krasznaa (approved):
Looks good.

PR #1297, opened 2026-04-16
Fix for alpaka after #1272

StewMH:
Fixes the throughput executables on CPU. Have yet to test with CUDA/HIP backends.

Add a quick run of the throughput example to the unit tests. Possibly this could be done better in some way? I think it's useful though.

### DIFF
--- a/examples/run/alpaka/CMakeLists.txt
+++ b/examples/run/alpaka/CMakeLists.txt
@@ -46,3 +46,6 @@ traccc_add_executable( throughput_mt_alpaka "throughput_mt.cpp"
 
 add_test( NAME throughput_st_alpaka
           COMMAND traccc_throughput_st_alpaka --finding-run-mbf-smoother=0 --read-bfield-from-file --cold-run-events=1 --processed-events=1)
+
+add_test( NAME throughput_mt_alpaka
+          COMMAND traccc_throughput_mt_alpaka --finding-run-mbf-smoother=0 --read-bfield-from-file --cold-run-events=1 --processed-events=1)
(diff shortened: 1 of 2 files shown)


stephenswat (approved):

examples/run/alpaka/CMakeLists.txt
```diff
@@ -43,3 +43,6 @@ traccc_add_executable( throughput_st_alpaka "throughput_st.cpp"
 
 traccc_add_executable( throughput_mt_alpaka "throughput_mt.cpp"
    LINK_LIBRARIES TBB::tbb indicators::indicators ${LIBRARIES} ${DETRAY} traccc_examples_alpaka )
+
+add_test( throughput_st_alpaka
+                 ${CMAKE_BINARY_DIR}/bin/traccc_throughput_st_alpaka --finding-run-mbf-smoother=0 --read-bfield-from-file --cold-run-events=1 --processed-events=1)
```
krasznaa: Let's rather use the "modern form" of `add_test(...)`:

https://cmake.org/cmake/help/latest/command/add_test.html

I.e. please specify `NAME` and `COMMAND` separately.

Then, rather than specifying the absolute path name of the executable, please rather just use the target name. `add_test(...)` is smart enough to recognize a target name as the first member of a `COMMAND`. So that if we change the layout of our build directory, the test doesn't break.

krasznaa (changes requested):

examples/run/alpaka/CMakeLists.txt
StewMH: Done, thanks!

StewMH:
Is it worth adding the other throughput executables?

krasznaa:
> Is it worth adding the other throughput executables?

I was thinking about that. I was just afraid of the CI possibly trying to run these tests in cases where it shouldn't.

But as you explained on Mattermost, in principle the CI's setup should be smart enough about this already. (We only run CUDA/SYCL/etc. runtime tests on GPU holding runners.)

So... Could you indeed add such `add_test(...)` calls to CUDA and SYCL as well? 😔

krasznaa (approved):
But even without the CUDA and SYCL additions, this looks good to me now. 👍

PR #1298, opened 2026-04-19
Fix CUDA greedy ambiguity resolver correctness bug

SorinBetisor:
# Summary

This PR fixes correctness bugs in the CUDA greedy ambiguity resolver and re-enables the previously disabled CPU-vs-CUDA tests.

The core issue was that the CUDA resolver could silently return incorrect outputs on some workloads, often by removing no tracks at all and returning the full input unchanged.

## Problem (how it was observed)

During a synthetic GPU benchmark sweep, the CUDA resolver produced incorrect results in 8 of 9 configurations. The dominant failure mode was:

* `n_removed = 0`
* GPU output = all input candidates
* CPU and GPU selected-track sets did not match

## Root Cause

The CUDA resolver had four correctness bugs:

* `terminate_device` was allocated but never initialized to `0`, so reused device memory could start with `terminate=1` and cause the removal kernel to exit immediately.
* `scanned_block_offsets_buffer` was allocated but never passed through `setup()`, corrupting the prefix-sum metadata used to maintain sorted track order during iterative removals.
* `max_shared_device` was initialized with the wrong `cudaMemcpy` direction (`HostToDevice` instead of `DeviceToDevice`) even though the source pointer came from device memory.
* The outer loop copied `terminate` and `n_accepted` back to host asynchronously but did not synchronize before checking the loop condition, so host-side control flow could observe stale values.

_This PR applies the minimal correctness fixes necessary_

### DIFF
--- a/device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu
+++ b/device/cuda/src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu
@@ -429,6 +429,7 @@ greedy_ambiguity_resolution_algorithm::operator()(
     int terminate = 0;
     vecmem::unique_alloc_ptr<int> terminate_device =
         vecmem::make_unique_alloc<int>(m_mr.main);
+    cudaMemsetAsync(terminate_device.get(), 0, sizeof(int), stream);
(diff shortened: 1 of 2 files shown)


stephenswat (approved):
These changes make a lot of sense to me; thanks, Sorin!

PR #1302, opened 2026-04-23
Decouple io from performance and simulation building

paradajzblond:


krasznaa (approved):
I was **convinced** that we would have some example executables depend on `traccc::simulation`. 🤔 But apparently not.

The logic seems good to me. If the CI is happy, I think this can go in.

PR #1307, opened 2026-05-04
Simplify GBTS connected component analysis

m-brettell:
Replaces the array of 32-bit edgeIdx for edges that need more CCA iterations with a smaller array of 8-bit flags for each edge.
Same output, negligible timing difference.
Primarily because this prevents strange errors when running with many streams / concurrent CCA kernels.


stephenswat:
I expanded the name on this to avoid confusion with the cell clustering CCA algorithm.

stephenswat (approved):

PR #1309, opened 2026-05-05
GBTS + Clang Fix, main branch (2026.05.05.)

krasznaa:
Fixed a warning/error that would show up when using Clang as the CUDA host compiler.


stephenswat:
Perhaps you could add a CI job that runs clang+CUDA to avoid future issues?

stephenswat (approved):

PR #1314, opened 2026-05-06
Fix swapped arguments in the Alpaka configuration

daltria:
Apparently running for single particles wasn't a problem with this bug but for pile-up it was crashing with an error in Athena showing zero clusters coming back from traccc.

This looks like it fixes the problem.

cc: @krasznaa @stephenswat  @StewMH


stephenswat (approved):

krasznaa:
👍

PR #1315, opened 2026-05-06
Fix Alpaka-HIP InvalidDeviceFunction error

daltria:
Without this the Alpaka-HIP in Athena crashes with the following error:

```
RAWtoALL 13:15:25 PixelRDO_ContainerCnv                                  0     0    INFO PixelRDO_ContainerCnv::initialize()
RAWtoALL 13:15:25 GPUClusteringAlg                                       0     0   FATAL Standard std::exception is caught in sysExecute
RAWtoALL 13:15:25 GPUClusteringAlg                                       0     0   ERROR parallel_for failed: hipErrorInvalidDeviceFunction: invalid device function
RAWtoALL 13:15:25 GPUClusteringAlg                                       0     0   ERROR Maximum number of errors ( 'ErrorMax':1) reached.
RAWtoALL 13:15:25 AlgTask                                                0     0 WARNING Execution of algorithm GPUClusteringAlg failed
RAWtoALL 13:15:25 BeamSpotCondAlg                                        0     0    INFO Read from condDB status 0 pos (0,0,0) sigma (0.012,0.012,50) tilt (0,0) sigmaXY 0
RAWtoALL 13:15:25 BeamSpotCondAlg                                        0     0    INFO Recorded new InDet::BeamSpotData to BeamSpotData with range {[350200,l:0] - [410000,l:0]} into conditions store.
RAWtoALL 13:15:26 AvalancheSchedulerSvc                                  0     0   ERROR *** Stall detected, event context: s: 0  e: 0
RAWtoALL 13:15:26 AvalancheSchedulerSvc                                  0     0   ERROR Event 0 on slot 0 failed
```

Given that the `CMAKE_CUDA_ARCHITECTURES` is defined in here I followed the lead and added the HIP one as well.
`gfx1101` is needed for the CERN tbed while `gfx90a` was suggested as an extra addition by Stewart.

cc: @stephenswat @krasznaa @StewMH

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,10 @@ endif()
 set( CMAKE_CUDA_ARCHITECTURES "75" CACHE STRING
    "CUDA architectures to build device code for" )
 
+# Set the HIP architecture to build code for.
+set( CMAKE_HIP_ARCHITECTURES "gfx1101;gfx90a" CACHE STRING
+   "HIP architectures to build device code for" )
+
 # Flag controlling whether warnings should make the build fail.
 option( TRACCC_FAIL_ON_WARNINGS
    "Make the build fail on compiler/linker warnings" FALSE )


krasznaa (approved):
Looks good!

https://cmake.org/cmake/help/latest/variable/CMAKE_HIP_ARCHITECTURES.html

I'm surprised though about the issue that you ran into. 🤔 Did you do the build on a machine that didn't have an AMD GPU? Because with a GPU present, CMake / ROCm should've chosen a valid value for this.

At least [in vecmem](https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-compiler-options-hip.cmake) I was also thinking of defining a default value for this variable, but since for my own tests so far I didn't need it, I shied away from it so far. (Because the behaviour where the VecMem tests would be built specifically for the architecture of the local GPU, is the desired behaviour. VecMem doesn't provide HIP device code in `vecmem::hip`...)

daltria:
> Looks good!
> 
> https://cmake.org/cmake/help/latest/variable/CMAKE_HIP_ARCHITECTURES.html
> 
> I'm surprised though about the issue that you ran into. 🤔 Did you do the build on a machine that didn't have an AMD GPU? Because with a GPU present, CMake / ROCm should've chosen a valid value for this.
> 
> At least [in vecmem](https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-compiler-options-hip.cmake) I was also thinking of defining a default value for this variable, but since for my own tests so far I didn't need it, I shied away from it so far. (Because the behaviour where the VecMem tests would be built specifically for the architecture of the local GPU, is the desired behaviour. VecMem doesn't provide HIP device code in `vecmem::hip`...)

The build was done on `ef-tb-c01` with the following arguments if that helps
```
        -DTRACCC_SUPPORTED_DETECTORS=itk_detector \
        -DBUILD_CUDA=OFF \
        -DBUILD_SYCL=OFF \
        -DBUILD_ALPAKA_HIP=ON \
        -DTRACCC_BUILD_ALPAKA=ON \
        -DTRACCC_BUILD_HIP=ON \
        -DTRACCC_BUILD_CUDA=OFF \
        -Dalpaka_ACC_GPU_HIP_ENABLE=ON \
        -DTRACCC_SETUP_ROCTHRUST=ON \
        -DTRACCC_USE_SYSTEM_THRUST=OFF \
        -DTRACCC_USE_ROOT=OFF \
        -DCMAKE_PREFIX_PATH=/opt/rocm/lib/cmake \
```

StewMH:
This is potentially useful if we want to generally build on machines without an AMD card, although we may want to keep it to the ATLAS code.

krasznaa:
> This is potentially useful if we want to generally build on machines without an AMD card, although we may want to keep it to the ATLAS code.

This update is definitely very useful. This project should absolutely set a value to this cache variable.

I was just surprised that it already came up as a problem. I would've expected to only run into it as an issue once we started actually building HIP code on machines without an AMD GPU.

But it's definitely good that the issue came up already!

daltria:
just to mention that `c01` has an AMD GPU (in case this wasn't clear)

PR #1317, opened 2026-05-07
Add accessors for strided vectors of structs

stephenswat:
This commit adds several new types that allow us to access vectors of structs in a fully strided way, which will allow us to better exploit the coalescing properties of CUDA (and other GPU) devices.


stephenswat:
Closing this to prep for the integration.

PR #1321, opened 2026-05-11
Disable CI tests for debug builds

stephenswat:
Running the full CI testing suite for debug builds takes too long. This commit ensures we run them only on release builds.


niermann999 (approved):
I guess, we will eventually put them into the ACTS nightly

PR #1327, opened 2026-05-13
Update ACTS to version 46.7.0 and use its detray

stephenswat:
This commit updates our ACTS dependency to version 46.7.0. It also removes our own external detray mechanisms and relies almost entirely on ACTS to provide detray. Until ACTS directly exposes the detray targets, we still need to FetchContent the detray content from the ACTS repository. Also automatically adds the ITk detector metadata if it is not provided.


niermann999 (approved):

niermann999 (approved):

stephenswat:
I believe this is currently broken by this change to short-circuiting logic between:

https://github.com/acts-project/detray/blob/7c3f18eec984ddf41dfca588ac2b8a59ac8138b2/core/include/detray/propagator/propagator.hpp#L378-L379

and

https://github.com/acts-project/acts/blob/2ac7b47cb984efc7c3faadbd1dccffc31abba6d6/Detray/core/include/detray/propagator/propagator.hpp#L369-L370

krasznaa (dismissed):
I see 2 options to move forward:
  - We first make some changes in Acts. Currently it is just way too hostile towards the setup that we would need from it.
  - We keep the fundamental setup in this repository as it's always been.
    * We keep `extern/acts` building just Acts;
    * We make `extern/detray` download the Acts TGZ, and build Detray out of that. (Using a `SOURCE_SUBDIR` argument to the existing `FetchContent` call. So that it would pick up just Detray out of the Acts source tree.)

Clearly N.1 is the better choice. But it's also what will take longer.

stephenswat:
> * We make `extern/detray` download the Acts TGZ, and build Detray out of that. (Using a `SOURCE_SUBDIR` argument to the existing `FetchContent` call. So that it would pick up just Detray out of the Acts source tree.)

Explain to me how this is any better than the current approach.

niermann999:
Should be fixed here https://github.com/acts-project/acts/pull/5452/changes

extern/acts/CMakeLists.txt
```diff
@@ -9,12 +9,14 @@ cmake_minimum_required( VERSION 3.25 )
 include( FetchContent )
 
 # Tell the user what's happening.
-message( STATUS "Building Acts as part of the TRACCC project" )
+message( STATUS "Building Acts (including Detray) as part of the TRACCC project" )
 
 # Declare where to get Acts from.
+#set( TRACCC_ACTS_SOURCE
```
niermann999: Should these lines be removed?

niermann999 (approved):

extern/acts/CMakeLists.txt
stephenswat: Done, thanks.

PR #1329, opened 2026-05-15
Fixes gcc16 build errors, added needed climits header

smuzaffar:
`climits` header is needed for `SHRT_MAX` and `UINT_MAX`


krasznaa (approved):
You're probably the first one to attempt a GCC 16 build. 😄 

As long as the CI is happy (it should be...), I'm very happy with this.

PR #1331, opened 2026-05-20
Remove unwanted files before ACTS integration

stephenswat:
This commit removes files which we do not want to integrate into ACTS.


stephenswat:
Comments welcome, but this is not intended to be merged until right before the integration.

.clang-format
krasznaa, line 1: I believe this file technically needs to stay. Otherwise Acts's formatting will not like the traccc files. If we keep it here, `clang-format` on Acts could just succeed as-is, no?

.gitlab-ci.yml
krasznaa, line 1: I don't think it's worth bothering about this file. 🤔 In a subdirectory it does not instruct GitLab CI for anything. So then is it not better to remove it as part of the cleanup in the Acts repository?

.gitmodules
krasznaa, line 1: This file we could just remove right now...

krasznaa (commented):
Generally I don't think we should do this. 🤔

We'll be archiving this repository once it gets merged into Acts. So I'd prefer to archive a version of it that actually works. So that people who may possibly look at the archive in the future, would see how things worked.

stephenswat:
> We'll be archiving this repository once it gets merged into Acts. So I'd prefer to archive a version of it that actually works. So that people who may possibly look at the archive in the future, would see how things worked.

Nota bene my plan is to put this commit into a branch other than main before the integration, so that the repository can be archived cleanly. :)

PR #1332, opened 2026-05-20
CMake Config Fix, main branch (2026.05.20.)

krasznaa:
During the Athena integration exercise I discovered this gem as well. That this configuration was never updated to account for `traccc::core` publicly depending on `Acts::Core`.

Note that `traccc::io` only privately depends on `Acts::PluginJson`, so setting up that component as a dependency is not necessary. 🤔 

Tagging @paradajzblond.


stephenswat (approved):
👍

PR #1336, opened 2026-05-22
Fix vecmem CMake language module usage

stephenswat:
This commit fixes two issues with the way we currently use the vecmem language functionality. Firstly, it moves the usage of this functionality inside the vecmem setup block such that setups which load vecmem later (i.e. if relying on another project to load it) do not break due to the module path not being set yet. Secondly, it ensures that the SYCL modules are made available properly, as is done in the installed vecmem configuration.


CMakeLists.txt
```diff
@@ -156,13 +156,14 @@ if( TRACCC_SETUP_VECMEM )
...
       add_subdirectory( extern/vecmem )
-      # Make the "VecMem language code" available for the whole project.
-      list( PREPEND CMAKE_MODULE_PATH "${VECMEM_LANGUAGE_DIR}" )
    endif()
+   # Make the "VecMem language code" available for the whole project.
+   list( PREPEND CMAKE_MODULE_PATH "${VECMEM_LANGUAGE_DIR}" )
+   list( PREPEND CMAKE_MODULE_PATH "${VECMEM_LANGUAGE_DIR}/sycl" )
+   # Including vecmem-check-language makes sure that the HIP and SYCL languages
+   # would be picked up from VecMem.
+   include( vecmem-check-language )
```
krasznaa: I'm still very confused about this whole thing. 🤔 
  - If `TRACCC_SETUP_VECMEM` is set to `FALSE`, the parent project **must have** set up vecmem previous to including traccc. It must have. Otherwise `enable_language(SYCL)` will not work inside this project.
  - If `TRACCC_SETUP_VECMEM` is `TRUE`, the code should work with the current setup. I believe you've used it with `TRACCC_USE_SYSTEM_VECMEM` set to `TRUE` since forever. And I always use `FALSE`.

In the Athena builds that I'm testing at the moment, I believe I test the exact code path that you talk about.
  - We build vecmem separately from Acts, before Acts's configuration would start. (https://gitlab.cern.ch/atlas/atlasexternals/-/blob/main/External/VecMem/CMakeLists.txt?ref_type=heads)
  - Then we tell Acts to find vecmem itself (https://gitlab.cern.ch/atlas/atlasexternals/-/merge_requests/1378/diffs#8659346d8be05f89369106a45557a9063e1c24a6_84_101), and Acts tells traccc not to look for vecmem (https://github.com/acts-project/acts/blob/main/CMakeLists.txt#L612).

So I'm really not sure what's going wrong in your setup. But I do not believe that this proposed change is the answer. 🤔

Rather, if we want to change something, changing to Acts's setup (https://github.com/acts-project/acts/blob/main/CMakeLists.txt#L576-L584) may be the most robust of all worlds. 🤔

stephenswat: If we go with https://github.com/acts-project/acts/pull/5519, I still think traccc needs to have its usage of this vecmem code reworked in a similar way to this PR does (which in turn is very similar to your ACTS PR): unconditionally move the module path setting into the setup block, and then moving the usage of that closer to where we compile SYCL code. What do you think?

krasznaa: Indeed, I believe all of these places need to be updated/fixed.

Unfortunately my code in https://github.com/acts-project/acts/pull/5519 is not perfect either, as it ends up calling `vecmem_check_language(SYCL)` even in places where we could do without.

But even with that, I think that's still better than to code up "internal knowledge" about VecMem's code layout in other projects. So I would propose that we do the same sort of thing that I proposed for Acts.
  - Extend `CMAKE_MODULE_PATH` with only `VECMEM_LANGUAGE_DIR`  in the same place that the PR already does;
  - Wherever in the code (practically in `device/sycl` and `tests/sycl`) we need `enable_language(SYCL)`, we would put:

```cmake
include(vecmem-check-language)
vecmem_check_language(SYCL)
enable_language(SYCL)
```

It's not great, as it's a bit verbose. But API-wise at least it would allow us to re-organize the vecmem code later on without clients having to update.

CMakeLists.txt
```diff
@@ -156,13 +156,14 @@ if( TRACCC_SETUP_VECMEM )
       find_package( vecmem 1.9.0 REQUIRED )
    else()
       add_subdirectory( extern/vecmem )
-      # Make the "VecMem language code" available for the whole project.
-      list( PREPEND CMAKE_MODULE_PATH "${VECMEM_LANGUAGE_DIR}" )
    endif()
+   # Make the "VecMem language code" available for the whole project.
+   list( PREPEND CMAKE_MODULE_PATH "${VECMEM_LANGUAGE_DIR}" )
+   list( PREPEND CMAKE_MODULE_PATH "${VECMEM_LANGUAGE_DIR}/sycl" )
+   # Including vecmem-check-language makes sure that the HIP and SYCL languages
```
StewMH: note that we don't have any HIP custom code any more so reference to HIP in the comments could go

CMakeLists.txt
stephenswat: Sure, updated. Please take a look!

CMakeLists.txt
stephenswat: Removed that comment, thanks.

krasznaa (approved):
I myself would've put the `vecmem_check_language(SYCL)` calls just before the existing `enable_language(SYCL)` calls. 🤔
  - https://github.com/acts-project/traccc/blob/main/device/sycl_utils/CMakeLists.txt#L8
  - https://github.com/acts-project/traccc/blob/main/device/sycl/CMakeLists.txt#L12
  - https://github.com/acts-project/traccc/blob/main/tests/sycl/CMakeLists.txt#L9

Though technically this setup, with calling `vecmem_check_language(SYCL)` in the main `CMakeLists.txt` file, works as well.

stephenswat:
> I myself would've put the `vecmem_check_language(SYCL)` calls just before the existing `enable_language(SYCL)` calls. 🤔

That works for me too, I'll make the change and then move to merge. :slightly_smiling_face:

PR #1231, opened 2026-01-16
feat: progressive kalman filter

niermann999:
Write the simplest Kalman track finder possible for debugging

PR #1337, opened 2026-05-29
feat: Define stream operators for the EDM types

niermann999:
Allow easy host side printing of tracking EDM result types for debugging. Additionally, adds a convenience function to the track container that prints the data of all track states or measurements for a given track


niermann999:
This is part of the change set in #1231

stephenswat (approved):
Looks sensible to me.

PR #1338, opened 2026-06-01
CUDA Flag Tweak, main branch (2026.06.01.)

krasznaa:
While working on https://gitlab.cern.ch/atlas/athena/-/merge_requests/88385 I had to realize that `nvcc` (at least in CUDA 12.8 in this case) is not great with warning us about our use of [std::array](https://en.cppreference.com/cpp/container/array) in device code. :thinking: 

To prevent other people from wasting as much time as I have done on Friday and today, this setup should make it a bit easier to make use of the traccc EDM in CUDA code. CUDA code that may not live in this project.

While I'm a bit worried about hardcoding compiler specific flags into our installation files (as it locks us into using appropriate CUDA versions), the alternative is that clients would have to know about this flag explicitly. Which is also not great. :thinking:

### DIFF
--- a/cmake/traccc-compiler-options-cuda.cmake
+++ b/cmake/traccc-compiler-options-cuda.cmake
@@ -1,6 +1,6 @@
 # TRACCC library, part of the ACTS project (R&D line)
 #
-# (c) 2021-2025 CERN for the benefit of the ACTS project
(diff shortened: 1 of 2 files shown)


stephenswat (approved):
This seems pretty sensible to me. 👍

PR #1339, opened 2026-06-02
expected layer pattern in track finding

zovelsanj:
Adding expected-layer pattern collection and propagation through the traccc tracking workflow, including host and device CKF paths and the Athena conversion chain.

Main changes:

* Collect and store expected-layer patterns during track finding.
* Propagate precomputed patterns through track conversion instead of recomputing them later at `TrackToTrackParticleCnvAlg` in CPU.
* Added mapping support and pattern aggregation for track candidates.
* Some minor fixes (empty-buffer checks, pattern initialization, cell ordering updates).
* Added debugging and timing checks to help validate the implementation. This was done for comparing expected layer pattern across different implementations, so can be removed later.

Details on the results can be found in [this](https://indico.cern.ch/event/1688349/) GPU roundtable meeting. This implementation sets a collector pipeline for expected layer pattern. As of now, it uses perigee_extrapolator from detray [#1099](https://github.com/acts-project/detray/pull/1099) which is a  linear step propagator from tracks to origin (0,0,0). Because of this, although the match percentage of expected layer patterns has increased by 17.1% after using perigee extrapolator, there is still room to improve by implementing the actual perigee surface along with back propagation (this is to be done in detray and used in traccc).

Adding log of this implementation validated with single muon data at `/eos/atlas/atlaslocalgroupdisk/dq2/rucio/mc21_14TeV/a2/de/RDO.41864945._000003.pool.root.1` with 100 events (tests also done with larger nEvents). 
[log_with_expectedLayerPattern_collector.log](https://github.com/user-attachments/files/28518372/log_with_expectedLayerPattern_collector.log)

The table shows the improvement after implementation of expected layer pattern collector in traccc along with perigee (0, 0,0) collector in detray. As expected, most of the mismatch comes from the endcap and transition layers.
| expectedLayerPattern in G200 | Match % wrt G200 + computation on CPU | Barrel match % | Endcap match % | Transition layer match % |
|------------------------------|----------------------------------------|----------------|----------------|--------------------------|
| **Without perigee<br>extrapolation** | 50 | 68.3 | 40 | 33.33 |
| **With perigee<br>extrapolation** | 67.1 | 83.92 | 50 | 52.63 |

### DIFF
--- a/core/include/traccc/finding/actors/expected_layer_pattern_collector.hpp
+++ b/core/include/traccc/finding/actors/expected_layer_pattern_collector.hpp
@@ -1,0 +1,171 @@
+/** TRACCC library, part of the ACTS project (R&D line)
+ *
+ * (c) 2026 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Project include(s).
+#include "traccc/definitions/qualifiers.hpp"
+
+// Detray include(s).
+#include <detray/geometry/barcode.hpp>
+#include <detray/propagator/base_actor.hpp>
+
+// System include(s).
+#include <array>
+#include <cstddef>
+#include <utility>
+#include <vector>
+
+namespace traccc {
+
+using expected_layer_pattern_type = std::array<unsigned int, 4>;
+
+struct expected_layer_mapping_entry {
+    detray::geometry::barcode::value_t barcode{0u};
+    unsigned int pattern_index{0u};
+    unsigned int layer_index{0u};
+};
+
+namespace details {
+
+inline thread_local std::vector<expected_layer_pattern_type>
+    g_last_expected_layer_patterns{};
+
+inline void clear_last_expected_layer_patterns() {
+    g_last_expected_layer_patterns.clear();
+}
+
+inline void set_last_expected_layer_patterns(
+    std::vector<expected_layer_pattern_type> patterns) {
+    g_last_expected_layer_patterns = std::move(patterns);
+}
+
+inline const std::vector<expected_layer_pattern_type>&
+last_expected_layer_patterns() {
+    return g_last_expected_layer_patterns;
+}
+
+}  // namespace details
+
+/// Default mapper that disables pattern updates.
+struct null_expected_layer_mapper {
+    struct result {
+        bool valid{false};
+        unsigned int pattern_index{0u};
+        unsigned int layer_index{0u};
+    };
+
+    TRACCC_HOST_DEVICE constexpr result
+    operator()(const detray::geometry::barcode&) const {
+        return {};
+    }
+};
+
+/// Mapper that uses a flat lookup table of barcode entries.
+struct expected_layer_table_mapper {
+    using entry_type = expected_layer_mapping_entry;
+
+    struct result {
+        bool valid{false};
+        unsigned int pattern_index{0u};
+        unsigned int layer_index{0u};
+    };
+
+    const entry_type* entries{nullptr};
+    std::size_t size{0u};
+
+    TRACCC_HOST_DEVICE result
+    operator()(const detray::geometry::barcode& barcode) const {
+        if (entries == nullptr || size == 0u) {
+            return {};
+        }
+
+        const auto value = barcode.value();
+        for (std::size_t i = 0u; i < size; ++i) {
+            const auto& entry = entries[i];
+            if (entry.barcode == value) {
+                return {true, entry.pattern_index, entry.layer_index};
+            }
+        }
+
+        return {};
+    }
+};
+
+/// Collector actor for expected-layer-pattern bitmasks.
+///
+/// The mapper must return:
+/// - valid = true if the barcode can be mapped
+/// - pattern_index in [0,3]
+/// - layer_index in [0,31]
+template <typename mapper_t = null_expected_layer_mapper>
+struct expected_layer_pattern_collector : detray::actor {
+
+    using pattern_type = std::array<unsigned int, 4>;
+    using mapper_type = mapper_t;
+    using mapping_result_type = typename mapper_type::result;
+
+    struct state {
+        /// Target bitmask to update.
+        pattern_type* pattern{nullptr};
+        /// Mapping from detray barcode to (pattern index, layer index).
+        mapper_type mapper{};
+
+        /// Optional guard against repeated updates on the same surface.
+        bool deduplicate_consecutive_surfaces{true};
+        detray::geometry::barcode last_surface{};
+        bool has_last_surface{false};
+
+        /// Lightweight counters for debugging.
+        unsigned int n_updates{0u};
+        unsigned int n_skipped{0u};
+    };
+
+    template <typename propagator_state_t>
+    TRACCC_HOST_DEVICE void operator()(state& actor_state,
+                                       propagator_state_t& propagation) const {
+
+        const auto& navigation = propagation._navigation;
+
+        // Keep behavior aligned with sensitive detector element collection.
+        if (!navigation.is_on_sensitive()) {
+            return;
+        }
+
+        const auto& sf_desc = std::as_const(navigation).current().surface();
+        const auto sf_barcode = sf_desc.barcode();
+
+        if (actor_state.deduplicate_consecutive_surfaces &&
+            actor_state.has_last_surface &&
+            sf_barcode == actor_state.last_surface) {
+            ++actor_state.n_skipped;
+            return;
+        }
+
+        actor_state.last_surface = sf_barcode;
+        actor_state.has_last_surface = true;
+
+        if (actor_state.pattern == nullptr) {
+            ++actor_state.n_skipped;
+            return;
+        }
+
+        const mapping_result_type mapping = actor_state.mapper(sf_barcode);
+        if (!mapping.valid || mapping.pattern_index >= 4u ||
+            mapping.layer_index >=
+                static_cast<unsigned int>(sizeof(unsigned int) * 8u)) {
+            ++actor_state.n_skipped;
+            return;
+        }
+
+        (*actor_state.pattern)[mapping.pattern_index] |=
+            (1u << mapping.layer_index);
+        ++actor_state.n_updates;
+    }
+};
+
+}  // namespace traccc
--- a/core/include/traccc/finding/details/combinatorial_kalman_filter.hpp
+++ b/core/include/traccc/finding/details/combinatorial_kalman_filter.hpp
@@ -73,6 +73,10 @@ combinatorial_kalman_filter(
            "overstep tolerance");
     assert(config.min_track_candidates_per_track >= 1);
 
+    // Clear buffer before starting the CKF, to avoid previous event's pattern 
+    // leaking into the current event's track finding
+    traccc::details::clear_last_expected_layer_patterns();
+
     /// The algebra type
     using algebra_type = typename detector_t::algebra_type;
     /// The scalar type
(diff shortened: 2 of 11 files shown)


core/include/traccc/finding/actors/expected_layer_pattern_collector.hpp
```diff
@@ -0,0 +1,171 @@
...
+#include <cstddef>
+#include <utility>
+#include <vector>
+
+namespace traccc {
+
+using expected_layer_pattern_type = std::array<unsigned int, 4>;
+
+struct expected_layer_mapping_entry {
+    detray::geometry::barcode::value_t barcode{0u};
```
niermann999, line 28: This seems to be using a fairly old detray version (the barcode has been renamed in a recent version). The geometry identifier can (and should) be used as a data state directly, because the underlying value is not human readable when printed (e.g. `std::cout << geo_id << std::endl`)
```suggestion
    detray::geometry::identifier geo_id{};
```

device/common/include/traccc/finding/device/impl/propagate_to_next_surface.ipp
```diff
@@ -104,35 +132,157 @@ TRACCC_HOST_DEVICE inline void propagate_to_next_surface(
...
-                                 momentum_aborter_state, ckf_aborter_state));
+    //NOTE: transporter not in device actor chain now, so not included in tie
+    // propagator.propagate(propagation, detray::tie(s0, s1, s2, s3, s4, s5, s6));
+    // Run a perigee prepass for step 0, initialize pattern, wire collector, 
+    // then propagate with new actor chain.
+    if (expected_layer_patterns.size() > link_idx) {
+        expected_layer_pattern_type perigee_pattern{};
+        bool perigee_pattern_ready = false;
+
+        if (payload.step == 0 && payload.expected_layer_map != nullptr &&
```
niermann999, line 157: I must admit, I don't really understand why this is part of the CKF? Once the CKF is finished and the track fit is complete, can we not run this backpropagation then? 

Because, if you do this here, you will run this code for every seed that goes into the CKF, but in case of the triplet seederm for example, only 1/3 or half of the seeds result in actual tracks. So that is a lot of extra work that needs to be done without a clear benefit. Besides, it would be way more precise to run the back-propagation starting from the fully fitted track parameters instead of the seed parameters, which are possibly quite inaccurate. 

Furthermore, as per separation of concerns, the `propagate_to_next` kernel should only propagate to the next sensitive surface for the CKF, not take on the additional task of finding the expected layer pattern.

Therefore, I would suggest, you put this code block into its own kernel and then run it after the track fit. This is also what was planned for the finding of the perigee parameters.

core/include/traccc/finding/actors/expected_layer_pattern_collector.hpp
```diff
@@ -0,0 +1,171 @@
...
+
+// System include(s).
+#include <array>
+#include <cstddef>
+#include <utility>
+#include <vector>
+
+namespace traccc {
+
+using expected_layer_pattern_type = std::array<unsigned int, 4>;
```
niermann999, line 25: What is the `expected_layer_pattern_type` and why does it consist of 4 values? Will this be the same for all experiments? If not, this should be configurable.

It would be nice to add a few comments here, explaining what the purpose of the expected layer pattern is and how it is defined

core/include/traccc/finding/actors/expected_layer_pattern_collector.hpp
```diff
@@ -0,0 +1,171 @@
...
+struct expected_layer_mapping_entry {
+    detray::geometry::barcode::value_t barcode{0u};
+    unsigned int pattern_index{0u};
+    unsigned int layer_index{0u};
+};
+
+namespace details {
+
+inline thread_local std::vector<expected_layer_pattern_type>
+    g_last_expected_layer_patterns{};
```
niermann999, line 36: Is this a per track data state? Then it should go into the `state` struct of the corresponding actor. If it is a per event data it should be defined in the scope where the algorithm is created and then passed into the algorithm, I think. 

Preferably, these vectors will be defined in such a way, that the actors in device code will use `device_vector`s instead of pointers

core/include/traccc/finding/actors/expected_layer_pattern_collector.hpp
```diff
@@ -0,0 +1,171 @@
...
+
+        return {};
+    }
+};
+
+/// Collector actor for expected-layer-pattern bitmasks.
+///
+/// The mapper must return:
+/// - valid = true if the barcode can be mapped
+/// - pattern_index in [0,3]
```
niermann999, line 103: How are these indices defined? Is this ATLAS specific?

core/include/traccc/finding/details/combinatorial_kalman_filter.hpp
```diff
@@ -513,46 +504,42 @@ combinatorial_kalman_filter(
...
+            typename detray::parameter_resetter<
+                typename detector_t::algebra_type>::state resetter_state{
+                prop_cfg};
+            typename detray::momentum_aborter<scalar_type>::state
                 momentum_aborter_state{};
+            // Configure expected layer pattern collector state
+            typename expected_layer_pattern_collector<
+                expected_layer_table_mapper>::state
+                expected_layer_pattern_collector_state{};
+            expected_layer_pattern_type propagated_expected_layer_pattern{};
```
niermann999, line 533: Could this not be a member of the actor instead?

core/include/traccc/finding/finding_config.hpp
```diff
@@ -92,6 +94,10 @@ struct finding_config {
     /// @note This parameter affects GPU-based track finding only.
     unsigned int initial_links_per_seed = 100;
     /// @}
+
+    /// Optional mapping table for expected-layer patterns.
+    const expected_layer_mapping_entry* expected_layer_map{nullptr};
```
niermann999, line 99: Once the config object is copied to device, is this pointer then still valid?

core/include/traccc/finding/details/combinatorial_kalman_filter.hpp
```diff
@@ -564,24 +551,26 @@ combinatorial_kalman_filter(
 
             // Propagate to the next surface
             TRACCC_DEBUG_HOST("Propagating... ");
+            // Add expected layer pattern collector to the actor chain
             propagator.propagate(
```
niermann999, line 555: The host CKF does not do a back propagation?

core/include/traccc/finding/details/combinatorial_kalman_filter.hpp
```diff
@@ -705,15 +711,34 @@ combinatorial_kalman_filter(
...
+        std::move(output_expected_layer_patterns));
+
+    // NOTE: For debugging purposes only
+    TRACCC_DEBUG_HOST("Published expected-layer patterns for "
+                      << traccc::details::last_expected_layer_patterns().size()
+                      << " tracks");
+    for (std::size_t i = 0;
+         i < traccc::details::last_expected_layer_patterns().size(); ++i) {
+        const auto& pattern = traccc::details::last_expected_layer_patterns().at(i);
+        (void)pattern;
```
niermann999, line 737: In order to avoid compiler warnings about unused parameters, please you the `[[maybe_unused]]` attribute

core/include/traccc/fitting/kalman_filter/kalman_actor.hpp
niermann999, line 380: There is a number of unrelated changes in the Kalman fitter. Can you please undo them?

core/include/traccc/finding/actors/expected_layer_pattern_collector.hpp
```diff
@@ -0,0 +1,171 @@
...
+#include <vector>
+
+namespace traccc {
+
+using expected_layer_pattern_type = std::array<unsigned int, 4>;
+
+struct expected_layer_mapping_entry {
+    detray::geometry::barcode::value_t barcode{0u};
+    unsigned int pattern_index{0u};
+    unsigned int layer_index{0u};
```
niermann999, line 30: Can you add a few comments on what these indices mean?

core/include/traccc/finding/actors/expected_layer_pattern_collector.hpp
```diff
@@ -0,0 +1,171 @@
...
+    };
+
+    template <typename propagator_state_t>
+    TRACCC_HOST_DEVICE void operator()(state& actor_state,
+                                       propagator_state_t& propagation) const {
+
+        const auto& navigation = propagation._navigation;
+
+        // Keep behavior aligned with sensitive detector element collection.
+        if (!navigation.is_on_sensitive()) {
```
niermann999, line 135: Would it also be possible to run the surface sequencer and construct the layer pattern from a vector of geo IDs? The surface sequencer is already part of the Kalman fitter and will be part of the CKF soon

device/cuda/src/finding/combinatorial_kalman_filter.cuh
```diff
@@ -0,0 +1,809 @@
...
+    copy.setup(expected_layer_patterns_buffer)->ignore();
+
+    // Copy mapping table for expected layer patterns to device
+    const auto expected_layer_map_size =
+        static_cast<vecmem::data::vector_buffer<
+            expected_layer_mapping_entry>::size_type>(
+            config.expected_layer_map_size);
+    vecmem::data::vector_buffer<expected_layer_mapping_entry>
+        expected_layer_map_buffer(expected_layer_map_size, mr.main);
+    const expected_layer_mapping_entry* expected_layer_map_ptr = nullptr;
```
niermann999, line 181: In order to pass information about device allocated memory to kernels, traccc generally uses views. From these device container types can be constructed to be used in the actors, which are also usable on device and easier to read than raw pointers. So usually the chain goes buffer->view->device vector. And the device vector is then used in the actor (see e.g. the surface sequencer)

device/common/include/traccc/finding/device/propagate_to_next_surface.hpp
```diff
@@ -69,7 +81,24 @@ struct propagate_to_next_surface_payload {
...
+    /**
+     * @brief View object to the per-link expected-layer patterns
+     */
+    vecmem::data::vector_view<expected_layer_pattern_type>
+        expected_layer_patterns_view;
+
+    /**
+     * @brief Device pointer to the expected-layer mapping table
+     */
+    const expected_layer_mapping_entry* expected_layer_map;
```
niermann999, line 93: can this be made a vector view?

device/cuda/src/finding/combinatorial_kalman_filter.cuh
```diff
@@ -0,0 +1,809 @@
...
+                    (n_candidates + nThreads - 1) / nThreads;
+                propagate_to_next_surface<
+                    traccc::details::ckf_propagator_t<detector_t, bfield_t>,
+                    bfield_t>(nBlocks, nThreads, 0, stream, config,
+                              host_payload);
+                TRACCC_CUDA_ERROR_CHECK(cudaGetLastError());
+
+                str.synchronize();
+                
+                //NOTE: For debugging purposes only
```
niermann999, line 575:
```suggestion
```

device/cuda/src/finding/combinatorial_kalman_filter.cuh
```diff
@@ -0,0 +1,809 @@
...
+                                     << step_to_link_idx_map[step]
+                                     << " map_ptr="
+                                     << (expected_layer_map_ptr != nullptr)
+                                     << " map_size="
+                                     << static_cast<unsigned int>(
+                                            config.expected_layer_map_size));
+
+                if (n_candidates > 0 && expected_layer_map_ptr != nullptr) {
+                    const unsigned int sample =
+                        std::min(3u, n_candidates);
```
niermann999, line 587: Where does the number `3` come from?

niermann999 (commented):
Thanks for opening PR!

stephenswat:
Hi Sanjay, as we discussed, this repository will be archived and for that we need to close your PR. Note that this has no effect on your branch, and you can keep rebasing your code. As we have also discussed, I will be available to help you transplant your changes to ACTS when you have rebased your code onto traccc 1.6.0.

Contact me if you need any help at all!

PR #1340, opened 2026-06-03
NVTX intrumentation fixes

ericcano:
This PR fixes multiple issues while trying to use NVTX instrumentation with the CUDA (FP32) preset:

`cmake -S traccc -B build -D TRACCC_ENABLE_NVTX_PROFILING:BOOL=ON --preset cuda-fp32`

During the investigation, I bumped up the CCCL version to 3.3.3 (latest). This is not proven necessary.

The fixed issues were:
- `-Werror=maybe-uninitialized'  option failed the compilation of CUB parallel with GCC as the host compiler.
-  `CCCL_TOPLEVEL_PROJECT` CMake options was necessary to compile with 3.X CCCL
- `CMAKE_CUDA_HOST_COMPILER` needed to be set explicitly (it was empty before, leading to CMake-time error).
- The instrumentation relies on GCC's `-f instrument-functions` to instrument all function calls. This included the functions used in the instrumentation call itself, and led to infinite recursion. This was corrected by excluding the function from any file matching `nvtx3/` in its path.


stephenswat:
The copilot review is on-point, please address those issues.

ericcano:
I addressed the copilot review and improved descriptions.

Unfortunately, the `maybe-uninitialized` warnings are now still showing up of being fully silenced. I'm not sure if this is fine to have such noise.

device/cuda/CMakeLists.txt
```diff
@@ -171,8 +171,8 @@ if(TRACCC_ENABLE_NVTX_PROFILING)
 
     target_compile_options(
         traccc_cuda
-        PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=-finstrument-functions>
-        $<$<COMPILE_LANGUAGE:CXX>:-finstrument-functions>)
+        PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=-finstrument-functions;-finstrument-functions-exclude-file-list=nvtx3/>
+        $<$<COMPILE_LANGUAGE:CXX>:-finstrument-functions;-finstrument-functions-exclude-file-list=nvtx3/>)
```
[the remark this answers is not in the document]
ericcano: Fixed in follow-up commit.

cmake/traccc-compiler-options-cuda.cmake
```diff
@@ -23,6 +23,12 @@ if( "${CMAKE_CUDA_COMPILER_ID}" MATCHES "NVIDIA" )
    traccc_add_flag( CMAKE_CUDA_FLAGS "-Wall" )
    traccc_add_flag( CMAKE_CUDA_FLAGS "-Wextra" )
    traccc_add_flag( CMAKE_CUDA_FLAGS "-Wconversion" )
+   # Suppress a GCC false-positive triggered in CCCL (CUB) headers
+   # (dispatch_scan.cuh, triple_chevron_launch.h).
+   if( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" )
+      traccc_add_flag( CMAKE_CUDA_FLAGS
+         "-Xcompiler -Wno-maybe-uninitialized" )
+   endif()
```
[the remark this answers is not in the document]
ericcano: Limited to just 3 compilation units in follow-up commit. The warning still shows up, strangely. Many options were tested.

extern/cccl/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set ( CCCL_ENABLE_INSTALL_RULES ON CACHE BOOL "Install CCCL as part of traccc."
 set ( THRUST_ENABLE_INSTALL_RULES ON CACHE BOOL "Install Thrust as part of traccc." )
 set ( CUB_ENABLE_INSTALL_RULES ON CACHE BOOL "Install CUB as part of traccc." )
 set ( libcudacxx_ENABLE_INSTALL_RULES ON CACHE BOOL "Install libcudacxx as part of traccc." )
+set ( CCCL_TOPLEVEL_PROJECT ON CACHE BOOL "Indicate that CCCL is being built as part of a larger project" )
```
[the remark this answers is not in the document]
ericcano: Comment improved in follow-up commit.

CMakeLists.txt
```diff
@@ -24,6 +24,12 @@ endif()
 set( CMAKE_CUDA_ARCHITECTURES "75" CACHE STRING
    "CUDA architectures to build device code for" )
 
+# Set the CUDA host compiler to use. This is needed for CCCL
+set ( CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER} CACHE STRING
+   "Host compiler to use for CUDA code" )
+# Print the value of the CUDA host compiler, to make sure that it's set correctly.
+message( STATUS "CMAKE_CUDA_HOST_COMPILER=${CMAKE_CUDA_HOST_COMPILER}" )
```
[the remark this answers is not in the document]
ericcano: Made conditional in follow-up commit.

ericcano:
As an extra insight, running this example takes 1+hour (still waiting) with `TRACCC_ENABLE_NVTX_PROFILING` on while it takes 9 seconds without.
```
./build/bin/traccc_throughput_mt_cuda  --input-directory odd-simulations-20240509/geant4_ttbar_mu20 --track-candidates-range=5:100 --seedfinder-vertex-range=-150:150 --initial-links-per-seed=6 --input-events=200  --cpu-threads 1
```
A different approach will be needed for the multi threaded tests.

device/cuda/CMakeLists.txt
```diff
@@ -111,6 +111,19 @@ traccc_add_library( traccc_cuda cuda TYPE SHARED
...
+# Unfortunately, the warning remains, but the error is suppressed.
+if( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" )
+  set_source_files_properties(
+     "src/ambiguity_resolution/greedy_ambiguity_resolution_algorithm.cu"
+     "src/finding/combinatorial_kalman_filter_algorithm.cu"
+     "src/fitting/kalman_fitting_algorithm.cu"
+     PROPERTIES
+     COMPILE_FLAGS "-Xcompiler=-Wno-maybe-uninitialized"
+  )
+endif()
```
krasznaa: No... 🤔

The way we deal with suppressing **all** warnings from headers that we download with [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html), but which we don't control/develop ourselves, is to include those headers using `-isystem`.

We have one exception to this, namely: https://github.com/acts-project/traccc/blob/main/device/sycl/src/utils/oneDPL.hpp But that's just because of technicalities related to what I consider a oneAPI bug.

And apparently now we'll need a second exception. 😦 Since as you can read in https://github.com/acts-project/traccc/blob/main/extern/cccl/CMakeLists.txt#L19-L20, there is a good reason why we didn't use `SYSTEM` for CCCL. 🤔

So... I believe we need to introduce an intermediate header file in `traccc::cuda_utils`, which would take care of including Thrust "in the right way". (I believe it would need to be `traccc::cuda_utils` instead of `traccc::cuda` to let Alpaka use the same trick.)

CMakeLists.txt
```diff
@@ -24,6 +24,13 @@ endif()
 set( CMAKE_CUDA_ARCHITECTURES "75" CACHE STRING
    "CUDA architectures to build device code for" )
 
+# Set the CUDA host compiler to use if not previousy set. This is needed for CCCL
+# CCCL will complain if the host compiler is not the same as CMAKE_CXX_COMPILER
+if ( NOT CMAKE_CUDA_HOST_COMPILER )
+   set ( CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER} CACHE STRING
+      "Host compiler to use for CUDA code" )
+endif()
```
krasznaa: No, none of this.

It is not the responsibility of a CMake project to actively set `CMAKE_CUDA_HOST_COMPILER`. Ever.

It is the responsibility of the user who's compiling the project. Either by doing `-DCMAKE_CUDA_HOST_COMPILER=...`, or better yet, by setting:

```sh
export CUDAHOSTCXX=...
```

I do not intend to add anything like this to the project...

krasznaa (changes requested):
I'll be curious to also talk to you about how you've been building the code exactly. 🤔

stephenswat:
> As an extra insight, running this example takes 1+hour (still waiting) with `TRACCC_ENABLE_NVTX_PROFILING` on while it takes 9 seconds without.
> 
> ```
> ./build/bin/traccc_throughput_mt_cuda  --input-directory odd-simulations-20240509/geant4_ttbar_mu20 --track-candidates-range=5:100 --seedfinder-vertex-range=-150:150 --initial-links-per-seed=6 --input-events=200  --cpu-threads 1
> ```
> 
> A different approach will be needed for the multi threaded tests.

I would honestly advocate for us removing the automated NVTX function instrumentation and going with either manual NVTX annotations or just removing the NVTX functionality. As much as it is nice, I don't think it has been bringing us much.

krasznaa:
The manual annotations, as in the things that we have in

https://github.com/acts-project/traccc/blob/main/performance/src/performance/timer.cpp

, I definitely want to keep. Seeing in NSys which part of the profile refers to initialization, warm-up and the actual measurement, has been quite convenient.

But I agree, marking each and every host function call with NVTX I never really needed. 🤔

m-fila:
> The manual annotations, as in the things that we have in
> 
> https://github.com/acts-project/traccc/blob/main/performance/src/performance/timer.cpp
> 
> , I definitely want to keep. Seeing in NSys which part of the profile refers to initialization, warm-up and the actual measurement, has been quite convenient.
> 
> But I agree, marking each and every host function call with NVTX I never really needed. 🤔

My experience is similar: NVTX was useful for async development, but I always kept automatic instrumentation disabled and added manual annotations for a few things I wanted to observe

CMakeLists.txt
ericcano: But then why not have the same requirement for the CMAKE_CXX_COMPILER, which cmake sets automatically? Is there a use case to compile the CUDA code with a host compiler different from the one for the rest of the C++ code?

That said, CCCL sides with me in fact. Failing to set it, the error was:
```
CMake Error at /scratch/traccc-PR1259/build/_deps/cccl-src/cmake/CCCLBuildCompilerTargets.cmake:56 (message):
  CCCL developer builds require that CMAKE_CUDA_HOST_COMPILER matches
  CMAKE_CXX_COMPILER when using nvcc:

  CMAKE_CUDA_COMPILER:
  /cvmfs/sft.cern.ch/lcg/releases/cuda/12.5-3429b/x86_64-el9-gcc13-opt/bin/nvcc


  CMAKE_CUDA_HOST_COMPILER:

  CMAKE_CXX_COMPILER:
  /cvmfs/sft.cern.ch/lcg/releases/gcc/13.1.0-b3d18/x86_64-el9/bin/g++

  Rerun cmake with
  "-DCMAKE_CUDA_HOST_COMPILER=/cvmfs/sft.cern.ch/lcg/releases/gcc/13.1.0-b3d18/x86_64-el9/bin/g++".


  Alternatively, configure the CUDAHOSTCXX and CXX environment variables to
  match.

  /scratch/traccc-PR1259/build/_deps/cccl-src/cub/cmake/CubBuildCompilerTargets.cmake:38 (cccl_build_compiler_interface)
  /scratch/traccc-PR1259/build/_deps/cccl-src/cub/CMakeLists.txt:22 (cub_build_compiler_targets)
```
On a more general perspective, I think any statement starting with "the user must..." should be avoided as much as possible. "The user can override...." is a different story, but making people jump through hoops is not productive.

ericcano:
I reverted this PR to its bare minimum but still need to figure out this system/non system include strategy to address the 'maybe-uninitialized' issue in the proper way.

device/cuda/CMakeLists.txt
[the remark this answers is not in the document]
krasznaa: So... This is the "main part" of this PR then? 🤔 To add `-finstrument-functions-exclude-file-list=nvtx3` to the build?

I don't have any strong feelings about this part. Happy to add this, as I think you'd be the only one actively exercising this code path at the moment anyway.

device/cuda/CMakeLists.txt
krasznaa: But this I would really want to avoid adding.

So as I described earlier, you should introduce `device/cuda/src/utils/thrust.hpp`. Which would surround the "usual Thrust includes" (that `traccc::cuda` needs) with appropriate GCC and Clang pragmas to suppress all warnings from those headers.

And then switch all `<thrust/...>` includes in the code to including that local/private `thrust.hpp` file. 🤔

krasznaa (changes requested):
So this all works with the currently used CCCL version as well? 🤔 Then the warning-related changes should really not be necessary for this PR. All that is only needed for the CCCL version update, no?

CMakeLists.txt
krasznaa: On Windows `nvcc.exe` can not really use alternative host compilers, just MSVC. This I have some experience with from using the Intel compiler on Windows.

Let's not try to re-invent the wheel here. On Linux, in order to use CUDA with CMake, one just needs to set `CUDAHOSTCXX`. That's just something that one  needs to get used to. :thinking:

krasznaa (approved):
Okay, I could live with this.

device/cuda/CMakeLists.txt
m-fila:
```suggestion
        PRIVATE CUDA::nvtx3 ${CMAKE_DL_LIBS}
```

Should this be updated? Relevant cmake [docs](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#nvtoolsext)
I see a long warning messge:
```
CMake Warning (dev) at device/cuda/CMakeLists.txt:238 (target_link_libraries):
  The library that is being linked to, CUDA::nvToolsExt, is marked as being
  deprecated by the owner.  The message provided by the developer is:

  nvToolsExt has been superseded by nvtx3 since CUDA 10.0 and CMake 3.25.
  Use CUDA::nvtx3 and include <nvtx3/nvToolsExt.h> instead.

This warning is for project developers.  Use -Wno-dev to suppress it.
...
```

m-fila (commented):
~~This doesn't build for me. With `TRACCC_ENABLE_NVTX_PROFILING=ON`, CUDA 12.8, gcc 14.2.0  I get a lot of errors like the following:~~
~~/home/mafila/traccc/build/_deps/cccl-src/thrust/thrust/cmake/../../thrust/system/cuda/detail/core/triple_chevron_launch.h:143:19: error: 'dispatch' may be used uninitialized [-Werror=maybe-uninitialized]
  143 |     NV_IF_TARGET(NV_IS_HOST, (return doit_host(k, args...);), (return doit_device(k, args...);));~~
Nevermind, builds fine

device/cuda/CMakeLists.txt
stephenswat: Is this still an open issue or can we merge this? Unclear given that the original message is crossed out.

m-fila: It's resolved

PR #1341, opened 2026-06-03
GBTS runtime improvements

mimodekjaer:
These are multiple performance improvements for GBTS, on an H100 I see an improvement of ~0.7 ms per event, going from ~2.8 to ~2.1 ms per event for the kernels. While producing the same results.
For the multithreaded performance, the difference is ~0.4 ms per event.

The changes are:
1. The graph_building_parameters has been split into smaller structs which are passed to individual kernels.
2. Four of the kernels which where use for binning the spacepoint have been merge into one kernel the bin_sp_kernel.
3. An option to use the LUT for the tau cuts has been added, to make it easier in the future to compare/debug against cpu.
4. const has been added to variables that are constant
5. The graph making has been changed so the check for the boundary in phi happens before the loop, which gives room for compiler optimizations
6. Also for the graph making the node params have been split into a block of tau min and max, r and z, and one with phi, which makes one vector of 128 bit per index and one with 32 bit per index.
7. For the seed edge bidding the initial removing of the seeds have been moved from the reset bids kernel to the seed rebid kernel, to remove on kernel launch with the benefit that the kernel it is moved into is the same speed.


stephenswat (approved):
Nice that this improves performance _and_ simplifies the code. 👍

PR #1342, opened 2026-06-04
Language Code Update, main branch (2026.06.04.)

krasznaa:
I didn't pay appropriate attention to the latest updates in #1336, this is to fix a minor thing.

As described in https://github.com/acts-project/traccc/pull/1336#pullrequestreview-4409986243 and https://github.com/acts-project/traccc/pull/1336#discussion_r3340883604, I didn't want to refer to `"${VECMEM_LANGUAGE_DIR}/sycl"` directly in this project. But rather to make use of `vecmem_check_language(SYCL)` wherever we wanted to call `enable_language(SYCL)`. So here I went for that.

At the same time I realized that in #1193 I didn't update Alpaka to make use of the newly added `traccc-compiler-options-hip.cmake` file. (Since Alpaka was in a very different state when I first wrote #1193.) That seemed worth fixing as well.


stephenswat:
When I tried this without the `/sycl` line it didn't configure (because it cannot find `CMakeSYCLInformation`), but it would be nice if it works now.

stephenswat:
FWIW I think the issue is that the SYCL code is not exported properly at

https://github.com/acts-project/vecmem/blob/4748631747973452112db84bd72f81e109e14fcc/CMakeLists.txt#L39

Contrast with

https://github.com/acts-project/vecmem/blob/4748631747973452112db84bd72f81e109e14fcc/cmake/vecmem-config.cmake.in#L22-L24

Which explains why dropping the `/sycl` path works for installed vecmems but not for FetchContent'ed ones. But you're the CMake expert so I'll let you judge.

stephenswat (approved):
Anyway, seems to work so I'm happy for this to go in.

krasznaa:
Argh... Apparently HIP's `clang++` is finding some new complaints about Detray with the additional warnings that this PR activates. (In the Alpaka+HIP build.)

Let's see how easy that one is to fix... 🤔

PR #1343, opened 2026-06-07
Add guard to avoid FPE crash

daltria:
This is a patch in case the number of terminus edges ends up being zero.

Without this there is a FPE error that ends up in a crash with a message pointing towards `traccc::cuda::gbts_seeding_algorithm::operator()`

I think it can be reproduced with any EF Tracking pipeline that uses GBTS when running on single mu? In my case it was triggered before the 50th event when running on `/eos/project/a/atlas-eftracking/RDOFiles_EFTrkPerf/list_mc21_14TeV.900495.PG_single_muonpm_Pt10_etaFlatnp0_43.recon.RDO.e8557_s4422_r16128_tid41864947_00.txt` (10GeV single mu)

Since I'm not really confident with the solution given the lack of familiarity with the code, @m-brettell could you have a look?

### DIFF
--- a/device/cuda/src/gbts_seeding/gbts_seeding_algorithm.cu
+++ b/device/cuda/src/gbts_seeding/gbts_seeding_algorithm.cu
@@ -709,10 +709,16 @@ gbts_seeding_algorithm::output_type gbts_seeding_algorithm::operator()(
         &path_sizes[1],
         &ctx.d_counters[traccc::device::gbts_counter::nTerminusEdges],
         sizeof(unsigned int), cudaMemcpyDeviceToHost, stream);
-    unsigned int pathsPerTerminus = 1 + (path_sizes[0] - 1) / path_sizes[1];
 
     TRACCC_DEBUG(path_sizes[0] << "size of path store | nTerminusEdges "
                                << path_sizes[1]);
+
+    if (path_sizes[1] == 0) {
+        TRACCC_DEBUG("No paths found long enough to form seeds");
+        return {0, m_mr.main};
+    }
+
+    unsigned int pathsPerTerminus = 1 + (path_sizes[0] - 1) / path_sizes[1];
 
     cudaMalloc(&ctx.d_path_store,
                (path_sizes[0] + path_sizes[1]) * sizeof(int2));


daltria:
oh I see some conflicts so maybe this has been already solved after v1.3.0?

m-brettell:
> This is a patch in case the number of terminus edges ends up being zero.
> 
> Without this there is a FPE error that ends up in a crash with a message pointing towards `traccc::cuda::gbts_seeding_algorithm::operator()`
> 
> I think it can be reproduced with any EF Tracking pipeline that uses GBTS when running on single mu? In my case it was triggered before the 50th event when running on `/eos/project/a/atlas-eftracking/RDOFiles_EFTrkPerf/list_mc21_14TeV.900495.PG_single_muonpm_Pt10_etaFlatnp0_43.recon.RDO.e8557_s4422_r16128_tid41864947_00.txt` (10GeV single mu)
> 
> Since I'm not really confident with the solution given the lack of familiarity with the code, @m-brettell could you have a look?

Hi, this a good fix. 

The problem is that terminus edges are only counted for paths long enough to be seeds (3 edges by default), but I only checked for a connected edge (a path of 2 edges).

It might be neater to move the terminus check above the pathsPerTerminus calculation:

```c++
    cudaMemcpyAsync(
        &path_sizes[1],
        &ctx.d_counters[traccc::device::gbts_counter::nTerminusEdges],
        sizeof(unsigned int), cudaMemcpyDeviceToHost, stream);

    TRACCC_DEBUG(path_sizes[0] << "size of path store | nTerminusEdges "
                               << path_sizes[1]);

    if (path_sizes[1] == 0) {
         TRACCC_DEBUG("No paths found long enough to form seeds");
         return {0, m_mr.main};
     }
    
    unsigned int pathsPerTerminus = 1 + (path_sizes[0] - 1) / path_sizes[1];

    cudaMalloc(&ctx.d_path_store,
               (path_sizes[0] + path_sizes[1]) * sizeof(int2));
```

Thanks for fixing this.

daltria:
tagging @stephenswat and @krasznaa for further reviews

krasznaa (approved):
Seems very reasonable.

daltria:
the CI failure doesn't seem related to this PR, right?

PR #1344, opened 2026-06-09
fix: Propagate traccc loglevel setting to detray

niermann999:
Make sure the loglevel set by traccc is propagated to the detray build. By default, this is `NONE`, while detray would otherwise default to `INFO`


krasznaa (approved):
I believe this is a good setting for when traccc builds Acts/detray.

But we probably also want to update how Acts builds traccc. Which I believe we should set `TRACCC_DEVICE_LOG_LVL` for in:

https://github.com/acts-project/acts/blob/main/thirdparty/traccc/CMakeLists.txt

PR #1345, opened 2026-06-09
fix: position of momentum aborter in simulator

niermann999:
If the random generator in the random scatterer produces an energy loss so large that the stepper calculates an infinite integration error, the momentum aborter should have stopped that particle. In order for that to work, the aborter needs to run directly after the scatterer


krasznaa (approved):
Your description seems sound. And the implementation is simple enough. 😄

PR #1346, opened 2026-06-09
fix: Use traccc internal IO includes

niermann999:
Use traccc IO headers, so that the detector IO log output remains turned off. Also removes some dead code.


stephenswat (approved):
Yes, good idea.

PR #1347, opened 2026-06-12
Gbts vecmem refactor with vecmem update

mimodekjaer:
Here is the entire GBTS code in traccc re written, to use vecmem.

This uses atomic max, which make vecmem 1.25.0 needed, which is why it is included here.

This gives the same results with a speed up on both my laptop and an H100. 

I have removed synchronization that does not seem necessary, while I tested on both an H100(multithreaded) and my laptop(singlethreaded), and I get the same results as before, I have marked the removed synchronization with // in the gbts_algorithm.cu.

Other than that I have restructured the entire GBTS code, so it fits into the common kernels and runtime. This should not include any algorimic changes.

I have added some comments, to hopefully make the algorithm more understandable.

I have tried to have the types of the variables which cannot got negative to be unsigned, to have a the newest unneassery conversions.

Lastly I have made it a structure which should make it easier to align it with the CPU code after traccc goes into ACTS.

Initially the fp16 was still in but it is now removed to use fp32 on all devices which is ~0.03 ms per event slower. 
EDIT:
I have chosen to add the difference between CPU and CUDA with CUDA using fp16 for the edges. Using fp16 over fp32 gives a minimal increase in thoughput, so it needs further testing on realistic hardware to see if it enough to warrant the difference.


core/include/traccc/gbts_seeding/gbts_types.hpp
```diff
@@ -0,0 +1,162 @@
...
+    unsigned int x, y;
+};
+struct TRACCC_ALIGN(16) uint4 {
+    unsigned int x, y, z, w;
+};
+struct TRACCC_ALIGN(4) short2 {
+    short x, y;
+};
+
+inline TRACCC_HOST_DEVICE float2 make_float2(const float x, const float y) {
```
krasznaa: All of these very many explicit functions... Do they actually help us? 🤔 Could the code not just use the implicit constructors of those structs directly? I.e. rather than writing

```c++
auto foo = make_float2(a, b);
```

, why not just write

```c++
auto foo = float2{a, b};
```

? It could save us on writing a **lot** of annoying-to-maintain code. 🤔

core/include/traccc/gbts_seeding/gbts_types.hpp
```diff
@@ -0,0 +1,162 @@
...
+// System include(s).
+#include <cmath>
+
+// ---------------------------------------------------------------------------
+// Edge precision toggle
+// ---------------------------------------------------------------------------
+// 1 = use a fp16 edge type where a device compiler provides one (CUDA and
+// SYCL); 0 = use float everywhere. Flip this 1 -> 0 to fall back to float and
+// test whether fp16 is needed. A host / CPU build is built with float
+#define TRACCC_GBTS_EDGE_HALF 1
```
krasznaa: I agree with SonarCloud that I'm not too thrilled about this. 🤔 If we want to use a different precision for certain operations on a GPU and on a CPU, could that part of the code not be implemented in a templated way?

That way we would only need to include the appropriate CUDA and SYCL (and later HIP, Alpaka, etc.) headers where they're actually available.

device/common/include/traccc/gbts_seeding/device/add_seed_proposal.hpp
```diff
@@ -0,0 +1,47 @@
...
+/// @param[in]  depth                      Optional maximum walk depth (-1 ==
+/// unlimited)
+///
+TRACCC_HOST_DEVICE
+inline void add_seed_proposal(
+    const int qual, const int path_idx, const unsigned int prop_idx,
+    const collection_types<char>::view d_seed_ambiguity_view,
+    const collection_types<int2>::view d_seed_proposals_view,
+    const collection_types<unsigned long long int>::view d_edge_bids_view,
+    const collection_types<int2>::const_view& d_path_store_view,
```
krasznaa: Let's not use `traccc::collection_types`!

In this case something like `collection_types<char>::view` is literally just `vecmem::data::vector_view<char>`. So let's use that type directly.

As far as I can see, you should be able to use vecmem types directly in all these places.

One could think of whether some [vecmem::edm](https://acts-project.github.io/vecmem/namespacevecmem_1_1edm.html) SoA types could be useful in the GBTS code. 🤔 But the `traccc::collection_types` and `traccc::container_types` classes have been on death row since a while.

device/common/include/traccc/gbts_seeding/device/bin_sp.hpp
```diff
@@ -0,0 +1,44 @@
...
+    const collection_types<unsigned int>::view layerCounts_view,
+    const collection_types<unsigned short>::const_view& spacepointsLayer_view,
+    const collection_types<unsigned int>::view original_sp_idx_view,
+    const collection_types<std::pair<unsigned int, unsigned int>>::const_view&
+        layer_info_view,
+    const collection_types<std::pair<float, float>>::const_view& layer_geo_view,
+    const collection_types<unsigned int>::view node_eta_index_view,
+    const collection_types<unsigned int>::view node_phi_index_view,
+    const collection_types<unsigned int>::view eta_phi_histo_view,
+    const unsigned int nPhiBins);
```
krasznaa: Multiple things:
  - Let's use some more expressive function names. I fear that if we don't make the names more reasonable in this rewrite, then we'll be stuck with them for a real long time.
  - For basically all "kernel functions" that take so many parameters, please introduce "payload types". In this case something like:

```c++
struct gbts_bin_spacepoints_payload {
   vecmem::data::vector_view<float4> sp_params;
   ...
};
```

At which point you could write the main function like:

```c++
inline void gbts_bin_spacepoints(
    const global_index_t globalIndex,
    const gbts_bin_spacepoints_payload& payload);
```

Note that in this case you can leave off the `_view` postfix from all the variable names. Since they will not clash with the `vecmem::device_vector` variables that you'll be creating in the function's body.

device/common/include/traccc/gbts_seeding/device/fill_path_store.hpp
```diff
@@ -0,0 +1,62 @@
...
+/// @param[in,out] nPathStoreSizeCounter  Global atomic write cursor
+/// @param[in]  nTerminus                 Number of terminus edges
+/// @param[in]  nTerminusPerBlock         Terminus edges processed per block
+/// @param[in]  max_num_neighbours        Maximum neighbours per edge
+/// @param[in]  nPaths                    Upper bound on the number of paths
+///
+/// Shared-memory scratch for fill_path_store: the block-local stack of live
+/// paths being walked and its running length.
+struct fill_path_store_shared_payload {
+    traccc::uint2* live_paths;
```
krasznaa: You should use a vecmem type here as well.

```suggestion
    vecmem::data::vector_view<traccc::uint2> live_paths;
```

device/common/include/traccc/gbts_seeding/device/fill_path_store.hpp
```diff
@@ -0,0 +1,62 @@
...
+/// @param[in]  d_levels_view             Per-edge CCA level array
+/// @param[in,out] nPathStoreSizeCounter  Global atomic write cursor
+/// @param[in]  nTerminus                 Number of terminus edges
+/// @param[in]  nTerminusPerBlock         Terminus edges processed per block
+/// @param[in]  max_num_neighbours        Maximum neighbours per edge
+/// @param[in]  nPaths                    Upper bound on the number of paths
+///
+/// Shared-memory scratch for fill_path_store: the block-local stack of live
+/// paths being walked and its running length.
+struct fill_path_store_shared_payload {
```
krasznaa: The "shared payload" type is good. In this place, as per my previous comment, let's also add a "normal payload" type as well. So that the function would become:

```c++
template <concepts::thread_id1 thread_id_t, concepts::barrier barrier_t>
TRACCC_HOST_DEVICE inline void fill_path_store(
    const thread_id_t& thread_id, const barrier_t& barrier,
    const fill_path_store_shared_payload& shared,
    const fill_path_store_payload& payload);
```

device/common/include/traccc/gbts_seeding/device/gbts_seeding_algorithm.hpp
```diff
@@ -0,0 +1,699 @@
...
+        const edm::spacepoint_collection::const_view& spacepoints,
+        const edm::measurement_collection::const_view& measurements)
+        const override;
+
+    protected:
+    /// @name Payloads & kernel launchers (to be implemented by backends)
+    /// @{
+
+    /// Payload for the count_sp_by_layer_kernel function
+    struct count_sp_by_layer_kernel_payload {
```
krasznaa: So, since you have these payload types already, just move them into the headers of the common functions. And make the common functions use these same types as well.

device/common/src/gbts_seeding/graph_making.cpp
```diff
@@ -0,0 +1,221 @@
+/** TRACCC library, part of the ACTS project (R&D line)
```
krasznaa: If you absolutely insist on splitting the different functions of the algorithm into separate files (which on first order I wouldn't do), then at least let's name the files in a descriptive way. In this case as `gbts_seeding_algorithm_make_graph.cpp`.

krasznaa (commented):
I'll stop at this point. 🤔

I really like the direction! Let's just go "all the way" then, as long as we're trying to sync things up with all other parts of our code!

core/include/traccc/gbts_seeding/gbts_types.hpp
m-brettell: As a heads up, I’m planning on moving from 4xfp16 to use 4x16-bit ints for edge matching in future. Using shorts should help avoid this problem of different types being available on CPU and GPU? So would it be simpler to use fp32 for everything now, then move to shorts with my edge connection update later? Thanks.

mimodekjaer: Okay then I will just move it to fp32 now. The difference in performance from what I can see between fp16 and fp32 is anywhy minimal on the gpu's I have tested.

mimodekjaer: Just to give a little more information on the H100 when testing using fp16 and fp32 the difference is 0.03 ms per event. But I have now removed the fp16 type.

device/common/include/traccc/gbts_seeding/device/bin_sp.hpp
mimodekjaer: Yes I have made it so all the host device kernels pass the payloads though now.

device/common/include/traccc/gbts_seeding/device/add_seed_proposal.hpp
mimodekjaer: Yes I think using an edm SoA might be a benefit, but this will be a later change, which might also help the cpu code which would make it easier when we have the traccc code in ACTS.

device/common/include/traccc/gbts_seeding/device/bin_sp.hpp
mimodekjaer: And I have change the name of all the kernels

core/include/traccc/gbts_seeding/gbts_types.hpp
```diff
@@ -0,0 +1,53 @@
...
+inline constexpr float PI_F = traccc::constant<float>::pi;
+inline constexpr float TWO_PI_F = 2.0f * traccc::constant<float>::pi;
+
+// Wrap an angle into (-pi, pi], matching the reference (round-to-nearest).
+// This can be generalised using if and floor which is in
+// trigonometric helper functions.
+// TODO: Test whether this is faster than the trigometric helper function.
+TRACCC_HOST_DEVICE inline float phi_wrap(const float phi) {
+    return phi - TWO_PI_F * rintf(phi * (1.0f / TWO_PI_F));
+}
```
krasznaa: We don't need to implement a new function for this. We have a version here already:

https://github.com/acts-project/traccc/blob/main/core/include/traccc/utils/trigonometric_helpers.hpp#L34

Though it would probably make more sense to switch those helpers to be implemented on top of both `float` and `double`, and not use `traccc::scalar` to define them. 🤔 (That's a separate issue though. For now, you should just avoid introducing this function.)

device/common/include/traccc/gbts_seeding/device/details/gbts_create_seed_candidate.hpp
```diff
@@ -0,0 +1,49 @@
...
+/// unlimited)
+///
+TRACCC_HOST_DEVICE
+inline void gbts_create_seed_candidate(
+    const int qual, const int path_idx, const unsigned int prop_idx,
+    const vecmem::data::vector_view<char> d_seed_ambiguity_view,
+    const vecmem::data::vector_view<int2> d_seed_proposals_view,
+    const vecmem::data::vector_view<unsigned long long int> d_edge_bids_view,
+    const vecmem::data::vector_view<const int2>& d_path_store_view,
+    char depth = -1);
```
krasznaa, line 45: Now, this is not an exact science **at all**. But since as far as I can see this function is only used internally by the other "user facing" GBTS functions, it could be good to hide it a little more from the API of this code. 🤔

I could see it being put into the `traccc::device::details` namespace, and the header moved to `traccc/gbts_seeding/device/details/`. (With the `.ipp` file staying where it is.)

At first I was hoping that even more of these functions would be like this. That they would not need to all be called from separate kernels. But I think this is the only function that is actually like this. 🤔

device/common/include/traccc/gbts_seeding/device/gbts_seeding_algorithm.hpp
```diff
@@ -0,0 +1,315 @@
...
+    /// Outputs of the graph-making stage that are consumed by seed extraction.
+    struct graph_making_output {
+        /// Compacted, row-major graph
+        vecmem::data::vector_buffer<unsigned int> output_graph;
+        /// Number of edges that survived re-indexing (0 == nothing to do)
+        unsigned int nConnectedEdges = 0;
+    };
+
+    /// Stage 1: count, bin, sort and characterise nodes.
+    node_making_output gbts_make_nodes(
```
krasznaa: 😄 You could skip the `gbts_` prefix on these member functions. For the standalone functions I think having the `gbts_` prefix is a good idea. But for member functions it just makes us type more characters. 😉

device/common/src/gbts_seeding/gbts_seeding_algorithm.cpp
```diff
@@ -0,0 +1,602 @@
...
+         cfg.gbts_count_spacepoints_by_layer_params});
+
+    // CPU: prefix-sum the per-layer counts on the host (turns counts into the
+    // layer-ordered write offsets used by gbts_bin_spacepoints).
+    vecmem::vector<unsigned int> layerCounts(cfg.nLayers + 1, mr().host);
+    copy()(vecmem::get_data(layerCounts_buf), layerCounts)->wait();
+    for (unsigned int layer = 0; layer < cfg.nLayers; layer++) {
+        layerCounts[layer + 1] += layerCounts[layer];
+    }
+    copy()(vecmem::get_data(layerCounts), layerCounts_buf)->wait();
```
krasznaa: I think you don't need to explicitly wait for this copy to finish here. 🤔

It's a bit curious that the code does this operation with a round-trip to the host. Would a Thrust/DPL algorithm not be available to do this on the device directly?

With that said, all of this is performance optimisation. So, not my main concern for this PR. 😉

device/cuda/src/gbts_seeding/gbts_seeding_algorithm.cu
```diff
@@ -5,847 +5,510 @@
...
-                    m_config.nLayers * sizeof(unsigned int),
-                    cudaMemcpyHostToDevice, stream);
-
-    ctx.nNodes = static_cast<unsigned int>(layerCounts[m_config.nLayers]);
-    if (ctx.nNodes == 0) {
-        TRACCC_WARNING(
-            "No nodes found after spacepoint counting, returning empty result");
-        return {0, m_mr.main};
+    for (unsigned int sp_idx = details::global_index1(); sp_idx < payload.nSp;
+         sp_idx += blockDim.x * gridDim.x) {
```
krasznaa: Why don't you put these for-loops inside of the `device::` functions? Generally the `__global__` functions should just call the `device::` functions without any real logic.

krasznaa (commented):
I continue to like how this all looks. The structure is at this point, how I would prefer this to look. So it's really just fine-tuning comments at this point.

device/cuda/src/gbts_seeding/gbts_seeding_algorithm.cu
mimodekjaer: So I can most definitely put this into the kernels, my biggest reason to not do it, is that I do not know if this is a "good" way of doing it outside CUDA.

device/common/src/gbts_seeding/gbts_seeding_algorithm.cpp
mimodekjaer: I will test if the wait is needed here.
Initially I tested thrust on device to reduce the copy, but this was slower. But I will test thrust scan on host.

core/include/traccc/gbts_seeding/gbts_types.hpp
mimodekjaer: Sure this makes sense, and I will change it to the helper.
But just for my edification, is there a reason to do the if and floor instead of an rint?

device/common/include/traccc/gbts_seeding/device/details/gbts_create_seed_candidate.hpp
mimodekjaer: There a couple of places like in the graph making where there is a detail function currently I have change the namespace from gbts_detail to detail, but when they are only used in one kernel should I have it as a separate file or should it live in the with the kernel that uses it?

core/include/traccc/gbts_seeding/gbts_types.hpp
krasznaa: The code proposed here may actually be better. 🤔 So I'm not saying that we shouldn't update the existing function to work like this instead.

device/common/include/traccc/gbts_seeding/device/details/gbts_create_seed_candidate.hpp
krasznaa: As we discussed, I think those can stay in the `.ipp` files. Though as long as we're on it, I'd skip on introducing the `traccc::device::gbts_detail` namespace. And just use `traccc::device::details::`, which we have already. Here for instance:

https://github.com/acts-project/traccc/blob/main/device/common/include/traccc/clusterization/device/ccl_kernel_definitions.hpp

device/cuda/src/gbts_seeding/gbts_seeding_algorithm.cu
krasznaa: On first order let's use the same logic that seems to work well with CUDA. This is what the `thread1` types are for after all.

If it turns out that AMD cards indeed respond better to a different pattern, we can re-visit how we should do this. 🤔

device/common/src/gbts_seeding/gbts_seeding_algorithm.cpp
mimodekjaer: Okay  so I did the test again using thrust on the device side and it is faster, so I have included that.

device/common/include/traccc/gbts_seeding/device/impl/gbts_make_graph_edges.ipp
```diff
@@ -0,0 +1,243 @@
...
+
+        float min_phi1 = phi2 - deltaPhi;
+        float max_phi1 = phi2 + deltaPhi;
+
+        if (min_phi1 < -traccc::device::PI_F) {
+            min_phi1 += traccc::device::TWO_PI_F;
+        }
+        if (max_phi1 > traccc::device::PI_F) {
+            max_phi1 -= traccc::device::TWO_PI_F;
+        }
```
krasznaa, line 176: I guess this should rather be:

```suggestion
        float min_phi1 = traccc::detail::wrap_phi(phi2 - deltaPhi);
        float max_phi1 = traccc::detail::wrap_phi(phi2 + deltaPhi);
```

🤔

device/common/include/traccc/gbts_seeding/device/impl/gbts_make_graph_edges.ipp
```diff
@@ -0,0 +1,243 @@
...
+        const bool boundary = max_phi1 < min_phi1;  // +/- pi wraparound
+
+        // expand over nearest bin boundary
+        max_phi1 += phi_bin_width;
+        min_phi1 -= phi_bin_width;
```
krasznaa, line 181: And then I wonder if we should protect these values with `wrap_phi(...)` as well. 🤔

device/cuda/src/gbts_seeding/gbts_seeding_algorithm.cu
```diff
@@ -5,847 +5,487 @@
...
-    kernels::eta_phi_counting_kernel<<<nBlocks, nThreads, 0, stream>>>(
-        ctx.d_eta_phi_histo, ctx.d_eta_node_counter, ctx.d_phi_cusums,
-        m_config.n_eta_bins, m_config.nPhiBins);
+/// CUDA kernel for running @c traccc::device::gbts_count_spacepoints_by_layer
+__global__ void gbts_count_spacepoints_by_layer(
+    const device::gbts_count_spacepoints_by_layer_payload payload) {
 
-    cudaStreamSynchronize(stream);
+    const details::thread_id1 thread_id;
+    device::gbts_count_spacepoints_by_layer(thread_id, payload);
```
krasznaa: Technically here you could've gone for a bit more compact formalism. 🤔

```suggestion
    device::gbts_count_spacepoints_by_layer(details::thread_id1{}, payload);
```

But this is really just nitpicking from me...

krasznaa (approved):
I think it's practically ready at this point. It would be worth thinking a bit about the "phi wrapping" in the part that I'm pointing out, but I'm generally happy with the PR.

device/common/include/traccc/gbts_seeding/device/impl/gbts_make_graph_edges.ipp
mimodekjaer: Here they are just used as cuts, so if we do the `wrap_phi` the logic for the cut would be worse, and there are no actual issue if this goes out of the[-pi,pi) range

device/common/include/traccc/gbts_seeding/device/impl/gbts_make_graph_edges.ipp
mimodekjaer: I did this initially, but give the wraps where one directional, I found it to be slower.
So yes I can include it, it is just the tradeoff with speed.

krasznaa: It's shades of grey then. :thinking:

I found this code because I was searching for uses of `PI_F` and `TWO_PI_F`. As I thought they would not be needed directly anymore.

If this is really having a measurable performance effect, then it's definitely not the end of the world to keep the code as-is. Though it's a bit worrisome that `wrap_phi(...)` is so heavy that this change makes a measurable effect. :thinking:

PR #1348, opened 2026-06-17
CUDA Stream Handling Simplification, main branch (2026.06.17.)

krasznaa:
As @paradajzblond found, `traccc::cuda::stream` currently must own its own CUDA stream. It cannot be told to use a stream that is managed by somebody else. Like a framework. (Athena...)

So here I'm greatly simplifying the `traccc::cuda::stream` class, and renaming it to `traccc::cuda::stream_wrapper`. So that it would finally be a bit more in sync with [traccc::sycl::queue_wrapper](https://github.com/acts-project/traccc/blob/main/device/sycl_utils/include/traccc/sycl/utils/queue_wrapper.hpp). In both naming and behaviour.

This also allowed me to simplify the code in the `traccc::cuda` library a little. Since `traccc::cuda::stream_wrapper`, unlike `traccc::cuda::stream`, can be trivially copied.

The examples and tests needed to be adjusted a little more. There I had to introduce [vecmem::cuda::stream_wrapper](https://acts-project.github.io/vecmem/classvecmem_1_1cuda_1_1stream__wrapper.html) objects to now take care of managing the lifetime of streams. While `traccc::cuda::stream_wrapper` would just view/use those streams managed by VecMem. (The logic here is that I believe it's enough to have complicated logic with a "smart stream object" either owning or not owning a stream, just in vecmem. In this project we just need to be able to use an externally managed stream.)


device/cuda_utils/include/traccc/cuda/utils/stream_wrapper.hpp
```diff
@@ -0,0 +1,52 @@
...
+///
+/// Note that unlike @c vecmem::cuda::stream_wrapper, this type can not own a
+/// stream of its own. It can only view a stream that is owned by
+/// "somebody else".
+///
+class stream_wrapper {
+
+    public:
+    /// Wrap an existing @c cudaStream_t object, without taking ownership
+    stream_wrapper(void* stream);
```
asalzburger: I agree with SonarCloud here, would be `explicit`.

For my education, is this `void*` business the recommended way to deal with these streams?

device/cuda/src/utils/utils.cpp
```diff
@@ -20,9 +20,9 @@ unsigned int get_warp_size(int device) {
     return static_cast<unsigned int>(warp_size);
 }
 
-cudaStream_t get_stream(const stream& stream) {
+cudaStream_t get_stream(const stream_wrapper& stream) {
 
-    return reinterpret_cast<cudaStream_t>(stream.cudaStream());
+    return static_cast<cudaStream_t>(stream.cudaStream());
```
asalzburger, line 25: Looks much safer, I do not understand why there was a `reinterpret_cast` at first place.

device/cuda_utils/src/stream_wrapper.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "traccc/cuda/utils/stream_wrapper.hpp"
+
+#include "cuda_error_handling.hpp"
+
+// CUDA include(s).
+#include <cuda_runtime_api.h>
+
+namespace traccc::cuda {
+
+stream_wrapper::stream_wrapper(void* stream) : m_stream{stream} {}
```
asalzburger, line 18: Ha, SonarCloud had the same question ... as I had before on the header file.

asalzburger (changes requested):
Can you make the constructors explicit ?

Also, I do not understand why the removal of "select_device" functionality is related to this PR ? Do I miss something ?

device/cuda/src/utils/utils.cpp
krasznaa: 🤷 Bad code writing on my part in the first place.

Don't be fooled though! Just because the text changed, this cast is as unsafe as it ever was.

device/cuda_utils/include/traccc/cuda/utils/stream_wrapper.hpp
krasznaa: I was thinking about it, but then went the lazy route. 🤔 But yes, technically making this `explicit` is the correct way to go.

device/cuda_utils/src/stream_wrapper.cpp
krasznaa: The whole point of `void*` is to not have the `traccc::cuda` library publicly depend on the CUDA SDK.

This is a very well founded boundary condition of ours. That we had quite some back-and-forth about with the GNN plugin as well in Acts with @benjaminhuth.

I can tell you the details in person. But the `void*` interface is definitely staying with us.

krasznaa:
As for the removal of the files: Those **private** source/header files were only needed to manage streams in memory. As long as we outsource the management to the caller, they are not needed anymore. (They were only used by `stream.cxx` so far. With that code gone, those classes are not needed anymore either.)

device/cuda_utils/src/stream_wrapper.cpp
asalzburger: Thanks, that makes sense to me.

device/cuda/src/utils/utils.cpp
asalzburger: Well, at least it looks better now.

device/cuda_utils/include/traccc/cuda/utils/stream_wrapper.hpp
asalzburger: Can go into the next round.

asalzburger (approved):
Got sufficient explanation- I can approve this now.

PR #1349, opened 2026-06-17
GBTS alpaka implementation

mimodekjaer:
Here is GBTS using alpaka.

It produces the same number of seed as the CUDA implementation.


krasznaa (approved):
👏 This was pretty quick.

Did you only test on the Alpaka CPU backend, or did you give a try to the CUDA backend as well?

In either case, I'm pretty happy with things already like this.

mimodekjaer:
> 👏 This was pretty quick.
> 
> Did you only test on the Alpaka CPU backend, or did you give a try to the CUDA backend as well?
> 
> In either case, I'm pretty happy with things already like this.

I have only did a proper test for CUDA backend. Because when I was not using the CUDA backend on my laptop, it defaulted to my integrated graphics which was really slow.

krasznaa:
I'm 99% sure that the test was running on your CPU in that case. (And not on any integrated GPU.) But that's not too important. I'm still happy with this PR.

PR #1350, opened 2026-06-17
CUDA Register Spilling Config, main branch (2026.06.17.)

krasznaa:
Changed the logic of enabling/disabling CUDA register spilling. Since it seems that at least in some cases with CUDA 13 it doesn't actually work. So this allows users to turn it off by hand if necessary.

With a local CUDA 13.2.1 + GCC 13.3.0 I observed the following:

```
[build] ptxas fatal   : Pragma 'enable_smem_spilling' is not allowed for dynamic SMEM
[build] gmake[2]: *** [device/cuda/CMakeFiles/traccc_cuda.dir/build.make:675: device/cuda/CMakeFiles/traccc_cuda.dir/src/finding/kernels/specializations/find_tracks.cu.in.default_detector.cu.o] Error 255
[build] gmake[2]: *** Waiting for unfinished jobs....
```

Which I couldn't track down to its root yet. And since the problem only shows up for certain detector types as it seems, I also couldn't figure out a good way for checking for this behaviour with [try_compile(...)](https://cmake.org/cmake/help/latest/command/try_compile.html). So for now I just did this dumb thing.

This is because I'm 99% sure that the Athena nightlies will need to make use of this knob. :thinking:

### DIFF
--- a/device/cuda/CMakeLists.txt
+++ b/device/cuda/CMakeLists.txt
@@ -198,3 +198,18 @@ if( ( "${CMAKE_CUDA_COMPILER_ID}" STREQUAL "NVIDIA" ) AND
   set_target_properties( traccc_cuda PROPERTIES
     CUDA_SEPARABLE_COMPILATION ON )
 endif()
+
+# Control whether to enable register spilling into shared memory. Also allowing
+# users to override our chosen default.
+set(TRACCC_CUDA_ENABLE_SPILL_TO_SHARED_MEMORY_DEFAULT OFF)
+if(CUDAToolkit_VERSION VERSION_GREATER_EQUAL 13.0)
+  set(TRACCC_CUDA_ENABLE_SPILL_TO_SHARED_MEMORY_DEFAULT ON)
+endif()
+option(TRACCC_CUDA_ENABLE_SPILL_TO_SHARED_MEMORY
+  "Enable spilling of registers to shared memory in CUDA kernels"
+  ${TRACCC_CUDA_ENABLE_SPILL_TO_SHARED_MEMORY_DEFAULT})
+mark_as_advanced(TRACCC_CUDA_ENABLE_SPILL_TO_SHARED_MEMORY)
+if(TRACCC_CUDA_ENABLE_SPILL_TO_SHARED_MEMORY)
+  target_compile_definitions(traccc_cuda
+    PRIVATE TRACCC_CUDA_ENABLE_SPILL_TO_SHARED_MEMORY)
+endif()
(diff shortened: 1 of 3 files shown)


krasznaa:
Let me also link this to #1132, which introduced this behaviour.

stephenswat:
Wrong solution remove line 29 of
https://github.com/acts-project/traccc/blob/main/device/cuda/src/finding/kernels/specializations/find_tracks_src.cuh

krasznaa:
I don't understand. 😕 So register spilling doesn't work in that function/kernel at all?

This would leave it just for `traccc::cuda::kernels::propagate_to_next_surface`. Is that what you propose? 🤔

krasznaa:
Let me split the difference then. With keeping the user-configurability, and also removing `TRACCC_CUDA_SPILL_TO_SHARED_MEMORY` from `traccc::cuda::kernels::find_tracks`. :thinking:

krasznaa:
@asalzburger, I'll need your help on this one as well. (It's this and the stream changes that we absolutely need for Athena. The const-correctness was more of a good-to-have.)

stephenswat (changes requested):
Please remove all this unnecessary CMake machinery and just go with the 1 line change...

If you have concrete indications that this is causing performance regressions and that it makes sense to be able to turn this off, then fine - but otherwise this is all unnecessary and moves decisions that the preprocessor is perfectly capable of making into CMake for no reason.

krasznaa:
My decision for adding the CMake code has nothing to do with performance considerations. It has to do with not wanting to hardcode code behaviour decisions into the source code.

Why would it be a bad thing to make the build of the code more flexible? I'm not changing any default behaviour here. I'm only leaving the door open to make the code usable even in an unforeseen compiler setup, which may not work in the way that we expect.

CMake-compiled code is fundamentally meant to be designed according to:

https://cmake.org/cmake/help/latest/guide/tutorial/In-Depth%20System%20Introspection.html

So that instead of hardcoding which version of a given library on a given platform does X, the build system should just check if "X" is available on the build platform as-is. As I described in my PR, I unfortunately couldn't find a good test for this piece of code. So for now I kept the decision to be made on the CUDA version. But at least I moved the decision to the CMake code. So that it could be overridden. Should it be necessary.

It's exactly for issues like this that we have to patch some projects in AthenaExternals for practically all new GCC versions that we switch to. Because the authors write those projects to be super sure that they will compile just fine in the rigid setup that their authors designed. And never considered that their code may be compiled on a platform that they never tested.

I really want to do better than that with our code!

stephenswat:
> As I described in my PR, I unfortunately couldn't find a good test for this piece of code.

Problem here is that you are failing to understand the underlying issue, which has nothing to do with introspection of the compiler toolchain. The issue is _not_ whether a specific PTX pragma is accepted by the PTX assembler (which would indeed be a toolchain issue) but rather that the use of that pragma is not allowed in kernels with dynamic shared memory, like `find_tracks`. Thus there exists no way and there will never exist any way to check this at the CMake level. The entire idea is void because this problem is not solvable through toolchain introspection.

The actual issue is patched in #1354, if you insist on having a CMake toggle for this spilling behaviour we can discuss that separately.

stephenswat:
I assume this can be closed now.

krasznaa:
Please go and prepare a new tag for me. See: https://its.cern.ch/jira/browse/ATLASRECTS-8402

PR #1352, opened 2026-06-18
Const Copy, main branch (2026.06.18.)

krasznaa:
As yet another thing that @paradajzblond noticed while integrating the algorithms/tools into Athena, we've been needlessly requiring non-const references to `vecmem::copy` objects in our device algorithms.

The idea here was that this would prevent people constructing a device algorithm with a temporary copy object. 🤔 But people will always be able to shoot themselves in the foot if they want to. And using a non-const reference for this purpose was not such a good idea anyway.

The changes are all super trivial luckily, as [vecmem::copy](https://acts-project.github.io/vecmem/classvecmem_1_1copy.html) already defines all of its functions/operators as `const` anyway.

### DIFF
--- a/device/common/src/device/algorithm_base.cpp
+++ b/device/common/src/device/algorithm_base.cpp
@@ -10,22 +10,13 @@
 
 namespace traccc::device {
 
-algorithm_base::algorithm_base(const memory_resource& mr, vecmem::copy& copy)
+algorithm_base::algorithm_base(const memory_resource& mr,
+                               const vecmem::copy& copy)
     : m_mr{mr}, m_copy{copy} {}
-
-memory_resource& algorithm_base::mr() {
-
-    return m_mr;
-}
 
 const memory_resource& algorithm_base::mr() const {
 
     return m_mr;
-}
-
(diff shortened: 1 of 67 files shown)


device/common/src/device/algorithm_base.cpp
```diff
@@ -10,24 +10,15 @@
...
 const memory_resource& algorithm_base::mr() const {
 
     return m_mr;
 }
 
-vecmem::copy& algorithm_base::copy() {
-
-    return m_copy.get();
-}
-
```
asalzburger, line 30: I see, you even disallow returning non-const copies of the algorithm

asalzburger (approved):
Trivial const additions, approve.

device/common/src/device/algorithm_base.cpp
krasznaa: Previously, as long as the base class was holding on to a non-const reference, there could be a reason to access either a const or non-const object, from different contexts.

But now that the base class just holds on to a const object, we only need to be able to return that const reference. So the non-const versions (which we technically didn't need so far either) cannot be there anymore.

PR #1353, opened 2026-06-18
GBTS sycl implementation

mimodekjaer:
Here is the SYCL version of GBTS. 

I have tested it on AMD and NVIDIA GPUs, which both run with the same results as the base CUDA version.

I found a small bug in the common kernel, where the address space for the shared memory atomics where not set correctly which is now set to the local space.


device/common/include/traccc/gbts_seeding/device/impl/gbts_fill_path_store.ipp
```diff
@@ -82,7 +82,9 @@ TRACCC_HOST_DEVICE inline void gbts_fill_path_store(
                 continue;
             }
             const unsigned int live_idx = static_cast<unsigned int>(
-                vecmem::device_atomic_ref<int>(shared.n_live_paths)
+                vecmem::device_atomic_ref<int,
+                                          vecmem::device_address_space::local>(
+                    shared.n_live_paths)
```
krasznaa, line 84: Excellent! Indeed, only SYCL is really sensitive to this at the moment. Since with CUDA and HIP we just use their overloaded `atomicAdd(...)` function.

I'm very happy that you managed to figure this one out, as I could've seen this causing some difficulty to debug. 🤔

krasznaa (approved):
Colour me impressed!

As long as the CI comes back happy, I'm happy to push this in.

device/sycl/include/traccc/sycl/gbts_seeding/gbts_seeding_algorithm.hpp
```diff
@@ -0,0 +1,96 @@
...
+    /// @param cfg The GBTS seed finding configuration
+    /// @param mr The memory resource(s) to use in the algorithm
+    /// @param copy The copy object to use for copying data between device
+    ///             and host memory blocks
+    /// @param queue The SYCL queue to perform the operations in
+    /// @param logger The logger instance to use
+    ///
+    gbts_seeding_algorithm(
+        const gbts_seedfinder_config& cfg, const memory_resource& mr,
+        vecmem::copy& copy, queue_wrapper& queue,
```
krasznaa: Since the AMD backend needs some changes anyway, now I'll leave this to you as well. (I was planning on making this tiny change myself if everything else would've gone well.)

Now that #1352 is in, let's do the same change to this new algorithm as well!


```suggestion
        const vecmem::copy& copy, queue_wrapper& queue,
```

krasznaa (changes requested):
Let me know if you need help with sorting out the SYCL-AMD issue!

mimodekjaer:
I have changed the functions in the common to use the traccc math header. 
For the CUDA implementation this does not change the kernel runtimes.

krasznaa (approved):
👍

PR #1356, opened 2026-06-24
Disable smem spilling for relocatable device code

stephenswat:
The shared memory register spilling pragma doesn't worm in relocatable CUDA device code for obvious reasons. This commit updates the compiler macro we use to invoke this PTX pragma so that it doesn't run when the CUDA compiler is set to emit relocatable code, or when it is running in debug mode (which also enabled RDC).


krasznaa (approved):
It looks correct. 🤞

PR #1357, opened 2026-06-26
Fix OOB write for halo threads in `find_tracks`

stephenswat:
This commit fixes a bug in the `find_tracks` kernel where halo threads would attempt to write outside the bounds of the parameter counting vector. This is easily fixed by wrapping the else clause in an additional check to establish whether the thread is actually part of the functional grid.


CMakeLists.txt
```diff
@@ -162,6 +162,7 @@ if( TRACCC_SETUP_VECMEM )
 endif()
```
niermann999, line 162: Changes in this file are debugging leftovers?

niermann999 (changes requested):

CMakeLists.txt
stephenswat: Whoops

niermann999 (approved):

